PR #76, opened 2014-12-31
Fix condition in build script

ptrv:


ptrv:
Without the patch I get the following error
```
./build.sh: line 147: [: too many arguments
```
You can see the line also in the Travis log.

The solution is to quote the $cmake_args within the square brackets. The second problem is the condition. `=` and `!=` do not work with pattern matching. The condition won't be true. You can try it by running this: `./build.sh --system-libclang`, which should call `usage` and exit, but doesn't.

Valloric:
You need to state you signed the Google CLA.

Valloric:
Ah, I saw on a different pull request that you said you signed it. In the future though, you always need to state you signed it in every PR.

Valloric:
Thanks for the PR!

ptrv:
> In the future though, you always need to state you signed it in every PR.

Ok, got it

PR #79, opened 2015-01-17
Improve Objective-C semantic completion triggering

alexdunn:
Related to Valloric/YouCompleteMe#84.

With these additions, triggers for Objective-C semantic completion behave like Xcode.  This should be a sensible default because 99% of Objective-C developers are going to be used to Xcode.

I added both positive and negative unit tests for every case I could think of and thoroughly tested it by hand in a real project.  All unit tests pass. It works very well!

All of the syntax situations presented [here](http://clang.llvm.org/docs/ObjectiveCLiterals.html) are covered. 

Without these changes, Objective-C semantic triggering is unusable.  It fails to trigger in 90% of the cases a developer will expect it.

### DIFF
--- a/ycmd/completers/completer_utils.py
+++ b/ycmd/completers/completer_utils.py
@@ -135,21 +135,6 @@ DEFAULT_FILETYPE_TRIGGERS = {
             r're!\[[_a-zA-Z]+\w*\s',    # bracketed calls
             r're!^\s*[^\W\d]\w*\s',     # bracketless calls
             r're!\[.*\]\s',             # method composition
-            # literals
-            r're!\@"\.*"\s',
-            r're!\@\w+\.*\w*\s',
-            r're!\@\(\w+\.*\w*\)\s',
-            r're!\@\(\s*',
-            r're!\@\[.*\]\s',
-            r're!\@\[\s*',
-            r're!\@\{.*\}\s',
-            r're!\@\{\s*',
-            r're!\@\'.*\'\s',
-            # variables
-            '#ifdef ',
-            r're!:\s*',
-            r're!=\s*',
-            r're!,\s*',
            ],
   'ocaml' : ['.', '#'],
   'cpp,objcpp' : ['->', '.', '::'],
--- a/ycmd/completers/completer_utils.py
+++ b/ycmd/completers/completer_utils.py
@@ -130,7 +130,27 @@ TRIGGER_REGEX_PREFIX = 're!'
 
 DEFAULT_FILETYPE_TRIGGERS = {
   'c' : ['->', '.'],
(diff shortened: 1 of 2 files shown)


alexdunn:
I signed the [Contributor License Agreement](https://developers.google.com/open-source/cla/individual).  Looking for a way to link to proof that I signed it, but I don't see it.

alexdunn:
If these changes are accepted, I'll also open a pull request to update Readme.md and youcompleteme.txt in the YouCompleteMe project with these new defaults.

Valloric:
Sorry for the delay in responding, I saw this PR a week ago but forgot about it.

Thanks for working on this! YCM has sub-par support for Objective-C (mostly because I know nothing about it) and any work that goes into improving this state of affairs is welcome.

I do have questions about some parts of this PR. If I'm reading the code correctly, it seems you are adding semantic triggers for literals and local variable names. 

YCM policy and behavior is to auto-trigger semantic completion only for object methods and the like where you have `<item-with-children><operator><child>` constructions. So for python, `foo.` will show semantic completion for methods/properties of `foo`; in C++, we'd do the same for `foo.`, `bar->` and `zoo::`, where `foo` is an object, `bar` is a pointer to an object and `zoo` is a namespace/class.

For "top-level" completions like when typing local identifiers/constants, the identifier completion engine kicks in and does the work. The reason for the split is that the ident engine is orders of magnitude faster than the semantic engines (which can be very slow in pathological cases so you don't want to trigger them too often).

I understand Objective-C is a bit special in this regard so also auto-triggering on function parameter in those bracketed calls might be warranted.

Am I reading the PR correctly? If I am, I'd appreciate it if you could remove the triggers that would make Objective-C support inconsistent with YCM behavior when completing other languages. We don't want to present different mental models of how YCM works depending on which language one is typing.

alexdunn:
Thanks for taking a look!  Yes, you read it correctly.  I noticed shortly after submitting my pull request that this support was different from the other languages supported by ycmd.

I think including these things is actually the right thing to do.  Very often you'll want to reference a variable that is defined in a different file.  Perhaps in the prefix header or perhaps in a third party framework.  That won't be completed by the ident engine.  Objective-C makes massive use of global constants. 

I think different mental models might be needed to support varied languages that do indeed have the programmer thinking differently from programmers of other languages.  However, it doesn't matter very much.  I'll simply keep my extra triggers in my .vimrc.  At least baseline support in ycmd will work correctly now.  I'll go ahead and fix this pull request.  Thanks!

alexdunn:
Alright, the triggers on variables and constants have been removed!

Valloric:
Thanks!

PR #80, opened 2015-01-18
make build.py choose python2

Gordin:
The build script will error out on distributions that have already switched their python binary to python3 (i.e. archlinux, not sure if there are others). Since 'The build script requires Python version > 2.6 and < 3.0.' I think it makes sense to use env python2 instead of env python. This makes the script work in archlinux out of the box.

### DIFF
--- a/build.py
+++ b/build.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 import os
 import os.path as p


Valloric:
The problem is that that doesn't work on systems that don't have a `python2`, like Mac OS X. On Yosemite, `/usr/bin/env python2` gives me an error.

Arch making `/usr/bin/python` point to python3 instead of python2 was IMO a terrible decision that breaks tons of stuff. It has caused me no end of pain on many projects.

Anyway, I don't know of a good solution here. For now, just call `python2 build.py`.

Gordin:
OK, I didn't know there were systems without a python2... Would making the install.sh from YouCompleteMe check for a python2 binary and run python2 "$build_file" "$@" instead of just "$build_file" "$@" be an acceptable change? That would also make this work.

Also, seems like one of the builds failed in a unittest for C# completion. Pretty sure that has nothing to do with my change, maybe you want to look at this: https://travis-ci.org/Valloric/ycmd/jobs/47406821#L1370

Valloric:
> OK, I didn't know there were systems without a python2... Would making the install.sh from YouCompleteMe check for a python2 binary and run python2 "$build_file" "$@" instead of just "$build_file" "$@" be an acceptable change? That would also make this work.

Yeah that's what I'm thinking of doing. I should get around to it tomorrow. Until then, stick to an older version that works.

> Also, seems like one of the builds failed in a unittest for C# completion.

That's just Travis being flaky, ignore it.

Valloric:
> Yeah that's what I'm thinking of doing. I should get around to it tomorrow.

You could also send a pull request for this if you wish. If you do, don't forget to sign the Google CLA and state that you have (details in CONTRIBUTING.md).

Gordin:
Done https://github.com/Valloric/YouCompleteMe/pull/1323

PR #81, opened 2015-01-22
Add CXTranslationUnit_Incomplete to the parse options.

r4nt:
This makes reparsing on large TUs > 10x faster. So far I haven't found
any diagnostics this might miss (the documentation says that it does not
do some template instantiations, but I have not yet been able to provide
an example where we would not diagnose the problem).

### DIFF
--- a/cpp/ycm/ClangCompleter/TranslationUnit.cpp
+++ b/cpp/ycm/ClangCompleter/TranslationUnit.cpp
@@ -38,7 +38,8 @@ namespace {
 unsigned editingOptions() {
   return ( CXTranslationUnit_DetailedPreprocessingRecord |
            clang_defaultEditingTranslationUnitOptions() ) |
-         CXTranslationUnit_IncludeBriefCommentsInCodeCompletion;
+         CXTranslationUnit_IncludeBriefCommentsInCodeCompletion |
+         CXTranslationUnit_Incomplete;
 }
 
 unsigned completionOptions() {


oblitum:
This is nice. I didn't test it but I wonder whether the `CXTranslationUnit_SkipFunctionBodies` flag can also have any impact, meaning not to skip the current scope but skip bodies of header only libraries with heavy template functions.

r4nt:
@oblitum - yea, playing around with more flags is probably a good idea (also, CacheComplationResults), but I'd take the 10x improvement first ;)

Valloric:
ycmd calls libclang's `clang_defaultEditingTranslationUnitOptions()` function for which the docs state:

>Returns the set of flags that is suitable for parsing a translation unit that is being edited.

>The set of flags returned provide options for clang_parseTranslationUnit() to indicate that the translation unit is likely to be reparsed many times, either explicitly (via clang_reparseTranslationUnit()) or implicitly (e.g., by code completion (clang_codeCompletionAt())). The returned flag set contains an unspecified set of optimizations (e.g., the precompiled preamble) geared toward improving the performance of these routines. The set of optimizations enabled may change from one version to the next.

This is exactly the use case for ycmd, which means that if the Incomplete flag is something we want to pass, it should probably already be returned from `clang_defaultEditingTranslationUnitOptions()`, no?

The whole point of `clang_defaultEditingTranslationUnitOptions()` is that clients can call it and if future version of libclang add new optimization flags, clients automatically get them.

Something is certainly wrong here; either the flag isn't all it's cracked up to be, or it should be included in `clang_defaultEditingTranslationUnitOptions()`. I don't see a third option.

oblitum:
@r4nt `CXTranslationUnit_CacheCompletionResults` is already part of `clang_defaultEditingTranslationUnitOptions()`:

```c++
unsigned clang_defaultEditingTranslationUnitOptions() {
  return CXTranslationUnit_PrecompiledPreamble |
         CXTranslationUnit_CacheCompletionResults;
}
```

I find it interesting since [the first time I tried to use this flag](https://github.com/Rip-Rip/clang_complete/pull/187#issuecomment-7844527) it provided unwanted side-effects. From that time I guess I also didn't find an issue using `CXTranslationUnit_Incomplete`.

oblitum:
My current try is:

```
unsigned editingOptions() {
  return clang_defaultEditingTranslationUnitOptions()          |
         CXTranslationUnit_Incomplete                          |
         CXTranslationUnit_SkipFunctionBodies                  |
         CXTranslationUnit_DetailedPreprocessingRecord         |
         CXTranslationUnit_IncludeBriefCommentsInCodeCompletion;
}
```

no issues so far...

oblitum:
Ah, `CXTranslationUnit_SkipFunctionBodies` indeed have a side-effect. I stopped to receive diagnostics.

r4nt:
Yea, skipping function bodies is certainly not what we need.

Valloric:
I've been using `CXTranslationUnit_Incomplete` for a ~week or so and I haven't had issues. I think there were one or two instances of peculiar diagnostics that the full compiler didn't complain about, but if that's the only issue I say we go with it. Everything is substantially faster with this flag.

cpp/ycm/ClangCompleter/TranslationUnit.cpp
```diff
@@ -38,7 +38,8 @@ namespace {
 unsigned editingOptions() {
   return ( CXTranslationUnit_DetailedPreprocessingRecord |
            clang_defaultEditingTranslationUnitOptions() ) |
-         CXTranslationUnit_IncludeBriefCommentsInCodeCompletion;
+         CXTranslationUnit_IncludeBriefCommentsInCodeCompletion |
+         CXTranslationUnit_Incomplete;
```
Valloric: Could you fix the formatting here while you're at it? The parens are unnecessary and the ordering is inconsistent. Make it `clang_defaultEditingTranslationUnitOptions()` and then the specific options each on their own line.

Valloric:
This is now implemented; thanks for the PR!

PR #82, opened 2015-01-23
Use the omnisharp port specified in user_options

akhosravian:
I saw this issue: 

https://github.com/Valloric/YouCompleteMe/issues/802

Obviously, two completion plugins active at the same time is a bad idea, but Omnisharp has functionality that isn't exposed in YCM (e.g. refactoring tools, find usages of a symbol, etc.). With this change, YCM will start the Omnisharp server listening on a fixed port. This value is already being set to 2000 (where Omnisharp's vim plugin expects the server) in default_settings.json, but was not being used by anything.

I've signed the CLA.

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -237,7 +237,7 @@ class CsharpCompleter( Completer ):
       raise RuntimeError( 'Autodetection of solution file failed.\n' )
     self._logger.info( u'Loading solution file {0}'.format( path_to_solutionfile ) )
 
-    self._omnisharp_port = utils.GetUnusedLocalhostPort()
+    self._omnisharp_port = int( self.user_options.get( 'csharp_server_port', utils.GetUnusedLocalhostPort() ) )
 
     # we need to pass the command to Popen as a string since we're passing
     # shell=True (as recommended by Python's doc)
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -237,7 +237,10 @@ class CsharpCompleter( Completer ):
       raise RuntimeError( 'Autodetection of solution file failed.\n' )
     self._logger.info( u'Loading solution file {0}'.format( path_to_solutionfile ) )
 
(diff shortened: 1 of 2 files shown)


Valloric:
Thanks for the PR!

While the Travis testrun is ongoing, some test configurations have already finished with failures. You need to get Travis to green.

Valloric:
Hm, that `csharp_server_port` option in default_settings.json seems to have snuck in. I don't see it documented anywhere in YCM's Readme and like you said, it was unused.

You also need to:
- send PRs to document the option in YCM
- change the default value of the option to 0
- change the logic you've implemented to use an unused local port if 0 is the port set.

akhosravian:
This change fixed tests for me, but I'm not sure if it is the right change:

The tests were failing because the code sends the OmniSharp server a stopserver command, and then would try to start a new server that would bind the same port. The OmniSharp server would error out due to failing to bind the port.

I fixed this by waiting for the OmniSharp server process to terminate after sending the stopserver command -- in addition to the CsharpCompleter's StopServer command this also affects its RestartServer command and the Shutdown method.

ycmd/completers/cs/cs_completer.py
```diff
@@ -237,7 +237,10 @@ def _StartServer( self, request_data ):
       raise RuntimeError( 'Autodetection of solution file failed.\n' )
     self._logger.info( u'Loading solution file {0}'.format( path_to_solutionfile ) )
 
-    self._omnisharp_port = utils.GetUnusedLocalhostPort()
+    self._omnisharp_port = int( self.user_options.get( 'csharp_server_port', 0 ) )
```
Valloric: Could you extract this small bit of port-selection logic into a helper method? `_StartServer` is already pretty big.

Valloric:
Just one small comment added and this should be good to go.

WRT waiting on the OmniSharp server to stop... I'm not sure what's the right approach here. Let's go with what you have and we'll see how it works out.

Valloric:
Thanks for the PR! For future reference, Github doesn't send notifications when new commits are added to a PR, so you'll have to post a new comment after adding commits if you want me to take another look. :)

PR #85, opened 2015-02-05
Fix CustomPythonCmakeArgs() on Python 2.6

ntkme:
This PR fix the following error in a rare case where Python 2.6 is system default:

```
  RAN: '/usr/bin/python -c import sys;i=sys.version_info;print "python%d.%d" % (i.major, i.minor)'

  STDOUT:


  STDERR:
  File "<string>", line 1, in <module>
AttributeError: 'tuple' object has no attribute 'major'
```
---
CLA signed.

### DIFF
--- a/build.py
+++ b/build.py
@@ -67,7 +67,7 @@ def CustomPythonCmakeArgs():
   else:
     which_python = sh.python(
       '-c',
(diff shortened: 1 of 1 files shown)


Valloric:
Thanks!

PR #87, opened 2015-02-09
Support for clang completer subcommands GetType and GetParent

puremourning:
## Requirement

Particularly in the context of c++11 and the pervasive use of `auto`, it is useful to have YouCompleteMe, via clang, deduce the type of a word within the source. Further, in large files or large projects, jumping around using searches and location lists, it can be useful to quickly determine the context of the current cursor position.

Typical example:

```c++
#include <third_party_api.h>

int main()
{
    auto x = third_party_init();
    // long, boring or otherwise complex code
    // ...
    // ... etc.
    do_something_with(x); // <-- what was x again?
}
```

## Specification

User can deduce the type of the word under the cursor by issuing a completer subcommand.
User can deduce the semantic parent of the cursor location by issuing a completer subcommand,
Commands can be bound using standard vim script or simple mappings.

### Deducing type

Clang provides simple APIs for this: `clang_GetTypeSpelling()` via `clang_GetCursorType()`. This returns the "lexical" type of the variable/method etc. It an be useful to know the "derived" type as well, and clang provides `clang_GetCannonicalType()` for that. So we return a simple string containing the type spelling of the type and append the canonical type to it if that is different from the simple type.

### Deducing semantic parent

Clang provies `clang_GetCursorSemanticParent()` for this; we just return the spelling returned by `clang_GetCursorDisplayName()`.

## Testing

Ran all existing regression tests for both ycmd and YouCompleteMe :  PASS
Added new regression tests for GetType and GetParent : PASS

The cyclomatic complexity check failed for clang_completer.py, so rather than suppressing it or changing the test suite, the implementation was changed from a `if: elif:` block to a dictionary mapping commands to a function to execute. Due to this change, tests were added for the existing commands too to ensure all changed code is exercised. Unfortunately, the behaviour of these commands seems strange, as commented. I have confirmed the behaviour of the tests is identical with and without the change to use a dictionary, so there is no regression.

As I had a number of problems getting the tests to work, i've also included a TESTS.md with some instructions on how to run the tests, at least on my platform (RHEL linux). Hopefully this can be extended by others with experience running them on other platforms.

## Foibles

As noted in the tests, and in the code, there is a clang bug that when executing GetType on the word auto, or on the line where a variable is declared, clang always returns `auto` rather than the deduced type. However, executing it on subsequent usage of the variable works roughly as expected, except that clang claims that the type and the canonical_type differ, though their spellings are the same. We could work around this, but it would be better for libClang to be fixed upstream.

[...]

PR #88, opened 2015-02-12
New completer subcommands GetType and GetParent

puremourning:
GetType - returns the type of the word at a specific location. This is used by
YouCompleteMe and echo'd in vim. Where the canonical type (the "derived" type,
e.g. for a typedef) is different from the declared type, that is also returned,
separated with " => ".
GetParent - returns the semantic parent of the word at a specific location.
This is also echo'd in vim by YouCompleteMe.

One other change was required to satisty the cylomatic complexity check in the
test suite: the  if:elif: pattern in OnUserCommand was changed to a pythonic
switch (a dictionary)

see also https://github.com/Valloric/ycmd/pull/87 for details and history.

### DIFF
--- a/cpp/ycm/ClangCompleter/TranslationUnit.cpp
+++ b/cpp/ycm/ClangCompleter/TranslationUnit.cpp
@@ -243,6 +243,92 @@ Location TranslationUnit::GetDefinitionLocation(
   return Location( clang_getCursorLocation( definition_cursor ) );
 }
 
+std::string TranslationUnit::GetTypeAtLocation(
+  int line,
+  int column,
+  const std::vector< UnsavedFile > &unsaved_files,
+  bool reparse ) {
+
+  if (reparse)
+    ReparseForIndexing( unsaved_files );
+
+  unique_lock< mutex > lock( clang_access_mutex_ );
+
+  if ( !clang_translation_unit_ )
+    return "Internal error: no translation unit";
+
+  CXCursor cursor = GetCursor( line, column );
+
+  if ( !CursorIsValid( cursor ) )
+    return "Internal error: cursor not valid";
+
+  CXType type = clang_getCursorType( cursor );
+
+  std::string type_description = 
+                        CXStringToString( clang_getTypeSpelling( type ) );
+
+  if ( type_description.empty() )
+    return "Unknown type";
+ 
+  // We have a choice here; libClang provides clang_getCanonicalType which will
+  // return the "underlying" type for the type returned by clang_getCursorType
+  // e.g. for a typedef
+  //     type = clang_getCanonicalType( type );
+  //
+  // Without the above, something like the following would return "MyType"
+  // rather than int:
+  //     typedef int MyType;
+  //     MyType i = 100; <-- type = MyType, canonical type = int
+  //
+  // There is probably more semantic value in calling it MyType. Indeed, if we
+  // opt for the more specific type, we can get very long or
+  // confusing STL types even for simple usage. e.g. the following:
+  //     std::string test = "test"; <-- type = std::string; 
+  //                                    canonical type = std::basic_string<char>
+  //
+  // So as a compromise, we return both if and only if the types differ, like
+  //     std::string => std::basic_string<char>
+
+  CXType canonical_type = clang_getCanonicalType( type );
+
+  if ( !clang_equalTypes( type, canonical_type ) )
+  {
+    type_description += " => ";
+    type_description += CXStringToString(clang_getTypeSpelling(canonical_type));
+  }
+
+  return type_description;
+}
+
+std::string TranslationUnit::GetEnclosingFunctionAtLocation(
+  int line,
+  int column,
+  const std::vector< UnsavedFile > &unsaved_files,
+  bool reparse ) {
+
+  if (reparse)
+    ReparseForIndexing( unsaved_files );
+
+  unique_lock< mutex > lock( clang_access_mutex_ );
+
+  if ( !clang_translation_unit_ )
+    return "Internal error: no translation unit";
+
+  CXCursor cursor = GetCursor( line, column );
+
+  if ( !CursorIsValid( cursor ) )
+    return "Internal error: cursor not valid";
+
+  CXCursor parent = clang_getCursorSemanticParent( cursor );
+
+  std::string parent_str = 
+                  CXStringToString( clang_getCursorDisplayName( parent ) );
+
+  if (parent_str.empty())
+    return "Unknown semantic parent";
+
+  return parent_str;
+}
 
 // Argument taken as non-const ref because we need to be able to pass a
 // non-const pointer to clang. This function (and clang too) will not modify the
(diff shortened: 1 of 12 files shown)


cpp/ycm/ClangCompleter/TranslationUnit.cpp
```diff
@@ -243,6 +243,100 @@ Location TranslationUnit::GetDefinitionLocation(
...
+  CXType type = clang_getCursorType( cursor );
+
+  std::string type_description = 
+                        CXStringToString( clang_getTypeSpelling( type ) );
+
+  if ( type_description.empty() )
+    return "Unknown type";
+ 
+  //
+  //
```
Valloric: Remove these two empty commented-out lines.

cpp/ycm/ClangCompleter/TranslationUnit.cpp
```diff
@@ -243,6 +243,100 @@ Location TranslationUnit::GetDefinitionLocation(
...
+  // opt for the more specific type, if that means that we can get very long or
+  // confusing stl types even for simple usage. e.g. the following:
+  //
+  // std::string test = "test"; <-- type = std::string; 
+  //                                canonical=std::basic_string<char>
+  //
+  // so as a compromise, we return both if and only if the types differ, like
+  //
+  // std::string => std::basic_string<char>
+  //
```
Valloric: Another empty comment.

cpp/ycm/ClangCompleter/TranslationUnit.cpp
```diff
@@ -243,6 +243,100 @@ Location TranslationUnit::GetDefinitionLocation(
...
+
+  std::string type_description = 
+                        CXStringToString( clang_getTypeSpelling( type ) );
+
+  if ( type_description.empty() )
+    return "Unknown type";
+ 
+  //
+  //
+  // we have a choice here. libClang provides clang_getCanonicalType which will
```
Valloric: "we" -> "We" :)

ycmd/completers/cpp/clang_completer.py
```diff
@@ -108,26 +108,72 @@ def DefinedSubcommands( self ):
...
-      return self._GoToDeclaration( request_data )
-    elif command == 'GoTo':
-      return self._GoTo( request_data )
-    elif command == 'GoToImprecise':
-      return self._GoToImprecise( request_data )
-    elif command == 'ClearCompilationFlagCache':
-      return self._ClearCompilationFlagCache()
-    raise ValueError( self.UserCommandsHelpMessage() )
-
+    #
```
Valloric: Another empty comment.

ycmd/completers/cpp/clang_completer.py
```diff
@@ -108,26 +108,72 @@ def DefinedSubcommands( self ):
...
              'GoTo',
              'GoToImprecise',
-             'ClearCompilationFlagCache']
+             'ClearCompilationFlagCache',
+             'GetType',
+             'GetParent']
 
 
-  def OnUserCommand( self, arguments, request_data ):
+  def OnUserCommand( self, arguments, request_data ):
```
Valloric: Down with trailing whitespace! :)

ycmd/completers/cpp/clang_completer.py
```diff
@@ -108,26 +108,72 @@ def DefinedSubcommands( self ):
...
-    elif command == 'ClearCompilationFlagCache':
-      return self._ClearCompilationFlagCache()
-    raise ValueError( self.UserCommandsHelpMessage() )
-
+    #
+    # An if/elif for each of the options fails the cyclomatic complexity check
+    # in the test suite. python doesn't have a switch statement, and
+    # recommendations typically point to using a dictionary. So we use a
+    # dictionary to map the entered command to the method which implements it
+    # via some (arguably ugly) metaprogramming.
```
Valloric: I don't think you need the above paragraph. The bottom part of the comment is fine.

ycmd/completers/cpp/clang_completer.py
```diff
@@ -185,11 +231,35 @@ def _GoToImprecise( self, request_data ):
...
+
+    type = getattr( self._completer, func )(
+        ToUtf8IfNeeded( filename ),
+        line,
+        column,
+        files,
+        flags,
+        reparse)
+
+    if type == "":
```
Valloric: Just `if type`. But don't use the name `type`, it's the name of a Python built-in function and you're overwriting it. You can see the blue color in the syntax-highlighting here..

ycmd/tests/subcommands_test.py
```diff
@@ -59,19 +59,8 @@ def foo():
...
-  char c;
-};
-
-int main()
-{
-  Foo foo;
-  return 0;
-}
-"""
+  contents = open( \
```
Valloric: You don't need the slash inside open parens. Python will parse this just fine.

ycmd/tests/subcommands_test.py
```diff
@@ -88,6 +77,281 @@ def RunCompleterCommand_GoTo_Clang_ZeroBasedLineAndColumn_test():
...
+    'line_num'          : 10,
+    'column_num'        : 3,
+    'contents'          : contents,
+    'filetype'          : 'cpp'
+  }
+  common_response = {
+    'filepath'  : '/foo',
+  }
+
+  for test in tests:
```
Valloric: What you probably want to use here is the Test Generators available in Nose. [More details here](http://nose.readthedocs.org/en/latest/writing_tests.html#test-generators). Also, search for "yield" in this file to see how it's used. Also look at usage of it in `identifier_utils_test.py` (search for "yield" again).

Valloric:
I added some more minor comments, but this is looking good!

cpp/ycm/ClangCompleter/TranslationUnit.cpp
puremourning: I have reformatted and proof-read this comment.

ycmd/completers/cpp/clang_completer.py
puremourning: fixed

ycmd/completers/cpp/clang_completer.py
puremourning: removed

ycmd/completers/cpp/clang_completer.py
puremourning: renamed to message

ycmd/tests/subcommands_test.py
puremourning: personally, I like the slash to tie the two lines together visually, but I have removed it as advised

ycmd/tests/subcommands_test.py
puremourning: I have changed this to use generators

puremourning:
Updated as a00fe22

Valloric:
Thanks again!

BatmanAoD:
This appears to require recompiling the support library, but I didn't get a notification when I ran "BundleUpdate"; is this a bug?

puremourning:
yes i think it might be. @Valloric how do you want to tackle this?

BatmanAoD:
I am so excited to be at the *bleeding edge of YCM functionality.*

Valloric:
Ah, seems we forgot to bump the internal API version numbers. I'll get on it.

Valloric:
OK, updated both ycmd and YCM. The message about recompiling should show up after syncing now.

PR #89, opened 2015-02-12
Enhance test suite

puremourning:
Enhance test suite

Update run_tests.sh to allow arguments to be passed directly to "nosetests".
This allows developers to run specific test scripts or specific test methods.

Ensure also that the test suite does not hang forever when problems with, say,
the installation cause OmniSharpServer to fail to start. Instead timeout after
20s.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -5,7 +5,7 @@ set -e
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 function usage {
-  echo "Usage: $0 [--no-clang-completer] [--skip-build]"
+  echo "Usage: $0 [--no-clang-completer] [--skip-build] [-- nose args]"
   exit 0
 }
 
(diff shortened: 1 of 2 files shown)


puremourning:
Updated again (hopefully last time) to only include the appropriate commit.

ycmd/tests/test_utils.py
```diff
@@ -85,5 +88,9 @@ def WaitUntilOmniSharpServerReady( app ):
     if result:
       raise RuntimeError( "OmniSharp failed during startup." )
     time.sleep( 0.2 )
+    retries = retries - 1
+
+  if not success:
+    raise RuntimeError( "Time out waiting for OmniSharpServer" )
```
Valloric: "Time out" -> "Timeout"

run_tests.sh
```diff
@@ -50,7 +56,15 @@ for directory in "${SCRIPT_DIR}"/third_party/*; do
 done
 
 if $use_clang_completer; then
-  nosetests -v "${SCRIPT_DIR}/ycmd"
+  if [ "$*" ]; then
+    nosetests -v $@
```
Valloric: This is peculiar to me; I'd assume we want to pass the path to ycmd to `nosetests` always, and that the user only adds new args. In which case would the user not want the path to ycmd passed?

Valloric: Actually ignore what I just said. The current approach is sensible.

ycmd/tests/test_utils.py
puremourning: ok done

puremourning:
Updated as 0fddbb9

Valloric:
Thanks!

PR #91, opened 2015-02-12
Allow ycmd to respect max filesize option

mwbrown:
Currently ycmd will fail on files larger than the hardcoded MEMFILE_MAX even when the client (i.e Vim) is set to use files larger than 1MB. This allows ycmd to pull the option provided by the client and set MEMFILE_MAX accordingly.

### DIFF
--- a/ycmd/handlers.py
+++ b/ycmd/handlers.py
@@ -41,6 +41,7 @@ from ycmd.responses import BuildExceptionResponse, BuildCompletionResponse
 from ycmd import hmac_plugin
 from ycmd import extra_conf_store
 from ycmd.request_wrap import RequestWrap
+from sys import maxint
 
 
 # num bytes for the request body buffer; request.json only works if the request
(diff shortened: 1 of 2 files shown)


ycmd/handlers.py
```diff
@@ -236,6 +237,14 @@ def UpdateUserOptions( options ):
   if not options:
     return
 
+  # Fix up MEMFILE_MAX for overridden max filesize.
+  if 'disable_for_files_larger_than_kb' in options:
+    max_size_kb = options['disable_for_files_larger_than_kb']
+    if(max_size_kb >= 0):
+      bottle.Request.MEMFILE_MAX = max_size_kb * 1024
+    else:
+      bottle.Request.MEMFILE_MAX = maxint
```
Valloric: This can't be what we set if no value is set for the option since this would mean that Bottle allocates a freaking huge buffer for every single request, no matter the size of the request. 

Note that the option for YCM is the size that the client is willing to _send_, and is different from the size ycmd is willing to _receive_. 

The actual solution would probably be to replace Bottle with Flask for ycmd which would remove this MEMFILE_MAX nonsense entirely and solve a bunch of other problems as well. A smaller (and easier) change would be to remove all calls to `request.json` in `handlers.py` and replace it with "manual" parsing of json instead of leaving it up to Bottle.

Valloric:
Thank you for the PR! I'll close this without merging as per rationale from my inline comment.

PR #95, opened 2015-02-20
Change build.py to also pipe the make's stderr

mispencer:
Currently, build.py only pipes the make's stdout to its stdout. The stderr
is included in the exception output, but truncated. The lack of full
stderr for make causes difficulty debugging make issues.


Valloric:
Thanks!

In the future, please make sure that you always state in your PRs that you've signed the Google CLA. You've sent PRs before and I just so happen to remember that you've signed it, but my memory should not be relied upon. :)

Issue #97, opened 2015-02-23
Add a Hack completer


cweagans:
For people working with Hack (Facebook's superset of PHP), it's nice to have autocomplete functionality. hhvm's typechecker provides this out of the box (`hh_client --auto-complete`). There's already a vim plugin that does this, but integrating with ycmd instead gets the functionality to more editors.

Valloric:
I agree with this and would like to see a pull request for it. I don't see much point to having an issue open for this since nobody is committed to doing the work. We could also have issues like "add a completer for language X" for every programming language out there and it wouldn't make much sense unless there was actual intent to go and build it.

So I'll close this for now; pull requests adding Hack support are welcome, but I don't see myself personally implementing this.

cweagans:
I'm working on this right now, actually.

I have a question for you, if you don't mind: Hack and PHP both use the same file extension (.php), but Hack files start with `<?hh` instead of `<?php`. Is there some way of telling a completer not to activate unless the file starts with `<?hh`?

Valloric:
ycmd doesn't care about the extension; the JSON request will have the filetype of the code that's supposed to be completed. So for Hack files, the filetype should be `hack`, and for PHP files it should be `php`.

In other words, differentiating between the two is a client-side issue, not a server-side one.

cweagans:
Ah, got it. That makes much more sense!

micbou:
Hack is too niche to be included and maintained in ycmd. Closing.

PR #98, opened 2015-02-26
[WIP] Add a Hack completer

cweagans:
Related to https://github.com/Valloric/ycmd/issues/97. Not entirely sure how to test this, so posting early for feedback and possibly a nudge in the right direction?

### DIFF
--- a/ycmd/completers/hack/hack_completer.py
+++ b/ycmd/completers/hack/hack_completer.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+#
+# Copyright (C) 2011, 2012  Stephen Sugden <me@stephensugden.com>
+#                           Google Inc.
+#                           Stanislav Golovanov <stgolovanov@gmail.com>
+#
+# This file is part of YouCompleteMe.
+#
+# YouCompleteMe is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# YouCompleteMe is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with YouCompleteMe.  If not, see <http://www.gnu.org/licenses/>.
+
+import subprocess
+import json
+from ycmd.utils import ToUtf8IfNeeded
+from ycmd.completers.completer import Completer
+from ycmd import responses
+
+class HackCompleter( Completer ):
+  """
+  A completer that offers integration with the HHVM typechecker.
+  """
+
+  def __init__( self, user_options ):
+    super( HackCompleter, self ).__init__( user_options )
+
+  def SupportedFiletypes( self ):
+    """ Just hack """
+    return [ 'hack' ]
+
+  def ComputeCandidatesInner( self, request_data ):
+    # Get the contents of the file and break it up by line.
+    filename = request_data['filepath']
+    contents = request_data['file_data'][filename]['contents']
+    contents = contents.splitlines(True)
+
+    # Lines are zero indexed.
+    line = request_data['line_num'] - 1
+    column = request_data['column_num']
+
+    # Insert the autocomplete token into the file.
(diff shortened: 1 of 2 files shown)


Valloric:
I see you are shelling out to `hh_client`; is there a way for ycmd to talk to `hh_server` directly? that would be much faster than going through the command-line and all the serialization/deserialization and process startup/teardown overhead.

cweagans:
I'm not sure. The hh_server API is undocumented as far as I'm aware, but I can take a look at the hh_client source and see what I can dig up.

Valloric:
Ah, so is hh_client the "correct" way to talk to hh_server then? That would be sad, since it's wildly inefficient. I decided against integrating support for Racer because it too had this "call a command-line client" approach instead of a library API.

cweagans:
Technically, yeah. That's how the official vim plugin does it. It's pretty dumb, but I'll look and see if there's a real API somewhere.

vheon:
@cweagans: Why just for Hack and not for PHP too? From what I understand Hack is a superset of PHP infact according to their github page they run Wordpress for example which, to my knowledge, is a PHP CMS.

Anyway I was looking at the `hh_client` the other night since I was bored so I will just write here what I found. I didn't know that the client and the server are written in ML, which being a declarative programming language is pretty easy to read :+1: 

* From what I understand the client will start a server when it is needed, so If you do two consecutive requests the server is started on the first and for the second request it is already running so It will be much faster. The sad thing about using the client is that the path of our request would be: from the vim client we preper a json for ycmd, then ycmd will call the hhvm completer which will shell out to `hh_client`, to call the `hh_client` we grab the json, put the token into the contents and pass it on stdin to the `hh_client` which then it will take stdin, make it a string, then repack it into a json and then send it to the server. It seems a lot of overhead.
* The server have to have a knowledge about the directory of the project from what I understand, in fact, when the client starts the server it do it like this: https://github.com/facebook/hhvm/blob/29ef150d79ace0c6a00de7d44d8b5501c56fc4d3/hphp/hack/src/client/clientStart.ml#L42-L45 which practically it will do `hh_server -d root_path something_that_i_dont_understand` and apparently that directory could go out of sync but I didn't look well the protocol there.
* The server can do a lots of things: https://github.com/facebook/hhvm/blob/29ef150d79ace0c6a00de7d44d8b5501c56fc4d3/hphp/hack/src/client/clientEnv.ml#L13-L34 these are the kinds of request that you can make to the server, but I would start small with just autocompletion.

cweagans:
@vheon I've already looked into this extensively. The Hack completer would be based on hh_client, and would use the hhvm typechecker's built-in autocomplete functionality. For PHP, we need to do something different because the typechecker doesn't support autocompleting normal PHP. To that end, I've hired someone to extract the CodeIntel library from Komodo Edit and write a completer that uses it in ycmd. That effort is ongoing and should hopefully be done next month.

Ideally, we'd write our own client for hh_server instead of just shelling out to hh_client, but that's going to be a ton of effort. Also, the existing vim plugin just shells out to hh_client and it seems to be fast enough. I'm inclined to go with the simple option for now, and then if it's a problem going forward, then implement our own client.

Valloric:
What's the state of this PR?

cweagans:
I haven't touched it in a while. I'm not sure how to talk directly to hh_server, so https://github.com/Valloric/ycmd/pull/98#issuecomment-76244869 is still an open question I guess.

Valloric:
That's perfectly fine; I'll close this PR for now then. We can reopen in the future if needed.

icholy:
The Go completer shells out to `gocode` for every completion request. `hh_client` seems pretty similar to me.

Valloric:
@icholy Correct, and that's pretty damn unfortunate. I've heard from gocode devs that this is something they want to change, and rightly so.

icholy:
gocode does have a client/server architecture (using [net/rpc](http://golang.org/pkg/net/rpc/)). I think they use some type of codegen to create the client methods from the server code. AFAIK it's in maintenance mode though, so I'd be surprised if any big changes came down the pipeline.

Regardless, I don't see a problem with it. And I mean that literally. I use gocode every day and have never had it choke or lag. Is it as efficient as it could be? No. Does it really matter? I don't think so.

Point being, end users would have appreciated a hack completer. The fact that it shelled out would (probably) not have made a noticeable difference in the UX.

Valloric:
If the perf hit is truly not noticeable (like with gocode), then I'd be fine with shelling out to hh_client.

icholy:
Ok, I might look into resurrecting this PR at some point.

PR #100, opened 2015-03-11
rename variables from UPPER_CASE to _lower_case

vheon:
I was looking in the code and say that TODO.

I have signed the CLA.

### DIFF
--- a/ycmd/handlers.py
+++ b/ycmd/handlers.py
@@ -47,26 +47,25 @@ from ycmd.request_wrap import RequestWrap
 # size is less than this
 bottle.Request.MEMFILE_MAX = 1000 * 1024
 
-# TODO: rename these to _lower_case
-SERVER_STATE = None
-LOGGER = logging.getLogger( __name__ )
+_server_state = None
+_logger = logging.getLogger( __name__ )
 app = bottle.Bottle()
 
 
 @app.post( '/event_notification' )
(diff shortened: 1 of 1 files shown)


Valloric:
I love what you're doing here. :)

I can see that the Travis run was actually fine, but one build errored out. That seems to be happening more often than usual...

vheon:
>I love what you're doing here. :)

Those are really simple things :)

> I can see that the Travis run was actually fine, but one build errored out. That seems to be happening more often than usual...

I ran the tests locally before I opened the PR and all went well. Could be travis under too much load or we saturate some resources? I mean it complaints that a test didn't complete in 10 minutes which seems strange since we wait for a maximum of 2 minutes on omnisharp.

Valloric:
> Those are really simple things :)

Yeah, but it's the grungy stuff nobody wants to do. Someone actually putting in the effort to do that is very much appreciated!

>  Could be travis under too much load or we saturate some resources? I mean it complaints that a test didn't complete in 10 minutes which seems strange since we wait for a maximum of 2 minutes on omnisharp.

I'm pretty sure it's something with our tests that talk to OmniSharp. It was an issue before, it's probably an issue again. Some recent PRs messed around with the OmniSharp test setup helper functions, so it's probably something there that's causing problems.

PR #104, opened 2015-03-29
fixed complete in objective-c.

SolaWing:
> when completing multi-param in objective-c files, the origin implementation only
> completed the last param. this commit fixed it. it will complete the first param,
> then after the user input param, call completion again to get second param, and
> so on.
> it also support completing declared methods or inherited methods. the user can
> call the completion after input '+ ' or '- '

here I fixed objective-c support in ycm. I have used it for months. 
I think a good support for objective-c can earn more iOS & MAC developers. 
Relative Issue: [YouCompleteMe #1221](https://github.com/Valloric/YouCompleteMe/issues/1221)
here is a preview:
img2
(I already sign **Google Contributor License Agreement**)

### DIFF
--- a/cpp/ycm/ClangCompleter/CompletionData.cpp
+++ b/cpp/ycm/ClangCompleter/CompletionData.cpp
@@ -32,6 +32,8 @@ CompletionKind CursorKindToCompletionKind( CXCursorKind kind ) {
 
     case CXCursor_ClassDecl:
     case CXCursor_ClassTemplate:
+    case CXCursor_ObjCInterfaceDecl:
+    case CXCursor_ObjCImplementationDecl:
       return CLASS;
 
     case CXCursor_EnumDecl:
@@ -43,6 +45,9 @@ CompletionKind CursorKindToCompletionKind( CXCursorKind kind ) {
       return TYPE;
 
     case CXCursor_FieldDecl:
+    case CXCursor_ObjCIvarDecl:
+    case CXCursor_ObjCPropertyDecl:
+    case CXCursor_EnumConstantDecl:
       return MEMBER;
 
     case CXCursor_FunctionDecl:
@@ -51,6 +56,8 @@ CompletionKind CursorKindToCompletionKind( CXCursorKind kind ) {
     case CXCursor_ConversionFunction:
     case CXCursor_Constructor:
     case CXCursor_Destructor:
+    case CXCursor_ObjCClassMethodDecl:
+    case CXCursor_ObjCInstanceMethodDecl:
       return FUNCTION;
 
     case CXCursor_VarDecl:
@@ -90,7 +97,8 @@ bool IsMainCompletionTextInfo( CXCompletionChunkKind kind ) {
     kind == CXCompletionChunk_SemiColon    ||
     kind == CXCompletionChunk_Equal        ||
     kind == CXCompletionChunk_Informative  ||
-    kind == CXCompletionChunk_HorizontalSpace;
+    kind == CXCompletionChunk_HorizontalSpace ||
+    kind == CXCompletionChunk_Text;
 
 }
 
@@ -161,12 +169,18 @@ CompletionData::CompletionData( const CXCompletionResult &completion_result ) {
   uint num_chunks = clang_getNumCompletionChunks( completion_string );
   bool saw_left_paren = false;
   bool saw_function_params = false;
+  bool saw_placeholder = false;
 
   for ( uint j = 0; j < num_chunks; ++j ) {
     ExtractDataFromChunk( completion_string,
                           j,
                           saw_left_paren,
-                          saw_function_params );
+                          saw_function_params,
+                          saw_placeholder );
+  }
+  if ( original_string_.size() > 0 && 
+      original_string_[ original_string_.size() - 1 ] == '(' ) {
+    original_string_.erase( original_string_.size() - 1, 1 );
   }
 
   kind_ = CursorKindToCompletionKind( completion_result.CursorKind );
@@ -194,7 +208,8 @@ CompletionData::CompletionData( const CXCompletionResult &completion_result ) {
 void CompletionData::ExtractDataFromChunk( CXCompletionString completion_string,
                                            uint chunk_num,
                                            bool &saw_left_paren,
-                                           bool &saw_function_params ) {
+                                           bool &saw_function_params,
+                                           bool &saw_placeholder ) {
   CXCompletionChunkKind kind = clang_getCompletionChunkKind(
                                  completion_string, chunk_num );
 
@@ -212,6 +227,8 @@ void CompletionData::ExtractDataFromChunk( CXCompletionString completion_string,
     }
 
     else if ( saw_function_params && kind == CXCompletionChunk_RightParen ) {
+      saw_left_paren = false;
+      saw_function_params = false; // fix objc have more than one ()
       everything_except_return_type_.append( " " );
     }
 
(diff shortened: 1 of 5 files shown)


SolaWing:
> $ sudo apt-get -qq update
W: Failed to fetch http://www.rabbitmq.com/debian/dists/testing/main/i18n/Translation-en_US  Connection failed

oblitum:
Interesting, it seems to look like #1300.

Valloric:
@SolaWing Could you go into more detail on what this change does?

@oblitum What's your take on this PR and how it relates to #1300?

SolaWing:
@Valloric The main change is in method `ExtractDataFromChunk` of `cpp/ycm/ClangCompleter/CompletionData.cpp`.

Suppose I want to complete a objc method `- (void)test:(int)arg1 withArg2:(int)arg2 withArg3:(int)arg3`, 
the clang CXCompletionString will have these chunks:

    void test:(int)arg1   withArg2:(int)arg2   withArg3:(int)arg3
    ret  i    pl       sp i        pl       sp i        pl

    ret: ret type     i: typed text, need to insert           
    pl:  placeholder          sp: HorizontalSpace

So there will have **multiple** typed text, this is why the original implementation will only insert last part, since the insertText get overwritten when encounter typed text. Though it is no problem in c&c++ files, since in those files only have one typed text.

So in my implementation, if I encounter placeholder, I won't append typed text to insert text any more. In this way, the user will get the first part he want to complete. Once he insert param and print space, he can call complete to get the second typed text. In this complete situation, the cursor is at front of second typed text, and previous chunk will change to *informative* chunk, so the second typed text become the first typed text and user can get the complete.

Another situation is completing the declared method in header or super class. as the gif's second completing situation. in this situation, the clang CXCompletionString will have these chunks:

    (void)test:(int)arg1   withArg2:(int)arg2   withArg3:(int)arg3
    (t   )i   i(t  )t   sp i        (t  )t   sp i        (t)  t

    t: Text that should be inserted as part of a code-completion result
    i:  Typed Text, a user would be expected to type.

In this situation, the user need all part to complete, so I use a empty insert string, and append each part sequencely. Notice the `(` and `)` is also part of insert. this will affect c&c++ original behavior, which will also contain a `(` in the insert text. (  `)` may not be contained, since the param is a placeholder chunk)

oblitum:
@Valloric Sorry, it's just the visuals when when getting options after typing an argument, but I guess it's quite different in behavior and implementation.

cpp/ycm/ClangCompleter/CompletionData.cpp
```diff
@@ -226,11 +241,27 @@ void CompletionData::ExtractDataFromChunk( CXCompletionString completion_string,
     }
   }
 
-  if ( kind == CXCompletionChunk_ResultType )
-    return_type_ = ChunkToString( completion_string, chunk_num );
-
-  if ( kind == CXCompletionChunk_TypedText )
-    original_string_ = ChunkToString( completion_string, chunk_num );
+    switch (kind) {
+        case CXCompletionChunk_ResultType:
```
Valloric: Indentation here is wrong; indent by 2 spaces, not 4.

cpp/ycm/ClangCompleter/CompletionData.cpp
```diff
@@ -226,11 +241,27 @@ void CompletionData::ExtractDataFromChunk( CXCompletionString completion_string,
...
+            break;
+        case CXCompletionChunk_TypedText:
+        case CXCompletionChunk_Text:
+        // need to add paren to insert string when implementing inherit methods or declare methods in objc.
+        case CXCompletionChunk_LeftParen:
+        case CXCompletionChunk_RightParen:
+        case CXCompletionChunk_HorizontalSpace:
+            if (saw_placeholder == false){
+                auto&& str = ChunkToString(completion_string, chunk_num);
+                original_string_ += str;
```
Valloric: Just do `original_string_ += ChunkToString(...)`, no need for a `str` temp.

cpp/ycm/ClangCompleter/CompletionData.cpp
```diff
@@ -226,11 +241,27 @@ void CompletionData::ExtractDataFromChunk( CXCompletionString completion_string,
...
+        case CXCompletionChunk_Placeholder:
+            saw_placeholder = true;
+            break;
+        case CXCompletionChunk_TypedText:
+        case CXCompletionChunk_Text:
+        // need to add paren to insert string when implementing inherit methods or declare methods in objc.
+        case CXCompletionChunk_LeftParen:
+        case CXCompletionChunk_RightParen:
+        case CXCompletionChunk_HorizontalSpace:
+            if (saw_placeholder == false){
```
Valloric: Don't do `if (x == false)`, do `if (!x)`. Also add a space between `){`.

cpp/ycm/ClangCompleter/CompletionData.cpp
```diff
@@ -152,7 +160,9 @@ std::string RemoveTwoConsecutiveUnderscores( std::string text ) {
 } // unnamed namespace
 
 
-CompletionData::CompletionData( const CXCompletionResult &completion_result ) {
+CompletionData::CompletionData( const CXCompletionResult &completion_result )
+    :original_string_()
```
Valloric: Adding `:original_string_()` isn't necessary. You can remove it and get the same behavior.

cpp/ycm/ClangCompleter/CompletionData.cpp
```diff
@@ -226,11 +241,27 @@ void CompletionData::ExtractDataFromChunk( CXCompletionString completion_string,
...
+    switch (kind) {
+        case CXCompletionChunk_ResultType:
+            return_type_ = ChunkToString( completion_string, chunk_num );
+            break;
+        case CXCompletionChunk_Placeholder:
+            saw_placeholder = true;
+            break;
+        case CXCompletionChunk_TypedText:
+        case CXCompletionChunk_Text:
+        // need to add paren to insert string when implementing inherit methods or declare methods in objc.
```
Valloric: Wrap lines to 80 columns.

Valloric:
>  Notice the ( and ) is also part of insert. this will affect c&c++ original behavior, which will also contain a ( in the insert text. 

Hm, that breaks the current behavior for C & C++. Any chance we could not break user's expectations here?

SolaWing:
for same habits, we need to do some check. for example, we can remove the last char if it is `(`. in situation that  c&c++ functions don't contain a param, it will insert () pair, I think this may not need to fix.

Valloric:
> for same habits, we need to do some check. for example, we can remove the last char if it is (. in situation that c&c++ functions don't contain a param, it will insert () pair, I think this may not need to fix.

Makes sense. If last char is `(`, remove it. Otherwise leave parens alone.

SolaWing:
I already fix format, and add `(` check

cpp/ycm/ClangCompleter/CompletionData.cpp
```diff
@@ -161,12 +169,18 @@ CompletionData::CompletionData( const CXCompletionResult &completion_result ) {
...
 
   for ( uint j = 0; j < num_chunks; ++j ) {
     ExtractDataFromChunk( completion_string,
                           j,
                           saw_left_paren,
-                          saw_function_params );
+                          saw_function_params,
+                          saw_placeholder);
+  }
+  if (original_string_.size()>0 &&
```
Valloric: You need spaces inside IF parens, inside `[]` and spaces around the `>` operator.

Similar for the `erase` call.

cpp/ycm/ClangCompleter/CompletionData.cpp
```diff
@@ -226,11 +243,27 @@ void CompletionData::ExtractDataFromChunk( CXCompletionString completion_string,
     }
   }
 
-  if ( kind == CXCompletionChunk_ResultType )
-    return_type_ = ChunkToString( completion_string, chunk_num );
-
-  if ( kind == CXCompletionChunk_TypedText )
-    original_string_ = ChunkToString( completion_string, chunk_num );
+  switch (kind) {
```
Valloric: Spaces inside paren.

cpp/ycm/ClangCompleter/CompletionData.cpp
```diff
@@ -226,11 +243,27 @@ void CompletionData::ExtractDataFromChunk( CXCompletionString completion_string,
...
+      saw_placeholder = true;
+      break;
+    case CXCompletionChunk_TypedText:
+    case CXCompletionChunk_Text:
+      // need to add paren to insert string 
+      // when implementing inherited methods or declared methods in objc.
+    case CXCompletionChunk_LeftParen:
+    case CXCompletionChunk_RightParen:
+    case CXCompletionChunk_HorizontalSpace:
+      if (!saw_placeholder) {
```
Valloric: Spaces inside paren. Here and line below.

Valloric:
I added a few more formatting comments. Other than that, I need to test this out on my machine a bit to see how it all works out. If everything looks fine, then I'll merge it.

Thanks for working on this! :)

SolaWing:
Ok, I add spaces and rebase the commit.

Valloric:
I tried this out to see how it affects C++ completions and it seems to work the same. The ObjC stuff I don't know enough about to verify it makes sense, but it seems to. :)

Anyway, merged! Thanks again!

PR #106, opened 2015-04-02
Add Gocode support using Completer API

### LINKED ISSUE #52: Add semantic engine for Go.
Would be an awesome feature :)

ekfriis:
Add Completer for semantic Golang completion using gocode. Fixes #52. 

Sorry in advance for my style mistakes, I have not written python in a long time!

A couple questions:
 * should gocode be installed automatically by build.py or can it just be found on the path?
 * I don't *think* I need to override ShouldUseNowInner - or am I missing something?

Thanks for YCM, it's awesome!

### DIFF
--- a/ycmd/completers/go/gocode_completer.py
+++ b/ycmd/completers/go/gocode_completer.py
@@ -19,7 +19,6 @@
 
 import logging
 import json
-import os
 import subprocess
 
 from ycmd import responses
--- a/ycmd/completers/go/gocode_completer.py
+++ b/ycmd/completers/go/gocode_completer.py
@@ -26,9 +26,9 @@ from ycmd import utils
 from ycmd.completers.completer import Completer
 
 GO_FILETYPES = set( [ 'go' ] )
-COMPLETION_ERROR_MESSAGE = "gocode call failed"
-PARSE_ERROR_MESSAGE = "gocode result parsing failed"
-NO_COMPLETIONS_MESSAGE = "gocode failed returned no completions"
+COMPLETION_ERROR_MESSAGE = "Gocode shell call failed."
+PARSE_ERROR_MESSAGE = "Gocode returned invalid JSON response."
+NO_COMPLETIONS_MESSAGE = "Gocode returned empty JSON response."
 
 _logger = logging.getLogger( __name__ )
 
@@ -58,25 +58,26 @@ class GoCodeCompleter( Completer ):
                          stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     stdoutdata, stderrdata = proc.communicate( contents )
     if proc.returncode:
-      _logger.error( "gocode failed with code %i stderr: %s",
+      _logger.error( COMPLETION_ERROR_MESSAGE + " code %i stderr: %s",
                     proc.returncode, stderrdata)
       raise RuntimeError( COMPLETION_ERROR_MESSAGE )
 
     try:
       resultdata = json.loads( stdoutdata )
     except ValueError:
-      _logger.error( "gocode failed to parse results json" )
+      _logger.error( PARSE_ERROR_MESSAGE )
       raise RuntimeError( PARSE_ERROR_MESSAGE )
-    if not resultdata:
-      _logger.error( "gocode got an empty response" )
+    if len(resultdata) != 2:
+      _logger.error( NO_COMPLETIONS_MESSAGE )
       raise RuntimeError( NO_COMPLETIONS_MESSAGE )
 
     return [ _ConvertCompletionData( x ) for x in resultdata[1] ]
 
 
 # Compute the byte offset in the file given the line and column.
+# TODO(ekfriis): If this is slow, consider moving this to C++ ycm_core,
+# perhaps in RequestWrap.
 def _ComputeOffset( contents, line, col ):
-  _logger.info("line %s col %s" % (line, col))
   curline = 1
   curcol = 1
   for i, byte in enumerate( contents ):
--- a/ycmd/completers/go/gocode_completer.py
+++ b/ycmd/completers/go/gocode_completer.py
@@ -54,30 +54,23 @@ class GoCodeCompleter( Completer ):
     offset = _ComputeOffset( contents, request_data[ 'line_num' ],
                             request_data[ 'column_num' ] )
 
-    filename_format = os.path.join( utils.PathToTempDir(), u'gocode_{std}.log' )
-    stdout_file = filename_format.format( std='stdout' )
-    stderr_file = filename_format.format( std='stderr' )
+    cmd = ['gocode', '-f=json', 'autocomplete', filename, str( offset )]
+    proc = self._popener( cmd, stdin=subprocess.PIPE,
+                         stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    stdoutdata, stderrdata = proc.communicate( contents )
+    if proc.returncode:
+      _logger.error( "gocode failed with code %i stderr: %s",
+                    proc.returncode, stderrdata)
+      raise RuntimeError( COMPLETION_ERROR_MESSAGE )
 
-    with open( stdout_file, 'w' ) as stdout:
-      with open( stderr_file, 'w' ) as stderr:
-        cmd = ['gocode', '-f=json', 'autocomplete', filename, str( offset )]
-        proc = self._popener( cmd, stdout=stdout, stderr=stderr, stdin=subprocess.PIPE )
-        proc.stdin.write( contents )
-        proc.stdin.close( )
-        retcode = proc.wait()
-        if retcode:
-          _logger.error( "gocode failed with code %i" % retcode )
-          raise RuntimeError( COMPLETION_ERROR_MESSAGE )
-
-    with open( stdout_file, 'r' ) as results_file:
-      try:
-        resultdata = json.load( results_file )
-      except ValueError:
-        _logger.error( "gocode failed to parse results json" )
-        raise RuntimeError( PARSE_ERROR_MESSAGE )
-      if not resultdata:
-        _logger.error( "gocode got an empty response" )
-        raise RuntimeError( NO_COMPLETIONS_MESSAGE )
+    try:
+      resultdata = json.loads( stdoutdata )
+    except ValueError:
+      _logger.error( "gocode failed to parse results json" )
+      raise RuntimeError( PARSE_ERROR_MESSAGE )
+    if not resultdata:
+      _logger.error( "gocode got an empty response" )
+      raise RuntimeError( NO_COMPLETIONS_MESSAGE )
 
     return [ _ConvertCompletionData( x ) for x in resultdata[1] ]
 
--- a/ycmd/completers/go/gocode_completer.py
+++ b/ycmd/completers/go/gocode_completer.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python
+#
+# Copyright (C) 2015 Google Inc.
+#
+# This file is part of YouCompleteMe.
+#
+# YouCompleteMe is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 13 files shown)


vheon:
@ekfriis As per [CONTRIBUTING.md](https://github.com/Valloric/YouCompleteMe/blob/master/CONTRIBUTING.md) you have to sign the [CLA](https://developers.google.com/open-source/cla/individual) and state here that you did.

> should gocode be installed automatically by build.py or can it just be found on the path?

I think you have to follow the convention that build.py should install gocode. If you think about it, you don't install `jedi` or `Omnisharp` by hand, you just install ycmd.

oblitum:
Just a tip, vim-go plugin does a great job integrating Go tools for vim, it's quite known, I use it myself. It offers options for installing binaries, like gocode. I believe it should be good to prepare to well-behave with it.

vheon:
@ekfriis is this using the gocode CLI? Isn't it better to use just the server part? I mean this way we spawn a new process for the client side of gocode at every request. It is not ideal and for the same reason other completers were not merged in. I don't know (I didn't check) if the API for the server is public and stable or if it changes whenever it like. Thoughts?

Valloric:
Agreed with @vheon that `build.py` should install gocode; @ekfriis you'll probably have to add it as a submodule in `third_party.py`. Note that the installation of gocode should happen only if a user passes a `--gocode-completer` flag (much like `--omnisharp-completer`, `--clang-completer` etc).

BTW ekfriis doesn't have to sign the CLA because he's a Google employee.

I haven't looked at the actual code yet; I'll try to do that as soon as I find the time.

ekfriis:
Thanks all.

 * I'll work on adding the installation it to build.py.
 * @oblitum Agree on vim-go, I'll figure out what needs to happen to make that seamless.
 * @vheon Yes, this uses the CLI, which according to [1] seems to be the current official way to integrate with an IDE. I agree directly calling the server would be preferable, but the server uses some auto-generated(?) interface backed by Golang net/rpc, which I don't think is straightforward to call from python. I would propose the way forward is to use the CLI for now, and send a future PR to gocode to add an HTTP json interface to the server.

[1] https://github.com/nsf/gocode/blob/master/docs/IDE_integration.md

Valloric:
@ekfriis Honestly I wouldn't worry too much about vim-go. YCM has always had the position that you shouldn't have other completion plugins installed if you're using YCM. For a select few other plugins, it supports their omnifunc if it is well-behaved.

Removing the need to rely on other Vim plugins for go support is the very reason why we want native gocode support in ycmd. Not to mention that emacs doesn't have vim-go.

>  I would propose the way forward is to use the CLI for now, and send a future PR to gocode to add an HTTP json interface to the server.

Yeah we're not big fans of going through the CLI; process startup/shutdown plus the serialization/deserialization round trip through the CLI is a nasty latency cost. But I'm fine with it if it's temporary. People already use vim-go which also goes through the CLI, so we'll be no worse.

ycmd/completers/go/gocode_completer.py
```diff
@@ -0,0 +1,98 @@
...
+import subprocess
+
+from ycmd import responses
+from ycmd import utils
+from ycmd.completers.completer import Completer
+
+GO_FILETYPES = set( [ 'go' ] )
+COMPLETION_ERROR_MESSAGE = "gocode call failed"
+PARSE_ERROR_MESSAGE = "gocode result parsing failed"
+NO_COMPLETIONS_MESSAGE = "gocode failed returned no completions"
```
Valloric: "failed returned" doesn't parse for me. :)

Valloric: Also, please use full sentences for error messages. Capital first letter, punctuation etc.

ycmd/completers/go/gocode_completer.py
```diff
@@ -0,0 +1,98 @@
...
+
+    return [ _ConvertCompletionData( x ) for x in resultdata[1] ]
+
+
+# Compute the byte offset in the file given the line and column.
+def _ComputeOffset( contents, line, col ):
+  _logger.info("line %s col %s" % (line, col))
+  curline = 1
+  curcol = 1
+  for i, byte in enumerate( contents ):
```
Valloric: This could be very slow on large files.

vheon:
> BTW ekfriis doesn't have to sign the CLA because he's a Google employee.

Yeah, I saw the github profile after writing the comment.

@Valloric the `contents` field in the json request that we pass to the various completer is always utf encoded? Or depends on the client?

ycmd/completers/go/tests/gocode_completer_test.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python
```
Valloric: Thank you for writing tests!

I'd like to see one small test in `get_completions_test.py` for Go completion, just so that we have something that exercises the full path from HTTP request to HTTP response. It ensures we don't break the plumbing that connects the go semantic completer to the rest of ycmd.

Valloric:
> @Valloric the contents field in the json request that we pass to the various completer is always utf encoded? Or depends on the client?

All strings in ycmd should be either Python unicode objects or utf8 `string`s (usually the latter). Python being Python, sometimes it's hard to tell when it's one vs the other...

ycmd/completers/go/gocode_completer.py
```diff
@@ -0,0 +1,98 @@
...
+    return GO_FILETYPES
+
+
+  def ComputeCandidatesInner( self, request_data ):
+    filename = request_data[ 'filepath' ]
+    _logger.info( "gocode completion request %s" % filename )
+    if not filename:
+      return
+
+    contents = utils.ToUtf8IfNeeded(
```
vheon: @Valloric I was asking for the encoding to check if this call was really necessary.

Valloric: If it isn't, `ToUtf8IfNeeded` will merely return the provided value without changing it. It doesn't hurt to call this. IMO this should stay.

oblitum:
> @ekfriis Honestly I wouldn't worry too much about vim-go. YCM has always had the position that you shouldn't have other completion plugins installed if you're using YCM. For a select few other plugins, it supports their omnifunc if it is well-behaved.

> Removing the need to rely on other Vim plugins for go support is the very reason why we want native gocode support in ycmd. Not to mention that emacs doesn't have vim-go, and we want to support it through the emacs-ycmd client.

vim-go does much more than completion and go-to-def, so YCM/ycmd by itself can't remove its need. I suspect this may go without a hassle but it won't cost much to be at last aware of it. vim-go claims to play well with YCM itself, so they have done their job in this sense. Since it's an integrator, the maintainer may be willing to support ycmd as an option over gocode.

ekfriis:
vim-go defines a gocode omnifunc in ftplugin/go.vim [1]. Would a complete solution to put a guard around this that checks if Go completion is supported by YCM and doesn't set if so? Or am I missing something?

[1] https://github.com/fatih/vim-go/blob/master/ftplugin/go.vim#L24

oblitum:
Pinging @fatih.

Valloric:
@oblitum I support the idea of not obstructing vim-go if at all possible, as long as everyone involved understands that we can't rely on it existing to provide Go semantic completion/GoTo etc. We want ycmd to support non-Vim editors.

@ekfriis YCM will handle this itself. It won't fall back to using the Vim omnifunc for any filetype for which it has native completion support. So vim-go can declare the omnifunc for Go just fine, but after your PR is merged, YCM won't use it.

ycmd/completers/go/gocode_completer.py
ekfriis: Added a TODO. I think a nice place for this would be a lazy function in RequestWrap.

ycmd/completers/go/gocode_completer.py
ekfriis: Done.

ycmd/completers/go/tests/gocode_completer_test.py
ekfriis: Good idea, will do. Do you have a handy example in one of the other tests for doing the E2E?

ekfriis:
I added features to auto-install gocode + a user option to override the binary location (I will need this), PTAL.

ekfriis:
PS - the installation of gocode is behind an opt-in flag in build.py, but since it is so trivial to build (yay Go) it might be nice to make the flag opt-out and just detect if the user has the "go" binary installed and build it if so.

ycmd/completers/go/tests/gocode_completer_test.py
vheon: If you look inside `get_completions_tests.py` you'll find E2E tests for every completion engine :+1:  https://github.com/Valloric/ycmd/blob/master/ycmd/tests/get_completions_test.py#L669 this is the basic one for python.

Valloric:
This looks fine. You only need that one extra E2E test.

> it might be nice to make the flag opt-out and just detect if the user has the "go" binary installed and build it if so.

I understand that it's trivial, but it could still fail in various ways and the fewer moving parts we have in the installation process by default, the better off we are.

Valloric:
Oh, you also need to get Travis to green. It's currently a bit flaky so if 3/4 runs in the test matrix work fine but one is timing out, that's good enough. 

I need to spend some time trying to make it not flaky again.

Valloric:
I pushed a commit that should unflake the tests, so you now need a full green run from Travis. :)

ekfriis:
OK, I think this should ready for a final inspection. My Travis build still had some timeout flakes from the omnisharp completer - is there away I can trigger a rebuild w/o a new commit?

https://travis-ci.org/Valloric/ycmd/builds/58107672

vheon:
@ekfriis unfortunately no. Only one with write access can re-trigger a build.

Valloric:
Ugh, those omnisharp flakes are getting seriously annoying. Sorry you had to go through that!

Let me take another look through the code.

ycmd/tests/get_completions_test.py
```diff
@@ -698,3 +698,22 @@ def GetCompletions_JediCompleter_UnicodeDescription_test():
...
+  completion_data = BuildRequest( filepath = filepath,
+                                  filetype = 'go',
+                                  contents = open( filepath ).read(),
+                                  force_semantic = True,
+                                  line_num = 9,
+                                  column_num = 11)
+
+  results = app.post_json( '/completions',
+                           completion_data ).json[ 'completions' ]
+  print results
```
Valloric: You probably left this in by accident. :)

ekfriis: Oops, thanks. Fixed.

Valloric:
Merged, thanks for the PR!

We need to update YCM docs to mention this though...

ekfriis:
Thanks all! Doc PRs incoming.

ycmd/completers/go/gocode_completer.py
```diff
@@ -0,0 +1,135 @@
...
+      curcol = 1
+  _logger.error( "GoCode completer - could not compute byte offset " +
+                "corresponding to L%i C%i", line, col )
+  return -1
+
+
+def _ConvertCompletionData( completion_data ):
+  return responses.BuildCompletionData(
+    insertion_text = completion_data[ 'name' ],
+    menu_text = completion_data[ 'name' ],
```
ptrv: Is there a reason to add the `name` information twice? There is also the `class` information int the `gocode` completion_data, which is missing here. Sent a PR for fixing the latter #112

ekfriis: As I understood it (possibly incorrectly), insertion_text is what goes into the buffer, and menu_text is what shows up on the pop-up.

ptrv: I see. Makes sense.

oblitum:
Just for reference, this is the end result from vim-go side: https://github.com/fatih/vim-go/commit/e757872d35393544d93ca7640752d53cb714cf35.

Valloric:
That's unfortunate, but we'll live.

PR #107, opened 2015-04-10
Allow to specify max number of identifier canditates in options

ptrv:
I have signed the Google CLA

### DIFF
--- a/ycmd/__main__.py
+++ b/ycmd/__main__.py
@@ -111,9 +111,10 @@ def SetupLogging( log_level ):
                        level = numeric_level )
 
 def SetupOptions( options_file ):
-  options = ( json.load( open( options_file, 'r' ) )
-              if options_file
-              else user_options_store.DefaultOptions() )
+  options = user_options_store.DefaultOptions()
+  if options_file:
+    user_options = json.load( open( options_file, 'r' ) )
+    options.update( user_options )
   utils.RemoveIfExists( options_file )
   options[ 'hmac_secret' ] = base64.b64decode( options[ 'hmac_secret' ] )
   user_options_store.SetAll( options )
--- a/ycmd/completers/all/identifier_completer.py
+++ b/ycmd/completers/all/identifier_completer.py
@@ -27,7 +27,6 @@ from ycmd import utils
 from ycmd.utils import ToUtf8IfNeeded
 from ycmd import responses
 
-MAX_IDENTIFIER_COMPLETIONS_RETURNED = 10
 SYNTAX_FILENAME = 'YCM_PLACEHOLDER_FOR_SYNTAX'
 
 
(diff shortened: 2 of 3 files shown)


ycmd/completers/all/identifier_completer.py
```diff
@@ -37,6 +37,10 @@ def __init__( self, user_options ):
     self._completer = ycm_core.IdentifierCompleter()
     self._tags_file_last_mtime = defaultdict( int )
     self._logger = logging.getLogger( __name__ )
+    try:
+      self._max_candidates = user_options[ 'max_num_identifier_candidates' ]
+    except KeyError:
+      self._max_candidates = MAX_IDENTIFIER_COMPLETIONS_RETURNED
```
ptrv: Not the nicest solution, but how should we else handle a config mismatch?

vheon: What do you mean by "config mismatch"? If I don't specify the option then `user_options[ 'max_num_identifier_candidates' ]` will be equal to the default one specified in `default_settings.json`. If I specify the option then `user_options[ 'max_num_identifier_candidates' ]` will be equal to what I specify. Where is the problem I didn't consider?

ptrv: I meant that maybe there will be some custom configs out there which don't have the new entry. I added catching the `KeyError` for that reason.

ptrv: I was not aware that if I use a custom config which does not have an entry, the default value from `default_settings.json` will be used.

vheon: If I'm not mistaken it works like that. Test it and see how it goes.

ptrv: Here is the logic for loading option files: https://github.com/Valloric/ycmd/blob/8e228643ffcab2aa274777ca8b9b0c8183f0ef24/ycmd/__main__.py#L114-L116

Either the option file from commandline argument gets loaded and if not specified, then `default_settings.json`
There is no fallback. You get a `KeyError` if the option is missing.

vheon: Oh sorry, I didn't understand what you were saying. Anyway the same argument could be done for every other option. Let's say that you provide an option file without the `min_num_of_chars_for_completion` entry, [ycmd/completers/completer.py#L104](https://github.com/Valloric/ycmd/blob/44eb4008ad17b48c7f1a1d08a52c6dc2b21f59f1/ycmd/completers/completer.py#L104) we pass the `user_options` that is basically what was inside in the option file and no check was done. Sorry if I still didn't understand and this was all noise.

Valloric: @ptrv Could you add a bit of code to merge the provided options with the default options? Basically have the provided options override the default ones, but still have a value for every field.

ycmd/__main__.py
```diff
@@ -111,9 +111,10 @@ def SetupLogging( log_level ):
                        level = numeric_level )
 
 def SetupOptions( options_file ):
-  options = ( json.load( open( options_file, 'r' ) )
-              if options_file
-              else user_options_store.DefaultOptions() )
+  options = user_options_store.DefaultOptions()
+  if options_file:
+    user_options = json.load( open( options_file, 'r' ) )
+    options.update( user_options )
```
ptrv: @Valloric I added code for loading the default values first and then update the dictionary with user_options if given as server commandline argument.

Valloric:
Thanks for the PR!

PR #110, opened 2015-04-14
now import can complete Frameworks

SolaWing:
hi, @Valloric, I add completion of `Frameworks` in iOS&Mac.

In Mac, Frameworks have the following structure:

`UIKit.framework/Headers/*.h`

And in Xcode, the user can import a framework file as `import <UIKit/UIKit.h>`, the user don't need to type .framework/Headers.   So I add this complete feature.

In detail, I add `-F`, `-iframework` flags in `UserIncludePaths`, so includePaths will contains framework's path.  then, in `GetPathsIncludeCase`, if the path which join `include_path` and user's inputting path is not a dir, I try append `.framework/Headers` and check if it's a dir. if so, this inputting path match a framework and return the files inside `Headers`. and upon return, if the complete basename ends with `.framework`, I remove it, and recognize it as a \[Lib](framework is long, so I use Lib instead)

the side effect is of standard file name completing which also use `_GenerateCandidatesForPaths`, so if file end with `.framework`, it also will be recognized as a framework and remove the extension. But I think this case is rarely happen.

(I already sign **Google Contributor License Agreement**)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -97,13 +97,15 @@ class Flags( object ):
       return sanitized_flags
 
 
-  def UserIncludePaths( self, filename ):
+  def UserIncludePaths( self, filename, has_framework = False ):
     flags = self.FlagsForFile( filename, False )
     if not flags:
       return []
 
     include_paths = []
     path_flags = [ '-isystem', '-I', '-iquote' ]
+    if has_framework: 
+        path_flags.extend( [ '-F' ,'-iframework' ] )
 
     next_flag_is_include_path = False
     for flag in flags:
--- a/ycmd/completers/general/filename_completer.py
+++ b/ycmd/completers/general/filename_completer.py
@@ -26,7 +26,8 @@ from ycmd.completers.cpp.flags import Flags
 from ycmd.utils import ToUtf8IfNeeded, ToUnicodeIfNeeded
 from ycmd import responses
 
-EXTRA_INFO_MAP = { 1 : '[File]', 2 : '[Dir]', 3 : '[File&Dir]' }
+EXTRA_INFO_MAP = { 1 : '[File]', 2 : '[Dir]', 3 : '[File&Dir]'
+        , 4 : '[Lib]', 5 : '[Lib&File]', 6 : '[Lib&Dir]', 7 : '[FDL]' }
 
 class FilenameCompleter( Completer ):
   """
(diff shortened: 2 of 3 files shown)


vheon:
As per [CONTRIBUTING.md](https://github.com/Valloric/YouCompleteMe/blob/master/CONTRIBUTING.md) you have to sign the [CLA](https://developers.google.com/open-source/cla/individual) and state here that you did.

Valloric:
I'm not sure this is the best approach. Doesn't it make more sense for the user to provide the `-F` and `-iframework` flags in their extra conf file?

I'm also not a fan of special-casing frameworks in ycmd. Seems like the completion can work fine if the user simply provided a path to `foo.framework/Headers`.

Valloric:
I'm not sure this is the best approach. Doesn't it make more sense for the user to provide the `-F` and `-iframework` flags in their extra conf file?

I'm also not a fan of special-casing frameworks in ycmd. Seems like the completion can work fine if the user simply provided a path to `foo.framework/Headers`.

SolaWing:
The user also need to provide `-F` and `-iframework`. the original `UserIncludePaths` function in flags.py only returns flags begin with `-isystem`, `-I`, `-iquote`. so even the user provides `-F` flags, file completer doesn't  return the paths. so I make `UserIncludePaths` also recognize framework paths. 

Secondly, according to my observation, All Frameworks is merely a structured folder, which contains a binary file, a `Header` folder, and some other folders and files such as `Resources`. So even `UserIncludePaths` returns `-F` and `-iframework` paths, it will return as, for example, `UIKit.framework`, And certainly, filecompleter will not recognize `UIKit/` as a framework folder and return the including header files, since the exist folder is `UIKit.framwork`, not the `UIKit`. and there even has a `Header` folder in it. so these are what I did.

Thirdly, I tried to call the clang completer for completing the import file. but it returns nothing. and originally, the include file is completed by filecompleter.py, so I simply extend it and make it support import framework.

After improving `UserIncludePaths` and making it recognize `-F` and `-iframework`, return the include paths, the completion did work fine if the user simply provided a path to `UIKit.framework/Headers`. But unfortunately, after I tried, the compiler didn't recognized this kind framework import pattern. It would complain with: `fatal error: 'UIKit.framework/Headers/UIKit.h' file not found`. It only accept 'UIKit/UIKit.h'.

Valloric:
This PR seems abandoned; closing.

PR #112, opened 2015-04-16
Add additionial class information to gocode completion data

ptrv:
`gocode` also returns `class` information for candidate. Add this to `detailed_info`

I have read and signed the Google CLA.


Valloric:
@ptrv I see that Travis doesn't like this commit, and it's not the usual flakiness that's the issue.

In related news, the C# completer test flakiness has become a serious problem; I'm putting more work into unflaking those tests, but tracking down the root cause is tricky.

ptrv:
@Valloric The test for the function is fixed now

Valloric:
Thanks for the PR!

PR #116, opened 2015-04-24
Continuous C# completions

mispencer:
While researching [NRefactory](https://github.com/icsharpcode/NRefactory), which our C# completer, [Omnisharp](https://github.com/OmniSharp/omnisharp-server), is a wrapper of, I noticed that it provided a "controlspace" parameter to a completion generator method. Upon investigation, it appears that [it should called with true for explicit completion (by, for example, pressing control-space), and with false for non-explicit completion (the completion-as-you-type that Ycmd is used for)](https://github.com/icsharpcode/SharpDevelop/blob/newNR/src/AddIns/BackendBindings/CSharpBinding/Project/Src/Completion/CSharpCompletionBinding.cs). This pull request implements this behavior.

When you are typing, the C# semantic completer is called in non-controlspace mode. Additionally, it is called everywhere, and is no longer restricted to locations matching the triggers. When it returns no completions, ycmd will fallback to the identifier completer. When you explicitly ask for completion (for example, by the completion key-bind in YouCompleteMe), the C# semantic completer is called in controlspace mode.

These two modes are not interchangeable, so this required altering the cache logic as well. We need to only return the cache if it matches the correct mode. I implemented this in a way which should not affect the other completers.

In my performance testing, the non-controlspace completions returns in sub-1/10 second in almost all cases. The only syntax which I noticed slowness is after a new keyword. The identifier completer results were rarely relevant in this case, and the semantic completions returns the constructors, so I feel this is acceptable.

Note also that this change required pulling forward our OmniSharpServer submodule. I have not noticed any regressions in my testing.

I have signed the CLA.

### DIFF
--- a/ycmd/completers/completer.py
+++ b/ycmd/completers/completer.py
@@ -110,6 +110,10 @@ class Completer( object ):
     self._completions_cache = CompletionsCache()
 
 
+  def CompletionType( self, request_data ):
+    return 0
+
+
   # It's highly likely you DON'T want to override this function but the *Inner
   # version of it.
   def ShouldUseNow( self, request_data ):
@@ -122,7 +126,8 @@ class Completer( object ):
     # data.
     cache_completions = self._completions_cache.GetCompletionsIfCacheValid(
         request_data[ 'line_num' ],
-        request_data[ 'start_column' ] )
+        request_data[ 'start_column' ],
+        self.CompletionType( request_data ) )
 
     # If None, then the cache isn't valid and we know we should return true
     if cache_completions is None:
@@ -165,7 +170,8 @@ class Completer( object ):
   def _GetCandidatesFromSubclass( self, request_data ):
     cache_completions = self._completions_cache.GetCompletionsIfCacheValid(
           request_data[ 'line_num' ],
-          request_data[ 'start_column' ] )
+          request_data[ 'start_column' ],
+          self.CompletionType( request_data ) )
 
     if cache_completions:
       return cache_completions
@@ -174,6 +180,7 @@ class Completer( object ):
       self._completions_cache.Update(
           request_data[ 'line_num' ],
           request_data[ 'start_column' ],
+          self.CompletionType( request_data ),
           raw_completions )
       return raw_completions
 
@@ -283,13 +290,15 @@ class CompletionsCache( object ):
     with self._access_lock:
       self._line = -1
       self._column = -1
+      self._completion_type = -1
       self._completions = []
 
 
-  def Update( self, line, column, completions ):
+  def Update( self, line, column, completion_type, completions ):
     with self._access_lock:
       self._line = line
       self._column = column
+      self._completion_type = completion_type
       self._completions = completions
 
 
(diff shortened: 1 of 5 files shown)


Valloric:
I'm not sure how I feel about this PR. The model we have with identifier-based completion when typing "normally" works surprisingly well. We also present a very consistent user experience across languages; this PR would change that so we end up doing something entirely different for C# for possibly not a lot of benefit. 

This whole "doing something entirely different for C#" is my main concern (after performance, but you say that isn't an issue). I'd prefer presenting a simple mental model to the user that's consistent across languages.

Thoughts?

mispencer:
In most cases, the semantic completer is going to be providing a list of identifiers, same as the identifier-based completer would. It's just that instead of the list being based on the buffers you have opened, it would be based on what identifiers the semantic completer knows are legal wherever you are completing. Without knowing the details of how ycmd works behind the scene, I'm not sure you could tell the difference.

The only exceptions are completing methods and constructors. Methods already are semantic completion, of course, and constructor completions are similar enough for semantic completions to be expected.

Having said that, if we want to continue with semantic completions on triggers, adding new trigger for new would get us most of the way there. I would especially like to see the completion for constructors.

In any case, I would love it if anyone who uses the ymcd C# completions regularly could check this out, and let us know what they think. I have been using this for awhile now, and I feel it's an improvement, but this is a subjective measure. Perhaps it is less of improvement, or even a setback on other code bases or programming styles.

Valloric:
I could see this being merged in, but we need to have this tested out a bit by those who use YCM with C# (I'm not one such user). 

@metatheos @akhosravian Could you take a look at this?

vheon:
@mispencer so did you tried with the regex trigger for the `new` keyword?

mispencer:
I have tried using the regex for ```new```, but I prefer this pull request's solution. Before I implemented the change in this pull request, I was using a ```.``` regex trigger.

akhosravian:
I'll try this out this week.

akhosravian:
I've noticed some very long waits (sometimes minutes) when running :YcmCompleter GoTo since pulling this change - I haven't been able to identify a particular pattern to when this happens. Here's the omnisharp stdout log from one such run: https://gist.github.com/akhosravian/199c31fceeea8039f4e4 The stderr file was empty, and I didn't grab the ycm server log file. 

What info would be useful here?

Valloric:
It could be related to the OmniSharp version bump.

Valloric:

mispencer:
I just rebased it.

The goto performance is probably due to the Omnisharp version increase. Possibly related to https://github.com/OmniSharp/omnisharp-server/pull/153?

akhosravian:
I've been running with this for a bit. The issue I previously mentioned appears to be related to using vimgrep on a relatively large project (i.e. several thousand files). I'll check whether the problem is actually related to this change or not later this week.

akhosravian:
Sorry for the delay - without this change it looks like there's a different set of bad behavior from OmniSharp when using vimgrep over a large number of files. In other words: this change seems fine to me.

Valloric:
OK, let's merge this in then. I'm wary of unintended breakage; not specifically the code, but the user's mental model.

But seeing reports of this being a solid change by those who have tested it out makes me more confident.

@homu r+

Valloric:
Bah, Travis flaked.

@homu retry

PR #117, opened 2015-04-24
Add python3 completer

ozon671games1:
I didn't add hook.py. I think you know better how to combine python2 and python3 completers.

jedi_server.py responses depend on version of interpreter that runs it. It easily can be made python2 and python3 compatible so this completer can replace already existing one.

### DIFF
--- a/ycmd/completers/python3/jedi_server.py
+++ b/ycmd/completers/python3/jedi_server.py
@@ -1,0 +1,110 @@
+import os
+import sys
+import json
+from urllib.parse import parse_qs
+from http.server import HTTPServer, BaseHTTPRequestHandler
+
+ABSPATH = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(ABSPATH + '/../../../third_party/jedi')
+import jedi
+
+
+class JediServerHandler(BaseHTTPRequestHandler):
+
+    supported_requests = ['gotodefinition',
+                          'gotodelcaration',
+                          'getcompletions']
+
+    def do_POST(self):
+        # cut '/'
+        cmd = self.path[1:]
+        if cmd in self.supported_requests:
+            length = int(self.headers['Content-Length'])
+            request_data = parse_qs(self.rfile.read(length).decode('utf-8'))
+            getattr(self, cmd)(request_data)
+        else:
+            self.send_response(404)
+            self.end_headers()
+
+    def gotodefinition(self, request_data):
+        script = self._get_script(request_data)
+        definitions = script.goto_definitions()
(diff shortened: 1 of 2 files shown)


vheon:
I was working on this on my spare time (which at the moment is very limited) but I have no prior experience with python so maybe It's better what you did here. My wrapper of Jedi is [JediHTTP](/vheon/JediHTTP) and it use bottle for the HTTP API and I was using weitress as the webserver (I kinda copy what ycmd itself does, not knowing any better) Is there any particular reason why you used bare `http.server`?

Anyway the test didn't pass because you imported some library but didn't use them, like `subprocess` and `time`.

Valloric:
Thanks for sending the PR!

Here's what I'd like to see: first, @vheon has the right idea that a Jedi completion server should be a separate repository. It should work transparently with python 2 and 3.

Our current jedi completer should then talk to two instances of it, one for python2 and one for python3. If the user is only editing python2, then we only have one instance running.

So start with a separate repo for the server and when you have that fully working, send in a PR that pulls it in and uses it from ycmd.

Thanks again for all of your work!

vheon:
> Our current jedi completer should then talk to two instances of it, one for python2 and one for python3. If the user is only editing python2, then we only have one instance running.

How should this part be handled? Is the client that send that info to ycmd?

Valloric:
> How should this part be handled? Is the client that send that info to ycmd?

Probably with the buffer filetype, yes. AFAIK to get python2 support with Jedi, one needs to run it in a python2 interpreter. To get python3 support, a python3 interpreter is needed. So you can't complete both from the same server.

This means that the jedi-server code itself will need to work in both python2 and python3.

vheon:
> Probably with the buffer filetype, yes.

So one can assume that in the request is specified the version of python?

> This means that the jedi-server code itself will need to work in both python2 and python3.

Yes, I know that, and in my wrapper I'm already testing that.

Valloric:
> So one can assume that in the request is specified the version of python?

For now, let's assume so. It's not something we need to decide right away. Either the editor client will specify the Python version, or ycmd will need to figure it out before sending the request to the jedi-server. But in either case, the jedi-server doesn't need to figure it out on its own.

ozon671games1:
> Is there any particular reason why you used bare http.server?

Why not? It's binded to local loopback and can't be accessed remotely. So it isn't insecure.

vheon:
@ozon671games1 
> Why not?

As I said before:

> but I have no prior experience with python so maybe It's better what you did here

So it was just a python noob question, good to know though :+1:

PR #118, opened 2015-04-25
Blacklist -Xclang flags for C/C++ completion

ehsan:
-Xclang flags are typically not used in the front-end, so they shouldn't affect anything that we care about for code completion.

Some background on why I made this change: I was trying to get YCM to work on the Mozilla code base.  We use a clang plugin to perform some static analysis checks when building our code.  When ycmd tries to use libclang, libclang fails to load the plugin so semantic code completion doesn't work.

When I was thinking about how to fix this, we realized that we should just ignore -Xclang options, since they are intended for cc1, which means they usually affect code generation but not syntax parsing and AST generation, which is all we need from libclang for code completion.

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -137,9 +137,30 @@ def _CallExtraConfFlagsForFile( module, filename, client_data ):
 
 
 def PrepareFlagsForClang( flags, filename ):
+  flags = _RemoveXclangFlags( flags )
   flags = _RemoveUnusedFlags( flags, filename )
   flags = _SanitizeFlags( flags )
   return flags
+
+
+def _RemoveXclangFlags( flags ):
+  """Drops -Xclang flags.  These are typically used to pass in options to
+  clang cc1 which are not used in the front-end, so they are not needed for
+  code completion."""
+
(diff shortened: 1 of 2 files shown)


vheon:
As per [CONTRIBUTING.md](https://github.com/Valloric/YouCompleteMe/blob/master/CONTRIBUTING.md) you have to sign the [CLA](https://developers.google.com/open-source/cla/individual) and state here that you did.

ehsan:
> As per CONTRIBUTING.md you have to sign the CLA and state here that you did.

I have already signed the Google CLA as part of my contribution to [Chromium] (https://code.google.com/p/chromium/codesearch#chromium/src/AUTHORS&q=Ehsan%20Akhgari&sq=package:chromium&type=cs&l=150) among other Google open source projects.

Valloric:
This looks fine. Please add a test for this though; there are already tests for the other flag filtering stuff so it shouldn't be hard to copy/figure out.

ehsan:
The travis-ci failure above doesn't seem to be related to this test...

Valloric:
_Sigh_, Travis CI is sad yet again.

Thanks for the PR!

PR #119, opened 2015-04-26
add python3 support

ozon671games1:
I tried to determine python version by 'g:path_to_python_interpreter' variable but when it's set to python3 path ycmd crashes... When 'g:path_to_python_interpreter' is unset /usr/bin/python used. Seems like it works for python 2 and python 3 but i don't really want to fuck with tests. Actually when I was developing this completer I realize that i don't even want to use vim as ide because it's too slow.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,3 +25,6 @@
 [submodule "third_party/gocode"]
 	path = third_party/gocode
 	url = https://github.com/nsf/gocode.git
+[submodule "third_party/jedihttpserver"]
+	path = third_party/jedihttpserver
+	url = https://github.com/ozon671games1/jedihttpserver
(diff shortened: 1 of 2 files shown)


vheon:
It doesn't work if you set the `path_to_python_interpreter` because that variable is used to specify which interpreter to use for ycmd itself which is not written for python3. What @Valloric was saying in the other PR is that the completer should start a server for jedi under pyton2 and one for python3 as soon as they are required so it can handle both at the same time. 

>  but i don't really want to fuck with tests

I don't know if this will ever be merged without at least some E2E tests. Other completer were put on hold until the E2E tests were pushed and travis was green (which btw at the moment travis failed).

@Valloric can correct me if I said anything wrong here.

Valloric:
Thanks for the PR!

> but i don't really want to fuck with tests. 

Tests are absolutely required. Closing PR until end-to-end tests are added.

Issue #96, opened 2015-02-23
Add CodeIntel as a completer for Go, PHP, Perl, Ruby, and Javascript


cweagans:
CodeIntel is the code completion engine from Komodo Edit, and it supports a ton of languages and does a pretty good job at most of them. I'd love to see it integrated with ycmd, and I may end up paying somebody to write the completers for it or I may spend the time to learn how to make it happen myself.

In [my fork](https://github.com/cweagans/ycmd), I've added CodeIntel as a third party library. I'm not sure how @Valloric will want to go about using it as a completer for so many different languages, so I'm opening this early with the hopes of getting architectural feedback on it.

vheon:
I'm not thrilled about the idea of using one big tool for autocompleting every language some of which are already integrated into ycmd, like python. But that is just me.

cweagans:
Clang is already kind of used for that? Also, we don't have to use CodeIntel for Python (and we shouldn't, IMO. Jedi is better). Using it to add as many languages as possible seems to make sense, though.

Go could also be omitted and provided by Gocode instead.

vheon:
Yeah, but you see where I'm going? You are saying: "we should use CodeIntel so we could bring more languages... but not for python because Jedi is better... and not for go because geocode is more appropriate..." and we could go on, like "and not for javascript because tern.js does already a great job at that..." and there you go that from the list of the good completion of CodeIntel we remain with 2-3 languages. Now I won't say that having the completion system for those 2-3 languages would be bad, but if we consider the effort that the developer of CodeIntel will put into the project will be shared by all those language and therefore would be less than if we had a completion system for each language. Again this is just my opinion, but considering that @Valloric already said that he would not prefer having the user configure which autocompletion system use for a specific language but deciding upfront which will be the better option for the user (so a sane default) I would go with caution before committing the project to a completer, thats all.

cweagans:
Yeah, I see what you're saying. However, if we're going to add CodeIntel for 2-3 languages that aren't provided by anything else, we might as well also add it for languages that *are* provided by other things that we don't have yet. If Gocode is a better option in the future, then we simply switch. There's also no reason we can't have both CodeIntel and gocode for Go completion. We can query both completers and merge the results.

[...]

PR #120, opened 2015-04-30
Add CodeIntel and a PHP completer

cweagans:
cc @jeteon
cc @Kronuz (thought you might like to know about this)

I mentioned a while ago that I had hired someone to work on a PHP completer for this project. This is the WIP version of the completer.

Things I think need to be done:

- [x] ~~Remove copyright header from the completer files (@Valloric can correct me if I'm wrong here - I think the CLA is also a copyright assignment?)~~
- [ ] Remove commented code -- @jeteon
- [x] ~~Sign the CLA (does @jeteon need to do this or should I since I paid for the work?)~~
- [x] ~~Decide what should happen if codeintel isn't installed with `pip` - should the ci_completer just not try to do anything? Or is there an automated way to `pip install codeintel` when ycmd is installed? @Valloric, this one is for you.~~ When the `--codeintel-completer` flag is passed to the build script, codeintel will be installed with pip.
- [ ] Testing on a large codebase -- @jeteon, can you please provide information about how to get this up and running to be able to test it?
- [ ] Verify that Codeintel supplies built-in PHP functions and classes for indexing/completion -- @jeteon, if you can show me how to get this working, I can do this testing.

As mentioned in https://github.com/Valloric/ycmd/issues/96, PHP is just the beginning here. We can use codeintel for other languages if needed. Ruby would be an easy next target, as would CSS/SCSS/SASS.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/120)


vheon:
You should also:

* reformat some code
* remove references to the jedi completer, like the `GoTo` family of functions.

vheon:
>  Decide what should happen if codeintel isn't installed with pip - should the ci_completer just not try to do anything? Or is there an automated way to pip install codeintel when ycmd is installed? @Valloric, this one is for you.

Regarding this: isn't https://github.com/SublimeCodeIntel/codeintel a suitable repo?? I'm just asking...

cweagans:
@vheon no, that's not the correct repo. `pip install codeintel` installs codeintel + all dependencies in one package. That repo is *only* the codeintel component from komodoedit. See https://github.com/SublimeCodeIntel/CodeIntelSources/issues/3#issuecomment-97900740

vheon:
@cweagans too bad. Thanks for the answer :+1:

jeteon:
Hi, the code really is still quite rough and is more of the frame than the picture for now. Code Intel does support `Go To` so I thought I would keep that code in there and just make it functional. However, this isn't as nuanced as in the Jedi completer.

Valloric:
Sorry for the delay in responding!

> Remove copyright header from the completer files (@Valloric can correct me if I'm wrong here - I think the CLA is also a copyright assignment?)

The copyright header needs to be what the rest of ycmd uses. The CLA is not a copyright assignment. 

> Sign the CLA (does @jeteon need to do this or should I since I paid for the work?)

Easy solution: both of you should sign it. :)

>  Decide what should happen if codeintel isn't installed with pip - should the ci_completer just not try to do anything? Or is there an automated way to pip install codeintel when ycmd is installed? @Valloric, this one is for you.
 Testing on a large codebase -- @jeteon, can you please provide information about how to get this up and running to be able to test it?
 Verify that Codeintel supplies built-in PHP functions and classes for indexing/completion -- @jeteon, if you can show me how to get this working, I can do this testing.

Valloric:
>  Decide what should happen if codeintel isn't installed with pip - should the ci_completer just not try to do anything? Or is there an automated way to pip install codeintel when ycmd is installed? @Valloric, this one is for you.

This one is difficult. The way ycmd currently works is that by default, it provides only some semantic completion out of the box. For semantic engines that need to be compiled, it includes their source code with a submodule and the user can compile them by providing a flag to `build.py`.

We can have `build.py` call `pip` with the right params to get everything installed if the user provides a `--php-completer` build flag; if `pip` isn't installed, then we bail out with an error message that the user needs to first install it.

> Testing on a large codebase -- @jeteon, can you please provide information about how to get this up and running to be able to test it?

Yeah we need to have full end-to-end tests for every semantic completer in ycmd. This might require tweaking the `travis.yml` file so that the correct dependencies are pulled in on the build machines.

examples/test_client.py
```diff
@@ -0,0 +1,456 @@
+#!/usr/bin/env python
```
Valloric: Why is this file being added?

ycmd/completers/php/ci_completer.py
```diff
@@ -0,0 +1,117 @@
+#/usr/bin/env python
```
Valloric: This file needs to be reformatted to follow the ycmd style. Plenty of work to do here.

cweagans:
>This one is difficult. The way ycmd currently works is that by default, it provides only some semantic completion out of the box. For semantic engines that need to be compiled, it includes their source code with a submodule and the user can compile them by providing a flag to build.py.

>We can have build.py call pip with the right params to get everything installed if the user provides a --php-completer build flag; if pip isn't installed, then we bail out with an error message that the user needs to first install it.

Looks like @jeteon took care of this already - https://github.com/jeteon/ycmd/commit/becbe0df94d351d3cdc0211e32c820bfcb55a051 and https://github.com/jeteon/ycmd/commit/c8f01269bfd514cf2711cbe52f9eb40b67a29d9a

jeteon:
I've signed the CLA. I'll work on removing any license information in lieu of this.

examples/test_client.py
jeteon: Including it was a bit sloppy. I've since removed it in favour of putting formal tests in (TBD).

Valloric:
> I'll work on removing any license information in lieu of this.

Don't forget to use the ycmd license header; just put your own name instead of "Google Inc."

cweagans:
So I got this up and running in vim, but I wasn't able to see any completions from codeintel. I checked, and the `codeintel` process wasn't running at all. How can I debug this further @jeteon?

jeteon:
Can you confirm that ycmd was built with the `--codeintel-completer` flag? In a short while it won't be possible to do this, but you can uncomment lines 42-50 in the _ci_completer.py_ file and comment out line 51. This will generate a log file entry for most operations of the completer in `/tmp/cicompleter.log` which you can inspect to make sure it is at least working.

In normal operation there shouldn't be a `codeintel` process as such. The process would be a Python one executing CodeIntel code.

cweagans:
Ah, I see. Okay, I'll give that a try. Thanks!

jeteon:
The latest set of changes cover the tests for built-in PHP functions and classes. I'm not sure what would qualify as a "large code-base". Would something like Laravel or WordPresss count?

Also, @Valloric, does `ycmd` create a `fileReadyToParse` event for all the files in a project that one is working on, even if one isn't editing them directly? I ask, because I'm not sure there is a mechanism for getting the information from `ycmd` to tell CodeIntel to index other files that aren't currently being edited but are needed.

vheon:
@jeteon `FileReadyToParse` events are sent for files currently open. In theory a client could send events for files not open but wouldn't make much sense IMHO. 

> I ask, because I'm not sure there is a mechanism for getting the information from ycmd to tell CodeIntel to index other files that aren't currently being edited but are needed.

Shouldn't be CodeIntel's job to find the relevant files that need to be indexed?

cweagans:
Yeah, that should definitely be a codeintel thing. @jeteon take a look at how SublimeCodeIntel triggers indexing in the oop branch. The standalone codeintel daemon has a built-in way to recurse through a directory and index everything.

jeteon:
I'm not the most proficient of PHP developers so please forgive any misunderstandings I may have. My concern is where we have something like autoloading where the locations of the relevant files for a project may be defined many levels up the directory tree from a file that is being edited. 

In this case, I suppose a possible thing to do is to recursively go up the directory tree looking for a `composer.json` and a corresponding `vendor` directory. But this makes a strong assumption about the use of Composer. There are things like CodeIgniter out there which can be somewhat opaque to static analysis. I suppose this kind of sniffing is possible; its something I hadn't considered.

jeteon:
@cweagans SublimeCodeIntel has user initiated actions or signals for these things: https://github.com/SublimeCodeIntel/SublimeCodeIntel#configuring. I'm having trouble mapping the notion of a "project root" to the information that `ycmd` would be giving the completer. I understand there is a lot we can do on our end but had hoped there is some consistent way that these things are defined. I guess I'm expecting too much from the abstraction of the code editor.

cweagans:
I think it would be reasonable to require devs to create a `.codeintel` file at the root of their project. composer.json doesn't always live at the top directory, so that's probably not a good thing to depend on. In future iterations, maybe we could even pass some config to Codeintel through that file for the project (for instance, Drupal puts PHP code in files with the `.module` extension. I'm not sure how to communicate that to Codeintel), but that's definitely not something that needs to happen now.

Valloric:
Sorry for the delay, I missed the github updates.

> Also, @Valloric, does ycmd create a fileReadyToParse event for all the files in a project that one is working on, even if one isn't editing them directly?

No, just the one the user is currently editing.

>  I'm having trouble mapping the notion of a "project root" to the information that ycmd would be giving the completer. 

This is something we could ask the user to provide as a path returned from a new function in `ycm_extra_conf.py`.

Valloric:
What's the state of this PR now? Is it moving forward or should I close this?

cweagans:
Still moving forward. @jeteon spoke about direction yesterday. Hopefully, there will be more progress soon.

jeteon:
I think the branch is very close to being able to be merged. I just have one major hurdle that is wracking my brain. The completion test for names defined within projects but outside the file being edited fails when run in the test suite. However, when I manually run through the series of tests from the command line keeping the same instance of the completer throughout, the tests seem to run just fine. In fact, if you run just the one test with:

```
./run_tests.sh --skip-build --no-clang-completer \
    -- ycmd.completers.php.tests.ci_completer_tests:CompleteProjectFunction_test
``` 

it also runs fine. Something weird just seems to happen within the test run.

jeteon:
I got that issue sorted out. I think that this PR is feature complete now. @cweagans, is there anything I'm missing?

I ended up going with having a `.codeintel` file as @cweagans originally suggested. I looked into using `.ycm_extra_conf.py` mechanism in place but it seemed quite involved and this use case didn't need what it provides. Its possible to switch to that easily anyway as the presence of a file with a particular name is all that I'm checking for.

cweagans:
@jeteon Is the Codeintel stuff run as part of the same process as ycmd still? Or does Codeintel work as a standalone daemon?

jeteon:
@Valloric would be in a better position to comment but as far as I can tell, none of the completers actually run within the `ycmd` process itself. I know this because `ycmd` returned results even when my completer had syntax errors. Further, from my reading of the source code, the CodeIntel methods delegate processing to internal asynchronous calls, but wrap them in a ping loop that will check for errors and time the request out and such things. 

I get the allure of having a separate process but given the amount of decoupling and threading already present in the system, I have my doubts about the possible gains.

cweagans:
@jeteon I tried getting this up and running locally and wasn't able to do it. I downloaded the YouCompleteMe plugin for vim, then swapped out the included ycmd for `master` in your ycmd repo. When I ran `./build.py --codeintel-completer`, I got this: https://gist.github.com/cweagans/4f3618900d14ed956d4a

Am I doing something wrong? Is there an easier way to test the completion suggestions from vim?

jeteon:
@cweagans did you see my response on the gist? I think it might be a setup specific thing since I can't replicate it and the CI build doesn't seem to either. It seems to involve an error with moving some files. It might be that the directory permissions don't permit it. 

I wonder whether it is caused by the directory not existing before you run it and that behaving slightly differently on Mac systems. Can you try running the command I suggested in the gist?

cweagans:
Just gave it a try and it's still not working :/

jeteon:
Please try running this command on its own and let me know if it works:
```
/usr/local/bin/pip install codeintel -U -t /tmp/ycmd_ci/codeintel
```

cweagans:
That does not work. https://gist.github.com/cweagans/c59839ac13760bae8cc2

It's worth noting that I'm doing this on a freshly installed Mac OS with Chrome, pip, and vim. I don't have any special config on this machine at all.

jeteon:
That seems like a Mac specific error in Python distuitls. The error message sent me to this [SO question](
http://stackoverflow.com/questions/24257803/distutilsoptionerror-must-supply-either-home-or-prefix-exec-prefix-not-both).

The "fix" suggested in the answer is rather broad. This version of the command seems to be required on Mac OSX from what I gather and doesn't seem to have as far reaching consequences:
```
/usr/local/bin/pip install -U -t /tmp/ycmd_ci/codeintel --install-option="--prefix=" codeintel
```

I can confirm that it works fine on a Linux machine. Can you try it on your machine too?

cweagans:
```
$ pip install -U -t /tmp/ycmd_ci/codeintel --install-option="--prefix=" codeintel
/usr/local/lib/python2.7/site-packages/pip/commands/install.py:198: UserWarning: Disabling all use of wheels due to the use of --build-options / --global-options / --install-options.
  cmdoptions.check_install_build_global(options)
Collecting codeintel
  Downloading CodeIntel-0.1.7.tar.gz (11.9MB)
    100% |████████████████████████████████| 11.9MB 43kB/s
Collecting six (from codeintel)
  Downloading six-1.9.0.tar.gz
Collecting zope.cachedescriptors (from codeintel)
  Using cached zope.cachedescriptors-4.1.0.tar.gz
Collecting inflector (from codeintel)
  Using cached Inflector-2.0.11.tar.gz
Collecting subprocess32 (from codeintel)
  Using cached subprocess32-3.2.6.tar.gz
Collecting setuptools (from zope.cachedescriptors->codeintel)
  Downloading setuptools-18.0.1.tar.gz (630kB)
    100% |████████████████████████████████| 630kB 360kB/s
Installing collected packages: six, setuptools, zope.cachedescriptors, inflector, subprocess32, codeintel
  Running setup.py install for six
  Running setup.py install for setuptools
  Running setup.py install for zope.cachedescriptors
  Running setup.py install for inflector
  Running setup.py install for subprocess32
  Running setup.py install for codeintel
Successfully installed codeintel-0.1.7 inflector-2.0.11 setuptools-17.1.1 six-1.9.0 subprocess32-3.2.6 zope.cachedescriptors-4.1.0
Exception:
  File "/usr/local/lib/python2.7/site-packages/pip/basecommand.py", line 223, in main
    status = self.run(options, args)
  File "/usr/local/lib/python2.7/site-packages/pip/commands/install.py", line 337, in run
    lib_dir = distutils_scheme('', home=temp_target_dir)['purelib']
  File "/usr/local/lib/python2.7/site-packages/pip/locations.py", line 192, in distutils_scheme
    i.finalize_options()
  File "/usr/local/lib/python2.7/site-packages/setuptools-17.1.1-py2.7.egg/setuptools/command/install.py", line 38, in finalize_options
    orig.install.finalize_options(self)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/command/install.py", line 264, in finalize_options
    "must supply either home or prefix/exec-prefix -- not both"
DistutilsOptionError: must supply either home or prefix/exec-prefix -- not both
```

jeteon:
Hmmm...do you happen to have a file called `~/.pydistutils.cfg` on your system? If so, could you let me know what its contents are?

cweagans:
Nope. No such file.

jeteon:
Please create it with the following contents
```
[install]
prefix=
```

and then try the command again.

hugochinchilla:
I've been trying to test this on a large php project, but I haven't managed to get the server started, in the docs it only explains how to build the project, but how do I proceed once it's been built?

jeteon:
Is this for `ycmd` in general or the CodeIntel completer specifically?

hugochinchilla:
I think it is for ycmd in general, as I've already built it with codeintel support.

jeteon:
I think that **ycmd** is a bit low-level for what you're trying to do. I'd highly suggest you take a look at the [documentation](http://valloric.github.io/YouCompleteMe/) for **YouCompleteMe**. Its quite thorough and is essentially the user-facing part of the project whereas **ycmd** is more what happens behind the scenes.

The only thing specific to the CodeIntel completer that you would have to do is to put a `.codeintel` file at the root of your project so that it knows how to find that.

hugochinchilla:
@jeteon I want to use the Atom plugin (https://github.com/Qusic/atom-youcompleteme) the fact is that it requieres a custom patched version to work, and does not include codeintel completion, I wanted to know how to launch my own version of the ycmd server so I can apply your pull request over the patched version from this proyect and try it over a PHP project.

jeteon:
@hugochinchilla, whilst in this case I suppose you could "patch" the server after it has been built, the recommended way of doing this is to build it with the patch already applied. In the instructions for ["Building the C++ components of YCMD"](https://github.com/Qusic/atom-youcompleteme#2-build-c-components-of-ycmd), after the first step:

```
cd ~/.atom/packages/you-complete-me/ycmd
```

You have to apply the "patch" at this point. I don't know whether the plugin pulls the code in via git or just fetches a copy of the files but this is the point at which you would have to merge in the code from this branch. After this, then you can run the modified build command:

```
./build.py --codeintel-completer  [--clang-completer] [--omnisharp-completer] [--gocode-completer]
```

<hr>
# Hack

I **do not** recommend this but as the code stands now, if you want a dirty alternative to get it working what you could basically do is copy the `ycmd/completers/php` directory from this branch into your `ycmd/completers/` directory and then from the root of your `ycmd` code (_~/.atom/packages/you-complete-me/ycmd_?) run the command:

```
pip install codeintel -U -t third_party/codeintel
``` 

This can be done before or after you build the server, **for now**. I can't guarantee this would work in future.

jeteon:
I can't make heads or tails of the build failures on the Python 2.6 tests. They both seem to be in areas I didn't touch. I can't imagine why the `ycmd.completers.completer_utils_test.FiletypeTriggerDictFromSpec_Works_test` would fail.

puremourning:
it does seem strange. But repeatable. i find the `eq_` checker a bit brute force. Maybe try changing the check to using `assert_that` and `has_entries`, `contains`, etc. from pyhamcrest. That might give you better diagnostics.

jeteon:
I'm reluctant to change that test code as this pull is largely to add a completer and that test is in the core of **ycmd**.

As far as I can tell, the failure is caused by the fact that the set under the key "foo" in that case, contains objects with differing identifies. All the other keys contain exactly the same objects (same type and location) between the two dictionaries being tested for equality. I would expect that the objects represent the same thing though.

I'm not sure changing from `eq_` would be a good idea since the test seems to be a check that the two methods result in the same objects being created.

puremourning:
I meant to help work out why it is broken. It is seemingly fine on everyone else's commits. It's not impossible that there is a flaw in what the test is checking for. But the output doesn't show the actual contents, so it is hard to identify the issue.

jeteon:
I know, and I really appreciate that. I'm sorry if my reluctance came across as a lack of appreciation. I'm also starting to think it may be something unrelated but somehow triggered by the particular configuration in this pull. 

I'm considering getting a Python 2.6 install to try to reproduce it myself. I would also like to get a look at the contents as well. In retrospect, my suggestion that the test is checking that the identical objects are returned is a little silly. The contents matter. I'm just guessing that `eq_()`, `contains_()`, etc, have no way other than object identify to check this. Perhaps a more thorough test is in order.

jeteon:
I can't see anything I've changed in the core from the diff of this branch to the main one. @Valloric do you happen to know if I may have accidentally done something to cause the test

`ycmd.completers.completer_utils_test.FiletypeTriggerDictFromSpec_Works_test`

to fail?

Valloric:
Huh, a failure specific to Python 2.6. We haven't had many of those; maybe one or two over the last few years.

I honestly don't have a clue what's causing it. Rewriting the test a bit so that it passes seems like a good bet forward. I understand that you didn't touch the code that's relevant to the test, but I don't have any other advice to offer, sadly.

jeteon:
Lol I had hoped catching up to the main branch would fix this. Thanks for looking at it, @Valloric. I'll try to see what I can do about that test.

cweagans:
@jeteon Sorry for the delay in getting back to you. Between work and travel, I haven't had much downtime.

IMO, the easiest path forward here is to *not* try to pull down a local copy of codeintel at all, and instead, utilize the socket connection capabilities that codeintel provides out of the box. If this means that a user must run `pip install codeintel`, then that should be fine. Again, look at how sublimecodeintel does it: it creates a socket server that listens on a given port, then starts the codeintel daemon with an argument that instructs codeintel to connect back to the socket server and wait for messages on the socket. The details for the message format can be found at https://github.com/Komodo/KomodoEdit/wiki/290-CodeIntel-OOP-Protocol. It should be very simple to do this, tell codeintel to index a directory, and then query completions from the codeintel daemon rather than trying to do it all inside the completer.

This also relies on a tried-and-true method of installation that will be used by hundreds/thousands of Sublime Code Intel users. *This problem has already been solved for us.*

jeteon:
Welcome back. 

I hear you with using a system install of CodeIntel for the Mac OSX installation issue and am considering add a `--global-codeintel` flag similar to what is done for the Clang completer. 

However, the failing test right now is due to a completely different issue.

cweagans:
I'm saying that a global codeintel daemon should be the *default*, rather than an option. The ycmd completer would be so much more simple that way.

vheon:
> If this means that a user must run `pip install codeintel`, then that should be fine.

I'm not a python developer so I have to ask for confirmation: can really a user do that? isn't `sudo` required in that case?

cweagans:
`sudo` shouldn't be required, no. At least, not if permissions are configured correctly. Users can also use a virtualenv if the path to the codeintel executable is configurable.

puremourning:
The other problem with global is that you can't control what version is being used. We have this problem all the time with Python. We'd be relying on code Intel to not break backward compatibility.

puremourning:
Also I can't run pip install anything globally on corporate systems.

vheon:
One thing that I'm also not entirely convinced is the `.codeintel` file. We already have `.ycm_extra_conf.py` as a configuration file and since `.codeintel` at the moment is an empty file I would prefer an empty `.ycm_extra_conf.py` than an extra file.

jeteon:
Running `pip install codeintel` on most Linux distributions requires `sudo`. This is because it is a global install by default. Changing the permissions of these directories to allow arbitrary writes by user level code would allow anything the user runs thereafter to change the Python core packages at will. Given the amount of Python a Linux user uses even without knowing it, bad things could happen very quickly. 

If you need to use a `virtualenv` to run `ycmd` with CodeIntel every time, I would argue that we would have made using CodeIntel with `ycmd` harder and not easier as you argue. If the path is configurable, then maybe you could install it somewhere where you don't need root privileges and then come back and change the configs but again, I don't see how this as a default would be making things easier for the user than just doing a local install by default and using that. Like I said before, this is something I will put in for users that have trouble with the default setup though.

@vheon The `.codeintel` file was used as a placeholder for some potentially CodeIntel specific flags later on (e.g. PHP files without .php extensions). But given that it is empty right now, it can just as well be an empty `.ycm_extra_conf.py`.

vheon:
> @vheon The .codeintel file was used as a placeholder for some potentially CodeIntel specific flags later on (e.g. PHP files without .php extensions). But given that it is empty right now, it can just as well be an empty .ycm_extra_conf.py.

I would like to know what the others think about this. @Valloric @puremourning @micbou @oblitum

vheon:
> If you need to use a virtualenv to run ycmd with CodeIntel every time, I would argue that we would have made using CodeIntel with ycmd harder and not easier as you argue. If the path is configurable, then maybe you could install it somewhere where you don't need root privileges and then come back and change the configs but again, I don't see how this as a default would be making things easier for the user than just doing a local install by default and using that. Like I said before, this is something I will put in for users that have trouble with the default setup though.

Thanks! That is exactly what I was thinking :+1:

puremourning:
If the purpose of the file is to set the project properties, then maybe we should have `def PropertiesForPHPProject` method in `.ycm_extra_conf.py`. That way the existing rules for finding the file, including having a global one, just work. It can return a dict which is extensible in future.

puremourning:
Maybe a better function name, but you get the idea :)

jeteon:
It would be nice to be able to reuse the existing machinery to find the file. Basically, I just use it to indicate the project root right now. Adding a function like that in future isn't a bad idea either.

puremourning:
But the method can return 

```python
'Project_root` : <path>
```

Then the files doesn't have to sit itself at the project root

jeteon:
That's true. Do you happen to know of anywhere else in the `ycmd` ecosystem where something like this is done already?

vheon:
@puremourning 
> If the purpose of the file is to set the project properties, then maybe we should have def PropertiesForPHPProject method in .ycm_extra_conf.py. That way the existing rules for finding the file, including having a global one, just work. It can return a dict which is extensible in future.

Once we do that, I think we should change the `FlagsForFile` function with a name relative to the C-family languages; of course we should leave `FlagsForFile` for backwards compatibility.

vheon:
> That's true. Do you happen to know of anywhere else in the ycmd ecosystem where something like this is done already?

@jeteon you should look how ycmd retrieve the flags from the `FlagsForFile` function in `.ycm_extra_conf.py`

Valloric:
Agreed with others that we shouldn't have a new config file. All configuration that requires the user to provide logic should go into `.ycm_extra_conf.py`. As @puremourning said, it should be in the form of a new top-level function in the file; if it's present, we use it to get the conf.

All this being said, if we can use the options file passed to ycmd on startup, that's the best option. The extra conf file is only used for "programmable" configuration, like the `FlagsForFile` method, not for key-value options and the like.

sekjun9878:
Just wondering, is there any progress on this?

jeteon:
Not as such. The branch is functional. There are just some architectural changes with regard to:

* Using the `.ycm_extra_conf.py` file to designate the project root instead of `.codeintel`
* Changing to using an out of process instance of CodeIntel to get the completions

Functionally, nothing will change though.

molovo:
Just merged this PR into the branched ycmd used by the atom you-complete-me plugin and can confirm the basics work.

Just an FYI: There are a few niggles, but I'm not sure if they are caused more by atom's completion than by this plugin. It can find function and variable references within classes, but doesn't autocomplete the full function call with parameter placeholders as it does in sublime. Haven't been able to get it to compile within vim's YouCompleteMe to know if you get the same results there.

jeteon:
Thanks for the feedback and sorry about the late response. I'm not sure that CodeIntel is able to provide doc or parameter information. At least not the way I coded it in the plugin. I'm going to look into whether this functionality is possible though.

Valloric:
This PR seems stalled so closing. A new PR can be opened if this work gets picked up.

tam5:
@Valloric has there been any progress made regarding a semantic completer for PHP?

puremourning:
We have a prototype using Language server protocol, but it is queues behind java support.

tam5:
Thanks @puremourning for your response.

Let me know if there is anything I can do to help out, just point me in a direction.

Issue #121, opened 2015-05-01
Change completion API response to a more general form


Qusic:
This seems too specific to YouCompleteMe vim client. Parameters list should not be merged into the function name.

```
{
    "detailed_info": "Foo & operator=( const Foo & )\nFoo & operator=( Foo && )\n",
    "extra_menu_info": "Foo &",
    "insertion_text": "operator=",
    "kind": "FUNCTION",
    "menu_text": "operator=( const Foo & )"
}
```

My desired completions look like:

```
{
    "kind": "function",
    "name": "myFun",
    "parameters": [{
        "type": "int",
        "name": "number"
    }, {
        "type": "std::string",
        "name": "text"
    }],
    "return_type": "Foo &",
    "docstring": "..."
}

{
    "kind": "variable",
    "name": "a",
    "type": "int"
}
```

So that clients decide what the insertion text, menu info or whatever is, and clients with different capabilities can all have a chance to make the best completion experience.
I am working on the Atom client and such completions can be used to build snippets and provide function completions with parameter placeholders.

Valloric:
Oh I agree with you completely. `detailed_info` is meant to be a suggestion. We should totally be returning a list of all the parameters (and function overloads, which are currently missing) in a structured format.

More work needs to be put into making this happen and I haven't had the time to do it so far.

Qusic:
Considering different argument formats for functions, templates, objc methods and even functions in other languages, my proposal in the original post seems to make things too complicated.
Do you have any ideas for a simple and flexible completion representation?

Valloric:
I don't currently have a specific design in mind beyond "we need to return structured data for parameters/overloads" so that stuff is broken out and that clients can pick and choose which data they want without having to parse detailed_info.

ptrv:
This change would be really useful since in `emacs-ycmd` we do parsing and text manipulation of detailed_info, to separate the parameters and return type. A more generic API would be really appreciated.

sanwv:
how about return a snippet?

```
{
    "snippet": "myFun(${1:int},${2:string}){\n return ${3:Foo}\n}",
   ....
}
```

richard1122:
Is there going to be any updates about this? 
Split parameters out would be great.

Also it's acceptable to just add parameters for backward compatible.

puremourning:
The new signature-help support will supersede this, I think.

Also we can't really break the API in this way

PR #124, opened 2015-05-07
Structured completion data format for parameters

Qusic:
See also issue #121

Completion objects have these properties:
* `typed_string`<br />Text that users would be expected to type to get this completion result. 
* `display_string`<br />Text that is displayed for users.
* `completion_parts`<br />An array of completion part objects.
* `result_type`<br />Such as 'int', 'void', 'unsigned long'.
* `kind`<br />Such as 'function', 'class', 'variable'.
* `doc_string`<br />Doc string.
* `extra_data`<br />An object contains completer-specific data.

Completion part objects have these properties:
* `part`<br />String value of the part.
* `literal`<br />Bool value that indicates whether this part should be inserted directly as a literal string or as a placeholder for users to insert code.

Two examples are below.
```
{
  "typed_string": "vector",
  "display_string": "vector<typename _Tp, typename _Alloc>",
  "kind": "CLASS",
  "completion_parts": [
    { "part": "vector<", literal: true },
    { "part": "typename _Tp", literal: false },
    { "part": ", ", literal: true },
    { "part": "typename _Alloc", literal: false },
    { "part": ">", literal: true }
  ]
}

{
  "typed_string": "withArg2:withArg3:",
  "display_string": "test:withArg2:(int) withArg3:(int)",
  "result_type": "void",
  "kind": "FUNCTION",
  "completion_parts": [
    { "part": "withArg2:", literal: true },
    { "part": "(int)", literal: false },
    { "part": " withArg3:", literal: true },
    { "part": "(int)", literal: false }
  ]
}
```

It is now working with identifier completer, clang completer and python completer. Tests are also updated. It should be able to put an end to major objc completion issues such as https://github.com/Valloric/YouCompleteMe/issues/1221.
However I haven't tested with cs completer and go completer because I am not familiar with those two languages.

(I have signed the Google CLA.)


Valloric:
Thanks for the PR! 

The "completion parts" section would be an improvement over the current state, but I feel like we can do better. I was thinking that ycmd should be splitting out function parameters and types. For something like `int foo(double bar)`, we'd return a structure like the following (along with the presently returned info):

 ```
{
  return_type: 'int', 
  parameters: [{type: 'double', name: 'bar'}]
}
```

Probably a list of such objects, so that function overloads can be provided as well. For Python, the `type` and `return_type` would be empty.

This would leave the logic of what should and shouldn't be displayed to the user up to the client.

**Comment by Qusic** [comment]
For function completions like `foo(double bar)`, `clang_getCompletionChunkText` actually returns `double bar` as one single chunk of kind `CXCompletionChunk_Placeholder`.
And if trying to extract out type and name, we will have problems with different parameter formats on both ycmd and clients.
`c_function_name(type2 name2)`
`cpp_template_class<type1 name1, type2 name2>`
`cpp_template_function<type1 name1>(type2 name2, type3 name3)`
`objc_method_name1:(type1)name1 name2:(type2)name2`
I think my current solution is a proper way to handle it.
As for function overloads, each one of them is a separate completion object and they all have the same `typed_string`.

**Comment by ptrv** [comment]
@Valloric @Qusic Once the data format is final I could help with the go completer. The `gocode` completer however does not return separate parts of a function signature but the whole function in one line.

See https://github.com/nsf/gocode/blob/master/docs/autocomplete_formats.md#json

**Comment by vheon** [comment]
@ptrv We could see if the maintainer of `gocode` thinks that could be useful and change the json formatter or make a new one which return the function signature with its arguments split up. Or we could parse the arguments in the ycmd completer itself since go is a really simple language to parse, but I would prefer to put this functionality in `gocode` itself.

**Comment by mispencer** [comment]
C# can be supported by passing true values for ``WantKind`` and ``WantSnippet`` in the Omnisharp Request. Then, ``kind`` can gotten directly from the ``Kind`` parameter in the result, and ``completion_parts`` can gotten by parsing the `Snippet` parameter, which should be something like ``Func<${1:in T}, ${2:out TResult}>(${3:T arg})$0``

**Comment by Valloric** [comment]
> For function completions like foo(double bar), clang_getCompletionChunkText actually returns double bar as one single chunk of kind CXCompletionChunk_Placeholder. 

My point is that for clang languages, we may want to skip the Placeholder chunks and actually go to the function itself through the clang cursor API and then read what the exact parameter names/types are (and then return that). I don't see why we should necessarily rely on the Placeholder chunks (other than "it's already there", but it's pre-formatted which is an issue).

I don't want ycmd to only provide such formatted placeholder data; we're just tying the hands of clients that way. We may want to provide this placeholder data _in addition to_ the structured data I mentioned. That could work.

Thoughts, everyone?

**Comment by Valloric** [comment]
Here's how we could include this. Rename the fields like so:

```
"completion_chunks": [
    { chunk: "vector<", placeholder: false },
    { chunk: "typename _Tp", placeholder: true },
    { chunk: ", ", placeholder: false },
    { chunk: "typename _Alloc", placeholder: true },
    { chunk: ">", literal: false }
  ]
```

Note that the logic of `literal` -> `placeholder` has been reversed. The point is to make it more obvious why we're making the distinction between the data types. Also, using the word "chunk" is consistent with the clang API.

We can then add structured info about param names and types in a separate section [like I mentioned](https://github.com/Valloric/ycmd/pull/124#issuecomment-99998831).

**Comment by oblitum** [comment]
> I don't want ycmd to only provide such formatted placeholder data; we're just tying the hands of clients that way. We may want to provide this placeholder data in addition to the structured data I mentioned. That could work.

> Thoughts, everyone?

Personally I agree, better to have a more extensible layout that can encompass more contrained APIs than the contrary, constrain ourselves and not be able to support nice features from others or when the limited one evolves.

**Comment by oblitum** [comment]
Anyway, as discussed before, it may be worth thinking about the flexibility of the all-in-one structured completion request vs the segmented, but still complete, completion request.

**Comment by Qusic** [comment]
If the param names and types are intended to be clang-specific, I think it is ok to be included in `extra_data`.

Btw I just looked into clang cursor API but couldn't find a way to obtain the cursor from each completion result. Any suggestions?

**Comment by Valloric** [comment]
> Btw I just looked into clang cursor API but couldn't find a way to obtain the cursor from each completion result. Any suggestions?

I can't see anything either. We might just be out of luck.

**Comment by Qusic** [comment]
This seems weird. https://travis-ci.org/Valloric/ycmd/jobs/62501431#L784
But it did succeed with python 2.6.

**Comment by Valloric** [comment]
@Qusic That looks like a flake. Ignore it.

File: cpp/ycm/ClangCompleter/ClangHelpers.cpp [hunk 1]
```diff
@@ -174,24 +175,7 @@ std::vector< CompletionData > ToCompletionDataVector(
...
-                                     completions.size() );
-
-    if ( index == completions.size() ) {
-      completions.push_back( boost::move( data ) );
-    }
-
-    else {
-      // If we have already seen this completion, then this is an overload of a
-      // function we have seen. We add the signature of the overload to the
-      // detailed information.
```
Valloric: Why is this logic going away?

cpp/ycm/ClangCompleter/CompletionData.cpp
```diff
@@ -158,121 +99,91 @@ std::string RemoveTwoConsecutiveUnderscores( std::string text ) {
...
-
-  original_string_ = RemoveTrailingParens( boost::move( original_string_ ) );
-  kind_ = CursorKindToCompletionKind( completion_result.CursorKind );
-
-  // We remove any two consecutive underscores from the function definition
-  // since identifiers with them are ugly, compiler-reserved names. Functions
-  // from the standard library use parameter names like "__pos" and we want to
-  // show them as just "pos". This will never interfere with client code since
-  // ANY C++ identifier with two consecutive underscores in it is
-  // compiler-reserved.
```
Valloric: Could you elaborate on why is this being removed as well?

ycmd/completers/go/tests/gocode_completer_test.py
```diff
@@ -84,10 +84,14 @@ def ComputeCandidatesInner_test( self ):
     result = self._completer.ComputeCandidatesInner(self._BuildRequest(10, 40))
     eq_(mock.cmd, ['THE_BINARY', '-f=json', 'autocomplete', PATH_TO_TEST_FILE, "292"])
     eq_(result, [{
-        'menu_text': u'Prefix',
-        'insertion_text': u'Prefix',
-        'extra_menu_info': u'func() string',
-        'kind': u'func'
+      'typed_string': u'Prefix',
```
Valloric: Changing all these names changes the public API; we'll have to update YCM itself plus notify emacs-ycmd people like @abingham if we go this route. All of that work may not be worth it.

Valloric:
@Qusic I seem to have misunderstood this PR; I thought you were adding the `completion_chunks` section to the responses, but having actually started reading through the code, you are changing parts of the public API along with other extensive changes. 

All of that is too much for a single PR. Break it up so it can be reviewed separately. Have one PR that just adds the completion chunks to the API; that can be reviewed and merged quickly since we discussed those changes here.

Changing the public API I'm not so sure about; in a vacuum it would be great, but ycmd has quite a few clients now and breaking compatibility (even before ycmd 1.0) comes with costs.

There also seem to be other, unrelated changes in this PR that change ycmd logic; those should be separate PRs on their own.

Qusic:
Field names such as `menu_text` and `extra_menu_info` are confusing to both clients and completers and there are no documentation on the API endpoints. Clients can point to a specific commit of ycmd and I presume it won't impact on end users and naming changes should not be that costly or scary.

I don't see much to split except some trivial ones like gitignore or code formating. Major logic changes are in the constructor of `CompletionData` and `ExtractDataFromString` and they are related to the new completion format.

The new implementation actually is simpler and easier to understand because of no separate `ChunkToString` and `OptionalChunkToString` functions, no `saw_left_paren`, `saw_function_params`, `saw_placeholder` flags and all kinds of chunks processed.

If you means adding the `completion_chunks` section and changing field names are necessary to be in separated PRs, I have to say it is a tedious job.

ptrv:
As a frequent committer to `emacs-ycmd` I really would like to see this PR merged. The current API is indeed a bit confusing and every completer assigns their information differently, which makes it necessary to parse differently. Also, there is extra parsing needed to extract function arguments. Additionally, you need to parse different objects for the function arguments. For C++ one has to parse `detailed_info` for function arguments, for Python the information is in `extra_menu_info`. I would really welcome a consistent API.

cpp/ycm/ClangCompleter/ClangHelpers.cpp
ptrv: I suppose also the overloads are added here as candidates and not only to the `detaqiled_info`

Valloric:
I'm not disagreeing with the changes that are being proposed here, just asking them to be split up into separate PRs. The smaller, the better while keeping the PRs self-contained. It makes it much easier to review such changes.

I eagerly await these separate PRs. :)

PR #125, opened 2015-05-08
Add path and line to python completion data

ptrv:
jedi provides the location of the completion candidate as `line` and `path`. In `company-ycmd` for Emacs this information can be used to jump to the location while completion popup is open.

I have read and signed the Google CLA

### DIFF
--- a/ycmd/completers/python/jedi_completer.py
+++ b/ycmd/completers/python/jedi_completer.py
@@ -56,12 +56,30 @@ class JediCompleter( Completer ):
       return jedi.Script( contents, line, column, filename )
 
 
+  def _GetExtraData( self, completion ):
+      location = {}
+      if completion.module_path:
+        location[ 'filepath' ] = ToUtf8IfNeeded( completion.module_path )
+      if completion.line:
+        location[ 'line_num' ] = completion.line
+      if completion.column:
+        location[ 'column_num' ] = completion.column + 1
+
+      if location:
+        extra_data = {}
+        extra_data[ 'location' ] = location
+        return extra_data
+      else:
+        return None
+
+
   def ComputeCandidatesInner( self, request_data ):
     script = self._GetJediScript( request_data )
     return [ responses.BuildCompletionData(
                 ToUtf8IfNeeded( completion.name ),
                 ToUtf8IfNeeded( completion.description ),
-                ToUtf8IfNeeded( completion.docstring() ) )
+                ToUtf8IfNeeded( completion.docstring() ),
+                extra_data = self._GetExtraData( completion ) )
              for completion in script.completions() ]
 
   def DefinedSubcommands( self ):
--- a/ycmd/tests/get_completions_test.py
+++ b/ycmd/tests/get_completions_test.py
@@ -42,6 +42,12 @@ def CompletionEntryMatcher( insertion_text ):
   return has_entry( 'insertion_text', insertion_text )
 
 
+def CompletionLocationMatcher( location_type, value ):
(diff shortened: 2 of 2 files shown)


ycmd/completers/python/jedi_completer.py
```diff
@@ -58,10 +58,20 @@ def _GetJediScript( self, request_data ):
 
   def ComputeCandidatesInner( self, request_data ):
     script = self._GetJediScript( request_data )
+
+    def get_extra_data(completion):
```
Valloric: Spaces inside parens.

Valloric:
This is fine. Needs a test though.

ptrv:
Updated PR with requests

ycmd/completers/python/jedi_completer.py
```diff
@@ -56,12 +56,22 @@ def _GetJediScript( self, request_data ):
       return jedi.Script( contents, line, column, filename )
 
 
+  def _GetExtraData( self, completion ):
+      extra_data = {}
+      if completion.module_path:
+        extra_data[ 'path' ] = ToUtf8IfNeeded( completion.module_path )
+      if completion.line:
+        extra_data[ 'line' ] = completion.line
```
Valloric: It just occurred to me that this is similar to the way diagnostic locations are reported in the API; [example](https://github.com/Valloric/ycmd/blob/4cbb1374af1eedfb3600e80828d4a6c41a9690f9/ycmd/tests/diagnostics_test.py#L67). We should probably emulate that structure. If Jedi provides a column number, we should return that as well. If not, leave the `column_num` field out; we can add it in the future when this data becomes available.

So something like this:

```
location: {
  line_num: 1,
  column_num: 1,  // this can be missing
  filepath: '/foo/bar.txt'  // must be full, absolute path
}
```

The point is to have the REST API structure and field names consistent across the API.

ptrv:
@Valloric Agree on the naming of the field names. Updated the PR.

ycmd/completers/python/jedi_completer.py
```diff
@@ -56,12 +56,24 @@ def _GetJediScript( self, request_data ):
       return jedi.Script( contents, line, column, filename )
 
 
+  def _GetExtraData( self, completion ):
+      extra_data = {}
+      if completion.module_path:
+        extra_data[ 'filepath' ] = ToUtf8IfNeeded( completion.module_path )
```
Valloric: Do also put the filepath/line_num/column_num under a `location` object.

ptrv:
@Valloric Yes, of course. I had overlooked the location. Updated PR.

ycmd/tests/get_completions_test.py
```diff
@@ -678,7 +678,20 @@ def GetCompletions_JediCompleter_Basic_test():
...
+                                   CompletionEntryMatcher( 'b' ),
+                                   has_entry( 'extra_data',
+                                              has_entry( 'location',
+                                                         has_entry( 'line_num', 3 ) ) ),
+                                   has_entry( 'extra_data',
+                                              has_entry( 'location',
+                                                         has_entry( 'line_num', 4 ) ) ),
+                                   has_entry( 'extra_data',
+                                              has_entry( 'location',
+                                                         has_entry( 'column_num', 10 ) ) ),
```
Valloric: Is this 80 columns? Seems to be too wide for that.

ptrv:
@Valloric Oh my, goodness! Sorry for this back and forth...

Upadted the PR

Valloric:
Thanks for the PR!

PR #129, opened 2015-05-13
Fix run_tests.sh for no clang case

mispencer:
The clang case has [] - the no clang case should too.

I've already signed the CLA.


Valloric:
Thanks!

PR #130, opened 2015-05-13
Add GetType completer command for CSharp completer

mispencer:
This command displays documentation on the type under the cursor.

This is the command I was using in the screenshot in  https://github.com/Valloric/YouCompleteMe/pull/1499.

I've already signed the CLA.

### DIFF
--- a/ycmd/tests/subcommands_test.py
+++ b/ycmd/tests/subcommands_test.py
@@ -603,11 +603,11 @@ def RunCompleterCommand_GoToImplementationElseDeclaration_CsCompleter_MultipleIm
 
 
 @with_setup( Setup )
-def RunCompleterCommand_TypeLookup_CsCompleter_test():
+def RunCompleterCommand_GetType_CsCompleter_test():
   app = TestApp( handlers.app )
   app.post_json( '/ignore_extra_conf_file',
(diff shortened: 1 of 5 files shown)


vheon:
Travis is sad.

mispencer:
It returns a different message on Mono than it does in Windows. And I should have been returning the type anyway. The test is fixed now.

Valloric:
We should use the same `GetType` name.

mispencer:
The command been renamed to ```GetType```.

ycmd/tests/subcommands_test.py
```diff
@@ -603,6 +603,130 @@ def RunCompleterCommand_GoToImplementationElseDeclaration_CsCompleter_MultipleIm
...
+
+  eq_( {
+        u'message': u"System.String"
+      },
+      app.post_json( '/run_completer_command', gettype_data ).json )
+
+  StopOmniSharpServer( app )
+
+
+@with_setup( Setup )
```
Valloric: This is fine for now, but soon we'll have to split up this file into file-per-semantic-completer since this is getting big.

Valloric:
Thanks for the PR!

PR #131, opened 2015-05-14
Remove executable mode

micbou:
Commit 5584388d15e8303ccab72d8597a677f2dbcb1295 made `jedi_completer.py` and `get_completions_test.py` files executable. It seems that `nosetests` does not like this and the `get_completions` tests was not run.

I signed the CLA.


Valloric:
Thanks!

PR #135, opened 2015-05-18
Remove print statement

micbou:
Someone forgot a print. It happens.

CLA signed.


Valloric:
Nice catch! Thanks for the PR!

PR #136, opened 2015-05-18
Remove deprecated --use-mirrors argument from pip

micbou:
Pip no longer supports --use-mirrors since version 1.5. See https://github.com/pypa/pip/pull/1098.

CLA signed.


Valloric:
Thanks!

PR #137, opened 2015-05-20
New completion format

Qusic:
See also https://github.com/Valloric/ycmd/pull/124

I have removed all unnecessary changes and it looks much cleaner now.

> Completion objects have these properties:
* `typed_string`<br />Text that users would be expected to type to get this completion result. 
* `display_string`<br />Text that is displayed for users.
* `completion_chunks`<br />An array of completion chunk objects.
* `result_type`<br />Such as 'int', 'void', 'unsigned long'.
* `kind`<br />Such as 'function', 'class', 'variable'.
* `doc_string`<br />Doc string.
* `extra_data`<br />An object contains completer-specific data.

> Completion chunk objects have these properties:
* `chunk`<br />String value of the chunk.
* `placeholder`<br />Bool value that indicates whether this chunk should be inserted directly as a literal string or as a placeholder for users to insert code.

> Two examples are below.
```
{
  "typed_string": "vector",
  "display_string": "vector<typename _Tp, typename _Alloc>",
  "kind": "CLASS",
  "completion_chunks": [
    { "chunk": "vector<", "placeholder": false },
    { "chunk": "typename _Tp", "placeholder": true },
    { "chunk": ", ", "placeholder": false },
    { "chunk": "typename _Alloc", "placeholder": true },
    { "chunk": ">", "placeholder": false }
  ]
}
{
  "typed_string": "withArg2:withArg3:",
  "display_string": "test:withArg2:(int) withArg3:(int)",
  "result_type": "void",
  "kind": "FUNCTION",
  "completion_chunks": [
    { "chunk": "withArg2:", "placeholder": false },
    { "chunk": "(int)", "placeholder": true },
    { "chunk": " withArg3:", "placeholder": false },
    { "chunk": "(int)", "placeholder": true }
  ]
}
```

(CLA signed)

### DIFF
--- a/cpp/ycm/CandidateRepository.cpp
+++ b/cpp/ycm/CandidateRepository.cpp
@@ -95,11 +95,11 @@ std::vector< const Candidate * > CandidateRepository::GetCandidatesForStrings(
 
     foreach ( const CompletionData & data, datas ) {
       const Candidate *&candidate = GetValueElseInsert( candidate_holder_,
-                                                        data.original_string_,
+                                                        data.typed_string_,
                                                         NULL );
 
       if ( !candidate )
-        candidate = new Candidate( data.original_string_ );
+        candidate = new Candidate( data.typed_string_ );
 
       candidates.push_back( candidate );
     }
--- a/cpp/ycm/ClangCompleter/ClangHelpers.cpp
+++ b/cpp/ycm/ClangCompleter/ClangHelpers.cpp
@@ -165,7 +165,6 @@ std::vector< CompletionData > ToCompletionDataVector(
     return completions;
 
   completions.reserve( results->NumResults );
-  unordered_map< std::string, uint > seen_data;
 
   for ( uint i = 0; i < results->NumResults; ++i ) {
     CXCompletionResult completion_result = results->Results[ i ];
@@ -174,24 +173,7 @@ std::vector< CompletionData > ToCompletionDataVector(
       continue;
 
     CompletionData data( completion_result );
-    uint index = GetValueElseInsert( seen_data,
-                                     data.original_string_,
-                                     completions.size() );
-
-    if ( index == completions.size() ) {
-      completions.push_back( boost::move( data ) );
-    }
-
-    else {
-      // If we have already seen this completion, then this is an overload of a
-      // function we have seen. We add the signature of the overload to the
-      // detailed information.
-      completions[ index ].detailed_info_
-      .append( data.return_type_ )
-      .append( " " )
-      .append( data.everything_except_return_type_ )
-      .append( "\n" );
-    }
+    completions.push_back( boost::move( data ) );
   }
 
   return completions;
--- a/cpp/ycm/ClangCompleter/CompletionData.cpp
+++ b/cpp/ycm/ClangCompleter/CompletionData.cpp
@@ -80,30 +80,6 @@ CompletionKind CursorKindToCompletionKind( CXCursorKind kind ) {
 }
 
 
-bool IsMainCompletionTextInfo( CXCompletionChunkKind kind ) {
-  return
-    kind == CXCompletionChunk_Optional     ||
-    kind == CXCompletionChunk_TypedText    ||
-    kind == CXCompletionChunk_Placeholder  ||
-    kind == CXCompletionChunk_LeftParen    ||
-    kind == CXCompletionChunk_RightParen   ||
-    kind == CXCompletionChunk_RightBracket ||
-    kind == CXCompletionChunk_LeftBracket  ||
-    kind == CXCompletionChunk_LeftBrace    ||
-    kind == CXCompletionChunk_RightBrace   ||
-    kind == CXCompletionChunk_RightAngle   ||
-    kind == CXCompletionChunk_LeftAngle    ||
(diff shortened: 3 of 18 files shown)


Valloric:
Either I'm missing something, or this PR is still mixing code refactoring with changing the API. It's also adding the completion chunks section. (Feel free to elaborate if I really am misunderstanding things here.)

Here's an idea of how to structure the changes you want to make, in a reasonable order:

1. PR for adding the completion_chunks section. This PR does that and only that; no other unrelated changes, only the minimum necessary to implement completion_chunks (with tests).
2. PR for changing the API. No extra refactoring, _just_ the API changes.
3. PR for the refactoring.

Qusic:
As for the third one, refactoring, the problem is these two functions:
* [CompletionData::CompletionData(const CXCompletionResult &completion_result)](https://github.com/Valloric/ycmd/blob/master/cpp/ycm/ClangCompleter/CompletionData.cpp#L177)
* [void CompletionData::ExtractDataFromChunk(...)](https://github.com/Valloric/ycmd/blob/master/cpp/ycm/ClangCompleter/CompletionData.cpp#L219)

I re-wrote them because
* they didn't process all kinds of chunks
* they were somewhat obscure

The proposed implementation to extract data from `CXCompletionString` is much shorter, simpler and easier to understand.
Completion chunks, display string and other completion data are all done at the same time and I have no way to split them up.
The related changes are all in these three files:
* cpp/ycm/ClangCompleter/CompletionData.h
* cpp/ycm/ClangCompleter/CompletionData.cpp
* cpp/ycm/ClangCompleter/ClangHelpers.cpp

----

Meanwhile, API changes accordingly and there are actually more than renaming. It doesn't make much sense to do renaming in a separate PR.

```
kind            ==> kind (unchanged)
extra_data      ==> extra_data (unchanged)
extra_menu_info ==> result_type (renamed)
menu_text       ==> display_string (renamed)
insertion_text  ==> (deleted)
detailed_info   ==> (deleted)
                ==> typed_string (added)
                ==> completion_chunks (added)
                ==> doc_string (added)
```

So you will see changes related to `BuildCompletionChunk` `BuildCompletionData`, and `BuildSimpleCompletionData`.
* ycmd/responses.py

`BuildCompletionChunk` is just like its name implies.
`BuildCompletionData` is changed to receive array of chunks as arguments and `BuildSimpleCompletionData` receives single non-placeholder string to handle simple occasions conveniently.

Now it is done.

----

The following changes are all to adopt the new names, to get code compiling or to get tests passing. Nothing important.

* cpp/ycm/ycm_core.cpp
* cpp/ycm/CandidateRepository.cpp
* cpp/ycm/tests/ClangCompleter/ClangCompleter_test.cpp
* ycmd/completers/completer.py
* ycmd/completers/all/identifier_completer.py
* ycmd/completers/cpp/clang_completer.py
* ycmd/completers/cs/cs_completer.py
* ycmd/completers/general/filename_completer.py
* ycmd/completers/general/ultisnips_completer.py
* ycmd/completers/go/gocode_completer.py
* ycmd/completers/python/jedi_completer.py
* ycmd/completers/general/tests/filename_completer_test.py
* ycmd/completers/go/tests/gocode_completer_test.py
* ycmd/tests/get_completions_test.py

Valloric:
@Qusic It seems like you have solid reasons for wanting to have the refactoring done before implementing the completion chunk stuff. Could you then send an PR with just the refactoring changes first? Those can then be reviewed and merged in before any of the API work that would have to depend on that.

If putting that together is truly the only thing that makes sense, then do that. But I can also see what appear to be logic changes in your PR (like removing the deduplication in `ToCompletionDataVector`) and it's hard for me to tease out which changes are necessary and which aren't. Note that while I might be OK with changing the API of ycmd and then updating all the clients, I am _not_ ok with changing the user experience of YCM (like what removing the deduplication seems to be doing, at a glance) without an _exceptionally_ good reason.

So we can add more stuff to the response and change the API between the client and server, but any change to user-facing functionality is to be avoided (without good reasoning).

In general, it's much easier to review PRs and analyze the potential fallout of them _the smaller they are_. So I again encourage you to split stuff out as much as you can if at all possible. Since the changes you want to make are extensive, if this blows up in some way, figuring out what went wrong will too benefit from smaller PRs that will need to be re-reviewed. 

So how are you planning to proceed?

Qusic:
I am quite free these days and want to continue working on this.

One question about removing the deduplication is which one to keep. Is it the one without any optional parameters or the one with all optional parameters?

I am going to split it into three PRs in this order:
1. (c) (d) (f)
2. (e) (g)
3. Refactor clang completer and adopt the `completion_chunks` field

    (a) kind            ==> kind (unchanged)
    (b) extra_data      ==> extra_data (unchanged)
    (c) extra_menu_info ==> result_type (renamed)
    (d) menu_text       ==> display_string (renamed)
    (e) insertion_text  ==> completion_chunks (converted)
    (f) detailed_info   ==> doc_string (merged)
    (g)                 ==> typed_string (added)

vheon:
> I am going to split it into three PRs in this order:
1. (c) (d) (f)
2. (e) (g)
3. Refactor clang completer and adopt the completion_chunks field

Look like a good plan :+1:

PR #138, opened 2015-05-21
Add kak-ycmd Kakoune client to README

mawww:


mawww:
Also, not sure it is needed for such a small change, but I did sign the Google CLA.

Valloric:
Thanks!

PR #140, opened 2015-05-22
Fix Gocode path on Windows

micbou:
This pull request fixes the default Gocode path on Windows by adding the `.exe` extension and using `os.path.join` for the full path. Otherwise, python does not find `gocode.exe` and there is no completion (and no exception raised).

There are other changes:
  - the logic in `hook.py` was moved to the `GoCodeCompleter` class. This avoids calling `FindGoCodeBinary` two times. An exception is raised when no binary is found (like in the `CsCompleter` class). 
  - `FindGoCodeBinary` is now a method of the `GoCodeCompleter` class since not used elsewhere. `os.path.exists` was replaced by `os.path.isfile`.
  - Add test for `FindGoCodeBinary` method (it does not test finding `gocode` binary in the `PATH`).
  - Fix code style and clean-up code.

There are still problems on Windows but it will be for another PR.

CLA signed.

### DIFF
--- a/ycmd/completers/go/gocode_completer.py
+++ b/ycmd/completers/go/gocode_completer.py
@@ -27,51 +27,33 @@ from ycmd import utils
 from ycmd.completers.completer import Completer
 
 GO_FILETYPES = set( [ 'go' ] )
-COMPLETION_ERROR_MESSAGE = "Gocode shell call failed."
-PARSE_ERROR_MESSAGE = "Gocode returned invalid JSON response."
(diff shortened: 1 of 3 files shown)


Valloric:
@ekfriis I'd appreciate you taking a look at this.

Valloric:
LGTM from me. Let's see what @ekfriis has to say; he wrote the original gocode completer.

ekfriis:
LGTM, thanks a lot!

micbou:
Forgot to convert some strings.

Valloric:
Thanks for the PR!

PR #142, opened 2015-05-28
Use CREATE_NO_WINDOW flag on Windows instead of shell = True

micbou:
In the C# completer, the `shell` option is set to `True` when starting the `OmniSharp` server to disable the creation of a console window on Windows. This is working but there are some issues:
 - this option can be a [security hazard](https://docs.python.org/2/library/subprocess.html#frequently-used-arguments).
 - this is useless on other platforms.
 - the creation of a console window is not specific to the `OmniSharp` server.

This PR fixes these issues by using the [CREATE_NO_WINDOW](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684863(v=vs.85).aspx) flag on Windows directly in the `SafePopen` function. No more pesky console window when starting any process!

A side-effect of this change is that the `pythonw` executable is no longer needed on Windows. Strangely enough, this makes the start of `gvim` smoother (no loading cursor when starting it).

CLA signed.

### DIFF
--- a/ycmd/utils.py
+++ b/ycmd/utils.py
@@ -31,6 +31,9 @@ import collections
 
 WIN_PYTHON27_PATH = 'C:\python27\pythonw.exe'
 WIN_PYTHON26_PATH = 'C:\python26\pythonw.exe'
+# Creation flag to disable creating a console window on Windows. See
+# https://msdn.microsoft.com/en-us/library/windows/desktop/ms684863.aspx
+CREATE_NO_WINDOW = 0x08000000
 
 
 def SanitizeQuery( query ):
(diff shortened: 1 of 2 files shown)


ycmd/utils.py
```diff
@@ -29,8 +29,10 @@
 import subprocess
 import collections
 
-WIN_PYTHON27_PATH = 'C:\python27\pythonw.exe'
-WIN_PYTHON26_PATH = 'C:\python26\pythonw.exe'
+WIN_PYTHON27_PATH = 'C:\python27\python.exe'
+WIN_PYTHON26_PATH = 'C:\python26\python.exe'
+# Creation flag to disable creating a console window on Windows
```
vheon: Wouldn't be better to link https://msdn.microsoft.com/en-us/library/windows/desktop/ms684863(v=vs.85).aspx in the comment? so one can figure out where this flag came from without looking at the issue tracker.

micbou: Done.

Valloric:
This looks really nice! Thanks for the PR!

PR #143, opened 2015-05-28
Support multiple C# solutions

mispencer:
Omnisharp only supports a single solution per instance. Therefore, to support multiple solutions, we have to have multiple instances of Omnisharp. We only need one instance per solution and we can tell which solution a file belongs to though FindSolutionPath, so for each request, lookup the solution, find a corresponding instance of Omnisharp, spawning it first if needed, and send it the request.

I implemented this by splitting the solution-specific parts of ``CsharpCompleter`` into a ``CsharpSolutionCompleter``. There is one instance of ``CsharpCompleter`` which manages a set of ``CsharpSolutionCompleter``, one per solution. The main section of ycmd calls ``CsharpCompleter``, which examines each request, determines the correct ``CsharpSolutionCompleter``, creating it if needed, and proxies the request to it.

It would probably be better to have this support in Omnisharp itself, but implementing this in Omnisharp seemed more difficult. Furthermore, Omnisharp is currently being completely rewritten, thus spending effort on the existing version seemed unwise, and the new version is not yet stable or feature complete.

I have signed the CLA.

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -71,40 +71,11 @@ class CsharpCompleter( Completer ):
   A Completer that uses the Omnisharp server as completion engine.
   """
 
-  subcommands = {
-    'StartServer': ( lambda self, request_data: self._StartServer(
-        request_data ) ),
-    'StopServer': ( lambda self, request_data: self._StopServer() ),
-    'RestartServer': ( lambda self, request_data: self._RestartServer(
-        request_data ) ),
-    'ReloadSolution': ( lambda self, request_data: self._ReloadSolution() ),
-    'ServerRunning': ( lambda self, request_data: self.ServerIsRunning() ),
-    'ServerReady': ( lambda self, request_data: self.ServerIsReady() ),
-    'ServerTerminated': ( lambda self, request_data:
-        self.ServerTerminated() ),
-    'SolutionFile': ( lambda self, request_data: self._SolutionFile() ),
-    'GoToDefinition': ( lambda self, request_data: self._GoToDefinition(
-        request_data ) ),
-    'GoToDeclaration': ( lambda self, request_data: self._GoToDefinition(
-        request_data ) ),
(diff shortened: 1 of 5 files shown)


ycmd/completers/cs/cs_completer.py
```diff
@@ -125,7 +97,18 @@ def SupportedFiletypes( self ):
     return [ 'cs' ]
 
 
+  def _GetSubCompleter( self, request_data ):
```
Valloric: Seems like this should be called `_GetSolutionCompleter`.

Valloric:
If only more PRs were like this... :) Excellent explanation of why this change is necessary & how it works, code that's well-written and very readable plus tests for the new feature. Very easy to review, thank you!

I added a minor optional bikeshedding comment, but other than that, this seems good to go if Travis ends up liking it.

ycmd/tests/diagnostics_test.py
```diff
@@ -182,7 +182,60 @@ def Diagnostics_CsCompleter_ZeroBasedLineAndColumn_test():
...
+                               event_name = 'FileReadyToParse',
+                               filetype = 'cs',
+                               contents = contents )
+
+    results = app.post_json( '/event_notification', event_data ).json
+
+    assert_that( results,
+                contains(
+                    has_entries( {
+                        'kind': equal_to( 'ERROR'),
```
micbou: Missing a space before `)`.

ycmd/tests/diagnostics_test.py
```diff
@@ -182,7 +182,60 @@ def Diagnostics_CsCompleter_ZeroBasedLineAndColumn_test():
...
+                        'text': contains_string(
+                            "Unexpected symbol `}'', expecting identifier" ),
+                        'location': has_entries( {
+                        'line_num': line,
+                        'column_num': 2
```
micbou: Indentation not the same as below.

mispencer:
I made all the requested changes.

Valloric:
Thanks for the PR!

PR #144, opened 2015-05-29
Handle serializing objects without __dict__

mispencer:
I encountered this with a network exception. Instead of receiving the
network exception, I received a AttributeError that __dict__ did not
exist in obj.

Instead of crashing, fallback to str in this case.

I have signed the CLA.


Valloric:
Thanks!

PR #145, opened 2015-05-29
Faster solution finder tests

mispencer:
Previously, starting the Omnisharp server generating the solution file,
and the SolutionFile method/command merely returned it. Now, the
SolutionFile command can generate it on its own. Since starting
Omnisharp is slow, no longer required and not the point of this test,
omit doing it.

I have signed the CLA.


Valloric:
Thanks!

mispencer:
This is ready.

PR #146, opened 2015-05-31
Add possibility to /ready for specific subserver

vheon:
I'm working on using my http wrapper for jedi ([JediHTTP](/vheon/JediHTTP)). While working on it I needed something like `WaitUntilJediHTTPIsReady` without depending on OmniSharpServer, so the `include_subservers` that currently works for `/ready` wasn't useful. This will simplify the testing of the readiness of server-backed semantic engine.

### DIFF
--- a/ycmd/handlers.py
+++ b/ycmd/handlers.py
@@ -114,10 +114,17 @@ def GetHealthy():
 @app.get( '/ready' )
 def GetReady():
   _logger.info( 'Received ready request' )
+  if request.query.subserver:
+    filetype = request.query.subserver
+    return _JsonResponse( _IsSubserverReady( filetype ) )
   if request.query.include_subservers:
-    cs_completer = _server_state.GetFiletypeCompleter( ['cs'] )
-    return _JsonResponse( cs_completer.ServerIsReady() )
+    return _JsonResponse( _IsSubserverReady( 'cs' ) )
   return _JsonResponse( True )
+
+
+def _IsSubserverReady( filetype ):
+  completer = _server_state.GetFiletypeCompleter( [filetype] )
+  return completer.ServerIsReady()
 
 
 @app.post( '/semantic_completion_available' )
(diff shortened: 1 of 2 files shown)


mispencer:
What is the use case of ``include_subservers``? As far as I know, it's only used in a single place in the tests. It always bothered me a little - it seems overly complex for a single usage. It would be more straightforward to remove ``include_subservers``, and, in the tests, directly call the C# completer command that ``include_subservers`` wraps.

vheon:
I don't know if there was any plans for `include_subservers` so I leaved it there. If the only use case for `include_subservers` is for testing purpose, then I could remove it and we could just use the `subserver` that I proposed.

> It would be more straightforward to remove include_subservers, and, in the tests, directly call the C# completer command that include_subservers wraps.

 If not through `/ready` how would you call the C# completer directly?

mispencer:
Though ``run_completer_command``, as the [function is doing already for ``ServerTerminated``](https://github.com/vheon/ycmd/blob/feature-ready-specific-subserver/ycmd/tests/test_utils.py#L89)

vheon:
I thought of using that approach, but that means I have to create a subcommand for `ServerReady` which from the user point of view is not relevant. It is true that the `/ready` with `subserver` query is public interface as well but is less visible for a user. Thoughts?

mispencer:
You're probably right. I assume you're planning on using this for Jedi? If so, my dislike for this parameter to ``/ready`` based on its limited usage would be lessened.

Also, I noticed while researching this that ``include_subservers`` is used in the example client. I assume that should be updated as well.

vheon:
Yes, I've proposed this changes while working on the http server for jedi

> Also, I noticed while researching this that include_subservers is used in the example client. I assume that should be updated as well.

When my work on the jedi server is done the `/ready` with `include_subservers` will check for readiness of jedi server as well as omnisharp. As it is right now I can see why that seems moot.

vheon:
Forgot to say that I already sign the CLA

Valloric:
Ugh, forgot about this PR. Sorry! Feel free to ping a PR thread if I don't respond in a few days.

Anyway, this looks good. I'm happy that your change is backwards compatible with the old API; the less breakage we introduce the better.

@homu r+

@mispencer While we do use the `/ready` stuff with subservers in tests, I can easily see a client wanting to know is just ycmd ready or is the whole ycmd "cluster" ready. So it seems generally useful. Also at this point unknown clients (beyond YouCompleteMe) could be depending on this in unknown ways, so keeping it seems sensible. It's not as if it's a big maintenance burden.

PR #147, opened 2015-05-31
Cache Clang download

### LINKED ISSUE #48: Cache Clang download
Everytime there's a change in ycmd's c++ files, and it's necessary to rebuilt/reinstall through `./install.sh --clang-completer`, it's necessary to wait for the clang download, even though it's the same version, since current approach downloads it to `/tmp` for building.

Suggestion is to let the compressed download in the plugin directory and building on `/tmp`, and download anew only when absent or newer version adopted.

micbou:
This fixes #48. Archives are downloaded in the `clang_archives` folder.

I bumped the cmake version to 2.8.7 cause the function to get the hash of a file is only available since 2.8.7 (fortunately, this is the version used by Travis).

I also replaced the `MD5` hash function by the more secure `SHA256` (and updated the hashes).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@
 *.la
 *.a
 
+# Clang archives
+clang_archives
+
 # CMake
 CMakeCache.txt
 CMakeFiles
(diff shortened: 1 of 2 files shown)


Valloric:
Thanks for sending a PR!

As per [CONTRIBUTING.md](https://github.com/Valloric/YouCompleteMe/blob/master/CONTRIBUTING.md), please sign the [Google CLA](https://developers.google.com/open-source/cla/individual).

Other than that, this looks good to merge.

micbou:
Oops, I forgot to say that I already signed it.

micbou:
By the way, Clang 3.6.1 was released a few days ago. I could send a PR to use the new version.

Valloric:
Feel free to send another PR updating to 3.6.1.

Issue #108, opened 2015-04-11
CMake minimum version


vheon:
From https://github.com/Valloric/YouCompleteMe/issues/1439#issuecomment-91825098

@asenac
> I think this should be reported in ycmd project to either set the minimum cmake version required to 2.8.12, currently it is set to 8.2, or use PATH component instead to support versions between 2.8 and 2.8.12.

shirro:
Wheezy has cmake version 2.8.9 and this change stopped me building the latest YCM. I installed with TRAVIS=1 ./install.sh

Is there a downside to doing that?

Valloric:
@shirro If you managed to compile it, then you're fine.

PR #148, opened 2015-06-01
Make minimum required cmake version as 2.8.11

tokuhirom:
Because CentOS 7 contains 2.8.11. This modification makes easier to install on CentOS 7.
And I guess this file doesn't depend on 2.8.12 feature.


Valloric:
Digging around the commit history, seems that 2.8.12 is required: https://github.com/Valloric/ycmd/issues/108

PR #149, opened 2015-06-01
Add TSServer completer

icholy:
TypeScript completion support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ install:
   - sudo apt-get -qq update
   - sudo apt-get -qq install monodevelop-current
   - sudo apt-get -qq install g++-4.8
+  - npm install -g typescript
   - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 90
   - MONO_PREFIX=/opt/monodevelop
   - export DYLD_LIBRARY_FALLBACK_PATH=$MONO_PREFIX/lib:$DYLD_LIBRARY_FALLBACK_PATH
--- a/ycmd/completers/typescript/hook.py
+++ b/ycmd/completers/typescript/hook.py
@@ -1,0 +1,23 @@
+#!/usr/bin/env python
+#
+# Copyright (C) 2015  Google Inc.
+#
+# This file is part of YouCompleteMe.
+#
+# YouCompleteMe is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# YouCompleteMe is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with YouCompleteMe.  If not, see <http://www.gnu.org/licenses/>.
+
+from ycmd.completers.typescript.typescript_completer import TypeScriptCompleter
+
+def GetCompleter( user_options ):
+  return TypeScriptCompleter( user_options )
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -1,0 +1,199 @@
+#!/usr/bin/env python
+#
+# Copyright (C) 2015 Google Inc.
+#
+# This file is part of YouCompleteMe.
+#
+# YouCompleteMe is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# YouCompleteMe is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with YouCompleteMe.  If not, see <http://www.gnu.org/licenses/>.
+
+import json
+import logging
+import os
+import subprocess
+
+from tempfile import NamedTemporaryFile
+
+from ycmd import responses
+from ycmd import utils
+from ycmd.completers.completer import Completer
+
+_logger = logging.getLogger( __name__ )
+
+
+class TypeScriptCompleter( Completer ):
+  """
+  Completer for TypeScript.
+
+  It uses TSServer which is bundled with TypeScript 1.5
+
+  See the protocol here:
+  https://github.com/Microsoft/TypeScript/blob/2cb0dfd99dc2896958b75e44303d8a7a32e5dc33/src/server/protocol.d.ts
+  """
+
+  def __init__( self, user_options ):
+    super( TypeScriptCompleter, self ).__init__( user_options )
+
+    binarypath = utils.PathToFirstExistingExecutable( [ 'tsserver' ] )
+    if not binarypath:
+      raise RuntimeError(
+        'tsserver not found. TypeScript 1.5 or higher is required' )
+
+    # Each request sent to tsserver must have a sequence id.
+    # Responses contain the id sent in the corresponding request.
+    self._sequenceid = 0
+    self._tsserver_handle = utils.SafePopen( binarypath,
+      stdout = subprocess.PIPE,
+      stdin  = subprocess.PIPE
+    )
+
+    _logger.info( 'Enabling typescript completion' )
+
+  def _SendRequest( self, command, arguments=None ):
+    """Send a request message to TSServer."""
+
+    seq = self._sequenceid
+    self._sequenceid += 1
+    request = {
+      'seq':     seq,
+      'type':    'request',
+      'command': command
+    }
+    if arguments:
+      request[ 'arguments' ] = arguments
+    self._tsserver_handle.stdin.write( json.dumps( request ) )
+    self._tsserver_handle.stdin.write( "\n" )
+    return seq
+
+  def _ReadResponse( self, expected_seq ):
+    """Read a response message from TSServer."""
+
+    # The headers are pretty similar to HTTP.
+    # At the time of writing, 'Content-Length' is the only supplied header.
+    headers = {}
+    while True:
+      headerline = self._tsserver_handle.stdout.readline().strip()
+      if not headerline:
+        break
+      key, value = headerline.split( ':', 1 )
+      headers[ key.strip() ] = value.strip()
+
+    # The response message is a JSON object which comes back on one line.
+    # Since this might change in the future, we use the 'Content-Length' header.
+    if 'Content-Length' not in headers:
+      raise RuntimeError( "Missing 'Content-Length' header" )
+    contentlength = int( headers[ 'Content-Length' ] )
+    message = json.loads( self._tsserver_handle.stdout.read( contentlength ) )
+
+    msgtype = message[ 'type' ]
+    if msgtype == 'event':
+      self._HandleEvent( message )
+      return self._ReadResponse()
+
+    if msgtype != 'response':
+      raise RuntimeError( 'Unsuported message type {0}'.format( msgtype ) )
+    if int( message[ 'request_seq' ] ) != expected_seq:
+      raise RuntimeError( 'Request sequence mismatch' )
+    if not message[ 'success' ]:
+      raise RuntimeError( message[ 'message' ] )
+
+    return message
+
+  def _HandleEvent( self, event ):
+    """Handle event message from TSServer."""
+
+    # We ignore events for now since we don't have a use for them.
+    eventname = event[ 'event' ]
+    _logger.info( 'Recieved {0} event from tsserver'.format( eventname ) )
+
+  def _Reload( self, request_data ):
+    """
+    Syncronize TSServer's view of the file to
+    the contents of the unsaved buffer.
+    """
+
+    filename = request_data[ 'filepath' ]
+    contents = request_data[ 'file_data' ][ filename ][ 'contents' ]
+    tmpfile = NamedTemporaryFile( delete=False )
+    tmpfile.write( contents )
+    tmpfile.close()
+    seq = self._SendRequest( 'reload', {
+      'file':    filename,
+      'tmpfile': tmpfile.name
+    })
+    self._ReadResponse( seq )
+    os.unlink( tmpfile.name )
+
+  def SupportedFiletypes( self ):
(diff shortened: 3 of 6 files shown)


vheon:
Where is `tsserver`? Every semantic engine usually should be bundled with ycmd itself and the user should decide if build it with some option from the `build.py` script. There are no test for this, and as per [CONTRIBUTING.md](https://github.com/Valloric/ycmd/blob/master/CONTRIBUTING.md) you have to sign the [CLA](https://developers.google.com/open-source/cla/individual) and state here that you did.

vheon:
I searched for `tsserver` and it seems to be bundled with the typescript environment itself, is that right?

icholy:
@vheon yes its going to be shipped with TypeScript 1.5.

icholy:
I've signed the CLA, and I'll look into adding some tests.

Valloric:
As @vheon said, we always bundle the semantic engine with ycmd (as a git submodule) so that users have everything set up and good to go. AFAIR there are no current exceptions to this; there would have to be some really good reasons to exempt that here.

If it's something that is commonly distributed with typescript installs and ~everyone who writes TS would have it installed, then we could possibly not bundle it. But you'll have to find a way to make it available to Travis CI so that the end-to-end tests that you're writing now can execute there.

icholy:
@Valloric the tsserver code is part of the primary TypeScript repository. Cloning it takes quite a while so your CI and install time would take a pretty big hit. With the 1.5 release everyone who has the typescript compiler will have tsserver as well. Your call.

vheon:
> Cloning it takes quite a while so your CI and install time would take a pretty big hit.

Isn't possible to install a prebundled version through `npm` or `apt-get`?? If you really have to use the github repo (https://github.com/Microsoft/TypeScript) just clone it with `--depth=1` and it is just 10Mb of data, so not so much.

> With the 1.5 release everyone who has the typescript compiler will have tsserver as well.

And everyone how doesn't? I mean if a user have TypeScript 1.4 would mean that the `tsserver` binary wouldn't be preinstalled? Than I think you should make sure that the user at least gets a message for that.

vheon:
BTW the tests are currently not passing because you have some unused variable and `flak8` isn't happy about that.

ycmd/completers/typescript/tsserver_completer.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+class TSServerCompleter( Completer ):
+
+  def __init__( self, user_options ):
+    super( TSServerCompleter, self ).__init__( user_options )
+
+    binarypath = utils.PathToFirstExistingExecutable( [ 'tsserver' ] )
+    if not binarypath:
+        raise RuntimeError( 'tsserver not found. TypeScript 1.5 or higher is required' )
```
icholy: @vheon this error isn't showing up for me in Vim. I'm thinking I should be using `responses.BuildExceptionResponse` but I can't figure out how to send that from the constructor.

vheon: If I'm not mistaken `RuntimeError` should be catched by `ycmd`. @Valloric could shed some light on this.

icholy:
Since this is just glue between ycmd and tsserver I'm not too sure about how to test it.

vheon:
you can look [ycmd/tests/get_completions_test.py](https://github.com/Valloric/ycmd/blob/master/ycmd/tests/get_completions_test.py) for examples.

ycmd/completers/typescript/tsserver_completer.py
```diff
@@ -0,0 +1,197 @@
...
+    message = json.loads( self.proc.stdout.read( contentlength ) )
+
+    # The tsserver can sends events, I don't know if they're useful
+    msgtype = message[ 'type' ]
+    if msgtype == 'event':
+      self._HandleEvent( message )
+      return self._ReadResponse()
+
+    if msgtype != 'response':
+      raise RuntimeError( 'Unsuported message type {}'.format( msgtype ) )
```
vheon: If I'm not mistaken the minimal python supported is 2.6 so positional arguments are required:

```python
raise RuntimeError( 'Unsuported message type {0}'.format( msgtype ) )
```

icholy:
@vheon I think it's ready.

vheon:
The code looks fine to me. I thought that we were going to assume that the user had to have the `tsserver` installed for this to work, but now I see that we're cloning the `TypeScript` repo by default.

icholy:
oh, I misunderstood. Should I remove it?

vheon:
I'm just asking :) Even thought I'm starting to worry about the amount of submodule that ycmd bring to the user and that the user might not won't (like this one for me :P).

.travis.yml
```diff
@@ -11,6 +11,8 @@ install:
   - sudo apt-get -qq update
   - sudo apt-get -qq install monodevelop-current
   - sudo apt-get -qq install g++-4.8
+  - sudo apt-get -qq install node
+  - sudo npm install -g typescript
```
icholy: @vheon @Valloric the version of `node` installed doesn't come with `npm`. Ideas?

micbou: From what I read, `npm` is already available. Try without `sudo`. Install `typescript` locally and add `tsserver` to the `PATH`.

icholy: Hopefully one of those builds works :stuck_out_tongue:

micbou: Nice. I'm betting on the last one!

icholy: Looks like you're right, thanks!

Valloric:
@icholy You made a strong case that we don't need to bundle tsserver since everyone who has the TS compiler after 1.5 will have tsserver as well. So no need to bundle it then.

ycmd/completers/typescript/tsserver_completer.py
```diff
@@ -0,0 +1,199 @@
...
+from tempfile import NamedTemporaryFile
+
+from ycmd import responses
+from ycmd import utils
+from ycmd.completers.completer import Completer
+
+_logger = logging.getLogger( __name__ )
+
+
+class TSServerCompleter( Completer ):
```
Valloric: Call it TypeScriptCompleter. The fact that it uses tsserver is an implementation detail. We violate this in other parts of ycmd (jedi_completer for example), but let's not add more cases of it.

ycmd/completers/typescript/tsserver_completer.py
```diff
@@ -0,0 +1,199 @@
...
+  See the protocol here:
+  https://github.com/Microsoft/TypeScript/blob/2cb0dfd99dc2896958b75e44303d8a7a32e5dc33/src/server/protocol.d.ts
+  """
+
+  def __init__( self, user_options ):
+    super( TSServerCompleter, self ).__init__( user_options )
+
+    binarypath = utils.PathToFirstExistingExecutable( [ 'tsserver' ] )
+    if not binarypath:
+      raise RuntimeError( 'tsserver not found. TypeScript 1.5 or higher is required' )
```
Valloric: It's hard to see, but is this line 80 columns? Seems like it might not be.

ycmd/completers/typescript/tsserver_completer.py
```diff
@@ -0,0 +1,199 @@
...
+
+    self.seq = 0
```
Valloric: Use something more descriptive than `seq`. At a glance, I don't know what this var is for. Also consider adding a comment above it describing what's it used for.

Valloric: Also, class private vars should start with `_`.

oblitum:
I'm with @vheon. IMO the current model is flawled but as @Valloric once said, this ship has already sailed... I think it's one of the reasons, if not the main one, that frequent users in the #vim channel, which I consider better vim users than me in general, are not drawn to YCM. I guess it's not considered modular in some sense.

ycmd/completers/typescript/tsserver_completer.py
```diff
@@ -0,0 +1,199 @@
...
+    self.proc = utils.SafePopen( binarypath,
```
Valloric: `proc` is also not a good idea; use `_tsserver_handle`

ycmd/completers/typescript/tsserver_completer.py
```diff
@@ -0,0 +1,199 @@
...
+      stdout = subprocess.PIPE,
+      stdin  = subprocess.PIPE
+    )
+
+    _logger.info( 'Enabling typescript completion' )
+
+  def _SendRequest( self, command, arguments=None ):
+    """ Send a request message to TSServer """
```
Valloric: No need for whitespace around the comment inside the quotes. Also use full sentences with periods.

ycmd/completers/typescript/tsserver_completer.py
```diff
@@ -0,0 +1,199 @@
...
+
+  def _ReadResponse( self, seq ):
+    """ Read a response message from TSServer """
+
+    # The headers are pretty similar to HTTP
+    # 'Content-Length' is the only provided header at the time of writing
+    headers = {}
+    while True:
+      headerline = self.proc.stdout.readline().strip()
+      if headerline == "":
```
Valloric: Use just `if headerline`; you don't need to compare with empty string.

ycmd/completers/typescript/tsserver_completer.py
```diff
@@ -0,0 +1,199 @@
...
+        break
+      [ key, value ] = headerline.split( ':', 1 )
```
Valloric: `key, value = ...` should work; no need for the brackets around them.

ycmd/completers/typescript/tsserver_completer.py
```diff
@@ -0,0 +1,199 @@
...
+      headers[ key.strip() ] = value.strip()
+
+    # The response message is a json object which always seems to be on one line.
+    # Since this might change in the future, use the 'Content-Length' header.
+    if 'Content-Length' not in headers:
+      raise RuntimeError( "Missing 'Content-Length' header" )
+    contentlength = int( headers[ 'Content-Length' ] )
+    message = json.loads( self.proc.stdout.read( contentlength ) )
+
+    # The tsserver can sends events, I don't know if they're useful
```
Valloric: Instead of saying "I don't know if they're useful", just put "we ignore events for now since we don't have a use for them".

ycmd/completers/typescript/tsserver_completer.py
```diff
@@ -0,0 +1,199 @@
...
+
+  def _Reload( self, request_data ):
+    """
+    Syncronize TSServer's view of the file to
+    the contents of the unsaved buffer
+    """
+
+    filename = request_data[ 'filepath' ]
+    contents = request_data[ 'file_data' ][ filename ][ 'contents' ]
+    tmpfile = NamedTemporaryFile( delete=False )
```
Valloric: Ugh, please don't tell me tsserver is not smart enough to be able to take the actual contents of the buffer but instead always takes a filename? That's terrible... having to round-trip through the disk kills performance.

Where can we complain about this to tsserver devs?

ycmd/completers/typescript/tsserver_completer.py
```diff
@@ -0,0 +1,199 @@
...
+    return [ 'GoToDefinition' ]
+
+  def OnUserCommand( self, arguments, request_data ):
+    command = arguments[ 0 ]
+    if command == 'GoToDefinition':
+      return self._GoToDefinition( request_data )
+    raise ValueError( self.UserCommandsHelpMessage() )
+
+  def _GoToDefinition( self, request_data ):
+
```
Valloric: Why the empty line here?

ycmd/completers/typescript/tsserver_completer.py
```diff
@@ -0,0 +1,199 @@
...
+    self._Reload( request_data )
+    seq = self._SendRequest( 'definition', {
+      'file':   request_data[ 'filepath' ],
+      'line':   request_data[ 'line_num' ],
+      'offset': request_data[ 'column_num' ]
+    })
+
+    filespans = self._ReadResponse( seq )[ 'body' ]
+    if len( filespans ) is 0:
```
Valloric: `if filespans` works the same, no need for calling `len()`.

Valloric:
@oblitum I think there's a balance to be had. In general, users don't care about how long it takes to check out a copy of YCM, but they do care about not having to go hunt down dependencies. The policy in my head has always been "bundle it if it's not too much of a hassle"; note that libclang itself is not bundled, but fetched. A long time ago before YCM was ever released to the public, there used to be a copy of all of LLVM + clang in the repo and it was built every time you built the YCM libs.

That got old _very_ quickly.

vheon:
> this ship has already sailed... 

I still don't understand why this ships has already sailed. Correct me if I'm wrong but currently we are in a situation where If I don't explicitly build the omnisharp server with `--omnisharp-completer` for example ycmd still say that a semantic engine is available for `cs` when it is not. From my point of view would be more sensible to make it available only if the flag is provided and at that point would be the right moment for cloning the submodule instead of doing it anyway. The only exception to this, right now, is for python where there is no need to build anything. So I'm not saying that bundle all together is a bad idea, I'm saying to not clone it until is really necessary.

oblitum:
Indeed, one thing is that there many developers out there that are interested solely in their main language of use, so the first impression this dev has about YCM is that it's bloated: download is big, to get stuff he'll never use.

Related to this, I guess UltiSnip had changed its curse, formally it included all the snippets in the UltiSnip repo, but now you have just the UltiSnip core and you can get a bundle of snipets elsewhere from another repo, or specialized from another users taste, but that follow the same UltiSnip standard, this model is much nicer.

oblitum:
Another issue with the current model besides the user experience is the supporting experience. In my view it seems better to have independent addons to ycm/ycmd being develop in their own repos with their own issues and their own forks than waiting maintainers of code sections in a single repo. As of now I see @Valloric as the owner, will not be handling C#, TypeScript or whatever that gets added with the same care as he can give to the completion he uses daily, so he needs to somewhat blind trust maintainers of the code-sections/ycmd-addons and be in need of them when issues on such addons shows up.

What I mean is that it seems easier of such addons to take a life of its own when developed apart (following a protocol/convention/api) than when bundled as a code section.

Valloric:
I appreciate and value both of your thoughts, but this PR is not the place to have this discussion.

ycmd/completers/typescript/tsserver_completer.py
icholy: Unless you have a really crappy filesystem with no caching, it's not really a round trip to the disk.

icholy:
@Valloric squashed and rebasd

Valloric:
This looks good to merge. Thanks for the PR!

icholy:
:tada: wohoo

Valloric:
We need to update the docs in YCM though so that users know this exists and how to use it.

icholy:
I think we should wait until TSServer is out of beta.

Valloric:
Sure, let's wait then.

PR #151, opened 2015-06-02
C# StopServer improvements

mispencer:
Sometimes ``StopServer`` fails. For example, if the server died,
it will fail with connection refused error.  Ignore error sending the
``StopServer`` requests, and retrying sending it every time we wait.
Similarly, if the server hasn't been started, logs files will be ``None``,
which will cause an error. Only delete log files if they exist.

This change made ``StopServer`` slower, thus there is a second change.

If the server is not running, both ``/stopserver`` and ``/checkalivestatus``
will sometime spin for the default network timeout of 2 seconds. This
normally doesn't matter, but in the tests, we spin up and shutdown
Omnisharp many times, and this accordingly adds many seconds to the
test suite run time. Similarly, the sleep time between still-running
checks in ``StopServer`` is a second, which is much longer than needed.

Add a timeout parameter to ``SentRequest``, and set the timeout to .1
seconds for ``/stopserver`` and .2 seconds for ``/checkalivestatus`` and
friends. Remove the no-longer-used silent parameter at the same time.
Instead calling ``ServerIsRunning`` once with a second wait between
each call of ``StopServer``, call it 10 times with a .1 wait between them.

These changes then required refactoring due to complexity limits.

I have signed the CLA.

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -355,11 +355,13 @@ class CsharpSolutionCompleter:
 
   def _StopServer( self ):
     """ Stop the OmniSharp server """
-    self._GetResponse( '/stopserver' )
-
     # Give OmniSharp 5 seconds to cleanly stop, then kill it if it's still up
     still_running = True
     for _ in range( 5 ):
+      try:
+        self._GetResponse( '/stopserver' )
+      except:
+        pass
       still_running = self.ServerIsRunning()
       if not still_running:
         break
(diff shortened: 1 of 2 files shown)


Valloric:
This looks fine, but Travis doesn't like it for some reason.

mispencer:
Well, that error is during the Omnisharp startup, and this change is to stopping it. The only effect on startup from this change is that ServerIsRunning could timeout, but the error was from ServerTerminated which wasn't affected.

The only thing I can think of is that due to this changes, it could be not stopping the server correctly, so it's running out of memory. It could occur if the ServerIsRunning check timed out, since that would lead to it not killing Omnisharp, and this change did lower that timeout. 

Upon examination of the stop logic, I'm not sure why it's even using ServerIsRunning. It should just call ServerTerminated directly to see if it stopped. That would be less error prone and slightly faster too.

Valloric:
Let's go with this. Hopefully it won't cause any new Travis flakiness. We had a bad bout of it a while ago and it was excruciating. If flakiness comes back, we can roll this back.

PR #150, opened 2015-06-02
Fix hash check when downloading Clang

micbou:
I made a mistake in PR #147. The `SHA256` option should be set without quotes when downloading Clang. I am not sure why it was working on Travis but it did not work on my linux machine. This is certainly the cmake version.

I also removed some trailing spaces.

Sorry for the trouble.

CLA already signed.

PR #152, opened 2015-06-03
Bump Clang version to 3.6.1

micbou:
I also updated the headers in `cpp/llvm/include/clang-c`. No changes in `clang_include` except I could not find `arm_neon.h` in new versions of Clang. Why do we need this file?

Important: we should merge PR #150 before merging this one. It fixes a bug I introduced in PR #147.

CLA signed.

### DIFF
--- a/cpp/llvm/include/clang-c/BuildSystem.h
+++ b/cpp/llvm/include/clang-c/BuildSystem.h
@@ -11,8 +11,8 @@
 |*                                                                            *|
 \*===----------------------------------------------------------------------===*/
 
-#ifndef CLANG_C_BUILD_SYSTEM_H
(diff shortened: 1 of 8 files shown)


mispencer:
This failure was probably caused by https://github.com/Valloric/ycmd/pull/151. https://github.com/Valloric/ycmd/pull/153 should resolve.

Valloric:
This probably also needs to update the clang built-in headers found in `clang_includes`; or were there no changes in that folder between 3.6.0 and 3.6.1?

micbou:
3.6.1 brings no changes in `clang_includes` folder except that I did not find the header `arm_neon.h` in the sources.

Valloric:
Thanks for the PR!

PR #153, opened 2015-06-03
Default timeout should be none

mispencer:
The default timeout in requests is actually None. We should use that to maintain
backward compatibility.

This is probably why https://github.com/Valloric/ycmd/pull/152 failed.

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -515,7 +515,7 @@ class CsharpSolutionCompleter:
     return 'http://localhost:' + str( self._omnisharp_port )
 
 
-  def _GetResponse( self, handler, parameters = {}, timeout = 2 ):
+  def _GetResponse( self, handler, parameters = {}, timeout = None ):
     """ Handle communication with server """
     target = urlparse.urljoin( self._ServerLocation(), handler )
     response = requests.post( target, data = parameters, timeout = timeout )


mispencer:
Should I revert adding a timeout to ServerIsReady and ServerIsRunning too?

micbou:
It should be enough for Travis. From my understanding, `ServerIsReady` is only used in `WaitUntilOmniSharpServerReady` with the `/ready` handler where we tell Travis to try forever. `ServerIsRunning` is not used in tests.

In my opinion, `ServerTerminated` (returning `True` when there is no handle) should be used instead of `ServerIsRunning` before stopping or not the server. It is more reliable to check if the process is up or not in this case. When using `ServerIsRunning` before starting or not the server, we could additionally use `ServerTerminated` to check if the process is still up (server is unresponsive) and kill it if this is the case. A `_KillServer` method could be added for this and also used in `_StopServer`.

Valloric:
> Should I revert adding a timeout to ServerIsReady and ServerIsRunning too?

If that improves backwards compatibility with code that calls it, yes. If code that calls it doesn't care, then no. Up to you.

mispencer:
Well, it makes the check more strict - Omnisharp needs to return within the timeout. I think this is fine; the normal use case is to poll until running/ready, and a false negative would just mean one more poll cycle.

ServerTerminated should probably replacing ServerIsRunning in Shutdown and Restart, but I don't think this pull request is the place for it.

Valloric:
Thanks for the PR!

PR #156, opened 2015-06-06
Fix TypeScript completer on Windows

micbou:
cc @icholy 

On Windows, TSServer executable is a `.cmd` file but `find_executable` works only for `.exe` files so we need our own implementation of the `find_executable` function. This new function tries to find executable with `.exe`, `.bat` and `.cmd` extensions on Windows. Its behavior should be identical to `find_executable` on others platforms.

This fix is not enough. We also need to set the [`universal_newlines` option](https://docs.python.org/2/library/subprocess.html#subprocess.Popen) to `True` (certainly because TSServer only accept `\n` for newlines) and to redirect `stderr` to `stdout` (not sure why but this is caused by the `creationflags = CREATE_NO_WINDOW` option in `utils.SafePopen`) when starting TSServer.

I also added a log when the binary is not found and formatted a 80+ line.

CLA signed.

### DIFF
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -28,6 +28,9 @@ from ycmd import responses
 from ycmd import utils
 from ycmd.completers.completer import Completer
 
+BINARY_NOT_FOUND_MESSAGE = ( 'tsserver not found. '
+                             'TypeScript 1.5 or higher is required' )
+
 _logger = logging.getLogger( __name__ )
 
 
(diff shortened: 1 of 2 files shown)


ycmd/completers/typescript/typescript_completer.py
```diff
@@ -46,16 +49,17 @@ def __init__( self, user_options ):
...
     # Responses contain the id sent in the corresponding request.
     self._sequenceid = 0
     self._tsserver_handle = utils.SafePopen( binarypath,
-      stdout = subprocess.PIPE,
-      stdin  = subprocess.PIPE
-    )
+                                             stdout = subprocess.PIPE,
+                                             stdin = subprocess.PIPE,
+                                             stderr = subprocess.STDOUT,
+                                             universal_newlines = True )
```
vheon: Why is this needed?

micbou: I am not sure but my explanation is that TSServer does not understand Windows newlines `\r\n` so we need to convert them to `\n` and this is what the `universal_newlines` option does.

vheon: Oh ok. Do you think is something that the TSServer devs might want to know? Maybe you should report it to them.

micbou: This is certainly the right thing to do but I need a minimal example. I hope this is not a Python bug...

icholy: My vote goes toward it being a python issue. You'd think Microsoft would make sure it worked on their own OS. I could be wrong though.

micbou: After investigation, the issue is that the value of `Content-Length` in the header is one off. What happens is:
  - first response is read until the character `\r`. 
  - second response reads the left character `\n` and fails.

The option `universal_newlines` solves this because the two characters `\r\n` are converted to one `\n`. Another solution would be to always add one to the contents length.

icholy:
In that case, it sounds like a tsserver bug.

micbou:
I opened an issue on TypeScript repo.

ycmd/utils.py
```diff
@@ -168,12 +169,33 @@ def PathToPythonInterpreter():
...
 def PathToFirstExistingExecutable( executable_name_list ):
   for executable_name in executable_name_list:
-    path = find_executable( executable_name )
+    path = FindExecutable( executable_name )
     if path:
       return path
   return None
 
 
+def FindExecutable( executable ):
```
Valloric: You need a comment explaining why we can't use the old `find_executable`.

micbou:
I added comments for the FindExecutable function and the SafePopen options used in the TypeScript completer.

Valloric:
Thanks!

PR #157, opened 2015-06-07
Introducing GoToInclude command for YcmCompleter

hstefan:
Inspired by Valloric/YouCompleteMe#334, this introduces the GoToInclude command. The main reason for adding this instead of just relying on vim's "gf" or similar commands in other editors, is that in this case the include paths for each file will be accounted for correctly. The implementation is relatively simple, relying on libclang's clang_getIncludedFile to get the file below the cursor.

From the implementations suggested on Valloric/YouCompleteMe#334, I decided to stick with inserting a standalone command. The idea of updating the editor search path doesn't seem ideal, as we wouldn't be able to support all clients with a single solution. The other possibility was baking the functionality in GoToDeclaration, but I don't think this behaviour would be very intuitive.

One issue that I found was in includes with angle brackets, but I *think* that's actually a bug in libclang. Quoting from subcommands_test.py:
>```python
#FIXME: clang_getIncludedFile behaves weirdly with angle brackets, if the cursor
#is *inside* the include argument it may not find a valid location to jump.
#Also, if there is more than one angle bracket include, some of them might work
#depending on their order in the file.```

Lastly, I also made the GoTo command try GoToInclude too. This enables users like myself to bind some key to GoTo and seamlessly use it to jump through the source code. This is different from the confusion that would be introduced by baking the feature inside GoToDeclaration, as in this case the it's just a generic action "go to".

I also confirm that I signed the "Google Contributor License Agreement".


ycmd/tests/subcommands_test.py
```diff
@@ -77,6 +77,78 @@ def RunCompleterCommand_GoTo_Clang_ZeroBasedLineAndColumn_test():
...
+    { 'request' : [3, 1], 'filepath' : 'test-gotoinclude/SomeDirectory/SomeHeader.hpp' },
+    { 'request' : [3, 9], 'filepath' : 'test-gotoinclude/SomeDirectory/SomeHeader.hpp' },
+    { 'request' : [3, 11], 'filepath' : 'test-gotoinclude/SomeDirectory/SomeHeader.hpp' },
+    { 'request' : [4, 1], 'filepath' : 'test-gotoinclude/OtherDirectory/NestedHeader.hpp' },
+    { 'request' : [4, 9], 'filepath' : 'test-gotoinclude/OtherDirectory/NestedHeader.hpp' },
+    { 'request' : [4, 11], 'filepath' : 'test-gotoinclude/OtherDirectory/NestedHeader.hpp' },
+  ]
+
+  for test in tests:
+    yield _RunCompleterCommand_GoToInclude_Clang, \
```
Valloric: Don't use `\`, wrap the expression in parens.

Valloric:
I think a better idea would be to just implement this at the Python layer and not involve libclang. We know what all the include paths are; we're already basically doing most of the work necessary inside [filename_completer](https://github.com/Valloric/ycmd/blob/master/ycmd/completers/general/filename_completer.py) for `#include` completion. It should be possible to extract the relevant common code from that and then build on top of it.

Access to libclang is single-threaded and we'd rather not depend on codepaths in it if possible; changing libclang logic is _hard_.

hstefan:
I can understand the argument for avoiding dependencies on libclang for less complicated tasks, though I'm not sure about the single-threaded issue. Unless I understood somethig wrong, we'd run into GIL anyway if we implement this in the python side.

One approach I can think of for implementing this in python would be extracting relevant utilities from [filename_completer](https://github.com/Valloric/ycmd/tree/master/ycmd/completers/general/filename_completer.py) to something like a include_utils file. With that we can probably implement the GoToInclude function in [clang_completer](https://github.com/Valloric/ycmd/blob/master/ycmd/completers/cpp/clang_completer.py).

Another possiblity would be implementing the feature in [filename_completer](https://github.com/Valloric/ycmd/blob/master/ycmd/completers/general/filename_completer.py), which might allow us to support objetive-c's ```#import```.

Valloric:
> I can understand the argument for avoiding dependencies on libclang for less complicated tasks, though I'm not sure about the single-threaded issue. Unless I understood something wrong, we'd run into GIL anyway if we implement this in the python side.

The GIL is a far more fine-grained lock than the clang access mutex, which will lock access to clang for possibly seconds while the clang operation is ongoing. The GIL won't do that.

> One approach I can think of for implementing this in python would be extracting relevant utilities from filename_completer to something like a include_utils file. With that we can probably implement the GoToInclude function in clang_completer.

That's what I was thinking as well.

Valloric:
What's the state of this PR now? Is it moving forward or should I close this?

hstefan:
I'm not working on it at the moment. I think you can close it and I'll
reopen when I get back to it.

On Mon, Jul 13, 2015 at 3:32 PM, Val Markovic <notifications@github.com>
wrote:

> What's the state of this PR now? Is it moving forward or should I close
> this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Valloric/ycmd/pull/157#issuecomment-121016695>.
>



-- 
[]'s,
Hugo Puhlmann

PR #161, opened 2015-06-18
Improve TypeScript completer

icholy:
* Fix bug with notification handling
* More detailed completion
* Prevent concurrent requests to tsserver

### DIFF
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -110,7 +110,7 @@ class TypeScriptCompleter( Completer ):
     msgtype = message[ 'type' ]
     if msgtype == 'event':
       self._HandleEvent( message )
-      return self._ReadResponse()
+      return self._ReadResponse( expected_seq )
 
     if msgtype != 'response':
       raise RuntimeError( 'Unsuported message type {0}'.format( msgtype ) )
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -188,13 +188,13 @@ class TypeScriptCompleter( Completer ):
                for e in detailed_entries ]
 
   def OnBufferVisit( self, request_data ):
+    filename = request_data[ 'filepath' ]
     with self._lock:
-      filename = request_data[ 'filepath' ]
       self._SendRequest( 'open', { 'file': filename } )
 
   def OnBufferUnload( self, request_data ):
+    filename = request_data[ 'filepath' ]
     with self._lock:
-      filename = request_data[ 'filepath' ]
       self._SendRequest( 'close', { 'file': filename } )
 
   def OnFileReadyToParse( self, request_data ):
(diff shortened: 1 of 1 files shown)


vheon:
I'm talking just after looking the code since I don't use typescript: So if we get 100 entries we do another 100 requests to the TSServer? Is it really going to be more responsive? (Also, remember that in every PR you have to state that you have signed the CLA)

icholy:
If there are **100 or less** entries, then we do **one** additional request to fetch details for them.

icholy:
I'm hitting a bug, don't merge yet.

vheon:
> If there are 100 or less entries, then we do one additional request to fetch details for them.

Oh sorry, I've missed the indentation of the request so I thought that you were doing 100 requests.

Valloric:
I'll hold off on reviewing this until you've resolved the bug you mentioned you found.

icholy:
It seems like sometimes the server doesn't send a `Content-Length` header. When this happens, the sequences get out of whack and no subsequent requests work.

Valloric:
@icholy Which server? ycmd or tsserver? If it's the former, we can fix that; if the latter, it needs to be reported upstream.

icholy:
@Valloric tsserver

Valloric:
@icholy Since you seem to have a firm grasp of the issue at hand, would you mind reporting this upstream so that tsserver devs can fix it?

icholy:
@Valloric yeah I'm just not able to consistently reproduce it yet.

icholy:
@Valloric ready for review.

Valloric:
> The bug was caused by sending concurrent requests to tsserver.

That sounds like something tsserver should support, doesn't it? Wrapping requests to it with a lock seems like something that might come back to haunt us in the future. If it's a hack to make this work until tsserver supports concurrent requests (which is something a server should be able to do), then we can live with it temporarily. But I'd like to know for a fact that this is something that the tsserver team is aware is an issue and wants to fix.

vheon:
@Valloric I think the problem lies in use of the `seq` number in the client not in the server not able to handle multiple requests concurrently

icholy:
@Valloric I need to control access to reads and writes to the process either way. I think acquiring it once for each "high level" operation is cleaner than doing it in `_SendRequest` and `_ReadResponse`. Especially because `_ReadResponse` calls itself recursively. At first I was thinking of implementing some type of event loop so you could register a callback for the response, but afaik ycmd doesn't support async responses.

ycmd/completers/typescript/typescript_completer.py
```diff
@@ -150,34 +156,60 @@ def SupportedFiletypes( self ):
...
+        'entryNames': names
+      })
+      detailed_entries = self._ReadResponse( seq )[ 'body' ]
+      return [ _ConvertDetailedCompletionData( e, namelength )
+               for e in detailed_entries ]
 
   def OnBufferVisit( self, request_data ):
-    filename = request_data[ 'filepath' ]
-    self._SendRequest( 'open', { 'file': filename } )
+    with self._lock:
```
Valloric: This seems like it's holding onto the lock for longer than necessary; you can move the `with` statement after the `filename` var is created.

In general, try to keep the lock for as short a time as possible.

ycmd/completers/typescript/typescript_completer.py
```diff
@@ -150,34 +156,60 @@ def SupportedFiletypes( self ):
...
 
   def DefinedSubcommands( self ):
     return [ 'GoToDefinition' ]
 
   def OnUserCommand( self, arguments, request_data ):
-    command = arguments[ 0 ]
-    if command == 'GoToDefinition':
-      return self._GoToDefinition( request_data )
-    raise ValueError( self.UserCommandsHelpMessage() )
+    with self._lock:
```
Valloric: It might be a better idea to move the `with` statement into `_GotoDefinition`; this way you'd hold the lock for a shorter time.

ycmd/completers/typescript/typescript_completer.py
```diff
@@ -209,3 +242,15 @@ def _ConvertCompletionData( completion_data ):
     kind           = utils.ToUtf8IfNeeded( completion_data[ 'kind' ] ),
     extra_data     = utils.ToUtf8IfNeeded( completion_data[ 'kind' ] )
   )
+
+
+def _ConvertDetailedCompletionData( completion_data, padding=0  ):
```
Valloric: You have two spaces before `)` instead of one and no spaces around `=`.

Valloric:
Sad that tsserver is designed in such a way that concurrent requests aren't possible. That greatly limits its potential. But such is life...

icholy:
@Valloric fixed.

icholy:
If people start running into performance issues I'll look into concurrent requests. I've been dogfooding it and haven't had any issues. I'd rather not complicate it until I know it's a problem.

ycmd/completers/typescript/typescript_completer.py
```diff
@@ -47,6 +50,9 @@ class TypeScriptCompleter( Completer ):
   def __init__( self, user_options ):
     super( TypeScriptCompleter, self ).__init__( user_options )
 
+    # Used to prevent concurrent requests to TSServer.
```
Valloric: You also need to explain in the comment _why_ concurrent requests cause problems. You've explained in the PR, but that won't be easily seen by those reading this code months from now.

icholy: @Valloric how about:

> Used to prevent threads from concurrently reading and writing to the tsserver process' stdout and stdin

Valloric: That sounds fine.

icholy: Done

Valloric:
Thanks!

@homu r+

PR #162, opened 2015-06-19
Allow use of CMake 2.8.11 on CentOS/RHEL 7.x

Xiol:
Hi,

CMake 2.8.12 was required due to YouCompleteMe issue #1439 which seems to be limited to OSX Yosemite. Requiring this version of CMake breaks compilation on CentOS / RHEL7 which comes with CMake 2.8.11 but compiles ycmd without issue.

I've adjusted `cpp/CMakeLists.txt` to detect CentOS/RHEL 7 and allow the use of CMake 2.8.11. All other operating systems and distributions will still require 2.8.12.

Thanks

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -20,7 +20,13 @@ if ( DEFINED ENV{TRAVIS} )
   # Travis for, the old version of CMake that it has is good enough.
   cmake_minimum_required( VERSION 2.8 )
 else()
-  cmake_minimum_required( VERSION 2.8.12 )
+  if ( APPLE )
+    # OSX requires CMake >= 2.8.12, see YCM issue #1439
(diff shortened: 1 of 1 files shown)


Xiol:
I've signed the CLA.

Valloric:
It seems simpler to me to invert the logic and detect running on a Mac. The check for that is just `if (APPLE)`, which is far shorter and easier to maintain that what you're proposing.

Xiol:
I had considered that, however I wasn't 100% sure that the bug *did* only affect OSX.

I'll look into this a bit more and update the PR later tomorrow.

Xiol:
I've updated the PR as requested and squashed the commits to keep it tidy.

Valloric:
Thanks!

PR #163, opened 2015-06-21
Add a instruction for build on windows with visual studio

rexdf:
```cmake
if( NOT WIN32 )
    set( Python_ADDITIONAL_VERSIONS 2.7 2.6 )
    find_package( PythonLibs 2.6 REQUIRED )
else()
    if( NOT ( DEFINED PYTHON_INCLUDE_DIRS ) OR
        NOT ( DEFINED PYTHON_LIBRARIES ) OR
        NOT ( DEFINED PYTHON_EXECUTABLE ) )
      set( Python_ADDITIONAL_VERSIONS 2.7 2.6 )
      find_package( PythonLibs 2.6 )
      if( NOT PYTHONLIBS_FOUND)
        message( FATAL_ERROR
            "PYTHON_LIBRARIES and PYTHON_LIBRARIES and PYTHON_EXECUTABLE is needed" )
      endif()
    endif()
endif()
```

The above code is to support pass in `PYTHON_INCLUDE_DIRS` etc.

Because my system python2 is python279 x86 version and cmake read python version from Windows Registry, it's difficult to build on my system. While the build progress only need Python.h and python27.lib etc.

And I provide two batches to build llvm/clang latest 3.7.0 svn source code and ycmd. Build_On_Windows.md is a building instructions in details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,13 @@ cpp/BoostParts/libs/*/test
 
 # Exclude auto generated vim doc tags.
 doc/tags
+
+# Windows build
+ycmd_build/
+# Visual studio in debug mode generate ycmd_core.pdb and ycm_core.ilk
+*.pdb
+*.ilk
+# build llvm/clang on windows
+llvm-src/
+llvm_build/
+
--- a/Build_On_Windows.md
+++ b/Build_On_Windows.md
@@ -17,6 +17,8 @@ Note: the relative path base in following text is ycmd git source code folder.
 
 ###1.5. Quick
 
+Edit `_build_ycmd.bat`, and change `PYTHON_PATH` to a correct one.
+
 Click `_run_both.bat` and go out for a walk.
 
 ###2. Build llvm/clang
@@ -25,13 +27,16 @@ Run `_build_llvm.bat` (Where? the relative path base..)
 
 This batch gets llvm/clang source code and generate a Visual Studio 2013 project files. So it maybe takes less than 10 minutes. The source code with .svn takes 360+MB disk space.
 
-If no error, the batch should quit silently.go to `llvm-build`
+If no error, the batch should quit silently.
 
-Now go to `llvm-build` folder, open `LLVM.svn` and change **Debug** mode to **Release** mode. In Solution Explorer, go to `ALL_BUILD`, right click on `ALL_BUILD` and choose build. It take about quite a long time to finish the build. CPU is 100% during the build. The `llvm-build` folder takes about 6.65GB disk space.
+Now go to `llvm_build` folder, open `LLVM.svn` and change **Debug** mode to **Release** mode. In Solution Explorer, go to `ALL_BUILD`, right click on `ALL_BUILD` and choose build. It take about quite a long time to finish the build. CPU is 100% during the build. The `llvm_build` folder takes about 6.65GB disk space.
+
+If error happens when svn clone code, we should delete `llvm-src` and run it again.
 
 ###3. Build ycmd
 
 **Note**: Edit `_build_ycmd.bat`, and change `PYTHON_PATH` to a correct one.
+
 Run `_build_ycmd.bat`
 
 This batch generates `YouCompleteMe.sln` in ycmd_build folder. Open it with Visual Studio 2013. Change **Debug** mode to **Release** mode. In Solution Explorer, select `ycm_support_libs` and right click on it choose build. (Note `ycm_core_tests` and `gmock_main` `gtest_main` etc will fail, so we cannot build the whole solution.) Now you should find `libclang.dll` and `ycm_client_support.pyd` and `ycm_core.pyd`.
@@ -55,6 +60,10 @@ You can follow:
 https://bitbucket.org/Haroogan/vim-youcompleteme-for-windows/
 
 In my case, it is `C:\Program Files (x86)\Intel\iCLS Client\msvcr90.dll`. Remove `C:\Program Files (x86)\Intel\iCLS Client` from path. Now everything goes quite well.
+
+###5. Cleanup
+
+You can safely delete `llvm_build\Release\lib`.
 
 ###note
 
--- a/Build_On_Windows.md
+++ b/Build_On_Windows.md
@@ -23,14 +23,14 @@ This batch gets llvm/clang source code and generate a Visual Studio 2013 project
 
 If no error, the batch should quit silently.go to `llvm-build`
 
-Now go to `llvm-build` folder, open `LLVM.svn` and change Debug mode to Release mode. In Solution Explorer, go to `ALL_BUILD`, right click on `ALL_BUILD` and choose build. It take about quite a long time to finish the build. CPU is 100% during the build. The `llvm-build` folder takes about 6.65GB disk space.
+Now go to `llvm-build` folder, open `LLVM.svn` and change **Debug** mode to **Release** mode. In Solution Explorer, go to `ALL_BUILD`, right click on `ALL_BUILD` and choose build. It take about quite a long time to finish the build. CPU is 100% during the build. The `llvm-build` folder takes about 6.65GB disk space.
 
 ###3. Build ycmd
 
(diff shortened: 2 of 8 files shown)


rexdf:
I already sign Google Contributor License Agreement.

vheon:
As you may know there is no official support for windows at the moment, even though some patches  and all the current support for windows was made by the community and the instructions for building YCM on Windows at the moment are in the [Windows-Installation-Guide in the wiki](https://github.com/Valloric/YouCompleteMe/wiki/Windows-Installation-Guide) so I don't know how @Valloric want to manage this, maybe move the new files to a `contrib` directory? update the wiki and put the `bat` scripts into some `gists` (btw I have so say that having those `bat` scripts would be very useful for windows user :+1: )?

micbou:
The CmakeLists.txt changes are unnecessary. CMake uses also the PATH environment variable on Windows to find Python. Just add the Python path to it.

I think that adding Windows support to the already existing Python script is the way to go instead of using batch scripts. You should check my [fork](https://github.com/micbou/ycmd).

rexdf:
It cannot on my system from path or cmake -D...  I install pythonxy which python279 x86. 
In `C:\Program Files (x86)\CMake\share\cmake-3.2\Modules\FindPythonLibs.cmake`:

```
foreach(_CURRENT_VERSION ${_Python_VERSIONS})
  string(REPLACE "." "" _CURRENT_VERSION_NO_DOTS ${_CURRENT_VERSION})
  if(WIN32)
    find_library(PYTHON_DEBUG_LIBRARY
      NAMES python${_CURRENT_VERSION_NO_DOTS}_d python
      PATHS
      [HKEY_LOCAL_MACHINE\\SOFTWARE\\Python\\PythonCore\\${_CURRENT_VERSION}\\InstallPath]/libs/Debug
      [HKEY_CURRENT_USER\\SOFTWARE\\Python\\PythonCore\\${_CURRENT_VERSION}\\InstallPath]/libs/Debug
      [HKEY_LOCAL_MACHINE\\SOFTWARE\\Python\\PythonCore\\${_CURRENT_VERSION}\\InstallPath]/libs
      [HKEY_CURRENT_USER\\SOFTWARE\\Python\\PythonCore\\${_CURRENT_VERSION}\\InstallPath]/libs
      )
  endif()
```

rexdf:
Batch is windows native support. It is easy to write and change.  @micbou  Yes, you repo is real great, everything is automatic. That should be the right way to go while it needs a lot of  efforts.

My batch script could exist with your python script at same time. It  should have no effect on any other build platform. It downloads latest llvm/clang svn source code from official site and builds ycmd. Following the official site llvm/clang svn code which is updated frequently is not suitable for a bot to build automaticly.

Anyway,it takes me less than half a day. If it is not fit this project, just close it.  Thanks.

rexdf:
If needed, I can just add the following code to batch to make the batch automaticly build it without open the GUI(But ycmd has some test fail case.):
```batch
call "C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\vcvarsall.bat" amd64
cd llvm_build
devenv LLVM.sln /build
```

micbou:
You are right. CMake does not use the PATH but the registry to find the Python library. However, it should still find the correct library because it uses the 32 or 64 bits registry depending on the generator used (`Visual Studio 12 2013` or `Visual Studio 12 2013 Win64`). You should check the value of these keys in the registry:
`HKEY_LOCAL_MACHINE\SOFTWARE\Python\PythonCore\2.7\InstallPath` and `HKEY_LOCAL_MACHINE\Wow6432Node\SOFTWARE\Python\PythonCore\2.7\InstallPath`. The first should point to the 64 bits version of Python and the second to the 32 bits version.

rexdf:
I know that different python versions can coexist. But I want a portable one python, so I don't install it, I startup gvim with a bat to set up PYTHONPATH PATH,etc. It's full feature gvim for USB drives.

rexdf:
@micbou you can call my `_run_both.bat` in python.

rexdf:
vim3

Valloric:
Sorry, YCM does not officially support Windows. Your changes are too extensive to be merged for a platform we don't support (nobody will be maintaining that).

vheon:
@micbou since I have to work on windows at the moment I'm trying out your fork and it seems to work (at least the basic YCM, I'm going to give a spin to the clang completer tomorrow) and it works great, so thanks! There was a problem though: Microsoft started including `msbuild.exe` since Visual Studio 2013, so for version prior to that, like Visual Studio 2012 you can't use `msbuild.exe` to find out which version of Visual Studio you have so line https://github.com/micbou/ycmd/blob/windows/build.py#L108 do not work. Now, this is the first time I'm developing on windows so I know basically nothing about the environment and could be a configuration problem, but `msbuild.exe` with Visual Studio 2012 is found at `C:\Windows\Microsoft.NET\Framework64\v4.0.xxxx` and if you run it from the command line you get `4.0.xxxx`. Anyway forcing the generator to be `Visual Studio 11 2012` compiles and it works. Now is it something with my environment or we can't rely on `msbuild /ver /nologo` for finding out versions of Visual Studio prior to 2013?

micbou:
@vheon Thanks for trying it. From what you are saying, it seems that `msbuild` is not the right way to check the version of Visual Studio. I should use the `cl` executable instead. I will update the script.

For the Clang completer, you will need to compile LLVM+Clang yourself since my fork automatically download the binaries from this [page](https://github.com/micbou/llvm/releases) which are built against Visual Studio 2013.

vheon:
> @vheon Thanks for trying it. From what you are saying, it seems that msbuild is not the right way to check the version of Visual Studio. I should use the cl executable instead. I will update the script.

Thanks!

> For the Clang completer, you will need to compile LLVM+Clang yourself since my fork automatically download the binaries from this page which are built with Visual Studio 2013.

Oh thanks for the heads up ;)

@micbou do you think it would be possible to set up a project on appveyor which would build LLVM+Clang with the various Visual Studio versions and download the binaries from the artifact that we generate? I just started with Windows so I don't really know if is a feasible idea but we could start a YCM-Contrib organization for a YouCompleteMe and ycmd for windows.

micbou:
Using the `cl` executable is more difficult than I thought. Apparently, it does not need to be in the PATH so I cannot rely on it. I decided to add a parameter to select the MSVC version. I think this is the safest approach.

Concerning your LLVM+Clang suggestion, I tried it but unfortunately it takes too long to compile LLVM+Clang on appveyor (more than 40 minutes) so, for the time being, uploading our own binaries is the only way (the official binaries does not work...)

vheon:
> Using the cl executable is more difficult than I thought. Apparently, it does not need to be in the PATH so I cannot rely on it. I decided to add a parameter to select the MSVC version. I think this is the safest approach.

I wanted to propose that myself.

> Concerning your LLVM+Clang suggestion, I tried it but unfortunately it takes too long to compile LLVM+Clang on appveyor (more than 40 minutes) so, for the time being, uploading our own binaries is the only way.

> (the official binaries does not work...)

That is strange, no?

vheon:
@micbou Oh btw, at the moment if I install you're fork of YouCompleteMe with vim-plug or bare git (git clone https://github.com/micbou/YouCompleteMe && cd YouCompleteMe && git submodule update --init --recursive) I get

```
fatal: reference is not a tree: 66ffb3c4e86c43fdd3347ef4b01b354019ec4551
Unable to checkout '66ffb3c4e86c43fdd3347ef4b01b354019ec4551' in submodule path 'third_party/ycmd'
```

micbou:
> That is strange, no?

If you compare the size, the official binaries are much smaller than mines so there are certainly some missing parts. There is also the only 32bit issue.

> fatal: reference is not a tree: 66ffb3c4e86c43fdd3347ef4b01b354019ec4551
> Unable to checkout '66ffb3c4e86c43fdd3347ef4b01b354019ec4551' in submodule path 'third_party/ycmd'

I need to update the `YouCompleteMe` repository.

Issue #165, opened 2015-06-25
hardcoded libclang.dll in ycmd\cpp\ycm\CMakeLists.txt


strrchr:
In Line 317:
COMMAND ${CMAKE_COMMAND} -E copy "${PATH_TO_LLVM_ROOT}/bin/libclang.dll" "$")
I think it should be:
COMMAND ${CMAKE_COMMAND} -E copy "${LIBCLANG_TARGET}" "$")
I made the modification and the build using cmake+ninja is OK.

I encounter error when compile ycm_support_libs using msys2.
There is clang.dll in msys2\mingw64\bin, but not libclang.dll.

Sorry, I haven't learned how to make a PR.
So I just commit this issue.

micbou:
`${LIBCLANG_TARGET}` is `libclang.lib` when compiling with MSVC and we need `libclang.dll` hence the hardcoded `libclang.dll`. I will send a PR to only do this with MSVC.

Anyway, I recommend you to compile ycmd with MSVC. Clang support is better than MinGW one.

Valloric:
Fixed by #166.

PR #166, opened 2015-06-25
Only copy libclang DLL when building with MSVC

micbou:
`LIBCLANG_TARGET` is correct when building with MinGW, so we should only copy `libclang.dll` when building with MSVC (where `LIBCLANG_TARGET` is `libclang.lib` instead of `libclang.dll`). See issue #165.

CLA signed.


Valloric:
Thanks!

@homu r+

PR #167, opened 2015-06-27
[RFC] Support for FixIt in clang completer

puremourning:
This change provides the server support in clang completer for providing fixit information in the diagnostic reply. The client uses this in a new :YcmFixIt command which applies changes to the vim buffer to implement trivial fixes to common coding errors. A note is added "(FixIt)" to the diagnostic text echo'd in vim when a FixIt is available.

While this may seem out of the scope of ycmd/YouCompleteMe (as a completion engine), I think it is a really useful feature, and provided straight out of libclang. so I'm sending this slightly incomplete PR (along with the associated YouCompleteMe change) to see what the appetite is within the user and maintainer community, and to get feedback on the approach.

I know that I need to write some tests, and perhaps tidy up some TODOs, but I would really appreciate any feedback on the approach. Also, I should confess that I haven't yet run the existing tests as I'm having trouble getting them to run on my mac. Hoping that travis will help me out :)

I know I also need to update the readme.

Testing has been performed manually by going through the list of .c, .cpp and .m files in clang's tests/FixIt folder and ensuring that by hitting :YcmFixIt (actually "@:" a lot!) on each diagnostic, the diagnostic went away (assuming a FixIt was available).

I also understand that adding the vim modelines might be divisive, but the standard in this project doesn't match my usual setting :) Happy to remove them if people react angrily.

Once again thanks for the time and a great (and improving!) product.

I have signed the CLA.


puremourning:
Adding some of my design notes to help explain the way it works (see also https://github.com/Valloric/YouCompleteMe/pull/1555)

ycmd

```
o libclang can provide the fixit data for a given diagnostic trivilly (clang_getDiagnosticFixItNum, clang_getDiagnosticFixIt), providing replacement text and a range into which to put it. ycmd can already deal with ranges from libclang

o so we add the fixit chunk details to the diagnostic object as a vector of
    replacement_text
    range

o the 'replacement_text' is the text to put in the file, replacing the text
  specified by 'range'

o this is exported in the python interface as normal (TODO using 1-based
  clang indexing? or should ycmd convert to 0-based indexing. Need to check the
  precedent)
```

puremourning:
As expected, this change breaks some of the tests. I will fix them, but please let me know any thoughts/comments.

puremourning:
ok latest update makes the travis build pass (though one of the runs failed due to being unable to download a pip package).

puremourning:
latest pushed change adds tests and got a clean run on travis

ycmd/responses.py
```diff
@@ -125,15 +125,25 @@ def BuildLocationData( location ):
...
+    return {
+      'replacement_text': fixit.replacement_text,
+      'range': BuildRangeData( fixit.range ),
+    }
+
   kind = ( diagnostic.kind_.name if hasattr( diagnostic.kind_, 'name' )
            else diagnostic.kind_ )
 
+  fixits = ( diagnostic.fixits_ if hasattr( diagnostic, 'fixits_' )
+                             else [])
```
vheon: you're missing a space before the closed paren.

puremourning: yup, thanks. Fixed.

mispencer:
Omnisharp has a support for "fixes" too. I think diagnostic output has that there is one, and it is a separate request for the fix itself. When I have time, I'll look into what it would requires to implement support for that as well. Keeping this as generic as possible would be helpful for C# support, when that come.

puremourning:
Interesting that it is a separate request. That might mean that @vheon's suggestion to use a completer subcommand might well be (annoyingly :)) spot on, and we should have the clang completer behave the same way. I don't know diddly about the omnisharp completer or server, so if you have any further info I could go digging around.

mispencer:
Upon further research, Omnisharp has two methods: a [fix issue action](https://github.com/OmniSharp/omnisharp-server/blob/master/OmniSharp/CodeIssues/CodeIssuesHandler.cs#L41) and a [run code action](https://github.com/OmniSharp/omnisharp-server/blob/master/OmniSharp/CodeActions/GetCodeActionsHandler.cs#L29).

/fixcodeissue takes the common request plus a line number, and returns the whole text of the file with the first code issue on that line fixed.

/runcodeaction takes the common request plus a index (into a list of all the code actions in the file), and returns the whole text of the file with the indexed code issue fixed.

It looks Omnisharp has more fine-grained changes chunks, but it isn't exposing them at the present.

puremourning:
Hmm that is quite different. I wonder what is more common amongst things-which-support-FixIts? In theory we could have the clang completer do a similar thing (i.e. get the complete text, patch it, then return the complete text), though applying the small changes in vim leads to somewhat nicer undo history for the user and a (theoretical) interactive mode when patching.
Then again, if it were completer-specific, the action returned could be "ReplaceChunk" for clang, and ReplaceFile for omnisharp.

I'm getting more and more convinced that this needs to be done server side with an action and a response instructing the client what to do, rather than just supplying the fixits on the diagnostic response. That's frustrating >_<.

Well... the more you know... I'll leave this open for now in case anyone else wants to chip in with ideas. Not sure if/when i'll find the time to convert this to a request/reply interface, though.

Valloric:
Thanks for sending in a PR!

A past version of me would tell you this is out of scope for ycmd. Today's version of me is more mellow about such things (I don't know is that good or bad).

If we were to go down the route of providing this feature, it seems like the simplest design is that the client sends the buffer contents and gets back a new version of the buffer, which should then replace the old version. 

But this then means that it's an all-or-nothing affair. You can either fix up your entire file, or nothing. By sending back info about each fix-up, the clients could present a UI for applying each fix individually or everything, as desired by the user.

We also seem to have differences between how OmniSharp and libclang expose this functionality; wrapping those differing approaches behind one API is basically the whole point of ycmd, so we need to do something sensible.

As usual, I'm open to ideas. Currently I'm leaning towards the idea of a new subcommand that would be common to completers that support fix-ups and that would return some sort of structured JSON with info about each available fix-it. The clients would then handle this in whatever way works best for them.

Thoughts?

> I also understand that adding the vim modelines might be divisive

I'm not a fan of any editor-specific artifacts in source code, and I say that as someone who uses Vim. Such artifacts don't scale (how many code editors are there?) and cruft up the code.

Valloric:
@abingham @ptrv I'm interested in your thoughts on how this feature should be exposed from ycmd. I don't want the design we come up with to be too YCM-centric, and even less so Vim-centric.

@r4nt You might also be interested. :)

abingham:
Like you, my instinct is that I'd rather get back a list of potential fixes and apply them locally. This integrates naturally with the standard undo/redo stacks, doesn't run the risk of stomping on text properties, etc. I also suspect that it will work better with any emacs refactoring packages/frameworks (notional though they are!).

r4nt:
@Valloric - yay \o/

In clang-format we support both: we support an unstructured output that gives you the full file, and also support outputting changes. For editor integration the changes are usually better (as mentioned, due to better fitting with the editors inherent capabilities).

One thing that was a rather important piece in clang-format was tracking of the cursor position (editors tend to get that wrong).

puremourning:
@abingham thanks, that's what I thought, and that''s how this PR is working now.

@r4nt thanks for the input. when you say "tracking cursor position" are you referring to the tricky problem of offsets changing as a result of patching the code? I have a bit of a hack in the client implementation right now which basically works, but i'm not super proud of it. From https://github.com/Valloric/YouCompleteMe/pull/1555

```
o special note:
  offsets in 'range' are pre-manipulation. this leads to a tricky problem where
  the serial application of replacements leads to the indices being out of sync
  clang has a whole library of complex code (libRewrite) for dealing with this.

  so to solve, we:
   - assume that each chunk is always on a single line
   - sort chunks by start line then column
   - maintain a delta of current line length and apply that to the columns
   - reset delta when chunk is on a new line

        example:
        [
            {
                u'range': {
                    u'start': {
                        u'line_num': 13, 
                        u'column_num': 10, 
                        u'filepath': u'/Users/ben/Development/tmp/test.cc'
                    }, 
                    u'end': {
                        u'line_num': 13,
                        u'column_num': 10,  
                        u'filepath': u'/Users/ben/Development/tmp/test.cc'
                    }
                },
                u'replacement_text': u'static_cast<int>('}, 
            {
                u'range': { 
                    u'start': {
                        u'line_num': 13,
                        u'column_num': 13,
                        u'filepath': u'/Users/ben/Development/tmp/test.cc'
                    },
                    u'end': {
                        u'line_num': 13,
                        u'column_num': 13,
                        u'filepath': u'/Users/ben/Development/tmp/test.cc'
                    }
                },
                u'replacement_text': u')'
            }
        ]
        FixIt with: 'static_cast<int>(' at (13,10) .. (13,10)

        line:
           switch(A()) { // expected-error{{explicit conversion to}}
        fixed:
           switch(static_cast<int>(A()) { // expected-error{{explicit conversion to}}

        FixIt with: ) at (13,13) .. (13,13)
        line:
           switch(static_cast<int>(()) { // expected-error{{explicit conversion to}}
        fixed:
           switch(static_cast<int>(A())) { // expected-error{{explicit conversion to}}
```

r4nt:
@puremourning - nope, I'm talking about tracking the actual cursor position (where the user types input). Usually users want the cursor at the same semantic position after the patch was applied.

The offset problem is a different problem. The easiest solution seems to be to re-query the fixits after one was fixed.

ptrv:
I am also in favor of ycmd providing a list of possible fixes and letting the client decide how to process it.

Valloric:
@puremourning Oh man, that offset changing problem is nasty. I think the approach you have makes sense, but consider also somehow marking several fix-up change chunks as being related (if they're all part of the same logic change) so that clients can apply all logically related changes at once instead of making it possible to apply half of a change.

Hm, this makes me think that the json structure for fix-ups should be multi-level. First, an array of logical change objects. Then each change object has an internal array of multiple (if necessary) line-level changes that are part of that change.

For the line-level changes, maybe we should include the "pristine" offsets in addition to the offsets that would be necessary if the other same-line changes are applied. I honestly don't know.  

What do you think?

Also, it seems that several of us here agree that sending back a list of fix-ups so that the client can then apply them would be the best approach. So let's go with that idea for now.

@r4nt Yeah, the client should preferably keep the user's cursor where it was even after a change. But IMO this is something that clients (YCM, emacs-ycmd etc) need to handle, not ycmd. We can't compute the cursor offset because we don't know which fix-ups will be applied by the user. What do you think?

Valloric:
We could also just skip the whole complexity with offsets by returning a list of fix-ups to apply that have nothing but "pristine" offsets against the original buffer but each fix-up gets an ID which the client can use to call into ycmd with whatever is the state of the current buffer and get back correct offsets to apply. 

So one fix-up "list" call and several "more info about this fix-up" calls that have the right data. This would probably massively simplify the amount of tricky logic for clients.

In fact, we're probably over-complicating things. You don't even need the extra calls about individual fix-ups. Just have one call for the list of fix-ups with only pristine offsets. When one fix-up is applied, the client should make the same "list of fix-ups" call again to get updated offsets. Hopefully that wouldn't be too slow, because if it's not, it's the simplest approach.

r4nt:
@Valloric - your final simple approach seems to match exactly what I meant when I wrote "The easiest solution seems to be to re-query the fixits after one was fixed."

Valloric:
@r4nt Ah, I missed the nuance of your suggestion. Yes, I'm suggesting the same thing. We might also want to include the text of the fully-fixed buffer in the response as well to optimize for the "apply all the fixes" use case, otherwise for 50 fix-ups there would be 50 calls to ycmd; that's fine when the user is applying each one by hand, but not otherwise. 

It would also then be just one undo point in the editor.

That's how I see it, since there are 2 use-cases: 
- The user wants to evaluate each fix-up and apply them one at a time.
- The user trusts the tool and just wants to apply all the suggestions. This covers things like clang-format since most users probably want to accept whatever it suggests for the whole file.

mispencer:
We should optimize for the first case, since it should be much more common usage. For this case, as an end user, what I want to see is a fix marker in the buffer, preferentially in the same spot as error markers, and be able to hit a key to have it be fixed. For trivial fixes it might take longer to run the fix than it would be to fix it manually, if the fix invoking was more complex than hover/select/move cursor and key press.

Thus, my preferred design would to return possible fixes in the diagnostic to leverage the existing error/warning display, both in implementation effect and learning curve. Then add a new completer command that takes the line number, and column index and/or index of fix applying to line, since the easiest way for users to indicate which fix they want would be to hover over/move the cursor to/whatever the line they want fixed.

Valloric:
@mispencer Oh I agree that we should definitely optimize for the common case of the user wanting to apply each fix separately, no disagreement there. We probably also need to have something that supports getting a fully-fixed buffer so that the client doesn't have to make a bunch of requests. This could (and probably should) be a separate request though, possibly as a new completer subcommand.

I agree with your suggested UI too.

r4nt:
Yep, this all sounds very awesome now (unrelated: note that for clang-format the common case is also to hit clang-format on a single statement and only format that statement - clang-format has the advantage of being really fast though).

puremourning:
OK wow. Quite a lot here, thanks. Let me summarise and respond ....

TL;DR, skip to end where i say actions :)

@Valloric:
> Hm, this makes me think that the json structure for fix-ups should be multi-level. First, an array of logical change objects. Then each change object has an internal array of multiple (if necessary) line-level changes that are part of that change

That's sort of what you get from libclang. The interface in the example above is pretty much a direct representation of what libclang tells us. That is, for a given diagnostic, a set of "chunks", each consisting: a range and some replacement text.

@r4nt:
> One thing that was a rather important piece in clang-format was tracking of the cursor position (editors tend to get that wrong).
> I'm talking about tracking the actual cursor position (where the user types input). 

I understand. I didn't give that too much thought because in the vim client implementation, the changes are all done in Normal mode, and the cursor does not move. Again, supplying the offsets and a series of edit commands to the client seems to be sensible insofar as each client's understanding of cursor position is likely to be domain-specific.

@Valloric:
> For the line-level changes, maybe we should include the "pristine" offsets in addition to the offsets that would be necessary if the other same-line changes are applied.

I'm not sure I fully followed this. The offsets in the current implementation are perhaps "pristine" in that they are the offsets returned by libclang. This offsetting problem is really only an issue for a given fix-it which has multiple edit "commands" (chunks). The way this works for multiple different diagnostic/fix-it chunk array pairs is that the vim client ensures that no fix-it chunk array is applied without first doing a "OnFileReadyToParse" loop synchronously (ensuring that all of the diagnostic data and offsets are accurate with respect to the current buffer contents).

@Valloric:
 > That's how I see it, since there are 2 use-cases: 1. The user wants to evaluate each fix-up and apply them one at a time. 2. The user trusts the tool and just wants to apply all the suggestions. This covers things like clang-format since most users probably want to accept whatever it suggests for the whole file.
> we should definitely optimize for the common case of the user wanting to apply each fix separately

That's what i've gone for in the vim client for PR. See the gif on https://github.com/Valloric/YouCompleteMe/pull/1555 - the user (in normal mode in vim) hits :YcmFixIt on the line with the diagnostic and the client applies the first fix-it.

By returning all fix-its supplied by libclang in the diagnostic response (queried whenever the file is "ready to parse"), the client is always up to date and is able to decide how and when to apply them. That is, it would not be hard to write a :YcmFixAll which applied all the fix-its to the file, just by iterating through the diagnostics and applying any fixits (in my client implementation that would be trivial).

My thought is that clients are *much* better at editing the text and applying the diffs than anything I could write in ycmd. Certainly, making the changes in vim's python interface is just [a few lines of code](https://github.com/puremourning/YouCompleteMe/blob/ff129009efc0ac1a1ff09137071958116a806d62/python/ycm/vimsupport.py#L475-L481).

So in summary of how this PR is set up:
* this implementation optimises for the "common" case of the user seeing a diagnostic and going "ok, just fix it for me"
* it returns fix-its, consisting of an array of "chunks" (range+replacement text) with the diagnostic response, queried effectively whenever the buffer changes, or when the user requests applying a FixIt
* most libclang fix-its consist of a single chunk. i have not seen any with more than 2.
* each fix-it chunk is applied individually by the client, when the user requests it - this can lead to the offsets in subsequent chunks being wrong (having modified the original buffer)
* i don't believe that you can "re-ask" libclang to update its offsets for you mid-application of a partial fix it chunk array
* maintaining the correct offsets for a given fix-it in the client is pretty simple (as show in my algorithm above), as compared to patching the file on the server (imho)
* Applying multiple fix-its requires effectively re-querying for the diagnostics, which effectively re-queries for the fix-its
* it assumes all completer behave like libclang.

And things which are still contentious:
* other completion engines supporting fix-its work differently from libclang, so this probably should be implemented on a per-completer basis. It doesn't seem fair or sane to mimic libclang's interface in the XYZ completer

Finally, i have been using eclim with YCM recently and found that the way it works backs up a lot of what we've discussed:
* there is a user command :JavaCorrect it presents the user with a split window showing diffs that can be applied based on the context
* applying the changes seems to get eclipse to write a new file, but the undo history is still sane (only the diffs, and undo works as you would expect). I haven't been able to quite divine how they did this, but it is certainly interesting.

So I think the actions are, unless anyone fundamentally objects to the above:
* provide a clang completer subcommand FixIt which takes the entire buffer text, re-parses the file and returns a structure containing all fix-its applicable to the current line/column
* the structure specifies a type of fix-it, where types are like (replace chunk, replace buffer). clang completer will always return replace chunk
* add a flag to the diagnostic response to indicate if a fix-it is available for the current diagnostic (as opposed to retuning the fix-it itself)
* change the client implemenation to apply the fix-it when it gets the response to this request, rather than expecting it in the diagnostic output.
* clients can still iterate the diagnostics with fix-it-available flag set and apply all fixits if they so desire

This leaves the door open for other completer's FixIt implementations to either use the same or similar approach, or return the entire buffer contents, and doesn't require significant complexity in the client or server.

However, it does pretty much amount to a re-write, so no promises on timescales, but if we're happy with that approach, I will have a go at changing it. Might make sense to raise a new PR for that implementation though.

Valloric:
@puremourning 

> other completion engines supporting fix-its work differently from libclang, so this probably should be implemented on a per-completer basis. It doesn't seem fair or sane to mimic libclang's interface in the XYZ completer

Agreed, which is why I believe we need to expose a "universal" interface. The fix-up interface (JSON structures in the response etc) we expose should also work with other completers. Let's not have one set of JSON structures for fix-ups coming from libclang, another from omnisharp etc.

The "replacement chunks" systems seems to do the trick. Like you said, fix-its are:

- A location range. Start line & column, end line & column.
- The replacement text for that range (possibly including whitespace like newlines).

This should be generic enough to work well with whatever other API other semantic engines expose. I can't see how we would have much difficulty adapting other APIs to this.

> provide a clang completer subcommand FixIt which takes the entire buffer text, re-parses the file and returns a structure containing all fix-its applicable to the current line/column

:+1: 

> the structure specifies a type of fix-it, where types are like (replace chunk, replace buffer). clang completer will always return replace chunk

I'm unsure are you suggesting the same structure I am. Here's what I have in mind:

```javascript
[ {
    'range': {
      'start': {
        'line_num': 3,
        'column_num': 16
      },
      'end': {
        'line_num': 3,
        'column_num': 21
      } 
    },
    'replacement_text': 'foobar'
  }, 
  ... 
]
```

This closely mirrors the structure we return for diagnostics.

> add a flag to the diagnostic response to indicate if a fix-it is available for the current diagnostic (as opposed to retuning the fix-it itself)

:+1: 

> change the client implemenation to apply the fix-it when it gets the response to this request, rather than expecting it in the diagnostic output.

:+1: 

> clients can still iterate the diagnostics with fix-it-available flag set and apply all fixits if they so desire

This one I'm unsure about. Why not have a separate `:FixAll` subcommand that just returns the fully fixed buffer contents? This can then be just one undo point in the editors, which makes sense from the user's perspective since they only issued one command (the `:FixAll` one). The easier we can make life for clients, the better.

I'm open to suggestions here. I'm also fine with punting on this; we could always add a `:FixAll` command later on.

Yeah, let's punt for now.

> However, it does pretty much amount to a re-write, so no promises on timescales, but if we're happy with that approach, I will have a go at changing it. 

No promises expected or needed. :) Thanks for doing this initial implementation, it has already produced a very useful discussion.

> Might make sense to raise a new PR for that implementation though. 

Up to you.

Thanks again for the PR!

puremourning:
> I'm unsure are you suggesting the same structure I am. Here's what I have in mind:
Yes, that's exactly it, which is neatly exactly the structure we have now in the PR. 

```python
                u'range': {
                    u'start': {
                        u'line_num': 13, 
                        u'column_num': 10, 
                        u'filepath': u'/Users/ben/Development/tmp/test.cc'
                    }, 
                    u'end': {
                        u'line_num': 13,
                        u'column_num': 10,  
                        u'filepath': u'/Users/ben/Development/tmp/test.cc'
                    }
                },
                u'replacement_text': u'static_cast<int>('}, 
```

> Why not have a separate :FixAll subcommand that just returns the fully fixed buffer contents?

Mostly because I don't want to write C code to manipulate the buffer upon the server :), I'd rather any changes are applied by the client. So, in addition to a `GetFixItsForLocationInFile`, we could add  `GetFixItsForEntireFile` (or whatever) and have the client apply them all join one go, though I think that would still end up with each chunk being a separate undo entry, unless the client has some magic which can squish buffer modifications into one "undo" entry. My vim-fu isn't quite at the level of knowing that (and my google-fu also failed me just now!). This would of course be much easier for the other completion engines where the underlying completer works exactly this way.

But you're right - let's leave that to see if the community really wants it :)

@Valloric One final question (slightly out of the blue) - is there a way in the ycmd c code to print to stdout or stderr? I might be being completely dumb but no matter what settings I try the stdout/stderr goes nowhere that I can find. again, sorry if that's a stupid question...

mispencer:
>>One thing that was a rather important piece in clang-format was tracking of the cursor position (editors tend to get that wrong).
>>I'm talking about tracking the actual cursor position (where the user types input).

>I understand. I didn't give that too much thought because in the vim client implementation, the changes are all done in Normal mode, and the cursor does not move. Again, supplying the offsets and a series of edit commands to the client seems to be sensible insofar as each client's understanding of cursor position is likely to be domain-specific.

We need to know how where to move the cursor to stay at the same logical location. For example, if the cursor is over a variable, it should still be over that variable after applying the fix, even if the variable is now three lines down and ten characters left from its pre-fix location. Else the cursor will seems to move semi-randomly when fixes are applied. So, the fix should tell us how to handle the cursor location too, probably by taking the current cursor, and returning where it should go after fix is applied.

> the structure specifies a type of fix-it, where types are like (replace chunk, replace buffer). clang completer will always return replace chunk

The replace buffer type is just a special case of the replace chunk spanning the whole buffer, so I don't know if we need fixit types. 

>>clients can still iterate the diagnostics with fix-it-available flag set and apply all fixits if they so desire

>This one I'm unsure about. Why not have a separate :FixAll subcommand that just returns the fully fixed buffer contents? This can then be just one undo point in the editors, which makes sense from the user's perspective since they only issued one command (the :FixAll one). The easier we can make life for clients, the better.

I'm not sure this is ideal. Fix are not necessary independent; I don't think we should assume that we can apply all fixes returned at once. Probably the only safe thing would be fix one, require, repeat until none left.

Valloric:
> Mostly because I don't want to write C code to manipulate the buffer upon the server 

Oh don't worry, I wouldn't want to have it in the C++ layer. It would be in the Python layer, which would get the fix-it list from the C++ layer and then munge the buffer. String manipulation in C++ is a pain.

But like I said, punt on it. No need to do it now, and we possibly won't need it later either. 

>  is there a way in the ycmd c code to print to stdout or stderr? 

We do crazy stuff to the stdin & stdout for ycmd. Check out [`__main__.py`](https://github.com/Valloric/ycmd/blob/master/ycmd/__main__.py).

Your best bet would be to just use the tests and not go through the whole ycmd infrastructure. Both gtest (which we use for C++ tests) and nose (which we use for Python tests) capture stdout in a nice way AFAIR.

puremourning:
well, I had a quick crack at this change to using a subcommand this evening - seems to work, but needs lots more testing. if you're curious, take a look. It will have broken the tests - i'll fix them later. Now i gotta sleep - work tomorrow :)

Valloric:
> Yes, that's exactly it, which is neatly exactly the structure we have now in the PR.

Ha, great! :)

Though you might want to move the `filepath` out of `start`/`end`; it's not as if we can have a range that spans files. You can have the `filepath` on the top level of the object, next to `range`.

puremourning:
the file path thing comes straight out of the existing YouCompleteMe::Range class. I didn't change it.

puremourning:
The `range` entry is a direct use of the existing `YouCompleteMe::Range` class, as used by the other diagnostics. It didn't seem sensible to create a new one, or change the existing one.

```c++
///
/// Information about a replacement that can be made to the source to "fix" a
/// diagnostic
///
struct FixItChunk {
  /// the replacement string. this string should replace the source range
  //represented by 'range'
  std::string replacement_text;

  /// the range within the file to replace with replacement_text;
  Range range; 

  bool operator == ( const FixItChunk &other ) const {
    return replacement_text == other.replacement_text &&
           range == other.range;
  }
};

```

r4nt:
Throwing in an anecdote of experience: I'd suggest to use byte offsets for ranges. Everything else is really hard to guess what exactly it means and lead to confusion. Lines and offsets are perfect when giving the information to the user. Between tools, 0-based byte offsets work well because they are  unambiguous.

Valloric:
@puremourning Ah, proceed then. :)

@r4nt It's a moot point since ycmd uses (line, column) pairs for locations in other parts of the interface and we're not going to break backwards compatibility now. And we do want to be consistent with the rest of the API.

puremourning:
OK so the latest pushed commit implements the actions + tests as discussed. Tests are clean and i've tidied up the commit.

Now:
* the diagnostic response includes fixit_available: True/False
* "FixIt" clang completer subcommand returns a list of fixits for the given line, ordered by proximity to the cursor
* Each fixit is, as before, a list of { replacement_text + range } pairs.

Client for vim also updated to apply this (with limitations as before)

r4nt:
@Valloric - is the API characters or bytes for the column?

Valloric:
@r4nt It's characters.

cpp/ycm/ClangCompleter/Diagnostic.h
```diff
@@ -33,6 +33,34 @@ enum DiagnosticKind {
   WARNING
 };
 
+/// Information about a replacement that can be made to the source to "fix" a
+/// diagnostic
+struct FixItChunk {
+  /// the replacement string. this string should replace the source range
```
Valloric: Please use capital letters for sentences in comments; it's extra jarring when you have periods etc. :)

cpp/ycm/ClangCompleter/Diagnostic.h
```diff
@@ -33,6 +33,34 @@ enum DiagnosticKind {
...
+  /// represented by 'range'
+  std::string replacement_text;
+
+  /// the range within the file to replace with replacement_text;
```
Valloric: Cap for start, no `;` at end.

cpp/ycm/ClangCompleter/Diagnostic.h
```diff
@@ -33,6 +33,34 @@ enum DiagnosticKind {
   WARNING
 };
 
+/// Information about a replacement that can be made to the source to "fix" a
+/// diagnostic
```
Valloric: Period at end.

cpp/ycm/ClangCompleter/Diagnostic.h
```diff
@@ -53,6 +81,8 @@ struct Diagnostic {
   std::string text_;
 
   std::string long_formatted_text_;
+
+  std::vector< FixItChunk > fixits_;
```
Valloric: I'm probably missing something, but why is this not a vector of `FixIt`s instead of `FixItChunk`s?

cpp/ycm/ClangCompleter/TranslationUnit.cpp
```diff
@@ -389,6 +388,68 @@ void TranslationUnit::UpdateLatestDiagnostics() {
...
 }
 
+namespace {
+  /// Sort a FixIt container by its location's distance from a given column
+  /// (such as the cursor location)
+  ///
+  /// PreCondition: all FixIts in the container are on the same line
+  struct sort_by_location {
+    sort_by_location( int column )
+      : column_(column) {
```
Valloric: You can write this as `sort_by_location( int column ) : column_( column ) {}`.

Unless the formatter is making it this way. If so, leave it.

cpp/ycm/ClangCompleter/TranslationUnit.cpp
```diff
@@ -389,6 +388,68 @@ void TranslationUnit::UpdateLatestDiagnostics() {
...
+
+    }
+
+    bool operator()( const FixIt& a, const FixIt& b ) {
+      int a_distance = a.location.column_number_ - column_;
+      int b_distance = b.location.column_number_ - column_;
+
+      return std::abs(a_distance) < std::abs(b_distance);
```
Valloric: Whitespace inside parens.

cpp/ycm/ClangCompleter/TranslationUnit.cpp
```diff
@@ -389,6 +388,68 @@ void TranslationUnit::UpdateLatestDiagnostics() {
...
+
+  if ( reparse )
+    ReparseForIndexing( unsaved_files );
+
+  std::vector< FixIt > fixits;
+
+  {
+    unique_lock< mutex > lock( diagnostics_mutex_ );
+
+    for ( std::vector< Diagnostic >::const_iterator it
```
Valloric: Oh God I hate C++98 so much... we need to move the minimum to C++11 some time soon.

ycmd/responses.py
```diff
@@ -110,30 +110,51 @@ def BuildCompletionResponse( completion_datas, start_column ):
...
-      'column_num': location.column_number_,
-      'filepath': location.filename_,
-    }
+def BuildDiagnosticData( diagnostic ):
 
   kind = ( diagnostic.kind_.name if hasattr( diagnostic.kind_, 'name' )
            else diagnostic.kind_ )
 
+  fixits = ( diagnostic.fixits_ if hasattr( diagnostic, 'fixits_' )
+                              else [] )
```
Valloric: The indentation seems weird here a bit. Maybe put the `if` and what follows it on all on one separate line?

Valloric:
Holy crap this is great work! Thanks for doing this!

I've made minor comments on the changes; nothing big.

Don't forget to also bump the ycm_core version number since we're changing the API. We do this even for backwards compatible changes to notify users of significant new features that are only available after recompiling. 

With these issues addressed, this is ready to merge.

Valloric:
For the version changes, you need to change `versioning.cpp` and `EXPECTED_CORE_VERSION` files.

cpp/ycm/ClangCompleter/Diagnostic.h
puremourning: No problem. I'm just so used to things correcting my lazy typing. You'd think there was a vim setting for that... *google*

cpp/ycm/ClangCompleter/Diagnostic.h
puremourning: done

cpp/ycm/ClangCompleter/Diagnostic.h
puremourning: done

cpp/ycm/ClangCompleter/Diagnostic.h
puremourning: It could be I suppose. In ```YouCompleteMe::Diagnostic``` I'm just cacheing the chunks returned by libClang (in ```YouCompleteMe::BuildDiagnostic``` (ClangHelpers.cpp)) and actually building the ```YouCompleteMe::FixIt``` structure on-demand. The ```FixIt``` structure is really just a lightweight ```Diagnostic``` suitable for the response to ```GetFixItsForLocationInFile```.

I'm not married to this approach, it just seemed the least amount of code to write at the time. For now, I've updated the comments to clarify the usage, if you feel strongly, happy to change this, e.g. so that ```Diagnostic``` contains 0 or 1 ```FixIt``` objects which is checked by ```GetFixItsForLocationInFile``` (probably a *shudder* std::auto_ptr given our c++98-ish-ness)

cpp/ycm/ClangCompleter/TranslationUnit.cpp
puremourning: Nope, just habit. Fixed.

cpp/ycm/ClangCompleter/TranslationUnit.cpp
puremourning: fixed

cpp/ycm/ClangCompleter/TranslationUnit.cpp
puremourning: no kidding :+1:

ycmd/responses.py
puremourning: sure can do that. i was lining up the if and else blocks.

cpp/ycm/ClangCompleter/Diagnostic.h
Valloric: I don't feel strongly either way, I just wanted to understand. Code comment clarifying this would ensure others don't get confused.

puremourning:
I've updated in line with your comments and pushed a new commit.

ycmd/tests/test_utils.py
```diff
@@ -105,3 +105,26 @@ def StopGoCodeServer( app ):
                  BuildRequest( completer_target = 'filetype_default',
                                command_arguments = ['StopServer'],
                                filetype = 'go' ) )
+
+
+# Prints an object in a way to show its structure and content. Useful when
+# debugging complex responses (such as the diagnostic response) in tests.
+def PrettyPrint( d, indent=0 ):
```
Valloric: Could you replace this with either the `pprint` module or `json.dumps` with an `indent` param set? There doesn't seem to be a need for us to have our own version.

Valloric:
Thanks for addressing the issues I raised; I think you missed my comment about the `PrettyPrint` function accidentally so I pointed it out again. :)

ycmd/tests/test_utils.py
puremourning: Sure thing - i totally did miss that! Not are what happened - i just saw the other comments in my inbox as a separate mail... Fixing now

puremourning:
ok pushed fix for pprint. This is much better. To my shame, i trusted stack overflow :(

Valloric:
Thanks for the PR!

@homu r+

PR #168, opened 2015-06-29
use user specified port for omnisharp if provided

akhosravian:
Looks like the user specified omnisharp server port functionality (https://github.com/Valloric/ycmd/pull/82) got lost in the refactor of the cs completer.

I have signed the CLA.


Valloric:
Oops, nice catch!

@homu r+

PR #173, opened 2015-07-02
Set stdin to PIPE on Windows only when necessary

micbou:
See PR Valloric/YouCompleteMe#1563.

CLA signed.


Valloric:
This actually needed a tweak to prevent a traceback on Linux: https://github.com/Valloric/ycmd/commit/bfcfa21fb87789e88c803ac1f2b1ab766961f0b4

micbou:
My bad. I didn't think of this.

Valloric:
Eh, don't worry about it. It happens.

PR #175, opened 2015-07-14
Implement FixIt for C#

mispencer:
Implement https://github.com/Valloric/ycmd/pull/167 for C#.

I have signed the CLA.

### DIFF
--- a/ycmd/tests/testdata/testy/FixItTestCase.cs
+++ b/ycmd/tests/testdata/testy/FixItTestCase.cs
@@ -1,0 +1,24 @@
+using Microsoft.CSharp;
+using System;
+using System.Text;
+
+namespace testy {
+    public class FixItTestCase {
+        public FixItTestCase() {
+            var str = "";
+            str.EndsWith("A");
+            var i = 5;
+            const int j = i + 5;
+        }
+
+        public int One() {
+            var self = this;
+            return self.GetHashCode();
+        }
+
+        public int Two() {
+            var self = this;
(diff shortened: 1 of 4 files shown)


ycmd/tests/testdata/testy/FixItTestCase.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using System.Text;
+
+namespace testy {
+	public class GetTypeTestCase {
+		public GetTypeTestCase() {
+			var str = "";
+			str.EndsWith("A");
+			var i = 5;
+			var j = i + 5;
```
Valloric: Consider adding a comment in the file describing the expected fix-it.

Valloric:
This looks nice, thanks for sending it in!

@puremourning I'm interested in your thoughts on this implementation since you implemented the libclang support.

mispencer:
I updated the pull request to resolve issues if there wasn't any text changes or the text change affected the first or last character in the buffer. I also added some tests.

Regarding comments in the test data file: I would rather not, since comments are not negligible here - they affect the fix output.

Valloric:
> Regarding comments in the test data file: I would rather not, since comments are not negligible here - they affect the fix output.

Oh, that makes sense.

puremourning:
@Valloric this looks ok to me. I've tested with the client changes and incorporated them into PR https://github.com/Valloric/YouCompleteMe/pull/1555

Valloric:
Thanks you both!

@homu r+

PR #176, opened 2015-07-17
Bump Clang version to 3.6.2

micbou:
No changes in the `clang_includes` and `cpp/llvm/clang-c/include` folders. And no worries, I kept the Ubuntu version to 14.04 this time.

CLA signed.


Valloric:
Thanks for the PR!

@homu r+

PR #178, opened 2015-07-21
Make ClangCompleter's GoToDeclaration work when on the definition

RedBeard0531:
Previously it would return the current position unchanged.

Based on comments in the test, I think this was considered a bug. This is also required for another pull request I am about to make.


Valloric:
Nice, thanks!

@homu r+

PR #179, opened 2015-07-21
Make ClangCompleter's GoTo(Imprecise) commands on definition go to declaration

RedBeard0531:
This allows using a single key binding to GoTo (or GoToImprecise). If it
takes you to the definition, but you wanted to go to the declaration (eg,
because it has some useful comment), you just GoTo again.

This is on top of and depends on #178.

The current implementation only works if on the first character of the definition name. You will need to hit GoTo twice if in the middle. I think the solution to this would require making the C++ code aware of GoTo in addition to the GoTo forms it already handles. Let me know if you prefer that I do that.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/179)

### DIFF
--- a/cpp/ycm/ClangCompleter/ClangCompleter.cpp
+++ b/cpp/ycm/ClangCompleter/ClangCompleter.cpp
@@ -128,6 +128,25 @@ ClangCompleter::CandidatesForLocationInFile(
 }
 
 
+bool ClangCompleter::IsLocationOnDefinition(
+  const std::string &filename,
+  int line,
+  int column,
+  const std::vector< UnsavedFile > &unsaved_files,
+  const std::vector< std::string > &flags,
+  bool reparse ) {
+  ReleaseGil unlock;
+  shared_ptr< TranslationUnit > unit =
+    translation_unit_store_.GetOrCreate( filename, unsaved_files, flags );
+
+  if ( !unit ) {
+    return false;
+  }
+
+  return unit->IsLocationOnDefinition( line, column, unsaved_files, reparse );
+}
+
+
 Location ClangCompleter::GetDeclarationLocation(
   const std::string &filename,
   int line,
--- a/cpp/ycm/ClangCompleter/ClangCompleter.h
+++ b/cpp/ycm/ClangCompleter/ClangCompleter.h
@@ -59,6 +59,14 @@ public:
     const std::vector< UnsavedFile > &unsaved_files,
     const std::vector< std::string > &flags );
 
+  bool IsLocationOnDefinition(
+    const std::string &filename,
+    int line,
+    int column,
+    const std::vector< UnsavedFile > &unsaved_files,
+    const std::vector< std::string > &flags,
+    bool reparse = true );
+
   Location GetDeclarationLocation(
     const std::string &filename,
     int line,
--- a/cpp/ycm/ClangCompleter/TranslationUnit.cpp
+++ b/cpp/ycm/ClangCompleter/TranslationUnit.cpp
@@ -184,6 +184,27 @@ std::vector< CompletionData > TranslationUnit::CandidatesForLocation(
   return candidates;
 }
 
+bool TranslationUnit::IsLocationOnDefinition(
+  int line,
+  int column,
+  const std::vector< UnsavedFile > &unsaved_files,
+  bool reparse ) {
+  if ( reparse )
+    Reparse( unsaved_files );
+
+  unique_lock< mutex > lock( clang_access_mutex_ );
+
+  if ( !clang_translation_unit_ )
+    return false;
+
+  CXCursor cursor = GetCursor( line, column );
+
+  if ( !CursorIsValid( cursor ) )
+    return false;
+
+  return clang_isCursorDefinition( cursor );
+}
+
 Location TranslationUnit::GetDeclarationLocation(
   int line,
   int column,
--- a/cpp/ycm/ClangCompleter/TranslationUnit.h
+++ b/cpp/ycm/ClangCompleter/TranslationUnit.h
@@ -66,6 +66,12 @@ public:
     int column,
     const std::vector< UnsavedFile > &unsaved_files );
 
+  YCM_DLL_EXPORT bool IsLocationOnDefinition(
+    int line,
+    int column,
+    const std::vector< UnsavedFile > &unsaved_files,
+    bool reparse = true );
+
   YCM_DLL_EXPORT Location GetDeclarationLocation(
     int line,
     int column,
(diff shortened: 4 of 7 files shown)


Valloric:
It's kinda hard to review this since it has changes from #178 mixed in; I can't easily tell which changes are specific to this PR. When #178 is merged by homu, please rebase so that it's easier to see what's going on.

> The current implementation only works if on the first character of the definition name.

That sounds unfortunate. I think most users would conclude that the feature doesn't work since having to put the cursor on the first char of the word is non-obvious. What causes this limitation and can we do something so that users don't have to deal with it?

RedBeard0531:
This pull request is just for this commit: https://github.com/RedBeard0531/ycmd/commit/de37ab7f28f0947639a171e825fdbead1a47508f. I created one pull request per commit, so you can just look at the last commit on each to see what it is about.

The reason for the first character limitation is that it is done in python since the C++ code doesn't know about 'GoTo' separate from 'GoToDefinition' and 'GoToDeclaration'. In python it can only check if the location returned by GoToDefinition is the same as the position from the request. For my use-case this is fine since I bind Alt-] (similar to goto tag) to run YCMCompleter GoTo, and I just repeatedly hit Alt-] until it takes me where I want to go. If in the middle of the identifier it first takes me to the first char, then to the declaration.

If you want to make it work on the first GoTo from anywhere in the identifier I can think of two solutions:

1. Use the heuristic that anywhere on the same line is likely to be the same thing and therefor go to the declaration.
2. Add a "CanonicalizePosition" method in CPP which returns the location of the cursor created by clang. Then check if GoToDefinition returns that position. This shouldn't increase latency if the CanonicalizePosition call does the reparse and the GoTo* calls don't (similar to what I did in #184).

If you agree that this behavior is desirable, let me know which of these solutions you'd prefer (or if you have another idea).

vheon:
I would prefer the second approach, specially if we reuse the generated AST.

Valloric:
@RedBeard0531 At a glance, it seems that the second approach might be better. 

Also, thanks for pointing out that you've separated this out by commit; it's indeed easy to review just one commit.

RedBeard0531:
Actually on second thought, there is probably a better way. Any objection to just adding an IsLocationOnDefinition method to the C++ code and using that directly? As with solution two, it will handle the reparse and the GoTo* methods won't.

Valloric:
@RedBeard0531 Sounds reasonable.

Valloric:
@RedBeard0531 What's the state of this PR now?

RedBeard0531:
@Valloric I've been a bit swamped so I haven't had a chance to make the discussed changes. I'll find some time to do it this weekend.

Valloric:
@RedBeard0531 No rush, just checking that it hasn't been abandoned. Take your time.

RedBeard0531:
Rebased to latest master and switched to using IsLocationOnDefinition as described in my last comment.

vheon:
LGTM

---

Reviewed 7 of 7 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/179#-:-K5luYykW88lNI9VoO8M:676834233)*

micbou:
We should implement the same strategy for other completers. I added some comments on the tests.

We'll need to bump the version in the `CORE_VERSION` file since it introduces changes in the API.

---

Reviewed 1 of 7 files at r1.

---

*[ycmd/tests/clang/subcommands_test.py, line 150 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/179#-K5lzkU7NMY0nBzbJzpb:-K5lzkU8y-xZA693Ox3s:-1667314178) ([raw file](https://github.com/valloric/ycmd/blob/c721f5b9934ea69094c67b64723c1c10add0d7bb/ycmd/tests/clang/subcommands_test.py#L150)):*
Last `definition` should be `declaration`, like above.

---

*[ycmd/tests/clang/subcommands_test.py, line 176 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/179#-K5lzvOq1KhTwlh4CsgW:-K5lzvOq1KhTwlh4CsgX:78199588) ([raw file](https://github.com/valloric/ycmd/blob/c721f5b9934ea69094c67b64723c1c10add0d7bb/ycmd/tests/clang/subcommands_test.py#L176)):*
Same comment.

---

*[ycmd/tests/clang/subcommands_test.py, line 178 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/179#-K5m-0MZ6TndwwswyczW:-K5m-0MZ6TndwwswyczX:78199588) ([raw file](https://github.com/valloric/ycmd/blob/c721f5b9934ea69094c67b64723c1c10add0d7bb/ycmd/tests/clang/subcommands_test.py#L178)):*
Same comment.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/179#-:-K5m07deZQjML6cqmyF6:227353177)*

Valloric:
Agreed with @micbou, it's critical we don't forget to bump `CORE_VERSION`. Otherwise LGTM.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/179#-:-K5mrecXqKH8SWoRiOfz:-1934012152)*

puremourning:
Couple of super pedantic comments, but otherwise LGTM too with the core version bump.

---

Reviewed 7 of 7 files at r1.

---

*[ycmd/completers/cpp/clang_completer.py, line 184 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/179#-K5wmFzb8gne0A7acde0:-K5wmFzb8gne0A7acde1:635393445) ([raw file](https://github.com/valloric/ycmd/blob/c721f5b9934ea69094c67b64723c1c10add0d7bb/ycmd/completers/cpp/clang_completer.py#L184)):*
whitespace inside bracket `if ( not location`

---

*[ycmd/completers/cpp/clang_completer.py, line 188 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/179#-K5wmIq4u3iFdS3rzhCX:-K5wmIq4u3iFdS3rzhCY:1804752656) ([raw file](https://github.com/valloric/ycmd/blob/c721f5b9934ea69094c67b64723c1c10add0d7bb/ycmd/completers/cpp/clang_completer.py#L188)):*
whitespace inside bracket `reparse = False ) )`

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/179#-:-K5wnP-cp4s2E4Pxe2kS:890430958)*

Valloric:

PR #180, opened 2015-07-21
POC Cross-TranslationUnit GoToDefinition support

RedBeard0531:
This is an initial POC of cross-TU support for GoToDefinition. It works by extracting the USRs (clang's Unified Symbol Resolution) for each definition in the file after parsing the TU. If GoToDefinition fails on the TU, it will query all TUs in the TUStore to see if they contain the definition.

There is currently no support for serializing this information. This means that it can only go to definitions in files that are currently open and have been parsed since the last restart of the server.

This is on top of and complements well with #178 and #179. In my use case, I often want to go to the actual definition, but I'd also like a way to get to the documentation if there is any.

I am putting up a PR before writing tests to get initial feedback on if this is desired functionality and if you have a more preferred way of achieving it.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/180)

### DIFF
--- a/cpp/ycm/ClangCompleter/ClangCompleter.cpp
+++ b/cpp/ycm/ClangCompleter/ClangCompleter.cpp
@@ -181,7 +181,12 @@ Location ClangCompleter::GetDefinitionLocation(
     return Location();
   }
 
-  return unit->GetDefinitionLocation( line, column, unsaved_files, reparse );
+  Location loc = unit->GetDefinitionLocation( line, column, unsaved_files, reparse );
+  if ( loc.IsValid() )
+      return loc;
+
+  std::string usr = unit->GetDefinitionUSR( line, column );
+  return translation_unit_store_.LocationForUsr(usr);
 }
 
 std::string ClangCompleter::GetTypeAtLocation(
--- a/cpp/ycm/ClangCompleter/ClangCompleter.cpp
+++ b/cpp/ycm/ClangCompleter/ClangCompleter.cpp
@@ -128,6 +128,25 @@ ClangCompleter::CandidatesForLocationInFile(
 }
 
 
+bool ClangCompleter::IsLocationOnDefinition(
+  const std::string &filename,
+  int line,
+  int column,
+  const std::vector< UnsavedFile > &unsaved_files,
+  const std::vector< std::string > &flags,
+  bool reparse ) {
+  ReleaseGil unlock;
+  shared_ptr< TranslationUnit > unit =
+    translation_unit_store_.GetOrCreate( filename, unsaved_files, flags );
+
+  if ( !unit ) {
+    return false;
+  }
+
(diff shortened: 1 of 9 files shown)


oblitum:
Cool. I've researched something like this [quite some time ago](http://stackoverflow.com/q/13789814/1000282) but the speed prospects for per-TU traversing didn't make me happy at the time. Do you think it's speedy enough? To counter this effect I've thought of turning the TUStore into a cache so that last queried and/or most frequently accessed TUs stays at the top of the list.

puremourning:
I thought about the speed of indexing too. I'll check it out at work and let you know. We have a big slow San, shared servers and big files, with huge include paths. This usually shows up YCM slowness issues if they are there. In particular when doing sub commands which require full re parse (like gettype, fixit). Adding extra work to re parse will likely slow those ops. I will see if I notice.

oblitum:
@puremourning cool, I'd like to know.

Valloric:
Thanks for sending in a PR!

I'm not sure this is a good idea. We'd still end up with GoTo working "sometimes," except that the now it would also work if you have opened the TU with the definition in the same session, which doesn't sound like it would cover a lot of new cases... and all this with the new cost of a failed GoTo now taking much longer because it has to parse all TUs. In large codebases that have huge TUs, this would result in quite a bit of pain I think.

I don't think this is a net win. :(

@r4nt As a Clang dev, what do you think about this?

r4nt:
I think at some point we'll want to integrate with something like Kythe (https://github.com/google/kythe) for cross-TU code browsing.

telemenar:
I wonder if providing support for something more like rtags would be appropriate.
https://github.com/Andersbakken/rtags

It has the methodology of creating a database of all of the TUs in a project based on the compile_commands.json file and then not only give you GoTo but also symbol references.

r4nt:
@telemar - kythe does something similar in a cross-platform way. But I think if we can get a common interface for such cross-tu solutions so we can add support with a small shim around them that would also be really useful.

RedBeard0531:
Sorry for the radio silence. I've been working on this during my company's quarterly(ish) skunkworks
periods. I've rebased on top of master and the latest version of #179 and removed the C++11-isms.
Again, only the top commit on this branch is about this pull request. I've kept the branch with both
commits because they complement each other well and it makes it easier for people (including myself)
to pull a useful version.

I've been using this for the past several months as have a few of my coworkers. I've found it very
useful since I tend to have very long running vim sessions (months sometimes) so there was a decent
chance that the TU containing the definition had already been loaded. When it wasn't, it still took
me to the correct header and [a.vim](https://github.com/vim-scripts/a.vim) took me to the correct
cpp file. My experience has been that the most notable cases where it couldn't take me where I
wanted to go were when dealing with virtual interfaces, and it probably wouldn't be able do that
right without presenting a menu like :tjump does. I've fallen back to using the built-in ctags
support for those cases.

Valloric:

PR #184, opened 2015-07-27
When GoTo fails to find the definition, don't reparse when finding declaration

RedBeard0531:
It just reparsed when looking for the definition, and nothing could have changed since then since this is a blocking operation.

This is on top of my #179 PR since both change nearby lines. If you want this one and not the others, it is a simple manual fix.


Valloric:
Thanks for sending in a PR!

Please rebase when/if #179 is merged and I'll take another look.

Valloric:

PR #185, opened 2015-07-31
Note TypeScript support in README

icholy:
TypeScript 1.5 was finally released. I've been using this completer on a daily basis without any issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,8 +75,8 @@ non-semantic.
 There are also several semantic engines in YCM. There's a libclang-based
 completer that provides semantic completion for C-family languages.  There's also a
 Jedi-based completer for semantic completion for Python, an OmniSharp-based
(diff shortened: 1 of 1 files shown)


Valloric:
@homu r+

TsumiNa:
Hi! I'm new with ycmd. how can use a tsconfig.json-file to configure the tsserver. In typescript@1.6.2 now, there are no need to write `/// <reference path="xxx/xxx">`s.

puremourning:
I think we may need to update YCM's README.md. for now, take a look at our test data directory in `ycmd/tests/testdata`. I thing there is example in there.

icholy:
@TsumiNa https://github.com/Microsoft/TypeScript/wiki/tsconfig.json

Valloric:
Agreed with @puremourning, seems like we need some more docs in the README. Maybe a bit of info including a link to the doc @icholy provided?

PR #187, opened 2015-08-03
Encode contents in Typescript completer

micbou:
Contents should be encoded to UTF-8 if containing non-ascii characters before writing it to a temporary file in Typescript completer.

Fix issue Valloric/YouCompleteMe#1616.

CLA signed.


Valloric:
Great, thanks!

@micbou You've become an excellent and reliable contributor; keep up the amazing work!

@homu r+

PR #188, opened 2015-08-04
use release omnisharp

corngood:
Using the release configuration of OmniSharp.

*** There are currently bugs with the Release configuration that need to be fixed with https://github.com/OmniSharp/omnisharp-server/pull/206, so this shouldn't be merged until the OmniSharp fix is merged and the submodule updated.  I'll update this PR if/when that happens.

### DIFF
--- a/build.py
+++ b/build.py
@@ -165,7 +165,7 @@ def BuildOmniSharp():
     sys.exit( 'msbuild or xbuild is required to build Omnisharp' )
 
   sh.cd( p.join( DIR_OF_THIS_SCRIPT, 'third_party/OmniSharpServer' ) )
-  sh.Command( build_command )( _out = sys.stdout )
+  sh.Command( build_command )( "/property:Configuration=Release", _out = sys.stdout )
 
 
 def BuildGoCode():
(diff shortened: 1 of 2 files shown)


corngood:
This change now contains the submodule update to include the fix in omnisharp-server.

vheon:
What I didn't understand is what this should bring to the user and at this point why we were using the debug version of Omnisharp.

Valloric:
Thanks!

@homu r+

@vheon AFAIR for .Net projects the distinction between debug/release isn't really significant. There isn't a large perf difference as with C++ projects. [More info](http://stackoverflow.com/questions/4043821/performance-differences-between-debug-and-release-builds).

corngood:
I'm not sure why it was originally using Debug, but it could be because of the solution configuration problem I referenced above.  Release in the omnisharp .sln was building the silverlight configuration of some dependencies, which stopped it from locating mscorlib etc.

As far as performance goes, I unfortunately don't have hard numbers at the moment, but I did this originally because I was trying to profile a severe performance problem in one of the code issue modules (https://github.com/icsharpcode/NRefactory/issues/476).  This sort of thing is CPU limited managed code, so there is a decent improvement.

Issue #171, opened 2015-07-02
Make ycmd check ycm_core version for compatibility


Valloric:
ycmd has always had a compiled component (ycm_core) and a Python component (the rest of ycmd). YCM has always been checking that the API version that was [compiled into the ycm_core library](https://github.com/Valloric/ycmd/blob/master/cpp/ycm/versioning.cpp#L23) matches the version that [the Python code expects](https://github.com/Valloric/YouCompleteMe/blob/master/python/ycm/base.py#L176).

For historical reasons, this API version check is currently being done in the YCM client and not ycmd. Since ycmd now has multiple clients, there's no guarantee that the other clients are performing this check; in cases where they aren't, a poor user experience (unintelligible Python tracebacks) will result in case of API mismatch which can easily happen if they pull in a more recent version of the ycmd code without recompiling ycm_core. 

The idea is to move this check into ycmd.

From now on, ycmd will verify the API version and will exit with a status code of 2 if there's a mismatch. Since it might be burdensome for clients to first start ycmd and then examine did it crash after a few seconds, a separate script (`check_core_version.py`) will be provided whose sole job is to quickly check the API version and exit with code 0 on sucess and 2 on mismatch. This script can then be called by editors in a blocking fashion on startup.

@abingham @ptrv @mawww @Qusic

micbou:
I think it would be easier to maintain if there was only one place to set the core version. One way to achieve this would be to get the core version with cmake and create the corresponding preprocessor variable which will be used in the C++ code. I tried this approach in commit 48e0d3b242e8215f4e0f9abc9915647a44f11f8a and it is working well. I could send a PR if you agree with this.

Valloric:
I did in fact think of that approach, but the thing is that YCM has a different build system for the Google-internal build and reproducing that preprocessor variable approach there would be a pain.

I appreciate you trying this out though; your commit also reminded me that I forgot to add the copyright comment. :)

micbou:
Oh, I see. I will keep the commit just in case. Could be useful later.

r4nt:
I think we'd be able to do this in our internal build system, as long as there's a generated header we need to #include to get at the version (we can generate headers just fine)

Valloric:
@r4nt Oh we could, it would just be a hassle. I don't think it's worth it. I could be wrong.

micbou:
@Valloric From issue #1620, I was wondering if we could reconsider the approach discussed here where we only need to maintain the core version in one place.

[...]

PR #193, opened 2015-08-08
Set core version in one place

micbou:
Improve #171 by using cmake to build `ycm_core` with the version in the `CORE_VERSION` file.

Even if the preprocessor variable is defined by cmake, I am still defining it in `versioning.h` (if not already defined) with a bogus value to avoid a diagnostic error when editing the `versioning.cpp` file.

I could not find a good place to put the test so I created a new file for it.


cpp/ycm/versioning.h
```diff
@@ -15,8 +15,17 @@
 // You should have received a copy of the GNU General Public License
 // along with YouCompleteMe.  If not, see <http://www.gnu.org/licenses/>.
 
+#ifndef VERSIONING_H_EEJUU0AH
+#define VERSIONING_H_EEJUU0AH
+
+#if !defined( YCMD_CORE_VERSION )
+  #define YCMD_CORE_VERSION 0
```
Valloric: Add a comment above this stating that the true value is set in a compiler command-line flag, and that we do so in the CMakeLists.txt file.

Valloric:
Thanks for the PR! 

This is pretty nice. I added a minor inline comment; when that's resolved, this will be ready to merge.

micbou:
I added the comment.

Valloric:
Thanks!

@homu r+

@r4nt This will need special care when you pull it in.

PR #194, opened 2015-08-08
Fixed example for installing requirements

czipperz:


micbou:
You should not add `sudo`. This is platform specific.

czipperz:
True.

PR #195, opened 2015-08-09
Added note to run as admin/super user

czipperz:


Valloric:
Thanks!

@homu r+

czipperz:
What does homu do? Test for bugs?

vheon:
http://homu.io/

czipperz:
Interesting

PR #196, opened 2015-08-14
make cmake use env python

amchoukir:
CustomPythonCmakeArgs forces CMake to pickup the environment python libraries and include but the interpreter is still picked using PythonInterp which picks up the system default instead of the environment python. This cause issues with vitualenv.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/196)

### DIFF
--- a/build.py
+++ b/build.py
@@ -84,7 +84,7 @@ def CustomPythonCmakeArgs():
   return [
     '-DPYTHON_LIBRARY={0}'.format( python_library ),
     '-DPYTHON_INCLUDE_DIR={0}'.format( python_include ),
-    '-DPYTHON_EXECUTABLE={0}'.format(python_executable)
+    '-DPYTHON_EXECUTABLE={0}'.format( python_executable )
   ]
 
 


Valloric:
Hm, this has the potential to break everything. The whole multiple-versions-of-python-on-one-machine thing is the single most frequent reason for failed YCM installations.

So we need to be _really_ sure this is doing the right thing. I'd appreciated it if various people could try out this change; the more OS's/distros we cover, the better.

@vheon @micbou @puremourning

micbou:
The `CustomPythonCmakeArgs` function is only called on OS X so Linux and Windows are not affected by this change.

puremourning:
happy to give this a spin as i have 2 macs - 1 which the install always works and 1 which i always have do battle with homebrew/python/etc.

However, @amchoukir could you elaborate on _exactly_ what problem this is solving? Does it relate to an issue on the tracker? Are there some obvious test steps we should go through?

Ta,
Ben

amchoukir:
Thanks for having a look and giving it a try. 

@puremourning, @Valloric 

`CustomPythonCmakeArgs` uses `python-config` to figure out where python is installed in the current environment. Unfortunately `virtualenv` does not copy `python-config` when setting up a new environment, I will take this issue separately to the `virtualenv` github repo. Copying `python-config` to my `virtualenv` takes care of the `PythonLibs`.

But there is still an issue with `PythonInterp`  as can be seen from the install log below:

```
-- Found PythonLibs: /usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/libpython2.7.dylib (found suitable version "2.7.10", minimum required is "2.6")
-- Found PythonInterp: /usr/bin/python2.6 (found suitable version "2.6.9", minimum required is "2.6")
```

`PythonInterp` picks up the system default instead of the interpreter in my `virtualenv`. The fix I proposed take care of the issue in a similar way as `PythonLibs`.

My setup is as follows:

OS X 10.10.3 (Maverick)
Python 2.7.10 installed via homebrew
Macvim 7.4 installed via homebrew

The test steps I went through are:
 * Make sure the same version of the lib and interpreter are picked during the install
 * Check that you don't get vim crashing upon startup.

vheon:
To my understanding `virtualenv` is a tool for python available not only on OSX, but the fix you're proposing is implemented in a function that is called only on OSX, so is the fix not needed on other platform?

build.py
```diff
@@ -79,10 +79,12 @@ def CustomPythonCmakeArgs():
     else:
       python_library = '{0}.dylib'.format( lib_python )
     python_include = '{0}/include/{1}'.format( python_prefix, which_python )
+    python_executable = '{0}/bin/python'.format( python_prefix )
 
   return [
     '-DPYTHON_LIBRARY={0}'.format( python_library ),
-    '-DPYTHON_INCLUDE_DIR={0}'.format( python_include )
+    '-DPYTHON_INCLUDE_DIR={0}'.format( python_include ),
+    '-DPYTHON_EXECUTABLE={0}'.format(python_executable)
```
Valloric: Whitespace inside parens. :)

amchoukir:
@vheon yes it might definitely be needed. I have access to linux servers on which I could try it out, if someone could have a look at windows.

vheon:
> @vheon yes it might definitely be needed.

Then this PR should be modified to run on linux as well right?

puremourning:
I think we should hold out on this change while I get the Travis OSX builds working. I've had to do some work to streamline the build which I think obsolete this change.

Valloric:
@puremourning Sure, np.

amchoukir:
Here are the results of my tests on Linux

Scenario 1: System default sanity check

-- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython2.7.so (found suitable version "2.7.6", minimum required is "2.6")
-- Found PythonInterp: /home/vagrant/.virtualenv/ycmd/bin/python2 (found suitable version "2.7.6", minimum required is "2.6")

Scenario 2: Vanilla virtualenv with 2.7.10

-- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython2.7.so (found suitable version "2.7.6", minimum required is "2.6")
-- Found PythonInterp: /home/vagrant/.virtualenv/ycmd2.10/bin/python2 (found suitable version "2.7.10", minimum required is "2.6")

Wrong library picked up. It picked up the  system default instead of the 2.7.10.

Scenario 3: virtualenv with 2.7.10 and python-config copied over

-- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython2.7.so (found suitable version "2.7.6", minimum required is "2.6")
-- Found PythonInterp: /home/vagrant/.virtualenv/ycmd2.10/bin/python2 (found suitable version "2.7.10", minimum required is "2.6")

Wrong library picked up. It picked up the  system default instead of the 2.7.10. So here python-config does not fix the issue. 

Here are the specifics of python-config

(ycmd2.10)vagrant@vvv:~/ycmd$ python-config --prefix
/usr/local/python/2.10
(ycmd2.10)vagrant@vvv:~/ycmd$ which python-config
/home/vagrant/.virtualenv/ycmd2.10/bin/python-config

Most likely this is due to the difference in paths between OS X and Linux

I will dig into build.py to check what happens.

puremourning:
> @amchoukir Unfortunately virtualenv does not copy python-config when setting up a new environment, I will take this issue separately to the virtualenv github repo. Copying python-config to my virtualenv takes care of the PythonLibs.

Yes, this is annoying. I found the same problem when making the ycmd build/test run on Travis OS X.  Did you raise this with virtualenv folks? If so could we get a reference?

Thanks,
Ben

amchoukir:
@puremourning yes Paul opened the following issue:
https://github.com/pypa/virtualenv/pull/783

amchoukir:
Hello Everyone, I am back from vacation and I will have a look at why the build does not work on virtualenv for Linux.

Meanwhile the fix to get the python-config copied in virtualenv has been merged:

https://github.com/pypa/virtualenv/pull/798

PR #197, opened 2015-08-15
Pass client_data parameter to the FlagsForFile function of .ycm_extra_conf.py file for #include completion

micbou:
Fix issue Valloric/YouCompleteMe#1631.

Although it concerns C-family completions, it does not depend on the Clang completer but on the filename completer so I added the test to the (new) FilenameCompleter group.

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -97,8 +97,8 @@ class Flags( object ):
       return sanitized_flags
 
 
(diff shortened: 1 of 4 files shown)


puremourning:
top work. this is pretty much exactly the change i had in mind. :+1:

vheon:
Looks good :+1:

Valloric:
Nice! 

@homu r+

Issue #67, opened 2014-12-19
Add option to fall-back to identifier completion during parsing


r4nt:
See:
https://github.com/abingham/emacs-ycmd/issues/121#issuecomment-67377618

puremourning:
there is some ongoing talk about this over here: https://github.com/Valloric/YouCompleteMe/issues/1506

@r4nt I'd appreciate your input on the approach suggested. It _does_ involve an API change, but it would also cover this use case.

r4nt:
Lots of good arguments for both approaches. I'm not sure what the right
approach is. Perhaps making c++ faster to parse *is* the right solution :)

vheon:
Shouldn't this be resolved?

puremourning:
Yup. Fixed by https://github.com/Valloric/ycmd/commit/74a62c3ee3911c5a703b22a1b1599336195eb39d

PR #198, opened 2015-08-17
Fall back to general completions when semantic engine has no suggestions

puremourning:
TL;DR
=====

- supply general completions (alongside any errors raised) when the semantic completer (`clang_completer`, `jedi_completer`, `omnisharp_completer`, etc.) returns no results for the query and semantic completion is not forced
- tag results from the identifier completer with `[ID]`
- add tests for the above cases

Background
=========

Submitted in reference to the following issues:

* https://github.com/Valloric/YouCompleteMe/issues/1506
* https://github.com/Valloric/YouCompleteMe/issues/1225
* https://github.com/Valloric/ycmd/issues/67

The semantic engine is called either if the user enters a semantic trigger, or if they _force_ semantic completion. There may be many reasons why the semantic engine might return no results, but most commonly:

* there are errors in the file, and the semantic engine cannot parse it
* the semantic engine is not sophisticated enough (or it is not possible) to present a meaningful, correct or complete set of semantic suggestions (e.g. c++ temples, c macros, python duck-typing)
* the user's query string does not match any semantic completion suggestion

Previously, in all of the above cases, YCMD would return either no suggestions, or an error. This could make it seem like YCMD was not functioning correctly. Indeed in some cases (as discussed in the issues) this can lead to awkward workflow problems.

API Changes
==========

The `get_completions` request may now optionally contain a list `errors` containing the details of 0 or more errors raised by the semantic completion engine in addition to (a possibly empty) list of completion suggestions. The contents of each entry is identical to the `get_completions` response when no completion suggestions are returned, and only an exception is returned. That is:

```
EXCEPTION :=
{
    'exception' : {
        'TYPE': "<python exeption type name>"
    },
    'message': "<error description>",
    'traceback': "<stack trace of where the exception was thrown>"
}

COMPLETION :=
{
    'insertion_text': "completion text to be inserted",
    'extra_menu_info': "additional display information, e.g. type, [ID], [file], <snip>",
    'menu_text': "alternative text to display in the menu, in place of insertion_text",
    'detailed_info': "additional info for separate display, such as documentation/overloads, etc."
    'kind': "character defining type of suggestion",
    'extra_data': "completer-specific additional data. currently not documented",
}

COMPLETION_SUGGESTIONS_RESPONSE :=
{
    'completions': [ <potentially empty list of COMPLETION> ],
    'completion_start_column': column index of start point for completion suggestion,
    'errors': [ <potentially empty list of EXCEPTION> ]
}

EXCEPTION_RESPONSE := EXCEPTION

GET_COMPLETIONS_RESPONSE := EXCEPTION_RESPONSE | COMPLETION_SUGGESTIONS_RESPONSE
```

Rationale
=======

YCMD's `README.md` says:

>There are two main use-cases for code-completion:
>
>1. The user knows which name they're looking for, they just don't want to type
>   the whole name.
>2. The user either doesn't know the name they need or isn't sure what the name
>   is. This is also known as the "API exploration" use-case.
>
>The first use case is the most common one and is trivially addressed with the
>identifier completion engine (which BTW is blazing fast). The second one needs
>semantic completion.

This change is pretty much 100% targeted at the first use-case.

I think it is arguable that it is in keeping with YCMD's ethos of 'the user knows what they want most of the time' that in all of the above listed cases for no semantic suggestions, offering identifier completions is better than offering nothing.

This behaviour works well for dynamic languages, in particular those using 'duck typing' or similar semantics, such as python. This naturally extends to compiled languages using compile-time duck typing (c++ templates, c macros).

However, it is totally legitimate concern that users typing a semantic trigger expect a certain amount of accuracy from the completion suggestions (the 2nd use case in README.md). This is particularly important in the case where the query string does not match any of the semantic completions supplied by the semantic engine, and we filter them out. In order to help with this mental model, we tag all identifier completions with `[ID]` in the completion menu. There is then a subtle, but clear indication that the suggestion is _not_ from the semantic engine. Existing "general" completer already tag their suggestions with `[File]`, `[Dir]` and `[snip]` etc.

What this change does
==================

When the semantic engine is called after a trigger (*not* after forced completion), and either no completion suggestions are returned or the completer throws an exception, we query the "general" completer (typically the identifier completer) and return those suggestions instead. 

In order to make this more obvious to the user, we mark suggestions from the
identifier completer with [ID] in completions menu. 

In order to preserve the behaviour of showing exceptions thrown by the completer to the user (e.g. in case they have an error in their `.ycm_extra_conf.py`, we return the exceptions as well as the identifier completions and print this exception to the log file.

Contentious issues
===============

As discussed on the above tracker issues, the general behaviour is somewhat divisive. I personally, having been using it for some time and *really* like it. I'm almost certainly going to maintain it in my fork. I thought about adding a switch (on by default), but thought I would avoid that because more switches is a _bad thing_(TM).

Also, this is a change in the YCMD API and as such requires minor changes in clients to display these additional errors. The `README.md` says that there is a SemVer API version, but i wan't able to find one other than the `CORE_VERSION`

### DIFF
--- a/ycmd/handlers.py
+++ b/ycmd/handlers.py
@@ -34,6 +34,7 @@ import logging
 import json
 import bottle
 import httplib
+import traceback
 from bottle import request, response
 import server_state
 from ycmd import user_options_store
@@ -89,17 +90,40 @@ def RunCompleterCommand():
 def GetCompletions():
   _logger.info( 'Received completion request' )
   request_data = RequestWrap( request.json )
-  do_filetype_completion = _server_state.ShouldUseFiletypeCompleter(
-    request_data )
+  ( do_filetype_completion, forced_filetype_completion ) = (
+                    _server_state.ShouldUseFiletypeCompleter(request_data ) )
   _logger.debug( 'Using filetype completion: %s', do_filetype_completion )
-  filetypes = request_data[ 'filetypes' ]
-  completer = ( _server_state.GetFiletypeCompleter( filetypes ) if
-                do_filetype_completion else
-                _server_state.GetGeneralCompleter() )
 
-  return _JsonResponse( BuildCompletionResponse(
-      completer.ComputeCandidates( request_data ),
-      request_data.CompletionStartColumn() ) )
+  errors = None
+  completions = None
+
+  if do_filetype_completion:
+    try:
+      completions = ( _server_state.GetFiletypeCompleter(
+                                  request_data[ 'filetypes' ] )
+                                 .ComputeCandidates( request_data ) )
+
+    except Exception as exception:
+      if forced_filetype_completion:
+        # user explicitly asked for semantic completion, so just pass the error
+        # back
+        raise
+      else:
+        # store the error to be returned with results from the identifier
+        # completer
+        stack = traceback.format_exc()
+        _logger.error( 'Exception from semantic completer (using general): ' +
+                        "".join( stack ) )
+        errors = [ BuildExceptionResponse( exception, stack ) ]
+
+  if not completions and not forced_filetype_completion:
+    completions = ( _server_state.GetGeneralCompleter()
+                                 .ComputeCandidates( request_data ) )
+
+  return _JsonResponse(
+      BuildCompletionResponse( completions if completions else [],
+                               request_data.CompletionStartColumn(),
+                               errors = errors ) )
 
 
 @app.get( '/healthy' )
--- a/ycmd/server_state.py
+++ b/ycmd/server_state.py
@@ -99,12 +99,30 @@ class ServerState( object ):
 
 
   def ShouldUseFiletypeCompleter( self, request_data ):
+    """
+    Determines whether or not the semantic completer should be called, and
+    returns an indication of the reason why. Specifically, returns a tuple:
+    ( should_use_completer_now, was_semantic_completion_forced ), where:
+     - should_use_completer_now: if True, the semantic engine should be used
+     - was_semantic_completion_forced: if True, the user requested "forced"
+                                       semantic completion
+
+    was_semantic_completion_forced is always False if should_use_completer_now
+    is False
+    """
     filetypes = request_data[ 'filetypes' ]
     if self.FiletypeCompletionUsable( filetypes ):
-      return ( ForceSemanticCompletion( request_data ) or
-               self.GetFiletypeCompleter( filetypes ).ShouldUseNow(
-                 request_data ) )
-    return False
+      if ForceSemanticCompletion( request_data ):
+        # use semantic, and it was forced
+        return ( True, True )
+      else:
+        # was not forced. check the conditions for triggering
+        return ( self.GetFiletypeCompleter( filetypes ).ShouldUseNow(
+                   request_data ), False )
+
+    # don't use semantic, ignore whether or not the user requested forced
+    # completion
+    return ( False, False )
 
 
   def GetGeneralCompleter( self ):
--- a/ycmd/tests/get_completions_test.py
+++ b/ycmd/tests/get_completions_test.py
@@ -27,14 +27,15 @@ from .test_utils import ( Setup, BuildRequest, PathToTestFile,
 from webtest import TestApp, AppError
 from nose.tools import eq_, with_setup
 from hamcrest import ( assert_that, has_item, has_items, has_entry, has_entries,
-                       contains_inanyorder, empty, greater_than,
+                       contains, contains_inanyorder, empty, greater_than,
                        contains_string )
 from ..responses import  UnknownExtraConf, NoExtraConfDetected
 from .. import handlers
+from ..completers.cpp.clang_completer import NO_COMPLETIONS_MESSAGE
 import bottle
+import pprint
 
 bottle.debug( True )
-
 
 def CompletionEntryMatcher( insertion_text, extra_menu_info = None ):
   match = { 'insertion_text': insertion_text }
@@ -48,6 +49,69 @@ def CompletionLocationMatcher( location_type, value ):
                     has_entry( 'location',
                                has_entry( location_type, value ) ) )
 
+
+def ErrorMatcher( cls, msg ):
+  return has_entries( {
+    'exception' : has_entry( 'TYPE', cls.__name__ ),
+    'message': msg,
+  } )
+
+NO_COMPLETIONS_ERROR = ErrorMatcher( RuntimeError, NO_COMPLETIONS_MESSAGE )
+
+@with_setup( Setup )
+def GetCompletions_RunTest( test ):
+  """
+  Method to run a simple completion test and verify the result
+
+  Note: uses the .ycm_extra_conf from general_fallback/ which:
+   - supports cpp, c and objc
+   - requires extra_conf_data containing 'filetype&' = the filetype
+
+  this should be sufficient for many standard test cases
+
+  test is a dictionary containing:
+    'request': kwargs for BuildRequest
+    'expect': {
+       'response': server response code (e.g. httplib.OK)
+       'data': matcher for the server response json
+    }
+  """
+  app = TestApp( handlers.app )
+  app.post_json( '/load_extra_conf_file',
+                 { 'filepath': PathToTestFile(
+                                  'general_fallback/.ycm_extra_conf.py' ) } )
+
+  contents = open( test[ 'request' ][ 'filepath' ] ).read()
+
+  def CombineRequest( request, data ):
+    kw = request
+    request.update( data )
+    return BuildRequest( **kw )
+
+  # Because we aren't testing this command, we *always* ignore errors. This is
+  # mainly because we (may) want to test scenarios where the completer throws
+  # an exception and the easiest way to do that is to throw from within the
+  # FlagsForFile function.
+  app.post_json( '/event_notification',
+                 CombineRequest( test[ 'request' ], {
+                                   'event_name': 'FileReadyToParse',
+                                   'contents': contents,
+                                 } ),
+                 expect_errors = True )
+
+  # We also ignore errors here, but then we check the response code ourself.
+  # This is to allow testing of requests returning errors.
+  response = app.post_json( '/completions',
+                            CombineRequest( test[ 'request' ], {
+                              'contents': contents
+                            } ),
+                            expect_errors = True )
+
+  print 'completer response: ' + pprint.pformat( response.json )
+
+  eq_( response.status_code, test[ 'expect' ][ 'response' ] )
+
+  assert_that( response.json, test[ 'expect' ][ 'data' ] )
 
 @with_setup( Setup )
 def GetCompletions_RequestValidation_NoLineNumException_test():
@@ -80,8 +144,8 @@ def GetCompletions_IdentifierCompleter_Works_test():
 
   eq_( 1, response_data[ 'completion_start_column' ] )
   assert_that( response_data[ 'completions' ],
-               has_items( CompletionEntryMatcher( 'foo' ),
-                          CompletionEntryMatcher( 'foogoo' ) ) )
+               has_items( CompletionEntryMatcher( 'foo', '[ID]' ),
+                          CompletionEntryMatcher( 'foogoo', '[ID]' ) ) )
 
 
 @with_setup( Setup )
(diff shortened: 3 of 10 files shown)


Valloric:
I'm fine with the idea of this change (haven't looked at the implementation yet).

> I thought about adding a switch (on by default), but thought I would avoid that because more switches is a bad thing(TM).

Your instinct to _not_ add Yet Another User-configurable Option was indeed correct. :)

> Also, this is a change in the YCMD API and as such requires minor changes in clients to display these additional errors.

Yes. It would be great if you could summarize such API changes here so that @abingham, @ptrv and others can quickly see what's changing.

> The README.md says that there is a SemVer API version, but i wan't able to find one other than the CORE_VERSION

We're still in the 0.x phase where anything goes. Still, we try not to break people.

P.S. I spell it as "ycmd", not "YCMD". So YCM is all-caps (since it's an acronym), but ycmd is lowercase (since it's a full name).

ycmd/handlers.py
```diff
@@ -89,17 +90,39 @@ def RunCompleterCommand():
...
-  filetypes = request_data[ 'filetypes' ]
-  completer = ( _server_state.GetFiletypeCompleter( filetypes ) if
-                do_filetype_completion else
-                _server_state.GetGeneralCompleter() )
 
-  return _JsonResponse( BuildCompletionResponse(
-      completer.ComputeCandidates( request_data ),
-      request_data.CompletionStartColumn() ) )
+  general_completer = _server_state.GetGeneralCompleter()
+  kw = {}
```
Valloric: `kw` is a bit short. I can't tell at-a-glance what this variable is about which is a solid hint that it should be longer.

ycmd/handlers.py
```diff
@@ -89,17 +90,39 @@ def RunCompleterCommand():
...
+    try:
+      completions = ( _server_state.GetFiletypeCompleter(
+                                  request_data[ 'filetypes' ] )
+                                 .ComputeCandidates( request_data ) )
+
+      if not completions and not forced_filetype_completion:
+        completions = general_completer.ComputeCandidates( request_data )
+    except Exception as e:
+      if not forced_filetype_completion:
+        completions = general_completer.ComputeCandidates( request_data )
```
Valloric: This repeats the code from above. I think you can restructure this to avoid the repetition.

Maybe something like:
```
completions = None
if do_filetype_completion:
  try:
    completions = <get semantic>
  except:
    if forced:
      raise
    else:
      <store error>
if not completions and not forced:
  completions = <get ident>
return completions + error data if any
```

ycmd/handlers.py
```diff
@@ -89,17 +90,39 @@ def RunCompleterCommand():
...
+        kw[ 'errors' ] = [ BuildExceptionResponse( e, tb ) ]
+      else:
+        raise
+  else:
+    completions = general_completer.ComputeCandidates( request_data )
+
+  return _JsonResponse(
+      BuildCompletionResponse( completions,
+                               request_data.CompletionStartColumn(),
+                               **kw ) )
```
Valloric: With the pseudo-code I posted above, you could probably also get rid of the whole `**kw` thing and just do `errors = errors` here. Should be easier for readers to understand.

ycmd/server_state.py
```diff
@@ -101,10 +101,12 @@ def ShouldUseGeneralCompleter( self, request_data ):
   def ShouldUseFiletypeCompleter( self, request_data ):
```
Valloric: Consider adding a comment here noting that a (bool, bool) is returned and what each of them is.

ycmd/tests/get_completions_test.py
```diff
@@ -330,9 +1129,17 @@ def GetCompletions_CsCompleter_NonForcedReturnsNoResults_test():
                                   column_num = 21,
                                   force_semantic = False,
                                   query = 'Date' )
-  results = app.post_json( '/completions', completion_data ).json[ 'completions' ]
-
-  assert_that( results, empty() )
+  results = app.post_json( '/completions', completion_data ).json
+
+  # there are no semantic completions. However, we fall back to identifier
+  # completer in this case.
```
Valloric: It's probably a better idea to not mix the fallback to identifiers into this. Use forced completion so that you can ignore it.

ycmd/tests/testdata/general_fallback/lang_c.c
```diff
@@ -0,0 +1,85 @@
...
+static void do_something( int );
+static void do_another_thing( int );
+
+// TESTCASE1: use of . on macro parameter. Macros used this way are compile-time
+// duck-typing. Note this in this particular instance, libclang actually
+// offers semantic completions if there is only call to this macro
+// (e.g. DO_SOMETHING_TO( a_test ) - clever little shrew), so we don't call it
+
+/*       1         2         3         4
+1234567890123456789012345678901234567890123456789 */
```
Valloric: What's the purpose of this list of numbers? Is it to facilitate seeing what column is where? If that's the case, readers should configure their editor to show them a column number in the status line. :) The numbers are kinda distracting.

ycmd/tests/get_completions_test.py
```diff
@@ -115,9 +123,800 @@ def GetCompletions_IdentifierCompleter_WorksForSpecialIdentifierChars_test():
...
+  Tests for:
+    - [X] multiple languages: (use cases)
+      - [X] C:      macros
+      - [X] C++:    templates
+      - [X] Python: properties set from external method
+    - [X] Forced vs. Triggered semantic completion
+    - [X] Identifiers tagged with '[ID]'
+    - [X] Filter/sort based on query
+    - [X] Min chars for completion (only applies to ID)
+    - [X] Extra conf throws exception
```
Valloric: I have a feeling you're testing too many things at once.

For instance, I'd have just one test that tests the fallback (for say C++). The codepath that implements the fallback is agnostic of the semantic engine, so testing that the fallback works for various semantic engines seems overkill to me.

Likewise, for semantic engine tests, just use force_semantic to not have to test the fallback logic. That's tested in the fallback test.

A good design idea for tests is that if one logical thing fails in the code, one test should fail. That's often not achievable, but something we should strive for. The idea is that if one logical component fails and 100 tests fail because of it, tracking down exactly which component failed because non-trivial.

ycmd/tests/testdata/general_fallback/lang_c.c
puremourning: Yeah, it's a ruler. I find reading the tests and the associated code quite a painfully slow and cumbersome when you have to read 2 numbers and then navigate to that exact spot in the file.

You can get vim to show column numbers, of course, but you still have to pretty much put the cursor there. The use case for the ruler is a vertical split where i'm writing (or reading) the test in one split and i can just visually see (via set nu and this ruler) the exact spot where the test needs to trigger (or is already triggering).

I'm happy to remove them if you feel strongly about it. Of course removing them changes all the offsets which is a bit tedious, but hey ho :)

ycmd/tests/get_completions_test.py
puremourning: Interesting perspective. Happy to change this.

FWIW my perspective has always been that it's really useful to write regression tests (in particular integration tests testing an interface - as these are) from the functional perspective, so that if the code is changed (arbitrarily) then the tests should still work. I've always extended that to mean "even if the underlying codepath is similar, then the functional test still stands". i.e. if we made a change to the python completer which somehow interfaced badly with this, then we'd catch it.

Like I said, happy to change this.

puremourning:
Thanks for the review, @Valloric. I'm back at work now, so might be a few days before i can finish it up.

puremourning:
> P.S. I spell it as "ycmd", not "YCMD". So YCM is all-caps (since it's an acronym), but ycmd is lowercase (since it's a full name).

Yeah, i know, but safari has a habit of auto-"correcting" ycmd to something else, so i've got in the habit of typing it in upper case. I suppose I should find a way of telling it ycmd is a real word instead :p

ycmd/tests/testdata/general_fallback/lang_c.c
Valloric: I'm not crazy enough to feel strongly about them. Feel free to leave them in. :)

ycmd/tests/get_completions_test.py
Valloric: That perspective is exemplary but in my experience leads to tests that can be hard to maintain and a general test suite that isn't as helpful as it could be. I've worked in a codebase with tons of tests all testing the same thing indirectly, and when one such important component failed, it was a nightmare to figure out what actually went wrong. You'd know _something_ went south because 100+ tests failed, but good luck figuring it out.

Since then I've focused on the mantra that "all functionality should be tested while trying to minimize test path duplication."

Think of it like a compiler error: if you forget a semicolon, would you rather see a spew of 100+ errors or just one saying "hey you forgot a semicolon here"? C++ for instance has compilers that spew out an ungodly mess of text on incorrect template usage and figuring out what exactly went wrong is hard.

> my perspective has always been that it's really useful to write regression tests (in particular integration tests testing an interface - as these are) from the functional perspective, so that if the code is changed (arbitrarily) then the tests should still work.

Black-box testing can be very useful, but isolated unit-tests IMO are more useful. But the principle you're espousing of "tests should still work if the code is changed" is absolutely something that's critical and IMO not at odds with my approach.

Valloric:
> Thanks for the review, @Valloric. I'm back at work now, so might be a few days before i can finish it up.

How _dare you_ sir have a life outside of open-source! :D

Seriously, take all the time you need. We're not in a rush.

ycmd/tests/get_completions_test.py
puremourning: Hmmm. The following tests seem to be explicitly testing "non-forced" semantic completions (or the interaction of forced and non-forced in the latter case). So i'm tempted to leave them as-is, so as not to interfere with the idea behind the original test.

```python
def GetCompletions_CsCompleter_NonForcedReturnsNoResults_test():
def GetCompletions_CsCompleter_ForcedDividesCache_test():
```

These ones test exceptions from the completer, and in particular the common-ish case where there is no .ycm_extra_conf (and therefore not compatible with my generic `GetCompletions_RunTest` function), but i think similar tests do exist, so i've reverted the changes:

```python
GetCompletions_ClangCompleter_ExceptionWhenNoFlagsFromExtraConf_test
GetCompletions_ClangCompleter_ExceptionWhenNoFlagsFromExtraConf_ID_test
```

What do you think?

ycmd/tests/get_completions_test.py
puremourning: I’m going to trim down the tests to just:

- one language (clang completer-based)
- forced vs triggered semantic (fallback only in 2nd case)
- exception thrown, vs. just no completions for query
- min chars for identifier completion

Would anyone object to keeping the testdata files as they are? This is not just laziness, but I do find it useful to open the testsdata/* files to do manual testing of a particular feature (almost as a demo). And the test files i wrote mostly for manual testing in the first instance to ensure the feature feels right and works sensibly.

puremourning: Done. I've trimmed the tests to just one each covering each relevant (IMO) code path.

ycmd/server_state.py
puremourning: Added method docstring.

ycmd/handlers.py
puremourning: Yep done.

ycmd/handlers.py
puremourning: done.

ycmd/handlers.py
puremourning: done.

puremourning:
> Yes. It would be great if you could summarize such API changes here so that @abingham, @ptrv and others can quickly see what's changing.

Done above. I'm not aware of any "official" ycmd interface docs, so i wrote a (very) quick spec. of the `get_completions` response above, with a summary of the changes. I suppose we should add "document the API" to the list of things to do in the future :)

I've pushed new commit addressing (most of) your comments.

Thanks,
ben

ycmd/tests/get_completions_test.py
Valloric: You've been looking at these tests more recently than I have and you make good points; I'll leave the decision of what tests should be changed up to you. I trust your judgment.

Valloric:
> I suppose we should add "document the API" to the list of things to do in the future :)

Ha! Right you are. I've been "meaning to do that" for quite a while now. :D

From my end, this is ready to merge. I'm not sure did you want to do something else on it as well before I merged it, so ping back if this is ready.

puremourning:
@Valloric yep this is ready to merge.

Valloric:
@homu r+

PR #199, opened 2015-08-19
Fix RemoveXclangFlags test

micbou:
This test was not run cause of the missing `_test` part and was not correctly written.


Valloric:
Ah crap. Thanks for fixing this!

@homu r+

PR #200, opened 2015-08-19
Add -imacros to the list of include flags

micbou:
Fix issue Valloric/YouCompleteMe#1637.


ycmd/completers/cpp/tests/flags_test.py
```diff
@@ -153,18 +153,27 @@ def RemoveUnusedFlags_RemoveFlagWithoutPrecedingDashFlag_test():
 
 
 def RemoveUnusedFlags_RemoveFilenameWithoutPrecedingInclude_test():
-  expected = [ '-I', '/foo/bar', '-isystem/zoo/goo' ]
+  def tester( flag ):
```
Valloric: Nice, this makes sense.

Valloric:
Thanks!

@homu r+

PR #201, opened 2015-08-22
[Windows support] Update install script references

micbou:
See PR Valloric/YouCompleteMe#1642.


Valloric:
Maybe roll these changes into that PR? I don't care too much though.

micbou:
Maybe I'm missing something but how I can do this? Those PRs are not targeting the same repository.

Valloric:
You're not missing anything, I was. :D

@homu r+

PR #202, opened 2015-08-23
[Windows support] Remove sh dependency

micbou:
Using sh module on Windows returns the following error:
```
ImportError: sh 1.11 is currently only supported on linux and osx. please install pbs 0.110 (http://pypi.python.org/pypi/pbs) for windows support.
```
so we drop it in favor of native and portable python functions.

We use the `--build` option of `cmake` to build the target because `make` is not available on Windows.


Valloric:
Thanks!

@homu r+

PR #203, opened 2015-08-23
[Windows support] Convert run_tests script to python

micbou:
This is needed to be able to run tests on Windows.

There are one small difference with the shell script: the double dashes are not needed to run specific tests.


Valloric:
Thanks!

@homu r+

PR #204, opened 2015-08-26
[Windows support] Join path components in a portable way

micbou:
Use the `os.path.join` function to join each component of a path (instead of using the `/` separator).

Fix 15 tests (11 failures and 4 errors) on Windows (listed in the commit message).


Valloric:
Thanks!

@homu r+

PR #205, opened 2015-08-27
[Windows support] Fix absolute paths in tests

micbou:
`ycmd` returns absolute platform-dependent paths in requests (example: `/foo.py` on Linux and `c:\\foo.py` on Windows) so we need to do the same in the tests.

Fix 26 tests (26 failures) on Windows (listed in the commit message).


Valloric:
Thanks!

@homu r+

PR #206, opened 2015-08-27
[Windows support] Normalize paths in _PathsToAllParentFolders

micbou:
We need to normalize paths in _PathsToAllParentFolders function to convert all `/` separators to `\` on Windows. We do the same in the tests since results are platform dependent.

Fix 2 tests (2 failures) on Windows: `PathsToAllParentFolders_Basic` and `PathsToAllParentFolders_WindowsPath`.


Valloric:
Thanks!

@homu r+

PR #207, opened 2015-08-27
Use SVG Travis badge instead of PNG

micbou:
Because badges have rights too (and it looks better).


vheon:
> Because badges have rights too

That is genius :laughing: 

LGTM :+1:

Valloric:
Nice, thanks!

@homu  r+

PR #208, opened 2015-08-28
 [OS X support] [Windows support] Do not depend on std::string implementation specifics

micbou:
Since it was the same issue, I merged the OS X fix from @puremourning and my Windows fix to this PR. Thanks to him.

Fix 3 tests (3 failures) on OS X and Windows.

### DIFF
--- a/ycmd/tests/subcommands_test.py
+++ b/ycmd/tests/subcommands_test.py
@@ -252,73 +252,75 @@ def _RunCompleterCommand_Message_Clang(filename, test, command):
 def RunCompleterCommand_GetType_Clang_test():
   tests = [
   # basic pod types
-    [{'line_num' : 12, 'column_num': 3} , 'Foo'],
+    [{'line_num' : 25, 'column_num': 3} , 'Foo'],
     [{'line_num' : 1,  'column_num': 1} , 'Internal error: cursor not valid'],
-    [{'line_num' : 4,  'column_num': 2} , 'Foo'],
-    [{'line_num' : 4,  'column_num': 8} , 'Foo'],
-    [{'line_num' : 4,  'column_num': 9} , 'Foo'],
-    [{'line_num' : 4,  'column_num': 10} , 'Foo'],
-    [{'line_num' : 5,  'column_num': 3} , 'int'],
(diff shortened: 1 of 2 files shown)


onuraslan:
I think this test needs improvements. LLVM in debian/ubuntu is also returning: `std::string => std::__cxx11::basic_string<char>`

puremourning:
Yeah. I suspect we should change the  test to exercise the behaviour without implementation specifics.

micbou:
I don't think this is an issue because we support only one version of Clang for each platform. That said, I agree that tests not depending on implementation will be easier to maintain.

Valloric:
I agree that we shouldn't have the test depend on platform specifics.

micbou:
So how do we solve this? I see the following approachs:
 - removing those tests;
 - replacing them by ones testing platform-independent types;
 - testing that the result contains `basic_string` which is common to all platforms.

puremourning:
I think the second. We can replicate the sort of equivalent with something like this:

```c++
namespace Test {
  template< typename T >
  struct _Test
  {
      _Test ( const char * );
      T test;
  };

  typedef _Test< int > Test;

  std::ostream& operator << ( std::ostream& o, const Test& test );
};
```

and

```c++
  Test::Test a = "hello";
  std::cout << a;
```

The tedious part is that all the offsets in the test will need tweaking

oblitum:
just for sake of information, beware of naming: http://stackoverflow.com/a/228797/1000282.

micbou:
So I followed @puremourning proposition. Seems to work fine.
Could someone check if the C++ code makes sense? I don't know much of C++.

puremourning:
@micbou nice work. LGTM

vheon:
@micbou LGTM too :+1:

puremourning:
@homu r+

Valloric:
Hm, homu seems to have not triggered on puremourning's comment. I need to figure out why.

Valloric:
OK, I think I figured it out and fixed it. @puremourning could you try the whole homu r+ thing again so we can test out it works?

puremourning:
Sure. 

@homu r+

PR #209, opened 2015-08-29
[Windows support] Update OmniSharpServer submodule

micbou:
OmniSharp Server was modifying the drive letter case of paths causing test failures on ycmd side. This is fixed upstream (see PR OmniSharp/omnisharp-server#207) so we need to update the submodule.

Fix 6 tests (1 error and 5 failures) on Windows (listed in the commit message).


vheon:
In that PR @nosami talked about the new [OmniSharp/omnisharp-roslyn](https://github.com/OmniSharp/omnisharp-roslyn) does someone know what improvements would it bring?

nosami:
@vheon https://github.com/OmniSharp/omnisharp-vim/issues/215

Another big advantage is that it uses Roslyn instead of NRefactory (hence the name :) ). It can handle C#6 and is better at handling malformed C# than NRefactory way. 

Omnisharp-Roslyn also has support for DNX projects (which are awesome because no csproj files to update :) )

vheon:
@nosami thanks for the infos and the link!

nosami:
It also has ScriptCS support (csx). VB.Net and F# coming soon too.

It also has a few other features not in the old server. Goto definition can jump into metadata-as-source ... and it has an endpoint for semantic syntax highlighting.



On the negative side, it has no endpoints for modifying csproj files (yet)

nosami:
ycmd could also benefit from snippet support https://github.com/OmniSharp/omnisharp-vim/pull/209

mispencer:
I have a implementation of ycmd using omnisharp-roslyn somewhere in my repository. As I recall, I had major problems getting omnisharp-roslyn running, and once I did, I ran into the Windows maximum path length...

nosami:
@mispencer fwiw, omnisharp-roslyn is the backend powering VSCode and Atom. Vim, emacs and sublime can be configured to use it.

The max_path issue was fixed some time ago. The main thing that would need changing to make YCM work would be to convert requests to json format from form-url-encoded.

Valloric:
Seems like a migration to omnisharp-roslyn lies in our future. :)

@mispencer A pull request would be highly appreciated. :) You mentioned having done the work in a personal fork.

@homu r+

mispencer:
I'll take another look at omnisharp-roslyn on Monday, time permitting.

PR #210, opened 2015-08-30
Use client working directory if supplied for filename completer

puremourning:
Background
====

Fixes: https://github.com/Valloric/YouCompleteMe/issues/1655

The documentation for YouCompleteMe says:

>The g:ycm_filepath_completion_use_working_dir option

> By default, YCM's filepath completion will interpret relative paths like ../ as being relative to the folder of the file of the currently active buffer. Setting this option will force YCM to always interpret relative paths as being relative to Vim's current working directory.

> Default: 0

>  let g:ycm_filepath_completion_use_working_dir = 0

But previously, only the working directory of `ycmd` was used. This meant that operations like changing the working directory of the client were not reflected in completion suggestions.

What this change does
====

The client can now (optionally) pass a `working_dir` property to the '/completions' request which is respected by the filename completer (and potentially any other completer should it be useful). If `working_dir` is not supplied, the previous behaviour remains.

API Changes
====

The `/completions` request data now allows an optional `working_dir` parameter containing the _absolute path_ to the working directory of the client.

Testing
====

Added new tests to the `filename_completer_test`. See also trivial manual test cases on the linked issue. The one remaining question is about the performance of the `os.getcwd` call on each request, but it probably pales into insignificance compared to the performance of the vimscript, and in particular the IO operations performed by the filename completer itself.

This only addresses the 'standard' case for filename completer, as the 'include' case already has its own logic for relative paths and there are already auto tests for that.

Related
====

There is a related (trivial) change to YouCompleteMe: https://github.com/Valloric/YouCompleteMe/pull/1659

### DIFF
--- a/ycmd/completers/general/filename_completer.py
+++ b/ycmd/completers/general/filename_completer.py
@@ -103,11 +103,16 @@ class FilenameCompleter( Completer ):
     path_dir = os.path.expanduser(
                 os.path.expandvars( path_match.group() ) ) if path_match else ''
 
+    # If the client supplied its working directory, use that instead of the
+    # working directory of ycmd
+    working_dir = request_data.get( 'working_dir' )
+
     return _GenerateCandidatesForPaths(
       _GetPathsStandardCase(
         path_dir,
         self.user_options[ 'filepath_completion_use_working_dir' ],
-        utf8_filepath ) )
+        utf8_filepath,
+        working_dir) )
 
 
   def GetPathsIncludeCase( self, path_dir, include_current_file_dir, filepath,
@@ -133,19 +138,47 @@ class FilenameCompleter( Completer ):
     return sorted( set( paths ) )
 
 
-def _GetPathsStandardCase( path_dir, use_working_dir, filepath ):
-  if not use_working_dir and not path_dir.startswith( '/' ):
-    path_dir = os.path.join( os.path.dirname( filepath ),
-                             path_dir )
+def _GetAbsolutePathForCompletions( path_dir,
+                                    use_working_dir,
+                                    filepath,
+                                    working_dir ):
+  """
+  Returns the absolute path for which completion suggestions should be returned
+  (in the standard case).
+  """
+
+  if path_dir.startswith( '/' ):
+    # Looks like an absolute path already, return it
+    return path_dir
+  elif use_working_dir:
+    # Return paths relative to the working directory of the client, if
+    # supplied, otherwise relative to the current working directory of this
+    # process
+    if working_dir:
+      return os.path.join( working_dir, path_dir )
+    else:
+      return os.path.join( os.getcwd(), path_dir )
+  else:
+    # Return paths relative to the file
+    return os.path.join( os.path.join( os.path.dirname( filepath ) ),
+                         path_dir )
+
+
+def _GetPathsStandardCase( path_dir, use_working_dir, filepath, working_dir ):
+
+  absolute_path_dir = _GetAbsolutePathForCompletions( path_dir,
+                                                      use_working_dir,
+                                                      filepath,
+                                                      working_dir )
 
   try:
     # We need to pass a unicode string to get unicode strings out of
     # listdir.
-    relative_paths = os.listdir( ToUnicodeIfNeeded( path_dir ) )
+    relative_paths = os.listdir( ToUnicodeIfNeeded( absolute_path_dir ) )
   except:
     relative_paths = []
 
-  return ( os.path.join( path_dir, relative_path )
+  return ( os.path.join( absolute_path_dir, relative_path )
            for relative_path in relative_paths )
 
 
--- a/ycmd/completers/general/tests/filename_completer_test.py
+++ b/ycmd/completers/general/tests/filename_completer_test.py
@@ -35,6 +35,19 @@ REQUEST_DATA = {
 }
 
 
+def _CompletionResultsForLine( filename_completer, contents, extra_data=None ):
+  request = REQUEST_DATA.copy()
+  request[ 'column_num' ] = len( contents ) + 1
+  request[ 'file_data' ][ PATH_TO_TEST_FILE ][ 'contents' ] = contents
+  if extra_data:
+    request.update( extra_data )
+
+  request = RequestWrap( request )
+  candidates = filename_completer.ComputeCandidatesInner( request )
+  return [ ( c[ 'insertion_text' ], c[ 'extra_menu_info' ] )
+          for c in candidates ]
+
+
 class FilenameCompleter_test( object ):
   def setUp( self ):
     self._filename_completer = FilenameCompleter(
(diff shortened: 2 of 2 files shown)


vheon:
To my understanding we were not following the doc only in the case of a user that has `let g:ycm_filepath_completion_use_working_dir = 1` in his configuration, am I right in this? If that is the case, we could not send the current directory at every request since is not needed.

puremourning:
That's certainly true. Arguably it is a slight abuse of the knowledge the client has though. Currently the client doesn't even really know the existence of that switch - it just passes it to the server blindly. OK it's in the docs, so it isn't 100% clear cut, but it is theoretically at least possible for other completers to require (or desire) the working directory of the client. If we hide the data behind the switch, we restrict the server to only having the data in certain configurations which isn't such a clean interface.

Happy to hear others' opinions on this. It's not a difficult change to make.

ycmd/completers/general/tests/filename_completer_test.py
```diff
@@ -233,3 +236,80 @@ def EnvVar_Undefined_Garbage_3_test( self ):
...
+    'filepath_completion_use_working_dir': 0
+  } )
+  completer = FilenameCompleter( options )
+
+  data = sorted( _CompletionResultsForLine( completer, 'ls ./include/' ) )
+  eq_( [
+        ( 'Qt',       '[Dir]' ),
+        ( 'QtGui',    '[Dir]' ),
+      ], data )
+
```
micbou: Missing a blank line.

ycmd/completers/general/tests/filename_completer_test.py
```diff
@@ -34,6 +34,17 @@
   'file_data' : { PATH_TO_TEST_FILE : { 'filetypes' : [ 'cpp' ] } }
 }
```
micbou: Missing a blank line.

ycmd/completers/general/tests/filename_completer_test.py
```diff
@@ -34,6 +34,17 @@
...
+  request = REQUEST_DATA.copy()
+  request[ 'column_num' ] = len( contents ) + 1
+  request[ 'file_data' ][ PATH_TO_TEST_FILE ][ 'contents' ] = contents
+  if extra_data:
+    request.update( extra_data )
+
+  request = RequestWrap( request )
+  candidates = filename_completer.ComputeCandidatesInner( request )
+  return [ ( c[ 'insertion_text' ], c[ 'extra_menu_info' ] )
+          for c in candidates ]
```
micbou: Missing a blank line.

ycmd/completers/general/tests/filename_completer_test.py
```diff
@@ -233,3 +236,80 @@ def EnvVar_Undefined_Garbage_3_test( self ):
...
+
+  try:
+    options = user_options_store.DefaultOptions()
+    options.update( {
+      'filepath_completion_use_working_dir': 1
+    } )
+
+    completer = FilenameCompleter( options )
+
+    # change current directory to DATA_DIR/include (path to which we expect
```
micbou: Trailing space.

puremourning: sure will fix

ycmd/completers/general/tests/filename_completer_test.py
puremourning: again, sorry I'm not sure where you're expecting another blank line. perhaps there's some style guideline i'm not aware of?!

ycmd/completers/general/tests/filename_completer_test.py
puremourning: This is copy/pasted and only changed to replace `self` with `filename_completer`

Where are you expecting another line?

ycmd/completers/general/tests/filename_completer_test.py
puremourning: where? I didn't change this line ._0

micbou: Two blank lines are expected before and after functions and classes. You can check this in other files. I think this is to follow the same guideline used in the C++ code. This is also part of the [official style guide](https://www.python.org/dev/peps/pep-0008/#blank-lines) (but we don't respect all of it).

micbou:
I would suggest to start comments with a capital letter.

Tests are passing on Windows. I also manually tried it and it was working like expected.

ycmd/completers/general/tests/filename_completer_test.py
puremourning: Gotya! Thanks.

puremourning:
Thanks @micbou - I've corrected those errors.

Valloric:
> But previously, only the working directory of ycmd was used.

Ha, another relic of when the ycmd code was running inside Vim.

I'm fine with always passing the `working_dir` property. Seems simpler than passing it sometimes.

Agreed with @micbou on capitals for comments.

ycmd/completers/general/filename_completer.py
```diff
@@ -133,10 +138,13 @@ def GetPathsIncludeCase( self, path_dir, include_current_file_dir, filepath,
     return sorted( set( paths ) )
 
 
-def _GetPathsStandardCase( path_dir, use_working_dir, filepath ):
-  if not use_working_dir and not path_dir.startswith( '/' ):
-    path_dir = os.path.join( os.path.dirname( filepath ),
-                             path_dir )
+def _GetPathsStandardCase( path_dir, use_working_dir, filepath, working_dir ):
+  if not path_dir.startswith( '/' ):
```
Valloric: This IF block smells like a separate function that returns `path_dir`. :)

ycmd/completers/general/tests/filename_completer_test.py
```diff
@@ -233,3 +235,81 @@ def EnvVar_Undefined_Garbage_3_test( self ):
 
     os.environ.pop( 'YCM_TEST_td' )
     eq_( [ ], data )
+
+
+def WorkingDir_Use_File_Path_test( ):
```
Valloric: If parens are empty, there should be no space inside them.

Valloric:
This is now LGTM from me (conditional on inline comments being resolved, of course).

puremourning:
Thanks for the review. I've:

- factored deciding which absolute directory to list into a (local) function. we can expose it should anything else need the logic. Added comments to make it clear how this works.
- corrected the style of `( )` and separation of methods

ycmd/completers/general/filename_completer.py
```diff
@@ -133,19 +138,39 @@ def GetPathsIncludeCase( self, path_dir, include_current_file_dir, filepath,
...
 
 
-def _GetPathsStandardCase( path_dir, use_working_dir, filepath ):
-  if not use_working_dir and not path_dir.startswith( '/' ):
-    path_dir = os.path.join( os.path.dirname( filepath ),
-                             path_dir )
+
+def _GetPathsStandardCase( path_dir, use_working_dir, filepath, working_dir ):
+
+  def GetAbsolutePath( path_dir, use_working_dir, filepath, working_dir ):
```
Valloric: If it doesn't need to be a inner function, make it a top-level one. The reason is that inner functions have access to all the vars from the enclosing function, and are thus worse at encapsulation. Even if the inner function only uses it's own params, the reader has to audit the whole function to figure that out.

puremourning: OK sure. Done.

micbou:
Thanks for fixing all those blank lines. LGTM.

puremourning:
OK I think this should be ready now.

Valloric:
Thanks!

@homu r+

PR #211, opened 2015-08-30
[Windows support] Convert paths to short path name on Windows

micbou:
Python 2 fails to execute a command with unicode characters on Windows. In ycmd case, only paths in command are likely to contain such characters so the trick is to convert them to short path name (also called 8.3 path or DOS path) which only contains ascii characters.

Fix GetCompletions_CsCompleter_PathWithSpace test.


mispencer:
Please apply if Cygwin as well, if possible.

micbou:
I don't use Cygwin so I cannot check this but I think it cannot be adapted to Cygwin because the Windows function `GetShortPathNameW` will not be available.

mispencer:
It should be available in cygwin, but it looks like python can't access it. Oh well, nevermind then.

Valloric:
Nice!

@homu r+

micbou:
@homu r+

PR #212, opened 2015-08-31
Fix flaky regular expression comparison test

jeteon:
The proposed change is to fix the test `ycmd.completers.completer_utils_test.FiletypeTriggerDictFromSpec_Works_test` which previoulsy used an equality check which relied on `_sre.SRE_Pattern` objects in the dictionary being the same instances. However, this is not reliable as demonstrated in spurious test failures in https://github.com/Valloric/ycmd/pull/120.

In the instances where this test works it is relying on the call to `re.compile()` caching results and returning the same objects for the same patterns. The Python [docs](https://docs.python.org/2/library/re.html#re.compile) state that:
> The compiled versions of the most recent patterns passed to `re.match()`, `re.search()` or `re.compile()` are cached, so programs that use only a few regular expressions at a time needn’t worry about compiling regular expressions. 

Note that there is no guarantee of the cached versions being returned each time. The update to the test instead compares the `sre.SRE_Pattern.pattern` strings which can be used to reasonably say whether two of the objects represent the same pattern without being the same object necessarily.

### DIFF
--- a/ycmd/completers/completer_utils_test.py
+++ b/ycmd/completers/completer_utils_test.py
@@ -22,19 +22,27 @@ from nose.tools import eq_, ok_
 from ycmd.completers import completer_utils as cu
 
 
+def _extractPatternsFromFiletypeTriggerDict( triggerDict ):
+  copy = triggerDict.copy()
+  for key, values in triggerDict.items():
+    copy[ key ] = set( [ sre_pattern.pattern for sre_pattern in values ] )
+  return copy
+
 def FiletypeTriggerDictFromSpec_Works_test():
+
   eq_( defaultdict( set, {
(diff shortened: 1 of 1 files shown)


ycmd/completers/completer_utils_test.py
```diff
@@ -22,19 +22,27 @@
 from ycmd.completers import completer_utils as cu
 
 
+def _extractPatternsFromFiletypeTriggerDict( triggerDict ):
```
Valloric: Capital "E" for function name start.

Also probably a comment explaining why this function is necessary. People reading the code can't easily refer to the rationale in the PR.

jeteon:
Thanks for the comments. I've made changes to the file to better comply with the project standards.

Valloric:
Thanks for the PR!

@homu r+

PR #213, opened 2015-09-01
Omnisharp roslyn

mispencer:
Allow using omnisharp-roslyn. See https://github.com/Valloric/ycmd/pull/209#issuecomment-136160670.

Omnisharp-roslyn tests do not currently pass on my system, and I'm still receiving a path too long error. Once it is built, it seems to work, though it seems slow.

This pull request does not make roslyn the default (yet). Rather, it must be switched to the roslyn implementation by running the `UseRoslynOmnisharp` completer subcommand. This is per-solution, and pre-runtime.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/213)

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -46,8 +46,7 @@ INVALID_FILE_MESSAGE = 'File is invalid.'
 NO_DIAGNOSTIC_MESSAGE = 'No diagnostic for current line!'
 PATH_TO_OMNISHARP_BINARY = os.path.abspath(
   os.path.join( os.path.dirname( __file__ ), '..', '..', '..',
-                'third_party', 'OmniSharpServer', 'OmniSharp',
-                'bin', 'Release', 'OmniSharp.exe' ) )
+                'third_party', 'omnisharp-roslyn', 'OmniSharp.exe' ) )
 LOGFILE_FORMAT = 'omnisharp_{port}_{sln}_{std}_'
 
 
@@ -414,7 +413,7 @@ class CsharpSolutionCompleter( object ):
       if self._ServerIsRunning():
         self._logger.info( 'Stopping OmniSharp server with PID {0}'.format(
                                self._omnisharp_phandle.pid ) )
-        self._GetResponse( '/stopserver' )
+        self._GetResponse( '/stopserver', returns_json = False )
         try:
           utils.WaitUntilProcessIsTerminated( self._omnisharp_phandle,
                                               timeout = 5 )
@@ -447,7 +446,8 @@ class CsharpSolutionCompleter( object ):
   def _ReloadSolution( self ):
     """ Reloads the solutions in the OmniSharp server """
     self._logger.info( 'Reloading Solution in OmniSharp server' )
-    return self._GetResponse( '/reloadsolution' )
+    self._GetResponse( '/reloadsolution', returns_json = False )
+    return True
 
 
   def CompletionType( self, request_data ):
@@ -550,13 +550,13 @@ class CsharpSolutionCompleter( object ):
   def _DefaultParameters( self, request_data ):
     """ Some very common request parameters """
     parameters = {}
-    parameters[ 'line' ] = request_data[ 'line_num' ]
-    parameters[ 'column' ] = request_data[ 'column_codepoint' ]
+    parameters[ 'Line' ] = request_data[ 'line_num' ]
+    parameters[ 'Column' ] = request_data[ 'column_codepoint' ]
 
     filepath = request_data[ 'filepath' ]
-    parameters[ 'buffer' ] = (
+    parameters[ 'Buffer' ] = (
       request_data[ 'file_data' ][ filepath ][ 'contents' ] )
-    parameters[ 'filename' ] = filepath
+    parameters[ 'FileName' ] = filepath
     return parameters
 
 
@@ -598,11 +598,17 @@ class CsharpSolutionCompleter( object ):
     return 'http://localhost:' + str( self._omnisharp_port )
 
 
-  def _GetResponse( self, handler, parameters = {}, timeout = None ):
+  def _GetResponse( self, handler, parameters = {}, timeout = None,
+                    returns_json = True ):
     """ Handle communication with server """
     target = urllib.parse.urljoin( self._ServerLocation(), handler )
-    response = requests.post( target, data = parameters, timeout = timeout )
-    return response.json()
+    self._logger.info( "Request: " + str(parameters) )
+    response = requests.post( target, json = parameters, timeout = timeout )
+    self._logger.info( "Response: " + response.text )
+    if returns_json:
+      return response.json()
+    else:
+      return None
 
 
   def _ChooseOmnisharpPort( self ):
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -44,9 +44,11 @@ SERVER_NOT_FOUND_MSG = ( 'OmniSharp server binary not found at {0}. ' +
                          '"./install.py --omnisharp-completer".' )
 INVALID_FILE_MESSAGE = 'File is invalid.'
 NO_DIAGNOSTIC_MESSAGE = 'No diagnostic for current line!'
+OMNISHARP_BINARY = ( 'OmniSharp.exe' if utils.OnWindows() or utils.OnCygwin()
+                      else 'OmniSharp' )
 PATH_TO_OMNISHARP_BINARY = os.path.abspath(
   os.path.join( os.path.dirname( __file__ ), '..', '..', '..',
-                'third_party', 'omnisharp-roslyn', 'OmniSharp.exe' ) )
+                'third_party', 'omnisharp-roslyn', OMNISHARP_BINARY ) )
 LOGFILE_FORMAT = 'omnisharp_{port}_{sln}_{std}_'
 
 
@@ -382,12 +384,6 @@ class CsharpSolutionCompleter( object ):
                   str( self._omnisharp_port ),
                   '-s',
                   u'{0}'.format( path_to_solutionfile ) ]
-
-      if not utils.OnWindows() and not utils.OnCygwin():
-        command.insert( 0, 'mono' )
-
-      if utils.OnCygwin():
-        command.extend( [ '--client-path-mode', 'Cygwin' ] )
 
       solutionfile = os.path.basename( path_to_solutionfile )
       self._filename_stdout = utils.CreateLogfile(
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,9 @@ doc/tags
 # Exclude tern installation area
 third_party/tern_runtime/node_modules
 
+# Exclude omnisharp-roslyn installation area
+third_party/omnisharp-roslyn/
+
 # Exclude vagrant directory
 .vagrant/
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "third_party/OmniSharpServer"]
-	path = third_party/OmniSharpServer
-	url = https://github.com/nosami/OmniSharpServer
 [submodule "third_party/argparse"]
 	path = third_party/argparse
 	url = https://github.com/bewest/argparse
--- a/build.py
+++ b/build.py
@@ -78,6 +78,9 @@ def OnMac():
 def OnWindows():
   return platform.system() == 'Windows'
 
+def OnCygwin():
+  return platform.system().startswith("CYGWIN")
+
 
 def OnTravisOrAppVeyor():
   return 'CI' in os.environ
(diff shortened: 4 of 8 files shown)


vheon:
Why the double backend? According to @nosami omnisharp-roslyn should be better then the one we use currently, am I missing something?

micbou:
I tried it and it is not really usable in its current state. There are too much completion timeouts. Maybe we are doing something wrong.

mispencer:
Yes, that's my experience too. Maybe it would be faster with the new stdio protocol, but that would be a considerable implementation effort.

The legacy omnisharp doesn't work on new-style projects, which is a use case for using this new roslyn omnisharp backend. But in general, the performance is not (yet?) there to use exclusively. 

@nosami Are we doing something wrong here?

nosami:
In my experience, it's marginally faster than the old server.

Are you fetching documentation on each request? That is known to be slow. There is caching on the server to make subsequent requests faster.

mispencer:
I'm making the exact same request. WantDocumentationForEveryCompletionResult set to true, and ForceSemanticCompletion and WantImportableTypes set to false. Maybe it's because omnisharp-roslyn doesn't seem to support either ForceSemanticCompletion or WantImportableTypes?

nosami:
`ForceSemanticCompletion` must be a purely ycm thing. OmniSharp _always_ returns semantic completions. 
`WantImportableTypes` hasn't yet been implemented.

Try `WantDocumentationForEveryCompletionResult` set to false. This is a huge time sink.

mispencer:
`ForceSemanticCompletion` was implemented in https://github.com/OmniSharp/omnisharp-server/pull/196.

nosami:
Ah I see. I was thinking of something else sorry. 

In that case the server will act as if that flag was set to true.

Valloric:
Thanks for sending out the PR!

I'd rather wait until we can fully switch over to omnisharp-roslyn than have a switchable implementation.

mispencer:
Yay, green checkmark from Travis. I assume I can ignore the AppVeyor failure.

Ok, so I gross violated the rule of small pull requests. Justification: [Travis is a jerk](https://travis-ci.org/mispencer/ycmd/builds)

To achieve good performance, we need to use the stdio interface of omnisharp-roslyn. This vastly complicated the C# completer. Additionally, roslyn did not play nicely with the legacy Travis, which necessitated switching to the new Travis infrastructure.

Regarding the switchable implementation, as you can see by all the tests I skipped while under it, omnisharp-roslyn is still not a feature complete replacement for OmnisharpServer. Yet, OmnisharpServer doesn't work on the new json-based C# project. So, where possible, I would prefer to use OmnisharpServer, but I still need omnisharp-roslyn available for new projects.

ycmd/completers/cs/cs_completer.py
```diff
@@ -513,11 +509,167 @@ def _DefaultParameters( self, request_data ):
...
+      try:
+        data = ""
+        while not stream.closed and self._omnisharp_phandle is not None:
+          data += os.read( stream.fileno(), 1024 * 1024 * 10 )
+          while "\n" in data:
+            (line, data) = data.split("\n", 1)
+            self._logger.info( "Omnisharp "+type+": " + line.rstrip() )
+      except Exception:
+        self._logger.error( "Read error: " + traceback.format_exc() )
+      #self._logger.info( "Log " + type + " Out done" )
```
Valloric: You probably left this by accident.

ycmd/completers/cs/cs_completer.py
```diff
@@ -545,30 +699,279 @@ def ServerTerminated( self ):
...
+    phandle = self._omnisharp_phandle
+    if phandle is not None:
+      self._logger.info( 'Killing OmniSharp server' )
+      for stream in [ phandle.stderr, phandle.stdout ]:
+        if stream is not None:
+          stream.close()
+      try:
+        phandle.kill()
+      except OSError as e:
+        if e.errno == 3:
```
Valloric: 3 is the best number. :)

In other words, you need to use a named constant here. The number alone is way too magic.

Valloric:
> I assume I can ignore the AppVeyor failure.

Yes, the breakage is expected, not everything is fully set up for it.

> Ok, so I gross violated the rule of small pull requests. 

Oh boy, did you ever. :) 

I'm going to need help reviewing this monster. @vheon @micbou @puremourning 

> Regarding the switchable implementation, as you can see by all the tests I skipped while under it, omnisharp-roslyn is still not a feature complete replacement for OmnisharpServer. Yet, OmnisharpServer doesn't work on the new json-based C# project. So, where possible, I would prefer to use OmnisharpServer, but I still need omnisharp-roslyn available for new projects.

I think the switchable implementation _might_ be ok if it's temporary. If work is happening upstream in Omnisharpland to get to feature parity and we might get there within 6 months to a year at most, then maybe. But I don't want to continue supporting both configurations. Also, we should now or some time soon declare (in the docs) the legacy Omnisharp support deprecated and on the way out.

I'm interested in what the rest of the YCM crew think of this.

vheon:
>  which necessitated switching to the new Travis infrastructure.

Did it worked? Isn't necessary to ask Travis to abilitate the repo for the new infrastructure? @Valloric did you ask for it in preparation for OSX testing?

> omnisharp-roslyn is still not a feature complete replacement for OmnisharpServer. Yet, OmnisharpServer doesn't work on the new json-based C# project. So, where possible, I would prefer to use OmnisharpServer, but I still need omnisharp-roslyn available for new projects.

How many feature are missing? maybe @nosami could shed some light on the roadmap for them or tell us if we're doing something wrong.

ycmd/completers/cs/cs_completer.py
```diff
@@ -215,7 +236,10 @@ def OnUserCommand( self, arguments, request_data ):
       raise ValueError( self.UserCommandsHelpMessage() )
 
     command = arguments[ 0 ]
-    if command in CsharpSolutionCompleter.subcommands:
+    del arguments[ 0 ]
```
vheon: can we use `command = arguments.pop( 0 )`?

nosami:
Roslyn is implementing a public intellisense API that should be ready any time now. It will be the exact same API that is used by Visual Studio. 

Currently, we are using the Roslyn Recommender API which isn't as fully featured yet as either VS or the older NRefactory API.

Like I said earlier though, fetching documentation for every single item in the completion list is always going to be slow on the first hit - there can be multiple sources for docs. It's much faster to fetch docs for an individual item on demand .. although you're probably constrained by the way that Vim does completion.

vheon:
> Roslyn is implementing a public intellisense API that should be ready any time now.

@nosami  so the API is going to change from the current one? If so we should wait...

nosami:
@vheon The OmniSharp API won't change, but the Roslyn API underneath it will.

vheon:
@nosami ok thanks. At this point though, if the roslyn work 

> should be ready any time now.

I would wait and see if we could use just the Roslyn version.

Valloric:
@mispencer What are your thoughts given the information @nosami presented?

---

Review status: 0 of 13 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/213)

mispencer:
@nosami The issue was the http interface was slow. When I switched to the stdio interface, the performance issues went away.

@vheon No, I didn’t have to do anything to use the new container infrastructure for travis. I briefly had it building on OS X, and that didn’t require me to request access either.

The only missing feature in omnisharp-roslyn that blocking switching to it for everything, for me, is the lack of support for WantImportableTypes.

@valloric I don’t urgently need this for anything. However, I’ve already had to fix merge conflicts twice, and I would prefer not have to do it again…

---

Review status: 0 of 13 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

<sup>**[ycmd/completers/cs/cs_completer.py, line 239 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/213#-Jzmm9G6DiuxxphzmdQ8-r5-239)** ([raw file](https://github.com/valloric/ycmd/blob/058d9fed6386b6f235faecfeca97fb35f39b0460/ycmd/completers/cs/cs_completer.py#L239)):</sup>
Yes, Done.

---

<sup>**[ycmd/completers/cs/cs_completer.py, line 620 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/213#-Jzmm9G6DiuxxphzmdQ8-r5-620)** ([raw file](https://github.com/valloric/ycmd/blob/058d9fed6386b6f235faecfeca97fb35f39b0460/ycmd/completers/cs/cs_completer.py#L620)):</sup>
It's been removed.

---

<sup>**[ycmd/completers/cs/cs_completer.py, line 723 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/213#-Jzmm9G6DiuxxphzmdQ8-r5-723)** ([raw file](https://github.com/valloric/ycmd/blob/058d9fed6386b6f235faecfeca97fb35f39b0460/ycmd/completers/cs/cs_completer.py#L723)):</sup>
Done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/213)

Valloric:
@mispencer Sorry about the merge conflicts, that sounds painful. We need to keep the bigger picture in mind though; if @nosami says that the new API should be ready "any time now," it might be best to just hold off on this PR until then. We could then just fully switch to the new implementation instead of supporting switchable confs.

So let's wait for now. And here's a cut-off date so it's not indeterminate: if by December of this year the new intellisense API in omnisharp is not ready to the point that we can fully switch to it, then let's go with the switchable approach. 

---

Review status: 0 of 13 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/213)

mispencer:
Well, I have resolved the merge conflict and updated to the latest OmnisharpRoslyn. However, OmnisharpRoslyn doesn't seem to build without errors in either Windows or OS X, so I guess this is still on hold. It also seems a little flakey.

vheon:
@mispencer did you tried it on your machine? Are the performance issues for the http interface resolved?

mispencer:
It does seem to be improved. I'm going to remove the stdio implemention then.

mispencer:
This appears to be blocked by https://github.com/OmniSharp/omnisharp-roslyn/issues/375.

willl:
@mispencer There's been a lot of work on omnisharp-roslyn recently (see the [DEV branch](https://github.com/OmniSharp/omnisharp-roslyn/tree/dev)) to work on Windows/OS X/Linux with the new .NET CLI bits. And that issue that you mentioned that you're blocked by has now been closed.

There's also been a new OmniSharp Slack team that has been created, and everyone is free to invite themselves into. Just visit the omnisharp-roslyn wiki if you would like an invite.

Valloric:
@mispencer What's the state of this PR now? @willl mentioned blockers being fixed.

mispencer:
I have updated to the latest release, but I am still having problems.

* I can't get it to work in Linux Travis. It depends on libunwind8, which seems to [conflict with cmake in travis](https://travis-ci.org/mispencer/ycmd/jobs/128368888).
* It doesn't seem to work for legacy project anymore - all the completions are empty, none of the goto work, etc.
* I'm getting odd errors for Roslyn projects as well.

Valloric:
Is this PR now abandoned? We probably need to migrate to the roslyn backend at some point, and the issues @mispencer encountered might not be present anymore.

mispencer:
I attempted to get this working a few months ago without any more success.

Valloric:
@mispencer Thanks for the heads-up. Also, ignore zzbot, it's misbehaving.

Valloric:
Let's close this PR then since it's not making progress.

willl:
@mispencer Is it omnisharp-roslyn related issues you're having?

If so, perhaps jump into their [slack team](https://github.com/OmniSharp/omnisharp-roslyn/blob/dev/README.md#help-wanted) and see if you could get some help.

There is work underway for omnisharp-roslyn to support the [language server protocol natively](https://github.com/OmniSharp/omnisharp-roslyn/issues/576#issuecomment-297582517) ([also see](https://github.com/OmniSharp/csharp-language-server-protocol/issues/1#issuecomment-299019872)), there are also plans underway for the server to run on top of a [embedded copy of mono](https://github.com/OmniSharp/omnisharp-roslyn/issues/666). The move to embedded mono should help resolve a lot of the runtime related issues. And LSP support may benefit if you if this project were an LSP client.

I don't know how far off these things are but I'm sure they will probably help in the long run when they're ready.

Valloric:
> And LSP support may benefit if you if this project were an LSP client.

We will eventually have support for consuming LSP APIs. There's work being done on this already.

mispencer:
Could this PR be re-opened? I have had a need for this new backend, so I have been working on getting this working.

bstaletic:
@mispencer You'll have to open a new pull request. "Reopen and comment is disabled because "The OmnisharpPoslyn" branch was force-pushed or recreated."

PR #217, opened 2015-09-03
[WIP] Bump Clang version to 3.7.0

micbou:
I encountered two issues with the new Clang version.

First one was that the `libclang.so` library in the Clang binaries for Ubuntu is a symlink of `libclang.so.3.7`. Since `cmake -E copy` resolves symlinks, `cmake` will copy the `libclang.so.3.7` file but with the wrong destination file name `libclang.so`. Thus, `ycmd` will fail to start because it will try to load the `libclang.so.3.7` file instead of `libclang.so`. I fixed this by setting the `LIBCLANG_TARGET` to the path of the symlinked library.

Second one is where I am stuck. One test is failing:
```
======================================================================
FAIL: ycmd.tests.subcommands_test.RunCompleterCommand_GetType_Clang_test('GetType_Clang_test.cc', [{'column_num': 16, 'line_num': 27}, 'Std::String => Std::BasicString<char>'], ['GetType'])
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/micbou/ycmd/ycmd/tests/subcommands_test.py", line 248, in _RunCompleterCommand_Message_Clang
    app.post_json( '/run_completer_command', request_data ).json )
AssertionError: {'message': 'Std::String => Std::BasicString<char>'} != {u'message': u'Unknown type'}
```
What is really strange is that it is returning the correct value when directly doing `:YcmCompleter GetType` at the location `(27,16)` in the `GetType_Clang_Test.cc` file.

I am still investigating the issue but I will certainly need help on this one.


Valloric:
Honestly I'd just change that test. There's no reason why we have to try to get the type of string; we could  get the type of something else entirely. Doesn't even have to be a type from std; in fact it's probably better that it isn't.

micbou:
Adding the system headers path to the Clang flags fixes the test but I don't think this is the right way to solve this:
 - too much trouble to find this path;
 - like you said, no reason to test the system headers.

So, I removed all the system headers from the C++ tests.

Valloric:
Thanks for doing this work!

@homu r+

PR #218, opened 2015-09-08
[Windows support] Add backslash trigger to complete path on Windows

micbou:
With this PR, paths can be completed using the `\` separator on Windows.

I also simplified and refactored the regular expression used to match a path.

Fix 5 tests (5 failures) on Windows (listed in the commit message).


Valloric:
Damn good work!

@homu r+

PR #219, opened 2015-09-11
Add soversion Clang library to .gitignore

micbou:
Ignore the `libclang.so.3.7` file. I forgot to do this when updating Clang to 3.7.0.


Valloric:
Thanks!

@homu r+

PR #221, opened 2015-09-12
[Windows support] Force C++03 in diagnostics FixIt test

micbou:
The `Diagnostics_ClangCompleter_FixIt_Available` test depends on the absence of a C++11 feature (explicit conversion functions) to succeed. By default, this is the case on Linux (and certainly OS X), not on Windows. This is solved by specifying an older version of C++ (C++03) in the test.


puremourning:
LGTM

vheon:
LGTM

Valloric:
Thanks!

@homu r+

PR #222, opened 2015-09-14
[request_wrap.py] _Query does not take UTF8 into account

Cocophotos:
Hi,

When designing a custom semantic completer for LaTeX (to complete labels and citations), I realised that the request_data['query'] is wrong when there are utf8 chars in the input line and works perfectly otherwise.

Changing l.80 in request_wrap.py by
```python 
return ToUtf8IfNeeded(self[ 'line_value' ])[
```
fixes the problem without making any tests failed. I also added some tests to ensure non regression with my modifications.

Hope this helps.

Best regards,

-- 
CR

### DIFF
--- a/ycmd/request_wrap.py
+++ b/ycmd/request_wrap.py
@@ -77,7 +77,7 @@ class RequestWrap( object ):
 
 
   def _Query( self ):
-    return self[ 'line_value' ][
+    return ToUtf8IfNeeded(self[ 'line_value' ])[
              self[ 'start_column' ] - 1 : self[ 'column_num' ] - 1 ]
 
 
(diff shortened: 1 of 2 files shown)


micbou:
Those tests pass on Travis without your change. You certainly need to test the `line_value` key, not the `query` one. Also, one test should be enough for this kind of thing.

Cocophotos:
No problem, I will do that. Thanks for the verification.

Regards,

Cocophotos:
Ok, after more careful evaluation, it happens that the problem is with column_num which is not correctly computed with some (weirdly not all) utf-8 chars. I don't know why using ToUtf8IfNeeded on l.80 is fixing the bug, but I need to dig more into the computation of column_num to be able to give a proper answer.

It won't be for now though, I let the PR opened until finding a complete explanation.

Best

Valloric:
@Cocophotos Thanks for sending a PR! Since it seems some further investigation is necessary and that the current approach doesn't quite address the problem, I'll close this PR. Feel free to open another when you've tracked this down.

Cocophotos:
@Valloric, thanks for that. I let you know when I find a good solution to this problem.

Best,

PR #223, opened 2015-09-15
[Windows support] Do not delay parsing of templates on Windows

micbou:
This PR is a delicate one.

On Windows, templates are not parsed until instantation time which makes commands like GetType and GetParent not returning the expected value when the cursor is inside a template. To fix this, we add the flag `-fno-delayed-template-parsing` on Windows (this is the default on other platforms). We also add it to the `RunCompleterCommand_GetParent_Clang` tests to fix two tests on Windows.

However, there are some concerns about adding automatically this flag. From this comment (see this [issue](https://code.google.com/p/include-what-you-use/issues/detail?id=129))
> When IWYU runs on Windows it defaults to MSVC compatibility, to be able to parse Microsoft's STL  and Platform SDK headers. Turning off delayed template parsing will likely make it fail to parse these system headers, in turn making IWYU useless.

using this flag could cause some problems, though I was not able to find an example where ycmd is not working because of this flag.

A more conservative approach would be to only use this flag for the tests and add a note about it in the documentation.

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -21,7 +21,7 @@ import ycm_core
 import os
 import inspect
 from ycmd import extra_conf_store
-from ycmd.utils import ToUtf8IfNeeded, OnMac
+from ycmd.utils import ToUtf8IfNeeded, OnMac, OnWindows
 from ycmd.responses import NoExtraConfDetected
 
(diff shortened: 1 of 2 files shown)


Valloric:
Let's go with including the flag by default for now and see if anyone complains. If they do, we can add an option to turn off automatic adding of the flag.

BTW Travis doesn't like your PR.

Valloric:
@homu r+

PR #226, opened 2015-09-19
[Windows support] Add Windows options to scripts

micbou:
Add the following options to `build.py` and `run_tests.py` scripts:
 - `--msvc`: pick the MSVC version between 11, 12 (default), and 14 (respectively 2012, 2013, or 2015);
 - `--arch`: force 32 or 64 bits architecture. Default is the python interpreter architecture used to start the script. Can be useful if both python 32 and 64 bits are installed. See discussion in issue Valloric/YouCompleteMe#1651.

### DIFF
--- a/build.py
+++ b/build.py
@@ -34,6 +34,10 @@ def OnMac():
   return platform.system() == 'Darwin'
 
 
+def OnWindows():
+  return platform.system() == 'Windows'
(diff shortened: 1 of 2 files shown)


Valloric:
Nice!

@homu r+

PR #227, opened 2015-09-19
[Windows support] Fix ycm_core_tests build

micbou:
Needed changes to successfully build ycm_core_tests and run it on Windows.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/227)

### DIFF
--- a/cpp/ycm/DLLDefines.h
+++ b/cpp/ycm/DLLDefines.h
@@ -1,0 +1,29 @@
+// Copyright (C) 2015 ycmd contributors
+//
+// This file is part of ycmd.
+//
+// ycmd is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// ycmd is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef DLLDEFINES_H_0IYA3AQ3
+#define DLLDEFINES_H_0IYA3AQ3
+
+// We need to export symbols for gmock tests on Windows.  The preprocessor
+// symbol ycm_core_EXPORTS is defined by CMake when building a shared library.
+#if defined( _WIN32 ) && defined( ycm_core_EXPORTS )
+  #define YCM_DLL_EXPORT __declspec( dllexport )
+#else
+  #define YCM_DLL_EXPORT
+#endif
+
+#endif /* end of include guard: DLLDEFINES_H_0IYA3AQ3 */
--- a/cpp/ycm/tests/CMakeLists.txt
+++ b/cpp/ycm/tests/CMakeLists.txt
@@ -33,6 +33,11 @@ if ( USE_SYSTEM_GMOCK )
   find_package( GTest REQUIRED )
   find_package( GMock REQUIRED )
 else()
+  if ( WIN32 )
+    # Override BUILD_SHARED_LIBS option in gmock and gtest CMakeLists
+    set( BUILD_SHARED_LIBS ON CACHE BOOL "Build shared libraries (DLLs)." )
+  endif()
+
   add_subdirectory( gmock )
   set( GTEST_INCLUDE_DIRS ${gtest_SOURCE_DIR} ${gtest_SOURCE_DIR}/include )
   set( GMOCK_INCLUDE_DIRS ${gmock_SOURCE_DIR} ${gmock_SOURCE_DIR}/include )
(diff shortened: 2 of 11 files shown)


cpp/ycm/DLLDefines.h
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2015  Google Inc.
```
Valloric: Not Google Inc, is it? :)

cpp/ycm/tests/CMakeLists.txt
```diff
@@ -77,13 +82,39 @@ add_executable( ${PROJECT_NAME}
...
+  set_target_properties( ${PROJECT_NAME}
+                         PROPERTIES
+                         COMPILE_DEFINITIONS
+                         "GTEST_LINKED_AS_SHARED_LIBRARY=1" )
+
+  # Preprocessor definition used to export symbols for gmock tests
+  set_target_properties( ${SERVER_LIB} PROPERTIES DEFINE_SYMBOL "TESTS_EXPORTS" )
+
+  # Fix gtest build on MSVC 11. See:
+  # http://stackoverflow.com/questions/8274588/c2977-stdtuple-too-many-template-arguments-msvc11
```
Valloric: SO has short links for all answers and questions through "share" links (bottom of each question/answer). Maybe use that?

cpp/ycm/tests/CMakeLists.txt
```diff
@@ -33,6 +33,11 @@ if ( USE_SYSTEM_GMOCK )
   find_package( GTest REQUIRED )
   find_package( GMock REQUIRED )
 else()
+  if ( WIN32 )
+    # Override BUILD_SHARED_LIBS option in gmock and gtest CMakeLists
+    set( BUILD_SHARED_LIBS ON CACHE BOOL "Build shared libraries (DLLs)." )
```
Valloric: Why? I've used cmake before with gtest/gmock without resorting to shared libs. [See the FlightCrew source code](https://code.google.com/p/flightcrew/source/browse/#git) (which I wrote years ago).

cpp/ycm/DLLDefines.h
micbou: Oops, old times.

cpp/ycm/tests/CMakeLists.txt
micbou: Didn't know that. I'll change it.

cpp/ycm/tests/CMakeLists.txt
```diff
@@ -77,13 +82,39 @@ add_executable( ${PROJECT_NAME}
...
+    string( TOUPPER ${OUTPUTCONFIG} OUTPUTCONFIG )
+    set_target_properties( ${GMOCK_LIBRARIES} PROPERTIES
+      RUNTIME_OUTPUT_DIRECTORY_${OUTPUTCONFIG} ${PROJECT_BINARY_DIR} )
+    set_target_properties( ${PROJECT_NAME} PROPERTIES
+      RUNTIME_OUTPUT_DIRECTORY_${OUTPUTCONFIG} ${PROJECT_BINARY_DIR} )
+  endforeach()
+endif()
+
 target_link_libraries( ${PROJECT_NAME}
+                       ${Boost_LIBRARIES}
```
Valloric: In theory this shouldn't be necessary since the dependency is picked up from server/client libs.

cpp/ycm/tests/CMakeLists.txt
micbou: I tried static linking (long time ago) and could not make it work. I could try again.

Valloric: No need to spend too much time on it, but taking a look at the Flightcrew `CMakeLists.txt` files should make things much easier. It's all CMake, C++ and gtest/gmock. 

These three files should be enough to get an idea: [top-level](https://github.com/Sigil-Ebook/flightcrew/blob/master/CMakeLists.txt), [library](https://github.com/Sigil-Ebook/flightcrew/blob/master/src/FlightCrew/CMakeLists.txt), [tests](https://github.com/Sigil-Ebook/flightcrew/blob/master/src/FlightCrew/tests/CMakeLists.txt).

cpp/ycm/tests/CMakeLists.txt
micbou: I am getting these errors without this:
```
BoostParts.lib(function.obj) : error LNK2005: "public: virtual unsigned int __cdecl boost::python::objects::py_function_impl_base::max_arity(void)const " (?max_arity@py_function_impl_base@objects@python@boost@@UEBAIXZ) already defined in ycm_core.lib(ycm_core.pyd) [c:\users\micbou\appdata\local\temp\ycm_build.ivs_js\ycm\tests\ycm_core_tests.vcxproj]
BoostParts.lib(function.obj) : error LNK2005: "public: virtual __cdecl boost::python::objects::py_function_impl_base::~py_function_impl_base(void)" (??1py_function_impl_base@objects@python@boost@@UEAA@XZ) already defined in ycm_core.lib(ycm_core.pyd) [c:\users\micbou\appdata\local\temp\ycm_build.ivs_js\ycm\tests\ycm_core_tests.vcxproj]
BoostParts.lib(errors.obj) : error LNK2005: "void __cdecl boost::python::throw_error_already_set(void)" (?throw_error_already_set@python@boost@@YAXXZ) already defined in ycm_core.lib(ycm_core.pyd) [c:\users\micbou\appdata\local\temp\ycm_build.ivs_js\ycm\tests\ycm_core_tests.vcxproj]
BoostParts.lib(errors.obj) : error LNK2005: "public: virtual __cdecl boost::python::error_already_set::~error_already_set(void)" (??1error_already_set@python@boost@@UEAA@XZ) already defined in ycm_core.lib(ycm_core.pyd) [c:\users\micbou\appdata\local\temp\ycm_build.ivs_js\ycm\tests\ycm_core_tests.vcxproj]
C:\Users\micbou\AppData\Local\Temp\ycm_build.ivs_js\ycm\tests\ycm_core_tests.exe : fatal error LNK1169: one or more multiply defined symbols found [c:\users\micbou\appdata\local\temp\ycm_build.ivs_js\ycm\tests\ycm_core_tests.vcxproj]
```
[Here](https://msdn.microsoft.com/en-us/library/72zdcz6f.aspx)'s the documentation of the error. I am trying to understand why it happens.

cpp/ycm/tests/CMakeLists.txt
micbou: There is a difference with the Flightcrew project: the `ycm_core` and `ycm_client_support` targets are built as shared libraries. If we want to link statically `ycm_core_tests`, we also need the static version of those libraries.

Valloric: Ha, excellent point. 

Feel free to proceed with the shared library approach.

cpp/ycm/tests/CMakeLists.txt
Valloric: That looks like BoostParts is getting doubly-linked, which is bad. If adding that line fixes it, just go with that.

cpp/ycm/DLLDefines.h
```diff
@@ -0,0 +1,28 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with YouCompleteMe.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef DLLDEFINES_H_0IYA3AQ3
+#define DLLDEFINES_H_0IYA3AQ3
+
+// We need to export symbols for gmock tests on Windows
+#if defined( _WIN32 ) && defined( TESTS_EXPORTS )
+  #define TESTS_EXPORT __declspec(dllexport)
```
Valloric: Maybe TESTS_EXPORTS is not the best name; it's saying what the symbols are used for instead of what the macro does.

Possibly use DLL_EXPORT instead?

micbou: I found out that CMake is already defining a preprocessor symbol (ycm_core_EXPORTS) for shared libraries if none are defined. I am updating the PR to use it instead of TESTS_EXPORTS.

Concerning the macro itself, I think DLL_EXPORT is too generic. It could already be defined (even if this is not the case). For example, Boost is defining BOOST_DLLEXPORT so I am proposing YCM_DLL_EXPORT.

micbou:
I replaced the preprocessor symbol TESTS_EXPORTS by the CMake predefined one. I also changed the TESTS_EXPORT name to YCM_DLL_EXPORT and I removed one unnecessary symbol export.

Valloric:
Reviewed 11 of 11 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/227)

Valloric:
Nice!

@homu r+

---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/227)

PR #228, opened 2015-09-19
[Windows support] Fix Gmock tests

micbou:
The filesystem boost library will not consider a path like `/foo/zoo` as absolute on Windows. As a result, `path.is_relative()` will return true and the computed path will be wrong. To obtain the right path, we need to use the [fs::absolute](http://www.boost.org/doc/libs/1_46_0/libs/filesystem/v3/doc/reference.html#absolute) function.
The [make_preferred](http://www.boost.org/doc/libs/1_46_0/libs/filesystem/v3/doc/reference.html#path-make_preferred) method converts slashes to backslahes on Windows.

Fix IdentifierUtilsTest.ExtractIdentifiersFromTagsFileWorks test.


Valloric:
Nice!

How did people ever use YCM successfully on Windows before? There seem to be soooo many little things that just plain didn't work there.

@homu r+

PR #229, opened 2015-09-19
Support GetDoc subcommand for Clang completer

puremourning:
References 
====

Feature request: https://github.com/Valloric/YouCompleteMe/issues/1653
YCM change: https://github.com/Valloric/YouCompleteMe/pull/1694

Overview
====

The GetDoc subcommand returns detailed information (sometimes known as
QuickInfo) for the identifier at a particular location.

For C-family languages, we display:

```
  <declaration of the identifier>
  <brief>
  Type: <canonical type>
  Name: <display name>
  ---
  <Detailed information from the doxygen comment>
```

Technical Information 
===

Clang actually has very rich support for parsing of doxygen comments. However,
this API is not strictly part of the "libclang" API - it is technically a
separate module. The core libclang module does include methods to get the raw
comment string (as typed in the buffer) and the "brief" as parsed by the full
documentation parser.

Annoyingly, we have to get a fully parsed XML document out of the non-core API
just to extract the declaration. There does not appear to be any other way to
get this out of libclang.

Regarding the detailed information, we actually use the raw comment text, rather
than the parsed infomation from the XML. This is because we don't want to write
a lot of complex code, and we also don't want to lose any actual data from the
comment. However, we do want to remove distracting indentation and comment
markets. We jsut use some basic regular expressions and python's textwrap to
remove the general syntax markers used on comments.

API Changes
====

Completer subcommands can now return 'detailed info' using `ycmd.responses.BuildDetailedInfoResponse`, as follows:

```
{
   detailed_info: <multi-line string, with OS-specific line endings>
}
```

### DIFF
--- a/cpp/ycm/ClangCompleter/Documentation.cpp
+++ b/cpp/ycm/ClangCompleter/Documentation.cpp
@@ -1,0 +1,49 @@
+// Copyright (C) 2015 YouCompleteMe Contributors
+//
+// This file is part of YouCompleteMe.
+//
+// YouCompleteMe is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// YouCompleteMe is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with YouCompleteMe.  If not, see <http://www.gnu.org/licenses/>.
+
+#include "Documentation.h"
+#include "ClangHelpers.h"
+
+#include <clang-c/Documentation.h>
+
+namespace YouCompleteMe {
+
+namespace {
+
+bool CXCommentValid( const CXComment & comment ) {
+  return clang_Comment_getKind( comment ) != CXComment_Null;
+}
+
+}
+
+DocumentationData::DocumentationData( const CXCursor& cursor )
+  : raw_comment( CXStringToString( clang_Cursor_getRawCommentText( cursor ) ) )
+  , brief_comment( CXStringToString(
+                                 clang_Cursor_getBriefCommentText( cursor ) ) )
+  , canonical_type( CXStringToString(
+                    clang_getTypeSpelling( clang_getCursorType( cursor ) ) ) )
+  , display_name( CXStringToString( clang_getCursorSpelling( cursor ) ) ) {
+
+
+  CXComment parsed_comment = clang_Cursor_getParsedComment( cursor );
+  if ( CXCommentValid( parsed_comment ) ) {
+    comment_xml = CXStringToString(
+                                clang_FullComment_getAsXML( parsed_comment ) );
+  }
+}
+
+} // namespace YouCompleteMe
--- a/cpp/ycm/ClangCompleter/TranslationUnit.cpp
+++ b/cpp/ycm/ClangCompleter/TranslationUnit.cpp
@@ -296,10 +296,10 @@ std::string TranslationUnit::GetTypeAtLocation(
 
   CXType canonical_type = clang_getCanonicalType( type );
 
-  if ( !clang_equalTypes( type, canonical_type ) )
-  {
+  if ( !clang_equalTypes( type, canonical_type ) ) {
     type_description += " => ";
-    type_description += CXStringToString(clang_getTypeSpelling(canonical_type));
+    type_description += CXStringToString(
+                                  clang_getTypeSpelling( canonical_type ) );
   }
 
   return type_description;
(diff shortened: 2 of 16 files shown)


puremourning:
Well that's embarrassing. Note to self: run make before submit :(

cpp/ycm/ClangCompleter/Documentation.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include "Documentation.h"
+#include "ClangHelpers.h"
+
+#include <clang-c/Documentation.h>
+
+namespace YouCompleteMe {
+
+namespace {
+  bool CXCommentValid( const CXComment & comment ) {
```
Valloric: No indent for namespace contents. Otherwise we'd go crazy. :)

cpp/ycm/ClangCompleter/TranslationUnit.cpp
```diff
@@ -452,6 +452,41 @@ std::vector< FixIt > TranslationUnit::GetFixItsForLocationInFile(
...
+
+  if ( !CursorIsValid( cursor ) )
+    return DocumentationData();
+
+  // If the original cursor is a reference, then we return the documentation
+  // for the type/method/etc. that is referenced
+  CXCursor referenced_cursor = clang_getCursorReferenced( cursor );
+  if ( CursorIsValid( referenced_cursor ) ) {
+    cursor = referenced_cursor;
+  }
```
Valloric: I don't care do you use braces around single-line blocks or not except that you keep it consistent inside a function. :)

Otherwise it's bedlam and Cthulhu.

ycmd/completers/cpp/clang_completer.py
```diff
@@ -399,3 +425,64 @@ def _ResponseForLocation( location ):
                                       location.column_number_ )
 
 
+# Strips the following leading strings from the raw comment:
```
Valloric: Thanks for commenting this. That regex looks mean.

ycmd/completers/cpp/clang_completer.py
```diff
@@ -399,3 +425,64 @@ def _ResponseForLocation( location ):
...
+                         re.sub( STRIP_LEADING_COMMENT, '', line ) )
+                                for line in comment.split( os.linesep ) ] ) )
+
+
+def _BuildGetDocResponse( doc_data ):
+  """Builds a "DetailedInfoResponse" for a GetDoc request. doc_data is a
+  DocumentationData object returned from the ClangCompleter"""
+
+  # Parse the XML, as this is the only way to get the declaration text out of
+  # libclang. It seems quite wasteful, but the while the contents of the XML
```
Valloric: "the while"?

ycmd/tests/subcommands_test.py
```diff
@@ -1327,3 +1330,368 @@ def RunCompleterCommand_GetType_HasNoType_TypescriptCompleter_test():
...
+      'detailed_info': """\
+char with_brief()
+brevity is for suckers
+Type: char ()
+Name: with_brief
+---
+
+This is not the brief.
+
+\\brief brevity is for suckers
```
Valloric: This had me laughing like a maniac! :D

ycmd/tests/subcommands_test.py
```diff
@@ -24,17 +24,20 @@
                           PathToTestFile,
```
Valloric: On a related note, we _really_ need to refactor this file so that it's smaller. Similar for the completions tests. Probably in per-language files.

Not in this PR though.

Valloric:
This looks nice! Minor inline comments added.

cpp/ycm/ClangCompleter/Documentation.cpp
puremourning: As you noticed in my test data, I might already *be* crazy. Anyway, yes, `4<<` done :)

cpp/ycm/ClangCompleter/TranslationUnit.cpp
puremourning: right. fixed.

ycmd/completers/cpp/clang_completer.py
puremourning: Yeah. I understand there is some pythonic regex magic which allows for setting these out more neatly, but I wasn't sure I fully understood the subtlety, so went with a single line and a hefty comment

ycmd/completers/cpp/clang_completer.py
puremourning: fixed

ycmd/tests/subcommands_test.py
puremourning: I may have had wine when writing some of this.

ycmd/tests/subcommands_test.py
puremourning: Yes. I thought about doing this, but wasn't sure of your appetite for the organisation. Proper organisation of tests is essential to avoid duplication, etc., though the correct organisation is always debatable. 

I suspect, for the subcommands at least, we should organise them around completer, then subcommand, e.g.

```
tests/subcommands_test_clang.py:
   class ClangSubcommandTest_GetType:
       ...
   class ...

tests/subcommands_test_jedi.py:
   ...
```

Up for debate ;)

Valloric: I don't have a preference to how we organize them as long as it results in files that _don't_ have 1k+ lines. :)

Your suggestion sounds reasonable.

Valloric:
This looks great now!

@homu r+

PR #230, opened 2015-09-19
[Windows support] Add Clang downloads for Windows

micbou:
`7-zip` is the best way I found to obtain the files from the Clang installer without installing it.

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -51,6 +51,17 @@ if ( USE_CLANG_COMPLETER AND
     set( CLANG_SHA256
          "057e56b445ac5dfbf0d5475badab374edebe8db5428001c8829340e40724b50d" )
     set( CLANG_FILENAME "${CLANG_DIRNAME}.tar.xz" )
+  elseif ( WIN32 )
+    if( 64_BIT_PLATFORM )
+      set( CLANG_DIRNAME "LLVM-${CLANG_VERSION}-win64" )
+      set( CLANG_SHA256
+           "04a393cf73bc7416c9903aa08f47f15c77cf475b22a607ac4b7493a658a11a3f" )
+    else()
+      set( CLANG_DIRNAME "LLVM-${CLANG_VERSION}-win32" )
+      set( CLANG_SHA256
+           "75933480c8f8e3ae3cf9281494bf0057e9e4cfae4315f59c468a4e850a86f544" )
+    endif()
+    set( CLANG_FILENAME "${CLANG_DIRNAME}.exe" )
   else()
     if ( 64_BIT_PLATFORM )
       # We MUST support the latest Ubuntu LTS release!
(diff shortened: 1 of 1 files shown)


cpp/ycm/CMakeLists.txt
```diff
@@ -103,16 +114,25 @@ if ( USE_CLANG_COMPLETER AND
     execute_process( COMMAND tar -xjf ${CLANG_FILENAME} )
   elseif( CLANG_FILENAME MATCHES ".+xz" )
     execute_process( COMMAND tar -xJf ${CLANG_FILENAME} )
+  elseif( CLANG_FILENAME MATCHES ".+exe" )
+    # NSIS tool is used to generate the Clang installer on Windows. Files
+    # from the installer can be extracted with 7-Zip.
+    execute_process( COMMAND 7z x ${CLANG_FILENAME} OUTPUT_QUIET )
```
puremourning: Does this make 7-zip a requirement for installation on Windows? If so do we need to document somewhere that 7-zip is a requirement for installing YCM?

micbou: Yes, it is a requirement for building YCM with Clang support. I will document it in the `Windows Installation` section of the YCM `README.md`.

Valloric: Any chance we could not have that extra dependency? I could live with another python-only dependency in `third_party` that could do the necessary work; but we'd be shipping that then and the user wouldn't have to install it themselves.

Valloric: In other words, it's the user-level dependency of "go install this other thing first" that I'd like to avoid.

micbou:
I removed the `7-zip` dependency. New approach is to (silently) execute the Clang installer, copy the installed files in a new directory, and uninstall it to clean up the registry. There are minor drawbacks:
 - when Clang is already installed, a dialog window will appear asking the user if he want to uninstall the previous installation;
 - 2 console windows appear a very short time during the build;
 - appveyor runs takes 1-2 more minutes.

Valloric:
@micbou I appreciate you trying to remove the 7zip depedency, but I think the new approach with running the installer is a step backwards from that. The issues with dialogs and the console windows plus the somewhat hacky approach of the whole thing seem... unfortunate.

Frankly, I'd rather have the 7zip dep if nothing better is available.

Valloric:
You know what? Let's just go with the installers.

vheon:
I'd rather not play with the user installed packages. I would prefer the 7-zip dep.

Valloric:
@vheon I agree with you, but we also have two suboptimal approach to chose from. At this point, I'm fine with whichever ends up being less hassle for us. I'll let @micbou decide since he'll be maintaining the windows support. Whatever works best for him, we'll go with that.

micbou:
Clearly, the 7-zip approach is cleaner. 7-zip is also one of the most popular file archiver on Windows so I don't think this is a problem to require it. What I could improve is to avoid the need to add it to the PATH. I will work on this tomorrow.

micbou:
I reverted the PR to the 7-zip approach and improved it. The 7-zip executable is now automatically found searching the PATH and using the registry key. An error is raised if no executable is found.

vheon:
I like this version much better.

Valloric:
Nice, thanks!

@homu r+

micbou:
@homu retry

micbou:
@homu r+

PR #236, opened 2015-09-20
Fix filename of comment_strip tests

micbou:
`ycmd/completers/cpp/tests/comment_strip.py` tests was not run automatically by nosetests because `_test` was missing at the end of the filename.


micbou:
We should also remove the `ClangCompleter` in the tests name. It makes nosetests ignore those tests when ycmd is built without Clang support even though Clang is not required to run them.

puremourning:
I added `ClangCompleter` to the name deliberately. You're right that _technically_ it doesn't rely on the c++ Clang stuff, but it still relates to the clang completer, which isn't used when the core doesn't support the C++ part.

puremourning:
Also, thanks for fixing this. LGTM

vheon:
LGTM too.

@homu r+

PR #234, opened 2015-09-20
Add GetDoc subcommand to typescript completer

puremourning:
This builds on the Clang completer support for GetDoc subcommand, and just
    returns the docuemntation supplied by tsserver in the quickinfo response.

References: 
* https://github.com/Valloric/YouCompleteMe/issues/1653
* https://github.com/Valloric/YouCompleteMe/pull/1694
* https://github.com/Valloric/ycmd/pull/229

PR #233, opened 2015-09-20
Add GetDoc support to Python completer

puremourning:
This builds on the Clang completer support and returns the docstring from any
    associated declarations for the context.

References: : 
* https://github.com/Valloric/YouCompleteMe/issues/1653
* https://github.com/Valloric/YouCompleteMe/pull/1694
* https://github.com/Valloric/ycmd/pull/229

PR #232, opened 2015-09-20
Add GetDoc subcommand for c-sharp completer

puremourning:
This builds on the support added for the Clang completer and simply returns the
    type information along with the documentation provided by OmniSharp-server.

    Change GetType to just return the type

    The documentation provided by OmniSharp-server was previously included in the
    GetType subcommand. In the Vim client, this is echo'd to the console and
    requires hitting a key to dismiss. Now that there is the GetDoc command, this
    has been changed to just return the type, for consistency with go completer and
    Clang completer.

References: 
* https://github.com/Valloric/YouCompleteMe/issues/1653
* https://github.com/Valloric/YouCompleteMe/pull/1694
* https://github.com/Valloric/ycmd/pull/229

PR #237, opened 2015-09-20
[Windows support] Do not use os.linesep

micbou:
Using `os.linesep` makes things complicated for no benefits:
 - Clang returns `\n` for newlines on Windows;
 - Python and editors should manage without issue `\n` on Windows;
 - there is a bug with `textwrap.dedent()` and `\r\n` newline;
 - tests are a pain to fix with platform-dependent newlines.

@puremourning Could you update your PRs (#232, #233, and #234) accordingly?

### DIFF
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -21,7 +21,6 @@ from collections import defaultdict
 import ycm_core
 import re
 import textwrap
-import os
 from ycmd import responses
 from ycmd import extra_conf_store
 from ycmd.utils import ToUtf8IfNeeded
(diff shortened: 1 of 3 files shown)


puremourning:
@micbou sure thing. I was trying (vainly it seems) to be platform agnostic.

puremourning:
done - all three pushed new commits

puremourning:
@micbou do we need to remove them from `comment_strip.py` too ?

```
./third_party/ycmd/ycmd/completers/cpp/tests/comment_strip.py:35:            + os.linesep
./third_party/ycmd/ycmd/completers/cpp/tests/comment_strip.py:37:            + os.linesep
./third_party/ycmd/ycmd/completers/cpp/tests/comment_strip.py:39:            + os.linesep
./third_party/ycmd/ycmd/completers/cpp/tests/comment_strip.py:41:            + os.linesep
./third_party/ycmd/ycmd/completers/cpp/tests/comment_strip.py:43:            + os.linesep
```

micbou:
There are removed by this PR for consistency.

puremourning:
This LGTM.

There are some arguably unnecessary string concatenations with `+` (which I actually wrote, and you actually educated _me_ about), but I don't see them as a big deal.

micbou:
Could not resist to remove them.

Valloric:
Nice!

@homu r+

PR #239, opened 2015-09-22
Set default MSVC version to 14

micbou:
Since this is the version that users will download and install if they don't already have it, it makes sense to set the default to MSVC 14. It will also simplify the Windows documentation (no need to tell the user to set the option `--msvc` to 14).

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/239)


vheon:
Reviewed 2 of 2 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/239)

Valloric:
Nice!

@homu r+

micbou:
@homu r+

micbou:
Don't know what is happening with the Homu bot. Trying again.

@homu r+

PR #243, opened 2015-09-26
Return JSON True instead of empty HTML for 2 requests

hydrargyrum:
/load_extra_conf_file and /ignore_extra_conf_file returned an empty HTML body
in case of success. Instead, return JSON body containing "true".

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/243)

### DIFF
--- a/ycmd/handlers.py
+++ b/ycmd/handlers.py
@@ -181,12 +181,16 @@ def LoadExtraConfFile():
   request_data = RequestWrap( request.json, validate = False )
   extra_conf_store.Load( request_data[ 'filepath' ], force = True )
 
+  return _JsonResponse( True )
+
 
 @app.post( '/ignore_extra_conf_file' )
 def IgnoreExtraConfFile():
   _logger.info( 'Received extra conf ignore request' )
   request_data = RequestWrap( request.json, validate = False )
   extra_conf_store.Disable( request_data[ 'filepath' ] )
+
+  return _JsonResponse( True )
 
 
 @app.post( '/debug_info' )
--- a/ycmd/tests/misc_handlers_test.py
+++ b/ycmd/tests/misc_handlers_test.py
@@ -21,8 +21,9 @@ from ..server_utils import SetUpPythonPath
 SetUpPythonPath()
 from webtest import TestApp
 from .. import handlers
-from nose.tools import ok_, with_setup
-from .test_utils import Setup, BuildRequest
+from hamcrest import assert_that, has_property
+from nose.tools import with_setup
+from .test_utils import Setup, BuildRequest, PathToTestFile
 import bottle
 
 bottle.debug( True )
@@ -32,14 +33,32 @@ bottle.debug( True )
 def SemanticCompletionAvailable_Works_test():
   app = TestApp( handlers.app )
   request_data = BuildRequest( filetype = 'python' )
-  ok_( app.post_json( '/semantic_completion_available',
-                      request_data ).json )
+  assert_that( app.post_json( '/semantic_completion_available', request_data ),
+               has_property('json') )
 
 
 @with_setup( Setup )
 def EventNotification_AlwaysJsonResponse_test():
   app = TestApp( handlers.app )
-  event_data = BuildRequest( contents = 'foo foogoo ba',
+  event_data = BuildRequest( filetype = 'python',
+                             filepath = PathToTestFile( '.ycm_extra_conf.py' ),
+                             contents = 'foo foogoo ba',
                              event_name = 'FileReadyToParse' )
+  assert_that( app.post_json( '/event_notification', event_data ),
+               has_property('json') )
 
-  app.post_json( '/event_notification', event_data ).json
+
+@with_setup( Setup )
+def LoadExtraConfFile_AlwaysJsonResponse_test():
+  app = TestApp( handlers.app )
+  data = { 'filepath': PathToTestFile( '.ycm_extra_conf.py' ) }
+  assert_that( app.post_json( '/load_extra_conf_file', data ),
+               has_property('json') )
+
+
+@with_setup( Setup )
+def IgnoreExtraConfFile_AlwaysJsonResponse_test():
+  app = TestApp( handlers.app )
+  data = { 'filepath': PathToTestFile( '.ycm_extra_conf.py' ) }
+  assert_that( app.post_json( '/ignore_extra_conf_file', data ),
+               has_property('json') )


vheon:
The code looks good. Is this just for consistency with the API (let's return always a JSON) or is this driven by some other issue? 

---

Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/243)

hydrargyrum:
Purely for consistency. I'm trying to make 401 errors (wrong HMAC signature) return JSON too.

Valloric:
It would be nice to get some tests for this. If it's part of the API, we should make sure it doesn't regress.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/243)

hydrargyrum:
I added tests for the 2 requests of the first commit.

For the HMAC handling, it seems a little harder as the HMAC verification is done in a "plugin" that doesn't seem enabled when testing. I lack knowledge if we can enable them and how to fix all tests then.

micbou:
Send another PR for the unauthorized request.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

<sup>**[ycmd/tests/misc_handlers_test.py, line 58 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/243#-K-9qZyzAxvW8WYsm8dt)** ([raw file](https://github.com/valloric/ycmd/blob/a257f2620c9f265d1de773de3bdcc87fde655fa5/ycmd/tests/misc_handlers_test.py#L58)):</sup>
Add a blank line or remove the one above.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/243)

vheon:
Reviewed 1 of 1 files at r3, 1 of 1 files at r4.

---

<sup>**[ycmd/tests/misc_handlers_test.py, line 53 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/243#-K-AAhB0xnfAAuASVrrT)** ([raw file](https://github.com/valloric/ycmd/blob/f221ad8711672e66308c8271c323a71b7d1d58b8/ycmd/tests/misc_handlers_test.py#L53)):</sup>
I know they work like they are, but to me it feels strange not having and assert there; something like:

```python
assert_that( app.post_json( '/load_extra_conf_file', data ).json, is_valid_json() )
```

But I assume that those are personal tastes. :)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/243)

Valloric:
Reviewed 1 of 1 files at r3, 1 of 1 files at r4.

---

<sup>**[ycmd/tests/misc_handlers_test.py, line 53 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/243#-K-AAhB0xnfAAuASVrrT)** ([raw file](https://github.com/valloric/ycmd/blob/f221ad8711672e66308c8271c323a71b7d1d58b8/ycmd/tests/misc_handlers_test.py#L53)):</sup>
I agree with you; lack of an assert-like-something makes it hard to see what's actually being asserted.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/243)

micbou:

---

<sup>**[ycmd/tests/misc_handlers_test.py, line 53 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/243#-K-AAhB0xnfAAuASVrrT)** ([raw file](https://github.com/valloric/ycmd/blob/f221ad8711672e66308c8271c323a71b7d1d58b8/ycmd/tests/misc_handlers_test.py#L53)):</sup>
Since the four tests of this file are similars, they should be written the same way. I suggest to use the same assertion as the first one: `ok_()`.  It will remove the need to import the `hamcrest` module. The second test should also be updated.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/243)

micbou:
Review status: 1 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

<sup>**[ycmd/tests/misc_handlers_test.py, line 45 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/243#-K-VRTJlxwd2l4uXSbtv)** ([raw file](https://github.com/valloric/ycmd/blob/3b9d6cd2b4669d2740490c4dadaddfc4a8bd9f78/ycmd/tests/misc_handlers_test.py#L45)):</sup>
It seems that using `ok_` was a bad idea because the response is an empty dictionary. We should actually check if the object contains the `json` property:
```python
assert_that( app.post_json( '/event_notification', request_data ), has_property( 'json' ) )
```
where `assert_that` and `has_property`come from the `hamcrest` module.
Could you update (again) the four tests? It would be greatly appreciated.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/243)

vheon:
Reviewed 1 of 1 files at r7.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/243)

vheon:
LGTM

Valloric:

---

<sup>**[ycmd/tests/misc_handlers_test.py, line 48 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/243#-K-a9ngyjE8Mgb_JPzM7)** ([raw file](https://github.com/valloric/ycmd/blob/eef39174197b81a410c012961d85cff4f8211efd/ycmd/tests/misc_handlers_test.py#L48)):</sup>
I'm not sure this does the right thing. It's verifying that the WebTest response object has the "json" property, which is not what we want to test for. We want to test that the _value_ of that json is True.

In other words,  you can probably use `ok_( app.post_json( '/event_notification', event_data ).json )`.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/243)

hydrargyrum:
handlers.py's EventNotification sometimes return an empty dict, which would not match ok_ and cause the test to fail

vheon:
What do you mean by "sometimes"?


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/243)

hydrargyrum:
I mean this: https://github.com/Valloric/ycmd/blob/master/ycmd/handlers.py#L75

Valloric:
> handlers.py's EventNotification sometimes return an empty dict, which would not match ok_ and cause the test to fail

Fair point, but I still don't see how checking for the `json` property on the WebTest object resolves that. We need to verify that we can parse the data coming from the handler as json.


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/243)

hydrargyrum:
It seems accessing the json property already checks if it's valid: https://github.com/Pylons/webtest/blob/master/webtest/response.py#L491

Valloric:
Interesting, but that's an implementation detail that's not immediately obvious to the reader of the test code. As such, we shouldn't really depend on it directly. Maybe write a small wrapper function that does this, like `EnsureValidJson` and that then has a comment explaining this bit of impl detail? That could work.


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/243)

PR #245, opened 2015-09-26
Improve Clang download robustness

micbou:
Some AppVeyor runs are failing because of network issues when downloading Clang. This PR should fix this by trying to download Clang multiple times until successful.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/245)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -100,10 +100,28 @@ if ( USE_CLANG_COMPLETER AND
     message( "Downloading Clang ${CLANG_VERSION}" )
 
     set( CLANG_URL "http://llvm.org/releases/${CLANG_VERSION}" )
-    file(
-      DOWNLOAD "${CLANG_URL}/${CLANG_FILENAME}" "${CLANG_LOCAL_FILE}"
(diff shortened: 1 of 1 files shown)


vheon:
Since this is specific to a CI issue should we do it only when testing on CI?

---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/245)

vheon:
Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/245)

micbou:
I don't think so. Clang download may fail for users too.

---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/245)

Valloric:
I'm not a huge fan of this; we're basically putting more code in CMakeLists to work around AppVeyor problems. We've never had a user come to us with "could you try the download in a loop, my network connection can fail," and even if they did, we'd tell them no, they should just retry the install.

I'll merge this because few things are worse than flaky CI, but I still don't like it.

@homu r+

---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/245)

PR #247, opened 2015-10-01
include `/usr/include` will make clang use incorrect macros

haifengkao:
If this flag is set,
TARGET_MAC will be 1
TARGET_OS_IPHONE will be 0
TARGET_IPHONE_SIMULATOR will be 0
in iOS projects,
It causes clang to include OS-X specific headers (which cannot be found in iOS projects)

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/247)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -45,7 +45,6 @@ MAC_INCLUDE_PATHS = [
  '/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1',
  '/usr/local/include',
  '/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include',
- '/usr/include',
  '/System/Library/Frameworks',
  '/Library/Frameworks',
 ]


Valloric:
But doesn't this change then break non-iOS development on OS X? If so, a different solution is necessary.


---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/247)

haifengkao:
The flag in  `ycmd/completers/cpp/flags.py` breaks the iOS development. There are no ways to "undo" the flags in `flags.py`. 
If OSX devs really need the flag, they can add the flag `-I/usr/include` in their own `.ycm_extra_conf.py`

The current solution is far from satisfactory. I believe a better and different solution is necessary for objc development. A possible direction is [YCM-Generator](https://github.com/rdnetto/YCM-Generator).
If we could use YCM-Generator to generate different `.ycm_extra_conf.py` for iOS/OSX projects, the problem would be solved perfectly.

haifengkao:
This patch is not required for the current libclang (3.7).

For the future visitors: 
if you find YouCompleteMe adds some incorrect objective-c flags, you can find these flags at 
`~/.vim/bundle/YouCompleteMe/third_party/ycmd/ycmd/completers/cpp/flags.py`

PR #248, opened 2015-10-03
Replace compiler with language flag if it's a C++ compiler.

clin18:
Currently, if we have a header file for which its corresponding C++ file exists
in the compilation database, we forget that it's supposed to be a C++ header by
stripping away the compiler.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/248)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -137,6 +137,7 @@ def _CallExtraConfFlagsForFile( module, filename, client_data ):
 
 
 def PrepareFlagsForClang( flags, filename ):
+  flags = _CompilerToLanguageFlag( flags )
   flags = _RemoveXclangFlags( flags )
   flags = _RemoveUnusedFlags( flags, filename )
   flags = _SanitizeFlags( flags )
@@ -187,24 +188,36 @@ def _SanitizeFlags( flags ):
   return vector
 
 
+def _CompilerToLanguageFlag( flags ):
+  """When flags come from the compile_commands.json file, the first flag is
+  usually the path to the compiler that should be invoked. We want to replace
+  it with a corresponding language flag.
+  E.g., -x c for gcc and -x c++ for g++."""
+
+  # First flag doesn't start with a '-', so it's probably a compiler.
+  if not flags[ 0 ].startswith( '-' ):
+
+    # If the compiler ends with '++', it's probably a C++ compiler
+    # (E.g., c++, g++, clang++, etc).
+    if flags[ 0 ].endswith( '++' ):
+        language = 'c++'
+    else:
+        language = 'c'
+
+    flags = [ '-x', language ] + flags[ 1: ]
+
+  return flags
+
+
 def _RemoveUnusedFlags( flags, filename ):
   """Given an iterable object that produces strings (flags for Clang), removes
   the '-c' and '-o' options that Clang does not like to see when it's producing
   completions for a file. Same for '-MD' etc.
 
-  Also removes the first flag in the list if it does not
-  start with a '-' (it's highly likely to be the compiler name/path).
-
   We also try to remove any stray filenames in the flags that aren't include
   dirs."""
 
   new_flags = []
-
-  # When flags come from the compile_commands.json file, the first flag is
-  # usually the path to the compiler that should be invoked. We want to strip
-  # that.
-  if not flags[ 0 ].startswith( '-' ):
-    flags = flags[ 1: ]
 
   skip_next = False
   previous_flag_is_include = False
--- a/ycmd/completers/cpp/tests/flags_test.py
+++ b/ycmd/completers/cpp/tests/flags_test.py
@@ -46,20 +46,6 @@ def RemoveUnusedFlags_Passthrough_test():
        flags._RemoveUnusedFlags( [ '-foo', '-bar' ], 'file' ) )
 
 
-def RemoveUnusedFlags_RemoveCompilerPathIfFirst_test():
-  def tester( path ):
-    eq_( expected,
-        flags._RemoveUnusedFlags( [ path ] + expected, filename ) )
-
-  compiler_paths = [ 'c++', 'c', 'gcc', 'g++', 'clang', 'clang++',
-                     '/usr/bin/c++', '/some/other/path', 'some_command' ]
-  expected = [ '-foo', '-bar' ]
-  filename = 'file'
-
-  for compiler in compiler_paths:
-    yield tester, compiler
-
-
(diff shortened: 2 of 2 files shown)


puremourning:
I think we need to add some module tests for this. 


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/248)

Valloric:
Thanks for sending a PR! :)

Agreed with @puremourning, tests needed. Also minor other comments added.


---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[ycmd/completers/cpp/flags.py, line 197 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/248#-K-zpP0Oysv3VlMtu07M)** ([raw file](https://github.com/valloric/ycmd/blob/95d0b356c901f516dee5bfa91eae5d44c8ff85c9/ycmd/completers/cpp/flags.py#L197)):</sup>
This should probably be named `first_flag`, not `compiler`. It might not contain a compiler name.


---

<sup>**[ycmd/completers/cpp/flags.py, line 200 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/248#-K-zpa1IkPq74ZRXVvbr)** ([raw file](https://github.com/valloric/ycmd/blob/95d0b356c901f516dee5bfa91eae5d44c8ff85c9/ycmd/completers/cpp/flags.py#L200)):</sup>
This needs a comment explaining why `++` is magic (I understand why, but the next reader might not).


---

<sup>**[ycmd/completers/cpp/flags.py, line 201 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/248#-K-zpgQqOzCQAx4XLH2N)** ([raw file](https://github.com/valloric/ycmd/blob/95d0b356c901f516dee5bfa91eae5d44c8ff85c9/ycmd/completers/cpp/flags.py#L201)):</sup>
You need tests for this behavior.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/248)

clin18:
Addressed comments and added tests with 954f782. Please take another look.

Some questions:
- There's a clang_helpers.py that contains similar code, should that be updated as well?
- Does it make sense to add a C language flag in case it's *not* a C++ compiler? E.g., `clang foo.cpp` should force `foo.cpp` to be parsed as a C file. I'm not sure if this would interfere with how YCM handles other languages (e.g., Objective-C, C#, etc)

clin18:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

<sup>**[ycmd/completers/cpp/flags.py, line 197 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/248#-K-zpP0Oysv3VlMtu07M)** ([raw file](https://github.com/valloric/ycmd/blob/95d0b356c901f516dee5bfa91eae5d44c8ff85c9/ycmd/completers/cpp/flags.py#L197)):</sup>
Done.


---

<sup>**[ycmd/completers/cpp/flags.py, line 200 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/248#-K-zpa1IkPq74ZRXVvbr)** ([raw file](https://github.com/valloric/ycmd/blob/95d0b356c901f516dee5bfa91eae5d44c8ff85c9/ycmd/completers/cpp/flags.py#L200)):</sup>
Done.


---

<sup>**[ycmd/completers/cpp/flags.py, line 201 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/248#-K-zpgQqOzCQAx4XLH2N)** ([raw file](https://github.com/valloric/ycmd/blob/95d0b356c901f516dee5bfa91eae5d44c8ff85c9/ycmd/completers/cpp/flags.py#L201)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/248)

Valloric:
Assuming `clang foo.cpp` means C code can be tricky. What if the user also passes `-x c++` as well to such an invocation? I'm assuming clang will do the right thing. If not, then it might make sense to do this.

WRT `clang_helpers.py`, that file is loaded directly by old `ycm_extra_conf.py` files. It's been obsolete for a while, but we keep it around for the sake of backwards compatibility. Probably best not to touch it.


---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/248)

clin18:
`clang -x c++ foo.cpp` would turn into `-x c -x c++ foo.cpp`. This is why I replaced the flag at the beginning.

Valloric:
Maybe we should add `-x c` then. Sounds like it wouldn't break users.


---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/248)

micbou:
Reviewed 2 of 2 files at r4.

---

<sup>**[ycmd/completers/cpp/flags.py, line 195 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/248#-K0D3s0tm4OyK7rUG_5M)** ([raw file](https://github.com/valloric/ycmd/blob/e80428fa51d19d721e44c70c9893611d2b3d8980/ycmd/completers/cpp/flags.py#L195)):</sup>
Maybe the docstring should talk about the use of the C language flag if not a C++ compiler.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/248)

clin18:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[ycmd/completers/cpp/flags.py, line 195 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/248#-K0D3s0tm4OyK7rUG_5M)** ([raw file](https://github.com/valloric/ycmd/blob/e80428fa51d19d721e44c70c9893611d2b3d8980/ycmd/completers/cpp/flags.py#L195)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/248)

Valloric:
This seems ready to go; thanks for the PR!

@homu r+


---

Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/248)

PR #249, opened 2015-10-08
Remove Vim-specific code

micbou:
This code is specific to Vim and should be moved to YouCompleteMe. See Valloric/YouCompleteMe#1719.

For @Valloric, is the Vim-related code really needed in `ycmd/completers/completer.py`?
```python
if RunningInsideVim():
  from ycm_client_support import FilterAndSortCandidates
else:
  from ycm_core import FilterAndSortCandidates
```
I tried it without the `RunningInsideVim` part and it was working so I am wondering if we can safely remove it. Ideally, ycmd should not contain any Vim specifics.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/249)

### DIFF
--- a/ycmd/utils.py
+++ b/ycmd/utils.py
@@ -21,15 +21,12 @@ import tempfile
 import os
 import sys
 import signal
-import functools
 import socket
 import stat
 import json
 import subprocess
 import collections
 
-WIN_PYTHON27_PATH = 'C:\python27\python.exe'
-WIN_PYTHON26_PATH = 'C:\python26\python.exe'
 # Creation flag to disable creating a console window on Windows. See
 # https://msdn.microsoft.com/en-us/library/windows/desktop/ms684863.aspx
 CREATE_NO_WINDOW = 0x08000000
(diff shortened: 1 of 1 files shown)


vheon:
That is actually used because the `OmniCompleter` runs on the vim side and not on the server side. That is the only place we should use that though... I don't know if it is possible but how about importing the 

```python
from ycm_core import FilterAndSortCandidates
```

in the `ycmd/completers/completer.py` and override the import just in the `OmniCompleter`?


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/249)

vheon:
Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/249)

Valloric:
If I recall correctly, the `completer.py` changes have to be there since like @vheon said, it's used by the OmniCompleter which runs within Vim.

The PR is nice, thanks for sending it in!

@homu r+


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/249)

puremourning:
Technically, this is a breaking change. I have merged the YouCompleteMe change, so that folk (like me!) using master (or equiv.) ycmd can continue to work. I've confirmed that the change also works with the current subproject commit of ycmd.

micbou:
Yes, I should have said to only merge this PR when the YouCompleteMe one is merged.

Valloric:
Yeah, I noticed the breakage myself. Sorry for the premature merge! At least "mainline" users weren't affected.

PR #251, opened 2015-10-23
Add options to configure dependencies

chubiei:
1. Add an option to configure the path of CMake
2. Add an option to configure the path of LLVM
3. Add an option to configure the Python headers to use
4. Add an option to configure the Python library to use

With these options, ycmd can be built easier on Windows:
- Setup PATH environment variables is no longer required. This will keep your environment clean.
- ycmd can be built with a single line of command. You don't have to configure cmake manually.

```
    %PYTHON% ./install.py --clang-completer --gocode-completer --python-include-dir=%PYTHON_INCLUDE_DIR% --python-library=%PYTHON_LIBRARY% --path-to-llvm-root=%LLVM% --cmake=%CMAKE% --msvc=12
```

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/251)

### DIFF
--- a/build.py
+++ b/build.py
@@ -56,8 +56,8 @@ def NumCores():
     return 1
 
 
-def CheckDeps():
-  if not PathToFirstExistingExecutable( [ 'cmake' ] ):
+def CheckDeps( args ):
+  if not PathToFirstExistingExecutable( [ args.cmake ] ):
     sys.exit( 'Please install CMake and retry.')
 
 
@@ -79,7 +79,7 @@ def _CheckOutput( *popen_args, **kwargs ):
   return output
 
 
-def CustomPythonCmakeArgs():
+def CustomPythonCmakeArgsForMac():
   # The CMake 'FindPythonLibs' Module does not work properly.
   # So we are forced to do its job for it.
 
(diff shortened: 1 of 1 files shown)


vheon:
I would be happy to have the `--cmake` flag (specially on Windows) but the other options are already covered by the `EXTRA_CMAKE_ARGS` env variable. We could think of making it a flag instead of an env variable (like `--extra-cmake-args="-D.....") but I don't think is necessary to break that into separate flags like you did.


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/251)

Valloric:
Like vheon said, the 2. - 4. can already be provided. For a cmake path, I can't see why that would be necessary. Can't the user just amend their PATH to include their custom cmake before calling `install.py`?


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/251)

chubiei:
Provide an option to configure cmake path is very helpful for systems with multiple cmake installed. For example, on Mac, you may have a couple versions of cmake installed by different tools such as MacPorts, homehrew, etc. This option provides the ability to explicitly assign the correct cmake that is necessary for build. Another reason to provide option is that changing the PATH environment variable on Windows may cause other programs to have unexpected behaviors. For example, mingw has its own system of build tools. Changing the PATH environment variable may cause mingw to use the unexpected cmake and leads to unwanted results.

For the other three options (I should have put them into another commit), here's my opinion:
1. We already have options `--system-boost` and `--system-libclang` to configure the linking behavior. Since ycmd depends on 3 libraries (clang, boost, and python) directly, having options to configure the rest of the libraries would be a good idea.
2. Using `--extra-cmake-args` is a great idea, but its hard to define its behavior. Note that in `build.py`, we have a function called `CustomPythonCmakeArgs` (which is renamed to `CustomPythonCmakeArgsForMac` in my commit) to detect the paths of Python library and include headers on Mac automatically. If you want to suppress this behavior but use `--extra-cmake-args` instead of `--python-include-dir` or `--python-library`, you have to parse the entire `--extra-cmake-args`, which is too much for a build script I think.

I admit that `--path-to-llvm-root` conflicts with `--system-libclang`. Since the default is to use the downloaded libclang unless `--system-libclang` is specified, `--path-to-llvm-root` could be removed.

Valloric:
@chubiei Thanks for the PR, but I think this functionality is too niche to be worth the long-term maintenance cost. As such, I'll close this PR.

PR #253, opened 2015-10-30
Ignore commands before the compiler path when replacing it with language flag

micbou:
PR #248 introduced an unexpected behavior when dealing with compilation database. When commands (like `ccache`) are used before the actual compiler, YCM considers the first of those commands as the compiler path, which leads YCM to use the wrong language flag (`-x c`) for C++ files. This was found in issue Valloric/YouCompleteMe#1737.

New behavior is to assume that the flag just before the first one starting with a dash is the compiler path. All flags preceding the compiler path are ignored.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/253)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -197,25 +197,35 @@ def _SanitizeFlags( flags ):
   return vector
 
 
+def _RemoveFlagsPrecedingCompiler( flags ):
+  """Assuming that the flag just before the first flag starting with a dash is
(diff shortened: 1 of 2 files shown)


Valloric:
Awesome, thanks!

We should bump up the linked ycmd commit that YCM is using.

@homu r+


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/253)

PR #255, opened 2015-11-03
clang_reparseTranslationUnit takes CXReparse_Flags, not CXTranslation…

Andersbakken:
Currently there's no flags even existing so one could argue whether it makes sense to incur the extra mutex lock but I didn't want to change anymore than necessary.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/255)

### DIFF
--- a/cpp/ycm/ClangCompleter/TranslationUnit.cpp
+++ b/cpp/ycm/ClangCompleter/TranslationUnit.cpp
@@ -42,6 +42,10 @@ unsigned editingOptions() {
          clang_defaultEditingTranslationUnitOptions();
 }
 
+unsigned reparseOptions( CXTranslationUnit translationUnit ) {
+    return clang_defaultReparseOptions( translationUnit );
+}
+
 
 unsigned completionOptions() {
   return clang_defaultCodeCompleteOptions() |
@@ -141,17 +145,6 @@ std::vector< Diagnostic > TranslationUnit::Reparse(
 }
 
 
-void TranslationUnit::ReparseForIndexing(
-  const std::vector< UnsavedFile > &unsaved_files ) {
-  std::vector< CXUnsavedFile > cxunsaved_files =
-    ToCXUnsavedFiles( unsaved_files );
-
-  Reparse( cxunsaved_files,
-           CXTranslationUnit_PrecompiledPreamble |
-           CXTranslationUnit_SkipFunctionBodies );
-}
-
-
 std::vector< CompletionData > TranslationUnit::CandidatesForLocation(
   int line,
   int column,
@@ -197,7 +190,7 @@ Location TranslationUnit::GetDeclarationLocation(
   const std::vector< UnsavedFile > &unsaved_files,
   bool reparse ) {
   if ( reparse )
-    ReparseForIndexing( unsaved_files );
+    Reparse( unsaved_files );
 
   unique_lock< mutex > lock( clang_access_mutex_ );
 
@@ -228,7 +221,7 @@ Location TranslationUnit::GetDefinitionLocation(
   const std::vector< UnsavedFile > &unsaved_files,
   bool reparse ) {
   if ( reparse )
-    ReparseForIndexing( unsaved_files );
+    Reparse( unsaved_files );
 
   unique_lock< mutex > lock( clang_access_mutex_ );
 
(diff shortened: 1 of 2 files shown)


puremourning:
Travis build failed. Need to fix that. 

Also out of curiosity was there a specific problem this resolved?

Andersbakken:
Ah. Clang somehow didn't complain about this on my system. Strange.

Should work better now.

No actual problem. clang_reparseTranslationUnit doesn't have any actual
flags atm but if it ever gets any this could have caused bugs.

Anders

puremourning:
Understood. I've checked the docs and the code and yes, there is only one reparse option ("none") and the options argument is ignored by the code. I've added some pedantic inline comments but otherwise this LGTM


---

---

<sup>**[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 46 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/255#-K2E4kOSlgxBFFm5kNsx)** ([raw file](https://github.com/valloric/ycmd/blob/6873a54ab3d6fb42f5697bef8d869005da8c9859/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L46)):</sup>
Looks like this will fit on one line within 80 chars.

Also space required before `)`


---

<sup>**[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 337 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/255#-K2E4zo7MNdhVxX-zpqu)** ([raw file](https://github.com/valloric/ycmd/blob/6873a54ab3d6fb42f5697bef8d869005da8c9859/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L337)):</sup>
arguably good style to initialise `options` on declaration.

also `reparseOptions()` is declared as returning `unsigned` and this is declared as `uint`. They are probably the same, but it makes sense to be consistent I suppose.


---

<sup>**[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 342 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/255#-K2E5JwzBSFjQRGeJCph)** ([raw file](https://github.com/valloric/ycmd/blob/6873a54ab3d6fb42f5697bef8d869005da8c9859/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L342)):</sup>
space after `(`


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/255)

puremourning:

---

<sup>**[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 342 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/255#-K2E5JwzBSFjQRGeJCph)** ([raw file](https://github.com/valloric/ycmd/blob/6873a54ab3d6fb42f5697bef8d869005da8c9859/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L342)):</sup>
and before `)` !


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/255)

Valloric:
I'm not sure I understand; what problem is this solving? What benefit do we get with this change?


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/255)

Andersbakken:
Future proofing. The function does not take the arguments that it's
currently being fed. It could have adverse effects in later versions.

Valloric:
OK, I understand what this is doing now. 

Resolve comments from @puremourning and this is good to go.


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/255)

Valloric:

---

<sup>**[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 339 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/255#-K2KKcmjuWIwkTYJBZcA)** ([raw file](https://github.com/valloric/ycmd/blob/6873a54ab3d6fb42f5697bef8d869005da8c9859/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L339)):</sup>
You don't need a mutex for this. In clang, it's unsafe to to access the translation unit object from multiple threads. Calling options func should be perfectly safe.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/255)

Andersbakken:
I pushed an update.

Anders

Valloric:
This looks good. Thanks for the PR!

@homu r+


---

Review status: 1 of 2 files reviewed at latest revision, 4 unresolved discussions.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/255)

Andersbakken:
My pleasure.

Anders

PR #256, opened 2015-11-05
fix throw UnicodeDecodeError when files are mix encoding

kkHAIKE:
eg:
a.h:"utf8",b.h:"gb2312",c.cpp:"include <a.h> <b.h>"

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/256)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 [submodule "third_party/gocode"]
 	path = third_party/gocode
 	url = https://github.com/nsf/gocode.git
+[submodule "third_party/chardet"]
+	path = third_party/chardet
+	url = https://github.com/chardet/chardet.git
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -32,6 +32,8 @@ from ycmd.completers.cpp.ephemeral_values_set import EphemeralValuesSet
 
 import xml.etree.ElementTree
 
+import chardet
+
 CLANG_FILETYPES = set( [ 'c', 'cpp', 'objc', 'objcpp' ] )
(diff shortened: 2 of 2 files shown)


vheon:
Isn't this true also for others completers?


---

Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

<sup>**[ycmd/completers/cpp/clang_completer.py, line 435 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/256#-K2OUDeAYwR6I0idfKf-)** ([raw file](https://github.com/valloric/ycmd/blob/2e532c3617fe7ab92067ef5dec17caff19ab5b15/ycmd/completers/cpp/clang_completer.py#L435)):</sup>
`t` is really a bad name for a variable even if the scope is limited. Why not simply `doc_string`?


---

<sup>**[ycmd/completers/cpp/clang_completer.py, line 438 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/256#-K2OURWJVGRyihLP_CHQ)** ([raw file](https://github.com/valloric/ycmd/blob/2e532c3617fe7ab92067ef5dec17caff19ab5b15/ycmd/completers/cpp/clang_completer.py#L438)):</sup>
spaces inside parentesis, also below.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/256)

Valloric:
Sorry, but we don't support any file encoding other than UTF-8.


---

Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/256)

Valloric:
In light of the above, I'll close this PR.

PR #261, opened 2015-11-22
Fix LLVM path for new 7-zip release

micbou:
With this PR, `PATH_TO_LLVM_ROOT` is correctly set on Windows with the new 7-zip release (15.12).

Issue was found on the [ycm-users](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/ycm-users/Rro51uaFooU/sR3Px2PDBwAJ) list.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/261)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -150,7 +150,14 @@ if ( USE_CLANG_COMPLETER AND
   # folder name the archive produced. It isn't the archive base name.
   # On Windows, the find command is not available so we directly set the path.
   if ( WIN32 )
-    set( PATH_TO_LLVM_ROOT ${CMAKE_CURRENT_BINARY_DIR}/../$_OUTDIR )
+    # Since 7-zip release version 15.12, files are not anymore extracted in
+    # $_OUTDIR folder but directly to the root. So, we check the existence of
+    # $_OUTDIR to appropriately set PATH_TO_LLVM_ROOT.
+    if ( EXISTS ${CMAKE_CURRENT_BINARY_DIR}/../$_OUTDIR )
+      set( PATH_TO_LLVM_ROOT ${CMAKE_CURRENT_BINARY_DIR}/../$_OUTDIR )
+    else()
+      set( PATH_TO_LLVM_ROOT ${CMAKE_CURRENT_BINARY_DIR}/.. )
+    endif()
   else()
     execute_process( COMMAND
       find ${CMAKE_CURRENT_BINARY_DIR}/.. -maxdepth 1 -type d -name clang*


puremourning:
LGTM. Thought it is pretty evil of 7-zip to change its default behaviour. Is there no way to force it to behave consistently, via command line options?

I actually had the exact same build error trying to install for the first time on windows yesterday (windows 10, 64 bit) and can confirm this resolves the build for me.

Doubt this is related, but ... unfortunately, c-family semantic completion (for ycmd/cpp/...) doesn't work properly for me, as it can't find the system headers. Is that "normal" ? If I point it at VC's include dir, via adding `-isystem C:\\Program Files (x86)\Microsoft Visual Studio 14\VC\include`, clang can't parse the system headers.

Is it required to have libc++ installed somewhere ? D'you think we should add something to the readme? Or am I being dense?

---

Reviewed 1 of 1 files at r1.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/261#-:-K3laSuguMoMCQBS_xOz:1185260987)_

micbou:
> Thought it is pretty evil of 7-zip to change its default behaviour.

Yes but the last stable version was released 5 years ago. This is certainly why.

> Is there no way to force it to behave consistently, via command line options?

Unfortunately, I could not find any option to change this behavior.

> Doubt this is related, but ... unfortunately, c-family semantic completion (for ycmd/cpp/...) doesn't work properly for me, as it can't find the system headers. Is that "normal" ? If I point it at VC's include dir, via adding -isystem C:\\Program Files (x86)\Microsoft Visual Studio 14\VC\include, clang can't parse the system headers.

> Is it required to have libc++ installed somewhere ?

For MSVC 14, you need to add the following flags:
```python
msvc_flags = [
 '-target',
  'x86_64-pc-windows-msvc19.00.23026',
  '-isystem',
  r'C:\Program Files (x86)\Windows Kits\10\Include\10.0.10150.0\ucrt'
]
```
Some explanation: the default target used by the Clang installer is `x86_64-w64-windows-gnu` (certainly because they didn't use MSVC to build it) so we need to specify it. `19.00.23026` is the compiler version. Can be found by running `cl.exe`.
Without including `C:\Program Files (x86)\Windows Kits\10\Include\10.0.10150.0\ucrt`, some headers will be missing like `corecrt.h`.

> D'you think we should add something to the readme?

Yes, but since no one raised the issue, I didn't take the time to do it. Now, you leave me no choice!

---







_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/261#-:-K3ledhuaBhJuw7fLys-:-1773977834)_

puremourning:
Legendary. That works perfectly. Thanks ;)

---

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/261#-:-K3lqq-k1aF9VssCaco0:-1398247736)_

Valloric:
Nice!

@homu r+

---

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/261#-:-K3qyZJ-0j5Zjcarzutq:1780035433)_

PR #262, opened 2015-11-27
Update flags.py for more complex compiler names

abigagli:
If you're using a versioned compiler name (e.g. g++-4.9, clang++-3.5) as it might happen when you have more than one compiler installed, current automatic detection of file-type fails as the name doesn't end with just '++'.
My proposed change simply relaxes the criteria a bit so that any name that contains '++' anywhere triggers the language to be c++

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/262)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -220,9 +220,10 @@ def _CompilerToLanguageFlag( flags ):
   if flags[ 0 ].startswith( '-' ):
     return flags
 
-  # If the compiler ends with '++', it's probably a C++ compiler
-  # (e.g., c++, g++, clang++, etc).
-  language = ( 'c++' if flags[ 0 ].endswith( '++' ) else
+  
+  # If the compiler contains '++' potentially followed by e.g. version numbers, it's probably a C++ compiler
+  # (e.g., c++, g++, clang++, c++-x.y, g++-x.y, clang++-x.y etc).
+  language = ( 'c++' if '++' in flags[ 0 ] else
                'c' )
 
   return [ '-x', language ] + flags[ 1: ]


vheon:
Since is a specific case, I would rather use a regex like `\+\+(?-\d.\d)$`.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/262#-:-K48hJaa1x7SN4phkwB7:1066709869)*

micbou:
I agree with @vheon. Tests should also be updated in `ycmd/completers/cpp/tests/flags_test.py` by adding candidates to `CompilerToLanguageFlag_ReplaceCCompiler_test` and `CompilerToLanguageFlag_ReplaceCppCompiler_test` functions.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/262#-:-K48j1srcSb-sB9mkLPE:1540856083)*

abigagli:
Yeah, I initially thought to use a regex, but it was getting a bit messy as versioning could be with 1, 2 or 3 digits and didn't have a right regex out of the top of my head (yes, I'm not a regex guru ;-)) so went for the lazy and simple approach.
Also, python is not my game so I was afraid of making it too complicated.
If you think that's the best way forward, I'll try to flesh out a proper regex-based solution later and resubmit the pull request (unless someone wants to come forward and volunteer, as I'm pretty sure it's pretty clear what I wanted to achieve.... ;-)).
Does that sound good?

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/262#-:-K48kQz5iSuNasquwFB1:1446137382)*

vheon:
Ok, so basically you want `'\+\+(-\d(\.\d){0,2})?$` as a regex. As @micbou said we need tests for this.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/262#-:-K498Cs-gaauxQfnrD9y:122708033)*

Valloric:
@vheon and @micbou have made excellent points. I'd merely like to add that we should pull out the "language" logic into a separate function now that it's going to have a regex in the mix.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/262#-:-K4EKDg36wz5RkwQUQXM:-818632205)*

abigagli:
Ok, I could only get back to this today, and after struggling a bit with python's regex basics, I came up with ```'\+\+(-\d+(\.\d+){0,2})?$'``` as the regex to detect names ending in 
-```++```
-```++-X```
-```++-X.Y```
-```++-X.Y.Z```

Now, I had a quick look at how tests are supposed to work and I should have also added the right tests for my patch, but the problem is that I don't have all the prerequisites listed in the *TESTS.md* document to run the tests myself (I'm on a company laptop on which installations of new software has to go through a lot of administrivia) .
The best I could do was to "manually" test my patch (by repeatedly manually modifying the *compile_commands.json* file and checking every time with *YcmDebugInfo* from vim that language was detected correctly).

I'm not sure what the best way to move forward would be: can I go ahead and create the a new pull request without running the tests myself as they'll be run anyway before accepting request? 
Not sure that's OK, so I'm also pasting my patch here below in case someone wants to have a look and see if I can go ahead with a new pull request for it.
Any suggestion is very much appreciated.

    diff --git a/ycmd/completers/cpp/flags.py b/ycmd/completers/cpp/flags.py
    index 86408f5..a1c9c6b 100644
    --- a/ycmd/completers/cpp/flags.py
    +++ b/ycmd/completers/cpp/flags.py
    @@ -20,6 +20,8 @@
     import ycm_core
     import os
     import inspect
    +import re
    +
     from ycmd import extra_conf_store
     from ycmd.utils import ToUtf8IfNeeded, OnMac, OnWindows
     from ycmd.responses import NoExtraConfDetected
    @@ -220,9 +222,13 @@ def _CompilerToLanguageFlag( flags ):
       if flags[ 0 ].startswith( '-' ):
         return flags
    
    -  # If the compiler ends with '++', it's probably a C++ compiler
    -  # (e.g., c++, g++, clang++, etc).
    -  language = ( 'c++' if flags[ 0 ].endswith( '++' ) else
    +  # If the compiler ends with ++ and optionally with some version digits,
    +  # it's probably a C++ compiler
    +  # (e.g., c++, g++, clang++, g++-4.9, clang++-3.7, etc).
    +
    +  cpp_pattern_regex = r'\+\+(-\d+(\.\d+){0,2})?$'
    +
    +  language = ( 'c++' if re.search( cpp_pattern_regex, flags[ 0 ] ) else
                    'c' )
    
       return [ '-x', language ] + flags[ 1: ]
    diff --git a/ycmd/completers/cpp/tests/flags_test.py b/ycmd/completers/cpp/tests/flags_test.py
    index 7c074d6..453299b 100644
    --- a/ycmd/completers/cpp/tests/flags_test.py
    +++ b/ycmd/completers/cpp/tests/flags_test.py
    @@ -205,7 +205,9 @@ def CompilerToLanguageFlag_ReplaceCCompiler_test():
    
     def CompilerToLanguageFlag_ReplaceCppCompiler_test():
       compilers = [ 'c++', 'g++', 'clang++', '/usr/bin/c++',
    -                '/some/other/path++', 'some_command++' ]
    +                '/some/other/path++', 'some_command++',
    +                'c++-5', 'g++-5.1', 'clang++-3.7.3', '/usr/bin/c++-5',
    +                '/some/other/path++-4.9.3', 'some_command++-5.1' ]
    
       for compiler in compilers:
         yield _ReplaceCompilerTester, compiler, 'c++'

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/262#-:-K4NeGq_PXU11blABK8N:-1948650810)*

puremourning:
If you enable travis for your fork, it should run them when you push the branch to your fork. It's a slow and painful way, but it is a way. 

Btw, you can use virtualenv to install the test prerequisites locally if that works. Ie 

```
$ virtualenv YCMtest
$ source YCMtest/bin/activate
$ pip install -r test_requirements.txt
```

This is what I do at work.

abigagli:
Ok, decided to go for the travis/CI route. Everything seems green, so I'm closing this in favor of new #266

PR #263, opened 2015-11-27
Fix IndexError in Typescript completer when querying a subcommand without argument

micbou:
When typing the command `:YcmCompleter` without argument in Vim, YCM will return the following error `IndexError: list index out of range`. This is because the Typescript completer, unlike other completers, does not check if `arguments` in `OnUserCommand` is defined.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/263)

PR #264, opened 2015-11-29
Move subcommands logic to Completer class

micbou:
Currently, each semantic completer reimplements the `OnUserCommand` and `DefinedSubcommands` in about the same way. To avoid duplication of code, this PR moves the logic of these methods to the `Completer` class. Completers only need to implement a new function `GetSubcommandsMap` that return a map between the subcommands name and the methods to call. The map structure is explained in the method documentation. 

Minor improvements:
 - `DefinedSubcommands` now returns an alphabetically sorted list (tests updated to reflect this);
 - Fix issue in PR #263.

Important: if this PR is merged, the `OmniCompleter` class will need to be updated in YCM by inheriting from `GeneralCompleter` instead of `Completer` or implementing the `GetSubcommandsMap` method.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/264)

### DIFF
--- a/ycmd/completers/completer.py
+++ b/ycmd/completers/completer.py
@@ -90,9 +90,10 @@ class Completer( object ):
   command :YcmCompleter and is passed all extra arguments used on command
   invocation (e.g. OnUserCommand(['first argument', 'second'])).  This can be
   used for completer-specific commands such as reloading external configuration.
-  When the command is called with no arguments you should print a short summary
-  of the supported commands or point the user to the help section where this
-  information can be found.
+  Do not override this function. Instead, you need to implement the
+  GetSubcommandsMap method. It should return a map between the user commands
+  and the methods of your completer. See the documentation of this method for
+  more informations on how to implement it.
 
   Override the Shutdown() member function if your Completer subclass needs to do
   custom cleanup logic on server shutdown."""
@@ -190,7 +191,19 @@ class Completer( object ):
 
 
   def DefinedSubcommands( self ):
-    return []
+    return sorted( self.GetSubcommandsMap().keys() )
+
+
+  def GetSubcommandsMap( self ):
+    """This method should return a dictionary where each key represents the
+    completer command name and its value is a lambda function of this form:
+
+      ( self, request_data ) -> method
+
+    where "method" is the call to the completer method with corresponding
+    parameters. See the already implemented completers for examples.
+    """
+    return {}
 
 
   def UserCommandsHelpMessage( self ):
@@ -242,7 +255,17 @@ class Completer( object ):
 
 
   def OnUserCommand( self, arguments, request_data ):
-    raise NotImplementedError( NO_USER_COMMANDS )
+    if not arguments:
+      raise ValueError( self.UserCommandsHelpMessage() )
+
+    command_map = self.GetSubcommandsMap()
+
+    try:
+      command = command_map[ arguments[ 0 ] ]
+    except KeyError:
+      raise ValueError( self.UserCommandsHelpMessage() )
+
+    return command( self, request_data )
 
 
   def OnCurrentIdentifierFinished( self, request_data ):
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -106,96 +106,39 @@ class ClangCompleter( Completer ):
     return [ ConvertCompletionData( x ) for x in results ]
 
 
-  def DefinedSubcommands( self ):
-    return [ 'GoToDefinition',
-             'GoToDeclaration',
-             'GoTo',
-             'GoToImprecise',
-             'GoToInclude',
-             'ClearCompilationFlagCache',
-             'GetType',
-             'GetParent',
-             'FixIt',
-             'GetDoc',
-             'GetDocQuick' ]
-
-
-  def OnUserCommand( self, arguments, request_data ):
-    if not arguments:
-      raise ValueError( self.UserCommandsHelpMessage() )
-
-    # command_map maps: command -> { method, args }
-    #
-    # where:
-    #  "command" is the completer command entered by the user
-    #            (e.g. GoToDefinition)
-    #  "method"  is a method to call for that command
-    #            (e.g. self._GoToDefinition)
-    #  "args"    is a dictionary of
-    #               "method_argument" : "value" ...
-    #            which defines the kwargs (via the ** double splat)
-    #            when calling "method"
-    command_map = {
-      'GoToDefinition' : {
-        'method' : self._GoToDefinition,
-        'args'   : { 'request_data' : request_data }
-      },
-      'GoToDeclaration' : {
-        'method' : self._GoToDeclaration,
-        'args'   : { 'request_data' : request_data }
-      },
-      'GoTo' : {
-        'method' : self._GoTo,
-        'args'   : { 'request_data' : request_data }
-      },
-      'GoToImprecise' : {
-        'method' : self._GoToImprecise,
-        'args'   : { 'request_data' : request_data }
-      },
-      'GoToInclude' : {
-        'method' : self._GoToInclude,
-        'args'   : { 'request_data' : request_data }
-      },
-      'ClearCompilationFlagCache' : {
-        'method' : self._ClearCompilationFlagCache,
-        'args'   : { }
-      },
-      'GetType' : {
-        'method' : self._GetSemanticInfo,
-        'args'   : { 'request_data' : request_data,
-                     'func'         : 'GetTypeAtLocation' }
-      },
-      'GetParent' : {
-        'method' : self._GetSemanticInfo,
-        'args'   : { 'request_data' : request_data,
-                     'func'         : 'GetEnclosingFunctionAtLocation' }
-      },
-      'FixIt' : {
-        'method' : self._FixIt,
-        'args'   : { 'request_data' : request_data }
-      },
-      'GetDoc' : {
-        'method' : self._GetSemanticInfo,
-        'args'   : { 'request_data'    : request_data,
-                     'reparse'         : True,
-                     'func'            : 'GetDocsForLocationInFile',
-                     'response_buider' : _BuildGetDocResponse, }
-      },
-      'GetDocQuick' : {
-        'method' : self._GetSemanticInfo,
-        'args'   : { 'request_data'    : request_data,
-                     'reparse'         : False,
-                     'func'            : 'GetDocsForLocationInFile',
-                     'response_buider' : _BuildGetDocResponse, }
-      },
+  def GetSubcommandsMap( self ):
+    return {
+      'GoToDefinition'           : ( lambda self, request_data:
+         self._GoToDefinition( request_data ) ),
+      'GoToDeclaration'          : ( lambda self, request_data:
+         self._GoToDeclaration( request_data ) ),
+      'GoTo'                     : ( lambda self, request_data:
+         self._GoTo( request_data ) ),
+      'GoToImprecise'            : ( lambda self, request_data:
+         self._GoToImprecise( request_data ) ),
+      'GoToInclude'              : ( lambda self, request_data:
+         self._GoToInclude( request_data ) ),
+      'ClearCompilationFlagCache': ( lambda self, request_data:
+         self._ClearCompilationFlagCache() ),
+      'GetType'                  : ( lambda self, request_data:
+         self._GetSemanticInfo( request_data, func = 'GetTypeAtLocation' ) ),
+      'GetParent'                : ( lambda self, request_data:
+         self._GetSemanticInfo( request_data,
+                                func = 'GetEnclosingFunctionAtLocation' ) ),
+      'FixIt'                    : ( lambda self, request_data:
+         self._FixIt( request_data ) ),
+      'GetDoc'                   : ( lambda self, request_data:
+         self._GetSemanticInfo( request_data,
+                                reparse = True,
+                                func = 'GetDocsForLocationInFile',
+                                response_builder = _BuildGetDocResponse ) ),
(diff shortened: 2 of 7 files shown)


vheon:
Review status: 0 of 10 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/completer.py, line 197 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IOF6syx0Li7uHmfAP:-K4IOF6syx0Li7uHmfAQ:-1855613381) ([raw file](https://github.com/valloric/ycmd/blob/4e7843f8b84f47d7c8bd23d9e4c4ee01b606f2d7/ycmd/completers/completer.py#L197)):*
I understand this guarantees us that a completer is well implemented but I would rather keep it as a default implementation, hence a completer which do not have any subcommands would not need to implement a dummy method like we have to do right now for the `filename_completer` the `general_completer` and the `general_completer_store`. If you think about it a very important method like `ComputeCandidatesInner` is not an abstract method.

---

*[ycmd/completers/completer.py, line 202 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IN-ofOWyOH4-Rjp1E:-K4IN-ofOWyOH4-Rjp1F:35655501) ([raw file](https://github.com/valloric/ycmd/blob/4e7843f8b84f47d7c8bd23d9e4c4ee01b606f2d7/ycmd/completers/completer.py#L202)):*
What do you think about a map of command to lambda functions? So instead of porting all the completer to the clang_completer format we do the opposite (IIRC is  also how @puremourning prototype the tern.js completer).

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264)*

micbou:
Review status: 0 of 10 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/completer.py, line 197 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IOF6syx0Li7uHmfAP:-K4IQ7e73ynZk2A3_Dwj:1796846855) ([raw file](https://github.com/valloric/ycmd/blob/4e7843f8b84f47d7c8bd23d9e4c4ee01b606f2d7/ycmd/completers/completer.py#L197)):*
I'll remove it in this PR but I think we need a new subclass for semantic completers (`SemanticCompleter` for example) and abstract the `SupportedFiletypes` and `GetSubcommandsMap` methods in it. Those are specific to semantic completers.

I also think that `FilenameCompleter` should derive from `GeneralCompleter` like `UltisnipsCompleter` and `IdentifierCompleter`. `GeneralCompleterStore` should not be a completer. This is just a container.

But this will be for another PR.

---

*[ycmd/completers/completer.py, line 202 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IN-ofOWyOH4-Rjp1E:-K4ISdxOpQ1xhoomvp6W:1445146284) ([raw file](https://github.com/valloric/ycmd/blob/4e7843f8b84f47d7c8bd23d9e4c4ee01b606f2d7/ycmd/completers/completer.py#L202)):*
I need to try it. See if it is simpler.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264)*

puremourning:
Review status: 0 of 10 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/completer.py, line 202 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IN-ofOWyOH4-Rjp1E:-K4IcJLyJcIOilPG1DeA:-1670063936) ([raw file](https://github.com/valloric/ycmd/blob/4e7843f8b84f47d7c8bd23d9e4c4ee01b606f2d7/ycmd/completers/completer.py#L202)):*
Yeah, i'm responsible for the crazy dictionary approach from clang_completer, and i'm not super proud of it. I think the map of lamda is more flexible, and probably neater.

For tern completer, I did something similar to cs completer:

```python
  subcommands = {
    'StartServer':     ( lambda self, request_data, args:
                                        self._StartServer() ),
    'StopServer':      ( lambda self, request_data, args:
                                        self._StopServer() ),
    'ConnectToServer': ( lambda self, request_data, args:
                                       self._ConnectToServer( args ) ),
    'GoToDefinition':  ( lambda self, request_data, args:
                                        self._GoToDefinition( request_data) ),
    'GoTo':            ( lambda self, request_data, args:
                                        self._GoToDefinition( request_data) ),
    'GetType':         ( lambda self, request_data, args:
                                        self._GetType( request_data) ),
    'GetDoc':          ( lambda self, request_data, args:
                                       self._GetDoc( request_data) ),
  }
```

Called like this:

```python
  def DefinedSubcommands( self ):
    return self.subcommands.keys()

  def OnUserCommand( self, arguments, request_data ):
    if not arguments or arguments[ 0 ] not in TernCompleter.subcommands:
      raise ValueError( self.UserCommandsHelpMessage() )

    return TernCompleter.subcommands[ arguments[ 0 ] ]( self,
                                                        request_data,
                                                        arguments[ 1: ] )
```

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264)*

micbou:
Updates:
 - `GetSubcommandsMap` is not an abstract method anymore;
 - Dictionaries are replaced by lambda functions in subcommands map.

---

Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[ycmd/completers/completer.py, line 202 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IN-ofOWyOH4-Rjp1E:-K4ItiPpLt04hAZFjLN6:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/4e7843f8b84f47d7c8bd23d9e4c4ee01b606f2d7/ycmd/completers/completer.py#L202)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264#-:-K4IIVWGLlHiTEpF1Fwb:-81788580)*

micbou:
Fix also a typo in the Clang completer.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264#-:-K4IuJiBduHi9TqHv-mh:-1990956454)*

vheon:
Reviewed 1 of 10 files at r1, 5 of 9 files at r2.
Review status: 6 of 7 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/completer.py, line 268 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IvFmyE5TFSW11s-Cb:-K4IvFmyE5TFSW11s-Cc:-1071073226) ([raw file](https://github.com/valloric/ycmd/blob/f45c847f3d23ddc4696e5b1ff0abf33670eeff53/ycmd/completers/completer.py#L268)):*
So we use only the first argument of `arguments`? Don't we support subcommands that take arguments?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264)*

puremourning:
Review status: 6 of 7 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/completer.py, line 268 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IvFmyE5TFSW11s-Cb:-K4IwYwKtgE8MHyQaS6K:-884771896) ([raw file](https://github.com/valloric/ycmd/blob/f45c847f3d23ddc4696e5b1ff0abf33670eeff53/ycmd/completers/completer.py#L268)):*
Yeah I think we should pass the remaining args, maybe via *args[1:] ?

Fwiw I need to pass an arg for tern, so maybe I can make that change on top (as for now it is yagni)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264)*

micbou:
*[ycmd/completers/completer.py, line 268 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IvFmyE5TFSW11s-Cb:-K4J02OW_naPDkZ9ya65:272040685) ([raw file](https://github.com/valloric/ycmd/blob/f45c847f3d23ddc4696e5b1ff0abf33670eeff53/ycmd/completers/completer.py#L268)):*
Yes, I forgot that. Let me change it.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264)*

vheon:
Review status: 6 of 7 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/completer.py, line 268 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IvFmyE5TFSW11s-Cb:-K4J01GnPtTtZCcHYCxD:-1927487138) ([raw file](https://github.com/valloric/ycmd/blob/f45c847f3d23ddc4696e5b1ff0abf33670eeff53/ycmd/completers/completer.py#L268)):*
If right now is not needed then we could leave it like this and add support later which would be covered by a test I imagine.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264)*

micbou:
*[ycmd/completers/completer.py, line 268 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IvFmyE5TFSW11s-Cb:-K4J2_TcqaDmGLT82yZS:323309314) ([raw file](https://github.com/valloric/ycmd/blob/f45c847f3d23ddc4696e5b1ff0abf33670eeff53/ycmd/completers/completer.py#L268)):*
Ok, not implemented in this PR then. For reference, adding `arguments[ 1: ]` to this line and a parameter to the lambda functions is enough.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264)*

vheon:
Reviewed 1 of 9 files at r2.

---

*[ycmd/completers/completer.py, line 197 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IOF6syx0Li7uHmfAP:-K4J8VLHnfpTt5zLGBsA:2059413157) ([raw file](https://github.com/valloric/ycmd/blob/4e7843f8b84f47d7c8bd23d9e4c4ee01b606f2d7/ycmd/completers/completer.py#L197)):*
> `GeneralCompleterStore` should not be a completer. This is just a container.

I don't agree with this, it is not just a container but basically a [Facade](https://en.wikipedia.org/wiki/Facade_pattern) which merges candidates from different completer, so you use it as a completer but you're actually receiving results from more than one. Maybe it doesn't have the best name but it is a Completer for sure.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264)*

micbou:
*[ycmd/completers/completer.py, line 197 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IOF6syx0Li7uHmfAP:-K4JBBPzynnxuAjL-cAs:-882717328) ([raw file](https://github.com/valloric/ycmd/blob/4e7843f8b84f47d7c8bd23d9e4c4ee01b606f2d7/ycmd/completers/completer.py#L197)):*
I see but the issue is that if you remove the inheritance, it works the same so what's the point to inherit?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264)*

vheon:

---

*[ycmd/completers/completer.py, line 197 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IOF6syx0Li7uHmfAP:-K4JGi10SMnq_PwszDVm:-1672705472) ([raw file](https://github.com/valloric/ycmd/blob/4e7843f8b84f47d7c8bd23d9e4c4ee01b606f2d7/ycmd/completers/completer.py#L197)):*
The reason that even if you don't inherit it continue to work is that we're working in a dynamic language as python and we define all the methods we call on this object; if we were working on a static type language we could not remove the inheritance (or better we would have to specify an interface). So the point to continue to derive from Completer is to be explicit about the interface we want to satisfy. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264)*

vheon:
This LGTM.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264#-:-K4JI0xaRiLzto87_AII:-1464129443)*

puremourning:
LGTM too. @homu r+

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264#-:-K4JJO4X7kKRQT1KgB1U:-1625574807)*

puremourning:
Reviewed 1 of 10 files at r1, 6 of 9 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264)*

Valloric:
This is really good stuff. Thanks for making the change.

---

---

*[ycmd/completers/completer.py, line 197 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/264#-K4IOF6syx0Li7uHmfAP:-K4JV_bjXvfWSXJx1Sd-:-1679060854) ([raw file](https://github.com/valloric/ycmd/blob/4e7843f8b84f47d7c8bd23d9e4c4ee01b606f2d7/ycmd/completers/completer.py#L197)):*
I agree with @vheon here; I prefer explicit interfaces over implicit ones. And like he said, GeneralCompleterStore is more than a container.

But @micbou has a good point in that there might be room for a SemanticCompleter class. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/264#-:-K4JWGmBzzEz_av2ZHzC:-2026891159)*

PR #269, opened 2015-12-09
Support typescript.tsx filetype for typescript completer

ianks:
`typescript.tsx` is a filetype to differentiate regular `.ts` files from `.tsx`. This is necessary to disambiguate typecasting (i.e. `<String>something`) from TSX syntax (i.e. `<SomeComponent />`)

This patch just adds support for ycmd to detect the `typescript.tsx` filetype.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/269)

### DIFF
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -172,7 +172,7 @@ class TypeScriptCompleter( Completer ):
 
 
   def SupportedFiletypes( self ):
-    return [ 'typescript' ]
+    return [ 'typescript', 'typescript.tsx' ]
 
 
   def ComputeCandidatesInner( self, request_data ):


vheon:

---

*[ycmd/completers/typescript/typescript_completer.py, line 175 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/269#-K57APNSr8_HRM6b16h7:-K57APNSr8_HRM6b16h8:320455053) ([raw file](https://github.com/valloric/ycmd/blob/cc6437fcbe195eef9ffea66b852947b347d0374e/ycmd/completers/typescript/typescript_completer.py#L175)):*
Isn't `tsx` the filetype?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/269)*

puremourning:
Need to add some tests :)

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/269#-:-K57LXUcohx2j_wa8WL9:839320411)*

micbou:
I am seeing two issues with this PR. First one is that `typescript.tsx` does not seem to be used as a filetype (see for example this [Vim plugin for Typescript](https://github.com/leafgarland/typescript-vim)). Second one is that even if the filetype is `typescript.tsx`, the Typescript completer will not be used because no folder `typescript.tsx` exists in the `completers` folder. You would need to had a hook to the Typescript completer like we do for the Clang completer.

puremourning:
@micbou could this be related to the other recent issue where we weren't properly parsing the filetype (splitting on '.')? Perhaps we're reading the filetype as `typescript` AND `tsx` when finding the semantic completer, but not when providing suggestions?

micbou:
@puremourning If you are talking about issue Valloric/YouCompleteMe#1766, no because it concerns diagnostics. But I am realizing now that it cannot work when the full filetype contains a `.` for the reason you said: we are splitting on `.`. In this case, the folder should be `tsx` instead of `typescript.tsx`.

ianks:
The Typescript VIM plugin does my support inline tsx highlighting, syntax, etc. I made a plugin which does: https://github.com/ianks/vim-tsx

I decided on using the typescript.tsx filetype since JavaScript does a similar thing with javascript.jsx. The point is that the filetypes need to be differentiated so that inline XML literals can be properly detected and highlighted.

Do you think I should rename the filetype? Also, please let me know any issues that I need to solve before possibly merging. 

And lest I forget, thanks for maintaining this repo!

puremourning:
I'd still like to see at least one integration test and a test file that uses this.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/269#-:-K5L5WXii6F-kbQPqtfk:1384748659)*

micbou:
Currently, the TypeScript completer is loaded when setting the filetype to `typescript.tsx` so I am not sure what your PR is trying to accomplish. Did you test it yourself?
That said, there is an issue when the filename doesn't end with the `.ts` extension. I'm getting the following RuntimeError (`.tsx` extension in this case):
```
2015-12-12 21:01:11,230 - ERROR - Exception from semantic completer (using general): Traceback (most recent call last):
  File "C:\Users\micbou\VIM~1\bundle\YOUCOM~2\THIRD_~1\ycmd\ycmd\..\ycmd\handlers.py", line 104, in GetCompletions
    .ComputeCandidates( request_data ) )
  File "C:\Users\micbou\VIM~1\bundle\YOUCOM~2\THIRD_~1\ycmd\ycmd\..\ycmd\completers\completer.py", line 164, in ComputeCandidates
    candidates = self._GetCandidatesFromSubclass( request_data )
  File "C:\Users\micbou\VIM~1\bundle\YOUCOM~2\THIRD_~1\ycmd\ycmd\..\ycmd\completers\completer.py", line 180, in _GetCandidatesFromSubclass
    raw_completions = self.ComputeCandidatesInner( request_data )
  File "C:\Users\micbou\VIM~1\bundle\YOUCOM~2\THIRD_~1\ycmd\ycmd\..\ycmd\completers\typescript\typescript_completer.py", line 186, in ComputeCandidatesInner
    entries = self._ReadResponse( seq )[ 'body' ]
  File "C:\Users\micbou\VIM~1\bundle\YOUCOM~2\THIRD_~1\ycmd\ycmd\..\ycmd\completers\typescript\typescript_completer.py", line 142, in _ReadResponse
    raise RuntimeError( message[ 'message' ] )
RuntimeError: Error processing request. Could not find file: 'C:/Users/micbou/tests/ycmd-pr-269/test.tsx'.
```
I think this is an issue with the tsserver because I don't see anything wrong with the Typescript completer.

Valloric:
I'm not sure I'm grasping everything here. Is the actual filetype set in Vim `typescript.tsx`? As in, `:set ft?` in vim returns `filetype=typescript.tsx`? If so, note that from Vim's perspective, the filetype then is NOT `typescript.tsx` but `typescript` AND `tsx`. Vim uses a dot to delimit between multiple simultaneous filetypes for a file. Us splitting on that dot to get the filetype list is not merely a YCM thing. Vim is encoding a list as a string for the sake of backwards compatibility. So `foo.bar` is really `['foo', 'bar']`.

So returning `['typescript', 'typescript.tsx']` doesn't really make any sense and isn't going to work (nor should it).

ianks:
Ok, sorry for the confusion all. My self testing was flawed, and I thought this PR was working on my end, when it was not.

So as @micbou stated, this problem seems to only be occurring with files that have the `*.tsx` extension. I tested this with these steps:

1. Set the filetype to `typescript.tsx` on a `test.ts` file
2. Run `:YcmRestartServer`
3. And completion then works

I am not sure where to go from here, as I am not familiar enough with the codebase to pinpoint what could be causing the issue here. Any ideas?

micbou:
It looks like this is fixed with the latest version of typescript. @ianks Could you confirm this?

micbou:
No answer, I am closing.

PR #270, opened 2015-12-11
[READY] Refactoring integration tests

micbou:
### Changes
- Move related completer tests to their own subdirectories:
```
ycmd
  └── tests        
      ├── clang
      │   ├── testdata
      │   ├── __init__.py
      │   ├── get_completions_test.py
      │   ├── diagnostics_test.py
      │   ├── subcommands_test.py
      │   └── utils.py 
      ├── cs ── ... 
      ├── python ── ...
      ├── go ── ...
      ...
```
- Regroup tests in classes and subclasses: take advantage of fixtures (`setUp` and `tearDown` methods) and inheritance.
- Rename tests: names are now shorter;
- Fix coding style;
- Update copyright notices;
- No tests were harmed during the making of this PR: 458 tests before, 458 tests after.

### Issues?
- Code to set the Bottle debug mode:
```python
import bottle

bottle.debug( True )
```
was removed from all tests inheriting from `Handlers_test` where it is set in the `setUp` method. Is it enough to enable debug mode in all tests? I don't know how to check this.
- `Basic_test` and `ZeroBasedLineAndColumn_test` tests in `Cs_Diagnostics_test` are near to be identical. Should we remove one of them?

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/270)

PR #272, opened 2015-12-12
[READY] JavaScript completer using Tern.js

puremourning:
# Overview

Adds support for a JavaScript completer using [Tern.js](http://ternjs.net).

The completer supports:

* Intelligent auto-completion
* Go to definition, references (`GoTo`, etc.)
* Semantic type information for identifiers (`GetType`)
* View documentation comments for identifiers (`GetDoc`)
* Management of `tern` server instance

This provides (for Vim at least) a replacement, for [tern_for_vim](https://github.com/ternjs/tern_for_vim) automagically in ycmd.

## Demo

Completions:

ycm-tern-demo

Subcommands:

ycm-tern-demo-subcommands

## Installation

Run `install.py --tern-completer`

## Caveats and differences vs. tern_for_vim

* Currently, to use it there must be a `.tern-project` configuration file somewhere in the directory tree above Vim's working directory when the first JavaScript file is opened (i.e. when the server starts). I believe this is also true for `tern_for_vim`. 
* Does not support refactoring (rename) which `tern_for_vim` does
* Methods from `Object.prototype` are always included. Tern has the ability to only include them when they match some number of typed characters, as they are otherwise "noise". This is incompatible with ycmd's cache and LCS matching.

## Status

I believe this to be fully working with the above caveats. I think the first one is the most frustrating, but it isn't obvious how to avoid it.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/272)


puremourning:
HMMMMMM.

The appveyor builds are failing, unable to find npm in build.py.

I've tried two approaches:
* [here](https://github.com/puremourning/ycmd-1/blob/7c9fa9aeb7a0e32c18d5fa23aa00b2f005e83c2e/build.py#L260-L267)
* and [here](https://github.com/puremourning/ycmd-1/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/build.py#L260-L272)

@micbou any clues why that might be?

I can see that npm must be there because [appveyor.yaml has uses it](https://github.com/puremourning/ycmd-1/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/build.py#L260-L272)

puremourning:
I think I got it. it's `npm.cmd` not `npm.exe`

puremourning:
Grumble. Very grumble.

The problem is https://bugs.python.org/issue2200

i.e. the extension `.cmd` cannot be used with `distutils.spawn.find_executable` so we have to do its job for it.

micbou:
Welcome to the joy of Windows.

> The problem is https://bugs.python.org/issue2200
> i.e. the extension .cmd cannot be used with distutils.spawn.find_executable so we have to do its job for it.

You can reuse the `FindExecutable` function in `ycmd/utils.py`.

I am not reviewing the tests because they will change with PR #270.

---

Review status: 0 of 27 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/javascript/tern_completer.py, line 164 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5Mfpsr63I9pwJ0S1eK:-K5MfpssVj7qkAA9Ab9H:-1511180357) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L164)):*
Using `localhost` on Windows makes the request to the server very slow (at least on my configuration). Replace it by `127.0.0.1`. Should always work.

---

*[ycmd/completers/javascript/tern_completer.py, line 199 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5MgQLAi8aeMs--k3Cd:-K5MgQLAi8aeMs--k3Ce:-2088993583) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L199)):*
Same comment as above. Maybe add a method to obtain the target (`_ServerLocation` for example).

---

*[ycmd/completers/javascript/tern_completer.py, line 280 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5MhJ3Y1pq6cYP0iu1B:-K5MhJ3Zt76kIOl4mprG:912084283) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L280)):*
Unfortunately, this will fail on Windows. You need to specifically call the Tern binary with the `node` executable.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5MfO8BgXQcSOPmMuAI:111004263)*

puremourning:
Thanks @micbou 'preciate it!

I've fired up my Windows VM to get this to work. :)

---

Review status: 0 of 27 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5MpTZiti9pJ859O1i7:1465479479)*

puremourning:
So strange.

Now i'm getting strange EPERM errors when tern is started by my completer, but not when run manually.

The completer runs it with the following command line:

```
2015-12-12 21:17:10,700 - DEBUG - Starting tern with the following command: C:\Program Files\nodejs\node.exe C:\Users\Ben\vimfiles\bundle\YOUCOM~1\THIRD_~1\ycmd\ycmd\completers\javascript\..\..\..\third_party\tern\bin\tern --port 49867 --host localhost --persistent --no-port-file --verbose
```

But the process dies immediately with:

```
events.js:142
      throw er; // Unhandled 'error' event
      ^

Error: EPERM: operation not permitted, read
    at Error (native)
```

If I run the same command from the command line, it works. If I run it in python interactive shell, it works too.

```
>>> node
'C:\\Program Files\\nodejs\\node.exe'
>>> tern = os.path.dirname( 'ycmd\\completers\\javascript\\tern_completer.py' )
>>> tern
'C:\\Users\\Ben\\vimfiles\\bundle\\YouCompleteMe-Clean\\third_party\\ycmd\\third_party\\tern\\bin\\tern'
>>> command = [ node, tern, '--host', 'localhost', '--persistent', '--verbose' ]
>>> import subprocess
>>> subprocess.Popen( command )
<subprocess.Popen object at 0x000000000241AF98>
>>> Listening on port 49881
```

The only difference seems to be the "shortened" path name with the tilde `~` in it. Could that be related?

Clues? Tips? ;_;

puremourning:
Even with the *same* path in command line python it works:

```
C:\Users\Ben\vimfiles\bundle\YouCompleteMe-Clean\third_party\ycmd>python
Python 2.7.10 (default, May 23 2015, 09:44:00) [MSC v.1500 64 bit (AMD64)] on win32
>>> import os, subprocess
>>> import ycmd.utils
>>> node = ycmd.utils.FindExecutable( 'node' )
>>> tern = 'C:\\Users\\Ben\\vimfiles\\bundle\\YOUCOM~1\\THIRD_~1\\ycmd\\ycmd\\completers\\javascript\\..\\..\\..\\third_party\\tern\\bin\\tern'
>>> tern
'C:\\Users\\Ben\\vimfiles\\bundle\\YOUCOM~1\\THIRD_~1\\ycmd\\ycmd\\completers\\javascript\\..\\..\\..\\third_party\\tern\\bin\\tern'
>>> command = [ node, tern, '--verbose' ]
>>> subprocess.Popen( command )
<subprocess.Popen object at 0x00000000024842B0>
>>> Listening on port 49907

^C
C:\Users\Ben\vimfiles\bundle\YouCompleteMe-Clean\third_party\ycmd>
```

I must be going mad. I even tried with raw Popen rather than `utils.SafePopen` - no joy.

More grumble :p

micbou:
Let me try your `tern-completer-locking-windows` branch.

---

Review status: 0 of 27 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5N7YyOptQ1ztcpl13f:-934907843)*

puremourning:
i pushed the fixed to build.py just now

puremourning:
thanks, btw. sorry to be a pain :(

micbou:
Redirecting stdout and stderr to the logfiles makes the server crash. Need to understand why.

---

Review status: 0 of 27 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5NF9tbQic3dxmcppm6:-2082044945)*

puremourning:
Interesting. It works if i reduce it all into one file and run it manually:

```python

import os
import sys

sys.path.append( os.path.join( os.path.dirname( __file__ ), '..', '..', '..' ) )

print sys.path

from ycmd.utils import ( PathToFirstExistingExecutable, SafePopen,
        GetUnusedLocalhostPort, PathToTempDir )


PATH_TO_TERNJS_BINARY = os.path.abspath( 
    os.path.join(
      os.path.dirname( __file__ ),
      '..',
      '..',
      '..',
      'third_party',
      'tern',
      'bin',
      'tern' ) )


node_path = PathToFirstExistingExecutable( [ 'node' ] )

if not node_path:
  raise RuntimeError( 'Unable to find node executable' )
else:
  print ( 'Using node binary from: ' + node_path )


server_port = GetUnusedLocalhostPort()

if True:
  extra_args = [ '--verbose' ]
else:
  extra_args = []

command = [ node_path,
            PATH_TO_TERNJS_BINARY,
            '--port',
            str( server_port ),
            '--host',
            '127.0.0.1',
            '--persistent',
            '--no-port-file' ] + extra_args

print( 'Starting tern with the following command: ' + ' '.join( command ) )

logfile_format = os.path.join( PathToTempDir(), u'tern_{port}_{std}.log' )

server_stdout = logfile_format.format(
    port = server_port,
    std = 'stdout' )

server_stderr = logfile_format.format(
    port = server_port,
    std = 'stderr' )

try:
  with open( server_stdout, 'w' ) as stdout:
    with open( server_stderr, 'w' ) as stderr:
      server_handle = SafePopen( command,
                                 stdout = stdout,
                                 stderr = stderr )
except Exception:
  print( 'Unable to start Tern.js server' )
  server_port = 0


if ( server_port > 0 and server_handle is not None 
    and server_handle.poll() is None ):
  print ( 'Tern.js Server started with pid: ' +
                str( server_handle.pid ) +
                ' listening on port ' +
                str( server_port ) )
  print ( 'Tern.js Server log files are: ' +
                server_stdout +
                ' and ' +
                server_stderr )

server_handle.wait()
```

Then i can use `postman` in chrome to talk to the server.

puremourning:
Hmmm. Yes. Removing stdout= and stderr= from Popen call fixes it, but the output from the server is lost in the ether. subprocess.PIPE also causes it to crash (and also loses the output).

puremourning:
I even hacked something within waitress/bottle and that works, so it isn't that either.

Valloric:
As usual, absolutely amazing work. Thank you!

I have only minor comments, otherwise LGTM.

You are almost certainly aware of this, but worth mentioning just in case: this will need a section on the Tern completer in the YCM readme.

WRT Popen, I vaguely recall issues with redirecting stdout/stderr from ycmd using Popen on Windows. Can't remember the details, but I think the root cause were bugs in Python's Popen. The comments in our SafePopen have links/mentions to issues, so might be worth checking that out.

---

Reviewed 27 of 27 files at r1.

---

*[test_requirements.txt, line 7 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NLDJnDQS3kwhM7yZU:-K5NLDJojTVD3-uwBAns:997587217) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/test_requirements.txt#L7)):*
Why do we need this?

---

*[ycmd/completers/javascript/tern_completer.py, line 51 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NKnJ1n-lrNHuK6uRO:-K5NKnJ1n-lrNHuK6uRP:1988411782) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L51)):*
Since Tern won't work without a separate tern config file, we should check somewhere does the user have that file set up correctly or not. (Similar to the way we do it for ycm_extra_conf.) Otherwise we'll have no end of "bug reports" saying how people can't get Tern to work.

Of course we'll mention the need for the conf file in the docs, but 0.1% of users will read that (and that's generous).

---

*[ycmd/completers/javascript/tern_completer.py, line 164 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5Mfpsr63I9pwJ0S1eK:-K5NKRjM4utdm3HVhOfX:-335095685) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L164)):*
@micbou That's rather peculiar. Is that true for all services listening on localhost on Windows or just this one? If it's a general thing, we should make sure we don't use localhost elsewhere in ycmd/YCM.

---

*[ycmd/tests/javascript/get_completions_test.py, line 266 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NM0EuO_Hj3reSWE9m:-K5NM0Evm60So0uqPMqK:-2009389862) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/tests/javascript/get_completions_test.py#L266)):*
I started keeping a per-coworker "cookie score" at work that gets incremented every time someone writes tests that are particularly easy to read & maintain. Every 10 points my wife makes a batch of delicious home-made cookies that they then get.

If I ever meet you IRL, remind me that I owe you a boatload of fresh cookies. :D

---

*[ycmd/tests/javascript/test_utils.py, line 27 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NMq_FiWM2gb53YT7d:-K5NMq_FiWM2gb53YT7e:1341584902) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/tests/javascript/test_utils.py#L27)):*
Ugh, why would they do that. Yucky yucky global state.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5NJ4Lm3vQsr0236Eie:1423117794)*

puremourning:

---

*[test_requirements.txt, line 7 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NLDJnDQS3kwhM7yZU:-K5NO5qe772PL9pZn1PV:-2136663917) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/test_requirements.txt#L7)):*
It is needed for `@with_cwd` in the tests. The contextlib package in Python 2.6 doesn't support the decorator syntax

`ycmd/tests/javascript/test_utils.py`:

```python
@contextlib2.contextmanager
def with_cwd( wd ):
...
  # contextlib2 is used because it works in python 2.6 (as a decorator)
```

used as:

```python
@with_setup( Setup )
@with_cwd( TEST_DATA_DIR )
def Subcommands_TernCompleter_Defined_Subcommands_test():
```

---

*[ycmd/completers/javascript/tern_completer.py, line 51 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NKnJ1n-lrNHuK6uRO:-K5NOXudiZXMO-c2kZY1:-427724878) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L51)):*
Fair enough.

I was sorta hoping  I could avoid searching for the project file, which basically duplicates what the tern server is about to do. I suspect if the completer just doesn't return any useful results we're likely to get the bug reports anyway, but you're right that we could at least _hint_ that there is a nonzero amount of setup required (sadly) and point at the docs.

I will take a look at what the least-coupled approach would be.

---

*[ycmd/tests/javascript/get_completions_test.py, line 266 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NM0EuO_Hj3reSWE9m:-K5NPGZizKkKQUl4SZVT:-1263886276) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/tests/javascript/get_completions_test.py#L266)):*
Haha. When you spend you life reviewing other people's code, so you pick up a few tricks here and there about what "non-data ink" is and about expressing key ideas in code succinctly.

Well I try anyway.

---

*[ycmd/tests/javascript/test_utils.py, line 27 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NMq_FiWM2gb53YT7d:-K5NPx9lMpToYcij3BpM:-695200964) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/tests/javascript/test_utils.py#L27)):*
Yeah. I have a laundry list of things to try out to see if this can be improved. But first, getting it to work on windows, second firefighting the million bug reports, then: improvements! :) Ship, ship, ship, right?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

Valloric:

---

*[test_requirements.txt, line 7 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NLDJnDQS3kwhM7yZU:-K5NSB6BJCMfTYpprvmD:-1189144525) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/test_requirements.txt#L7)):*
Ah, thanks for clarifying!

---

*[ycmd/completers/javascript/tern_completer.py, line 51 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NKnJ1n-lrNHuK6uRO:-K5NSLosR5ABDrMEbUgT:-552173560) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L51)):*
Yeah it's shitty that we have to duplicate the logic, but trust me the alternative is worse. We didn't use to have the nice error message for ycm_extra_conf being missing and "doesn't work for me" bug reports whose root cause was no extra conf file we're something like 50% of everything on the tracker.

Similar experience with installation failures because submodules weren't checked out. Now build.py/install.py check for that and tell the user what they need to do.

Both of these error messages cut down bug reports about those problems by _huge_ amounts.

If people can do something stupid, they will. They'll also think it's your fault.

I've since learned this lesson for good.

---

*[ycmd/tests/javascript/test_utils.py, line 27 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NMq_FiWM2gb53YT7d:-K5NTwVccxYSOQdg55PT:-1461877906) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/tests/javascript/test_utils.py#L27)):*
Completely agreed. :)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

micbou:

---

*[ycmd/completers/javascript/tern_completer.py, line 164 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5Mfpsr63I9pwJ0S1eK:-K5NSWWzKRmTGjhOkYuI:-1126047408) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L164)):*
So it was an issue with my configuration. The line
```
127.0.0.1 localhost
```
was missing in my `C:\Windows\System32\drivers\etc\hosts`.
But maybe we should still use `127.0.0.1` to avoid this kind of issue. I read that `localhost` is faster than `127.0.0.1` on Unix because it will use a Unix Domain Socket instead of TCP/IP. Does it apply in this case? Should we use `127.0.0.1` on Windows and `localhost` on Unix?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

micbou:
*[ycmd/completers/javascript/tern_completer.py, line 164 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5Mfpsr63I9pwJ0S1eK:-K5NYX4atwOTY2elfF8Y:1883144534) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L164)):*
For reference: http://stackoverflow.com/a/3715946

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

Valloric:

---

*[ycmd/completers/javascript/tern_completer.py, line 164 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5Mfpsr63I9pwJ0S1eK:-K5NY7LcDayxVArfa4CV:2142033961) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L164)):*
I remember we've had issues with localhost before BTW:
https://github.com/Valloric/YouCompleteMe/issues/1130
https://github.com/Valloric/YouCompleteMe/pull/987

Two instances is more than enough for me to say we should just use 127.0.0.1 everywhere.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

vheon:
Reviewed 27 of 27 files at r1.

---

*[ycmd/completers/javascript/tern_completer.py, line 79 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NWgENEndWG5QZhsGL:-K5NWgENEndWG5QZhsGM:-731551414) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L79)):*
this is never used

---

*[ycmd/completers/javascript/tern_completer.py, line 123 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NX6z6UqK3cs8J0FhY:-K5NX6z6UqK3cs8J0FhZ:-255697972) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L123)):*
Why don't we define the map here directly?

---

*[ycmd/completers/javascript/tern_completer.py, line 164 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5Mfpsr63I9pwJ0S1eK:-K5NXZ9F1aBNMjoYZIYi:123856040) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L164)):*
> Should we use 127.0.0.1 on Windows and localhost on Unix?

That is a very interesting question. How much slow are we talking about? is it noticeable?

---

*[ycmd/tests/javascript/subcommands_test.py, line 40 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5N_3snztZXnhDg6O9O:-K5N_3snztZXnhDg6O9P:-1743227498) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/tests/javascript/subcommands_test.py#L40)):*
Is this test really needed? I mean, don't we handle all the subcommand interface through the Completer class directly?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

puremourning:

---

*[ycmd/completers/javascript/tern_completer.py, line 79 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NWgENEndWG5QZhsGL:-K5Nhj6OfrIU-rimH8sI:-740147598) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L79)):*
Good spot! Thanks.

It was *supposed* to be used in `_Reset`, but there is a typo:

```python
  def _Reset( self ):
    """Callers must hold self._server_state_mutex"""

    if not self.user_options[ 'server_keep_logfiles' ]:
      if self._server_stdout:
        os.unlink( self._server_stdout )
      if self._server_stderr:
        os.unlink( self._server_stderr )
```

Hmm I'd half expect this to give an error of some type, but it is probably the way `not` works in this context.

---

*[ycmd/completers/javascript/tern_completer.py, line 123 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NX6z6UqK3cs8J0FhY:-K5NibI4wqZfCXjZA4Qe:88771670) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L123)):*
Well, this way it is only defined once I guess. It's metadata really... 

---

*[ycmd/completers/javascript/tern_completer.py, line 164 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5Mfpsr63I9pwJ0S1eK:-K5NixBIlNesZfrCT9gY:1030132458) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L164)):*
I didn't notice a difference. TBH i'm not convinced that requesting a tcp socket with a hostname of 'localhost' will actually yield a domain socket, but I could be wrong. I mean, particularly when the server end is actually going to open a real TCP socket, which takes up a real IPv4 port. If it were a unix domain socket, then it wouldn't consume an IPv4 port. Right? 

I think the link referred might be some specific magic in apache and/or PHP. It might require more investigation, but my gut is telling me that unless you ask for a domain socket (and give it a filesystem path) you're going to get an IP socket.

From: http://www.linuxhowtos.org/C_C++/socket.htm

> The address of a socket in the Unix domain is a character string which is basically an entry in the file system.
>
> The address of a socket in the Internet domain consists of the Internet address of the host machine (every computer on the Internet has a unique 32 bit address, often referred to as its IP address).
> In addition, each socket needs a port number on that host.

---

*[ycmd/tests/javascript/subcommands_test.py, line 40 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5N_3snztZXnhDg6O9O:-K5Nl09f6SnUkQGmBIJk:-2089886514) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/tests/javascript/subcommands_test.py#L40)):*
Fair point. Maybe not. It was worthwhile before merging the changes to `GetSubCommandsMap` to check that I was returning a sorted list.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

vheon:

---

*[ycmd/completers/javascript/tern_completer.py, line 79 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NWgENEndWG5QZhsGL:-K5NlpeB4yKByEjaKtGt:-1790194715) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L79)):*
since you just need that option why keep all the dictionary around?

---

*[ycmd/completers/javascript/tern_completer.py, line 123 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NX6z6UqK3cs8J0FhY:-K5NlzBiwECP4jbl8s7-:97523764) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L123)):*
We can't have two instance of this class anyway :confused: 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

puremourning:

---

*[ycmd/completers/javascript/tern_completer.py, line 79 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NWgENEndWG5QZhsGL:-K5NmRjtDoX9ZXg91o_v:2023981779) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L79)):*
iirc it is global anyway, and i'm only retaining a pointer, but you're right just storing the flag is cleaner.

---

*[ycmd/completers/javascript/tern_completer.py, line 123 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NX6z6UqK3cs8J0FhY:-K5NmeycAExNMdfQisC8:921474490) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L123)):*
I know, but

```python
def GetSubcommandMap( self, request_data):
    return {
        # object def here
    }
```

Creates a new dictionary every time that method is called, whereas mine only has one (static) version? Or am i missing something?

Given that the performance of this code path isn't critical, is your main objection that when you  come to find it, it's not on the screen and you have to go searching for the definition? That's a reasonable argument....

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

vheon:

---

*[ycmd/completers/javascript/tern_completer.py, line 123 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NX6z6UqK3cs8J0FhY:-K5Nnb-kUcMQIYNylS7Y:-579543078) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L123)):*
Oh now I see your point. Well yes my argument was that a method called `GetSubcommndMap` don't have a subcommand map in it.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

puremourning:

---

*[ycmd/completers/javascript/tern_completer.py, line 51 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NKnJ1n-lrNHuK6uRO:-K5NttMS_sT6qyasektq:-1528225052) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L51)):*
Well, this turns out to be somewhat harder than I expected because there isn't an obvious time/place to do the check. If we do it in OnFileReadyToParse, then we have the issue Joe found with YCM client eating exceptions generated by that for anything other than files supporting diagnostics. And if we do it in the initialisation, there's no obvious way to return the error to the user (throwing an exception just makes the Completer subclass fail to construct and you get stuck in a loop).

Simple options:
 - just add something to ycmd's log files, so that at least we can see it if people follow CONTRIBUTING.md (they won't)
 - add something to :YcmDebugInfo, so that at least we can see it if people follow CONTRIBUTING.md (they won't)
 - do the check on the first "completion" response - this would be intrusive, and doesn't really work without ':set noshowmode' because the exception text is hidden behind `-- INSERT --` in VIM client
 - any other options? open to suggestions :)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

vheon:

---

*[ycmd/completers/javascript/tern_completer.py, line 51 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NKnJ1n-lrNHuK6uRO:-K5Nwx8W4jPa5x4DIjDp:1876825873) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L51)):*
I think that doing it in `OnFileReadyToParse` is the right place for doing the check: a client send us a file and if we are not able to give it what it wants or if something is missing, then we should tell the user. The API seems clear and intuitive; the only thing that came to mind is that *maybe* we want to check for something missing only on the first of the `OnFileReadyToParse` but for the rest seems clean enough. The problem seems to be in our client not in the server.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

puremourning:

---

*[ycmd/completers/javascript/tern_completer.py, line 51 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NKnJ1n-lrNHuK6uRO:-K5NxpymL1WYw5yHw1lK:-65335275) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L51)):*
I agree. We certainly only want to do it once.

I'll put the server exception in and add some tests. I'll then spend some time on the client. I might require some refactoring.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

vheon:

---

*[ycmd/completers/javascript/tern_completer.py, line 51 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5NKnJ1n-lrNHuK6uRO:-K5NyNlg9ncoDviZTv6N:-1401389744) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L51)):*
If we cache the result we could decide not to change the API, so it would not cause the other clients to do massive changes.

> I'll then spend some time on the client. I might require some refactoring.

Start a thread on the ycm-dev; since @jwilm is also exploring that problem we  can exchange ideas and avoid to overlap with work on this :+1: 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

Valloric:
Review status: 25 of 27 files reviewed at latest revision, 2 unresolved discussions.

---

*[ycmd/completers/javascript/tern_completer.py, line 164 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5Mfpsr63I9pwJ0S1eK:-K5O-mL-UhyCKuj0I936:439648546) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L164)):*
I don't really buy the perf part either. Even if there was an actual difference, it would most likely be negligible. 

I'm more interested in the correctness aspect of always using the raw port vs localhost (see the links I provided above). Using localhost has caused real problems.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

puremourning:
Added a check for .tern-project in the working directory (or parents) in the OnFileReadyToParse event after any server startup. Note this doesn't really work due to the mentioned YCM client behaviour.

Also fixed the windows build/tests:
 - to prevent node from crashing, simply don't create the log files on windows. this isn't great, but i have no clue what causes the crash.
 - run all commands explicitly under node or npm
 - use our fixed version of "find_executable"

---

Review status: 25 of 28 files reviewed at latest revision, 2 unresolved discussions.

---

*[ycmd/completers/javascript/tern_completer.py, line 164 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5Mfpsr63I9pwJ0S1eK:-K5R2Jg93t9IyYo-SgCp:-1846895175) ([raw file](https://github.com/valloric/ycmd/blob/bc6fccbe161ac28fd4fdc59acc4eada18cf660a3/ycmd/completers/javascript/tern_completer.py#L164)):*
changed to use 127.0.0.1 in all cases

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5R2Pk551AIgYg1cb9a:-1119005148)*

micbou:
I'm still investigating on the Windows issue but it takes time. Meanwhile, we can ship the completer in this state and if I find a solution, I will send a PR to fix the issue.

---

Review status: 25 of 28 files reviewed at latest revision, 4 unresolved discussions.

---

*[build.py, line 30 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5R8NL0K8v2kxGIO5_a:-K5R8NL0K8v2kxGIO5_b:963720508) ([raw file](https://github.com/valloric/ycmd/blob/0abbf2d9481d4e7e805c68837a73d0c413377738/build.py#L30)):*
Is it not better to keep the scripts separated from the ycmd codebase? i.e. explicitly adding the `FindExecutable` function in this file instead of importing it. We are already doing that for `OnMac`, `OnWindows`, `PathToFirstExistingExecutable`, etc.

---

*[ycmd/completers/javascript/tern_completer.py, line 352 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5R83RAfNr91HPuWeZI:-K5R83RAfNr91HPuWeZJ:-1564308954) ([raw file](https://github.com/valloric/ycmd/blob/0abbf2d9481d4e7e805c68837a73d0c413377738/ycmd/completers/javascript/tern_completer.py#L352)):*
Don't seem we are doing this in other part of the code but adding a `FIXME` for this issue may be a good idea.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5R6ZATOlWmvc48i-Ih:-1981509877)*

puremourning:
Thanks!

---

Review status: 25 of 28 files reviewed at latest revision, 4 unresolved discussions.

---

*[build.py, line 30 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5R8NL0K8v2kxGIO5_a:-K5RArqhJ477H_MF0BWj:585196973) ([raw file](https://github.com/valloric/ycmd/blob/0abbf2d9481d4e7e805c68837a73d0c413377738/build.py#L30)):*
I wondered about that. Specifically, I wondered why we had the duplication in the first place. Can you remember? Some little bell in the back of my mind is now ringing saying there was an issue with imports before. 

I just really don't like the duplication if it can be avoided :)

---

*[ycmd/completers/javascript/tern_completer.py, line 352 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5R83RAfNr91HPuWeZI:-K5RB91Ly2RgNZ9uRDsv:-1902405889) ([raw file](https://github.com/valloric/ycmd/blob/0abbf2d9481d4e7e805c68837a73d0c413377738/ycmd/completers/javascript/tern_completer.py#L352)):*
OK yes you're right.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5RBCDAMMpNhLLKGija:-1834867277)*

vheon:
Reviewed 1 of 3 files at r3, 1 of 2 files at r4.
Review status: 27 of 28 files reviewed at latest revision, 6 unresolved discussions.

---

*[build.py, line 287 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5RLUlWI7C1YH4bFcyH:-K5RLUlWI7C1YH4bFcyI:298264612) ([raw file](https://github.com/valloric/ycmd/blob/1b9078fdb570709675c675bafc63c0dca92283f3/build.py#L287)):*
I don't understand this change :confused: Couldn't we just replace `PathToFirstExistingExecutable` with `FindExecutable`?

---

*[ycmd/tests/javascript/event_notification_test.py, line 56 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5RNd_UzMQ3rDgI1Qqz:-K5RNd_UzMQ3rDgI1Qr-:-1134990838) ([raw file](https://github.com/valloric/ycmd/blob/1b9078fdb570709675c675bafc63c0dca92283f3/ycmd/tests/javascript/event_notification_test.py#L56)):*
Should we keep those `print` statements?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

vheon:
Reviewed 1 of 2 files at r4.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

puremourning:

---

*[build.py, line 287 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5RLUlWI7C1YH4bFcyH:-K5ROU3mM3Z72i38wsWi:-518482709) ([raw file](https://github.com/valloric/ycmd/blob/1b9078fdb570709675c675bafc63c0dca92283f3/build.py#L287)):*
I flip-flopped on the implementation of this method, while trying to get it to actually work. I went eventually with the one that was closest to my original implementation.
Might be easier to review the code as-is, rather than a diff between revisions.

But FWIW, the method:
 - looks at a list of binaries which are required
 - if it finds it in the path, stores the path found
 - otherwise, raises an error.

Then it uses the npm path to run npm install in the right directory.

Maybe I'm missing your point :x ?

`PathToFirstExistingExecutable` is used to find the first such binary in the list which exists. I need all to exist.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

puremourning:

---

*[ycmd/tests/javascript/event_notification_test.py, line 56 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5RNd_UzMQ3rDgI1Qqz:-K5RPIn4tWOGktAnb_20:-1932368304) ([raw file](https://github.com/valloric/ycmd/blob/1b9078fdb570709675c675bafc63c0dca92283f3/ycmd/tests/javascript/event_notification_test.py#L56)):*
They are useful when the test fails. No-op otherwise.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

puremourning:

---

*[build.py, line 30 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5R8NL0K8v2kxGIO5_a:-K5RPTeZUfLMY3HAG1EV:-985529118) ([raw file](https://github.com/valloric/ycmd/blob/0abbf2d9481d4e7e805c68837a73d0c413377738/build.py#L30)):*
I decoupled them. Still not a fan of the duplication, but not having the coupling means build.py is likely to work on more systems, which means fewer entries on the tracker, which means less pain.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

vheon:

---

*[build.py, line 287 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5RLUlWI7C1YH4bFcyH:-K5RRuqLryRd3AjNMgGr:-1741965300) ([raw file](https://github.com/valloric/ycmd/blob/1b9078fdb570709675c675bafc63c0dca92283f3/build.py#L287)):*
I mean:

Your first approach was:

```python
node_path = PathToFirstExistingExecutable( [ 'node', 'node.exe' ] )
npm_path = PathToFirstExistingExecutable( [ 'npm', 'npm.exe' ] )
 
if not node_path:
  sys.exit( 'node is required for Tern.js completer' )
 
if not npm_path:
  sys.exit( 'npm is required for Tern.js completer' )
 
os.chdir( p.join( DIR_OF_THIS_SCRIPT, 'third_party', 'tern' ) )
 
subprocess.check_call( [ npm_path, 'install', '--production' ] )
```

So you try to find those executable and if one of them is not found then you raise an exception. Now I don't understand why don't simply replace `PathToFirstExistingExecutable` with `FindExecutable`
```python
node_path = FindExecutable( 'node' )
npm_path = FindExecutable( 'npm' )
 
if not node_path:
  sys.exit( 'node is required for Tern.js completer' )
 
if not npm_path:
  sys.exit( 'npm is required for Tern.js completer' )
 
os.chdir( p.join( DIR_OF_THIS_SCRIPT, 'third_party', 'tern' ) )
 
subprocess.check_call( [ npm_path, 'install', '--production' ] )
```

 instead of doing a map of paths and then use only one of the entry in the map :confused: 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

puremourning:

---

*[build.py, line 287 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5RLUlWI7C1YH4bFcyH:-K5RZSc2UvRcqYqdkyAu:960439867) ([raw file](https://github.com/valloric/ycmd/blob/1b9078fdb570709675c675bafc63c0dca92283f3/build.py#L287)):*
Actually, my first approach was:

```python
def SetUpTern():
  for exe in [ 'node', 'npm' ]:
    if not find_executable( exe ):
      sys.exit( '"' + exe + '" is required to set up ternjs' )

  os.chdir( p.join( DIR_OF_THIS_SCRIPT, 'third_party', 'tern' ) )

  subprocess.check_call( [ 'npm', 'install' ] )
```

I would argue that this approach is more general, as it sets out the "dependencies" in data (a list) rather than via hard-coded copy/pasted code for each one. In both cases, the path to node is not used.

How strongly do you feel about this? Retesting this across all the platforms etc. is quite painful :/

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

vheon:

---

*[build.py, line 287 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5RLUlWI7C1YH4bFcyH:-K5R_mfNTP3CeUoPqvOv:1126725335) ([raw file](https://github.com/valloric/ycmd/blob/1b9078fdb570709675c675bafc63c0dca92283f3/build.py#L287)):*
Don't worry I'm fine with the way it is right now, I will spare you the hastle ;)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

puremourning:

---

*[build.py, line 287 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5RLUlWI7C1YH4bFcyH:-K5RabGsEgyUqiKv7lA5:846763659) ([raw file](https://github.com/valloric/ycmd/blob/1b9078fdb570709675c675bafc63c0dca92283f3/build.py#L287)):*
:grin: 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

Valloric:
Reviewed 1 of 2 files at r4, 2 of 2 files at r5.

---

*[build.py, line 30 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5R8NL0K8v2kxGIO5_a:-K5Rxs1eII07Iyw5BYEw:1736062527) ([raw file](https://github.com/valloric/ycmd/blob/0abbf2d9481d4e7e805c68837a73d0c413377738/build.py#L30)):*
I remember there was a good reason why we didn't want to import stuff in the scripts, though I can't remember why.

The duplication is shitty but we'll live with it.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

puremourning:
I've marked this as [READY], though AppVeyor is my nemesis today - seemingly impossibly slow. I'll see if I can massage it.

Also, do you want me to rebase this into one commit ?

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5S-I6hjqH9o5tl2Cyy:246981660)*

micbou:
I think there is an issue with AppVeyor. All tests pass but the run does not finish. It could be that the Tern server is still running.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5S-unNCPxMxtqCTgZp:-823528889)*

puremourning:
that's what i was just thinking. I will check it out.

micbou:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[ycmd/completers/javascript/tern_completer.py, line 412 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5S0md2EtouM1v1-R5j:-K5S0md2EtouM1v1-R5k:1806210134) ([raw file](https://github.com/valloric/ycmd/blob/2c613fbcaaef40fd9f74df54b91e06e4aad44e43/ycmd/completers/javascript/tern_completer.py#L412)):*
Not sure but I think you need to use terminate instead of kill.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

puremourning:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[ycmd/completers/javascript/tern_completer.py, line 412 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5S0md2EtouM1v1-R5j:-K5S0yfcnwZVEDkLuLfC:-1553457775) ([raw file](https://github.com/valloric/ycmd/blob/2c613fbcaaef40fd9f74df54b91e06e4aad44e43/ycmd/completers/javascript/tern_completer.py#L412)):*
I think you might be right. There's a `utils.TerminateProcess` which looks much more aggressive and has windows-specifics.

I also have tons of node processes on my windows system, so it is obviously the problem!

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

micbou:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[ycmd/completers/javascript/tern_completer.py, line 412 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5S0md2EtouM1v1-R5j:-K5S0yQSaL5E-gh9bX7h:571240108) ([raw file](https://github.com/valloric/ycmd/blob/2c613fbcaaef40fd9f74df54b91e06e4aad44e43/ycmd/completers/javascript/tern_completer.py#L412)):*
Yes, I was writing a comment about this function. Certainly a good idea to use it.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

Valloric:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[ycmd/completers/javascript/tern_completer.py, line 412 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5S0md2EtouM1v1-R5j:-K5S1IrFq0ph146B-rSH:-1603550021) ([raw file](https://github.com/valloric/ycmd/blob/2c613fbcaaef40fd9f74df54b91e06e4aad44e43/ycmd/completers/javascript/tern_completer.py#L412)):*
Yep, we've fought this battle before. Use `utils.TerminateProcess`.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272)*

Valloric:
Since this is likely caused by us not killing the Tern server properly, this is great instance of AppVeyor saving our butt by catching a nasty issue before it went out to users! :)

---

Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5S23wsk53q626si_RF:-1046412612)*

puremourning:
There seems to be a wider issue. Basically the @atexit stuff isn't actually firing, so `.kill()` or not, `Shutdown()` isn't being called. I'll see what I can find out.

I think this project has increased my respect for the folk who develop for windows (or actually cross-platform stuff) in a more general capacity. It takes a special type of genius to work in this environment :grinning: 

---

Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[ycmd/completers/javascript/tern_completer.py, line 412 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/272#-K5S0md2EtouM1v1-R5j:-K5S1mZkqgC0uqq-fsKH:-1929002460) ([raw file](https://github.com/valloric/ycmd/blob/2c613fbcaaef40fd9f74df54b91e06e4aad44e43/ycmd/completers/javascript/tern_completer.py#L412)):*
I think this project has increased my respect for the poor folk who develop for windows (or actually cross-platform stuff) in a more general capacity. It takes a special type of genius to use `cmd.exe` and use the `\` key all the time. And a special type of patience required to do development without so many of the creature comforts of POSIX-like systems. Like a shell. Did I mention the shell? I hate the shell.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5S2Z2OYgACTbkTq5d0:-1872937318)*

puremourning:
status changed back to [WIP]

---

Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5S7qAmxrBEa8TX-bq2:1092888025)*

micbou:
> There seems to be a wider issue. Basically the @atexit stuff isn't actually firing, so .kill() or not, Shutdown() isn't being called. I'll see what I can find out.

Yes, I know about this issue. It was discussed in Valloric/YouCompleteMe#876 but no satisfactory solution was found. This is on my list of things to improve on Windows.

Anyway, I see that you decided to manually stop the server and this is the right thing to do. Even without this issue, you don't want to pile up all those Tern servers and kill them in one go at the end.

---

Review status: 22 of 28 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5SFPr5lbFtgAmvff1Q:1602575823)*

puremourning:
I see, so we have 2 known issues on windows:

- downstream logging is disabled
- the tern server is left running on vim exit.

The former is not (yet) well-understood and the latter is a "known" issue on windows.

It also looks like the AppVeyor tests are now passing (woo!). So based on that, switching this back to [READY]

---

Review status: 22 of 28 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5SOEQiT5_hNMQ_icbe:309167075)*

Valloric:
I'm fine with merging this with the tern server still running after vim exit on Windows because we have the same issue with OmniSharp, but that's a really nasty problem. It was sorta fine before because we didn't officially support Windows, but now that we do, it's just looks bad.

With that, @homu r+

---

Review status: 22 of 28 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/272#-:-K5TE8LNkMGcR95CWMDu:-719748858)*

vheon:
>  I'm fine with merging this with the tern server still running after vim exit on Windows because we have the same issue with OmniSharp, but that's a really nasty problem. It was sorta fine before because we didn't officially support Windows, but now that we do, it's just looks bad.

Should we write it somewhere?

PR #279, opened 2015-12-20
Remove coverage options on Travis because we can

micbou:
`run_tests.py` automatically uses coverage options if the `COVERAGE` environment variable is defined.

Maybe it would be better to change the name of this environment variable to `YCMD_COVERAGE`, `COVERAGE` is too generic.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/279)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ install:
   - source travis/travis_install.sh
 compiler:
   - gcc
-script: ./run_tests.py --with-coverage --cover-package=ycmd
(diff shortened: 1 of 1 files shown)


Valloric:
Nice!

@homu r+

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/279#-:-K5xZiwaVrcUXOmFGuTn:1780035433)*

PR #280, opened 2015-12-20
[READY] Recognise ~/.tern-config in Tern completer

### LINKED ISSUE #277: Tern completer: Recognise ~/.tern-config
## Problem 

Tern supports a "global" project file (`~/.tern-config`) which contains global defaults. In theory this can be used instead of `.tern-project`.

The ycmd tern completer does not check for this file and issues its warning anyway if no `.tern-project` file could be found.

## Requirement

Don't issue a warning when a `.tern-config` file exists, as this could well contain enough info for useful javascript completion for many projects.

puremourning:
Fixes #277 

We simply suppress the warning if we detect a `.tern-project` or a user's `~/.tern-config`.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/280)

### DIFF
--- a/ycmd/completers/javascript/tern_completer.py
+++ b/ycmd/completers/javascript/tern_completer.py
@@ -70,17 +70,7 @@ def ShouldEnableTernCompleter():
   return True
 
 
-def FindTernProjectFile( starting_directory ):
-  # We use a dummy_file here because AncestorFolders requires a file name and we
-  # don't have one. Something like '.' doesn't work because, while
-  # os.path.dirname( /a/b/c/. ) returns /a/b/c, AncestorFolders calls
-  # os.path.abspath on it, so the /. is removed.
-  starting_file = os.path.join( starting_directory, 'dummy_file' )
-  for folder in utils.AncestorFolders( starting_file ):
-    tern_project = os.path.join( folder, '.tern-project' )
-    if os.path.exists( tern_project ):
-      return tern_project
-
+def PathToGlobalTernConfig():
   # As described here: http://ternjs.net/doc/manual.html#server a global
   # .tern-config file is also supported for the Tern server. This can provide
   # meaningful defaults (for libs, and possibly also for require paths), so
@@ -92,6 +82,20 @@ def FindTernProjectFile( starting_directory ):
     return tern_config
 
   return None
+
+
+def FindTernProjectFile( starting_directory ):
+  # We use a dummy_file here because AncestorFolders requires a file name and we
+  # don't have one. Something like '.' doesn't work because, while
+  # os.path.dirname( /a/b/c/. ) returns /a/b/c, AncestorFolders calls
+  # os.path.abspath on it, so the /. is removed.
+  starting_file = os.path.join( starting_directory, 'dummy_file' )
+  for folder in utils.AncestorFolders( starting_file ):
+    tern_project = os.path.join( folder, '.tern-project' )
+    if os.path.exists( tern_project ):
+      return tern_project
+
+  return PathToGlobalTernConfig()
 
 
 class TernCompleter( Completer ):
--- a/ycmd/completers/javascript/tern_completer.py
+++ b/ycmd/completers/javascript/tern_completer.py
@@ -70,18 +70,11 @@ def ShouldEnableTernCompleter():
   return True
 
 
-def PathToGlobalTernConfig():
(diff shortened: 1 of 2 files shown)


puremourning:
OK so coveralls shamed me into writing a test. We already require that there is no user config, so we might as well abuse that.

vheon:
This :lgtm: :+1: 

---

Reviewed 2 of 2 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/280#-:-K6-J37inSBiOeZg8Y-o:-2087071121)*

Valloric:

---

*[ycmd/tests/javascript/event_notification_test.py, line 40 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/280#-K60gZLF5L81TecGFUHe:-K60gZLF5L81TecGFUHf:-305905334) ([raw file](https://github.com/valloric/ycmd/blob/48aadc1b32ca01ff9e375b44b5b58528e6f273b0/ycmd/tests/javascript/event_notification_test.py#L40)):*
Doesn't this mean that a `~/.tern-config` file will be overwritten if a developer runs the ycmd test suite on their machine? They might have such a file because they use Tern.js.

---

*[ycmd/tests/javascript/event_notification_test.py, line 86 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/280#-K60guGmmoEXAqzCxf2k:-K60guGmmoEXAqzCxf2l:906269391) ([raw file](https://github.com/valloric/ycmd/blob/48aadc1b32ca01ff9e375b44b5b58528e6f273b0/ycmd/tests/javascript/event_notification_test.py#L86)):*
I see, so you check for it first... still, this isn't ideal. 

How about this: instead of calling `os.path.expanduser( '~/.tern-config' )` directly in the code, create a wrapper `PathToGlobalTernConfig` function that calls that and then mock out that function in your test.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/280)*

Valloric:

---

*[ycmd/tests/javascript/event_notification_test.py, line 86 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/280#-K60guGmmoEXAqzCxf2k:-K60hTwmMS2bJUXj6mHL:-490955461) ([raw file](https://github.com/valloric/ycmd/blob/48aadc1b32ca01ff9e375b44b5b58528e6f273b0/ycmd/tests/javascript/event_notification_test.py#L86)):*
Or better yet, have a `GlobalTernConfigExists` function that returns a bool, and mock that out in tests.

In general, I've found that mocking out filesystem access for tests leads to more maintainable test code.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/280)*

puremourning:
I have applied the mocking approach for the tests.

---

---

*[ycmd/tests/javascript/event_notification_test.py, line 40 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/280#-K60gZLF5L81TecGFUHe:-K64EC0IC6BKjfyiO6nk:-1988458546) ([raw file](https://github.com/valloric/ycmd/blob/48aadc1b32ca01ff9e375b44b5b58528e6f273b0/ycmd/tests/javascript/event_notification_test.py#L40)):*
Yeah, this is unfortunate. It was the only way I could think of to test the actual code without invasive patching.

---

*[ycmd/tests/javascript/event_notification_test.py, line 86 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/280#-K60guGmmoEXAqzCxf2k:-K64ETPTZ0SKU0Q-AsLQ:-1883809049) ([raw file](https://github.com/valloric/ycmd/blob/48aadc1b32ca01ff9e375b44b5b58528e6f273b0/ycmd/tests/javascript/event_notification_test.py#L86)):*
Good point. It still leads to the same amount of uncoverable code, but testing the logic is more useful than not.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/280#-:-K64GVefz7tG85FkC8eX:-444252651)*

Valloric:
Nice!

@homu r+

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/280#-:-K66vBHtVELNulA4fSV0:1780035433)*

PR #281, opened 2015-12-20
[READY] Remove timeout from Tern completer

### LINKED ISSUE #278: Tern completer: Tern server crashes on moderate sized files
## Problem

On moderate to large sized projects (such as Tern itself), the Tern server is quitting, and `:YcmDebugInfo` reports the Tern server crashed. Users get no completions/etc. until issuing `:YcmCompleter StartServer`. The Tern server may then crash again.

## Cause

ycmd's Tern completer uses a (conservative) 500ms timeout for requests to the Tern server. The Tern docs indicate this should return a HTTP error when time-out is hit, but it does not. Instead the server crashes. See Tern issue: https://github.com/ternjs/tern/issues/702

## Requirement

Ideally, get a fix upstream in Tern and update the submodule. Otherwise, if the linked Tern issue is not resolved, ycmd tern completer needs to handle this error, potentially including any or all of the following:

- removing or increasing the timeout sent to Tern server 
- using a client-side session timeout instead
- automatically restarting the crashed server

puremourning:
Fixes #278 

As described on the issue, the Tern server exits when a timeout is hit. The Tern server's timeout is actually a guesstimate anyway, and ignores I/O time. We're better off just removing it and relying on ycmd client requests timing out. It doesn't appear that the other completers add any other time-out.

We can re-asses should https://github.com/ternjs/tern/issues/702 be fixed.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/281)

### DIFF
--- a/ycmd/completers/javascript/tern_completer.py
+++ b/ycmd/completers/javascript/tern_completer.py
@@ -254,7 +254,7 @@ class TernCompleter( Completer ):
     the files are being updated.
 
     The request block should contain the optional query block only. The file
-    data and timeout are are added automatically."""
+    data are added automatically."""
 
     if not self._ServerIsRunning():
       raise ValueError( 'Not connected to server' )
@@ -271,7 +271,6 @@ class TernCompleter( Completer ):
     full_request = {
       'files': [ MakeIncompleteFile( x, file_data[ x ] )
                  for x in file_data.keys() ],
-      'timeout': 500,
     }
     full_request.update( request )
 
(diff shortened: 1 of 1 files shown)


vheon:
:lgtm: 

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/281#-:-K6-JxGP5l-iqT_mbqRu:980077423)*

Valloric:
Nice!

@homu r+

PR #282, opened 2015-12-22
[READY] Properly shut down on Windows

micbou:
### Problem

When killing a process on Windows, it is terminated like it received a `SIGKILL` signal on Linux. This means that @atexit functions will not be called. In our case, `ServerCleanup` is not executed:
completers `Shutdown` methods are not called, neither the `Shutdown` function in the global extra conf. In particular, subservers like `OmniSharpServer` and `Tern.js` are left running.
This issue happens in two situations:
 - [a client like YCM terminates the ycmd server](https://github.com/Valloric/YouCompleteMe/blob/master/python/ycm/youcompleteme.py#L175);
 - [watchdog plugin pulls the trigger](https://github.com/Valloric/ycmd/blob/master/ycmd/watchdog_plugin.py#L92).

How to solve it?

### Solution

First situation should be solved by adding a new handler `/shutdown` to properly shut down the ycmd server from a client. Unfortunately, `waitress` and `bottle` don't provide an easy way to shut it down. Using `sys.exit()` in the `/shutdown` handler will not work because it is run in a thread. The solution is to use the same mechanism as in the watchdog plugin. After cleaning up, we start a thread to immediately kill the server.
Second situation is fixed by using the function associated to the `/shutdown` handler in the watchdog plugin.

### Tests

To test that the shutdown handler and the watchdog plugin are properly working, we need to start the ycmd server for real. We cannot use [webtest](https://webtest.readthedocs.org/en/latest/) here. I added 4 tests, which are a combination of shutting down ycmd without and with a subserver, from a request and from the watchdog plugin.
[psutil](http://pythonhosted.org/psutil/) module was added to the test requirements to work with processes in a portable way.
In order to test the watchdog plugin, I added a command-line argument to set the `check_interval_seconds` variable.
Since we are starting subprocesses in those tests, some changes were needed to enable coverage in Python subprocesses. Another change to improve coverage was to update Waitress to version 0.8.10. See the commit message for some explanation.
A benefit of those tests is a significant increase in coverage (~6%).

###  Caveats

- Tests may introduce flakyness. Robustness should be improved;
- Coverage seems to cause issue with the Gocode daemon. Not sure if it is related to this PR. Need to investigate.

### Next

- Update YCM to use the `shutdown` handler (fix issue Valloric/YouCompleteMe#876);
- Fix logfiles clean up on Windows.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/282)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,14 @@
 [report]
 omit =
+    */third_party/requests/*
     */tests/*
     */__init__.py
+
+[run]
+parallel = True
+source =
+    ycmd
+
+[paths]
+source =
+    ycmd/
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ pip-log.txt
 
 # Unit test / coverage reports
 .coverage
+.coverage.*
 cover/
 .tox
 nosetests.xml
--- a/ci/travis/travis_install.sh
+++ b/ci/travis/travis_install.sh
@@ -57,6 +57,11 @@ pip install -U pip wheel setuptools
 pip install -r test_requirements.txt
 npm install -g typescript
 
+# Enable coverage for Python subprocesses. See:
+# http://coverage.readthedocs.org/en/coverage-4.0.3/subprocess.html
+echo -e "import coverage\ncoverage.process_startup()" > \
+  ${PYENV_ROOT}/versions/${PYENV_VERSION}/lib/python${YCMD_PYTHON_VERSION}/site-packages/sitecustomize.py
+
 ############
 # rust setup
 ############
--- a/examples/example_client.py
+++ b/examples/example_client.py
@@ -113,10 +113,10 @@ class YcmdHandle( object ):
 
   def Shutdown( self ):
     if self.IsAlive():
-      self._popen_handle.terminate()
+      self.PostToHandlerAndLog( 'shutdown' )
 
 
-  def PostToHandlerAndLog( self, handler, data ):
+  def PostToHandlerAndLog( self, handler, data = None ):
     self._CallHttpie( 'post', handler, data )
 
 
--- a/run_tests.py
+++ b/run_tests.py
@@ -192,7 +192,9 @@ def NoseTests( parsed_args, extra_nosetests_args ):
       nosetests_args.extend( COMPLETERS[ key ][ 'test' ] )
 
   if parsed_args.coverage:
-    nosetests_args += [ '--with-coverage', '--cover-package=ycmd',
+    nosetests_args += [ '--with-coverage',
+                        '--cover-erase',
+                        '--cover-package=ycmd',
                         '--cover-html' ]
 
   if extra_nosetests_args:
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -6,6 +6,7 @@ WebTest>=2.0.20
 ordereddict>=1.1
 nose-exclude>=0.4.1
 unittest2>=1.1.0
+psutil>=3.3.0
 # This needs to be kept in sync with submodule checkout in third_party
 future==0.15.2
 codecov>=2.0.5
--- a/ycmd/__main__.py
+++ b/ycmd/__main__.py
@@ -36,16 +36,14 @@ import sys
 import logging
 import json
 import argparse
-import waitress
 import signal
 import os
 import base64
-from ycmd import user_options_store
-from ycmd import extra_conf_store
-from ycmd import utils
-from ycmd.watchdog_plugin import WatchdogPlugin
+
+from ycmd import extra_conf_store, user_options_store, utils
 from ycmd.hmac_plugin import HmacPlugin
 from ycmd.utils import ToBytes, ReadFile, OpenForStdHandle
+from ycmd.wsgi_server import StoppableWSGIServer
 
 
 def YcmCoreSanityCheck():
@@ -166,10 +164,11 @@ def Main():
 
   PossiblyDetachFromTerminal()
 
-  # This can't be a top-level import because it transitively imports
+  # These can't be top-level imports because they transitively import
   # ycm_core which we want to be imported ONLY after extra conf
   # preload has executed.
   from ycmd import handlers
+  from ycmd.watchdog_plugin import WatchdogPlugin
   handlers.UpdateUserOptions( options )
   handlers.SetHmacSecret( hmac_secret )
   SetUpSignalHandler( args.stdout, args.stderr, args.keep_logfiles )
@@ -177,10 +176,11 @@ def Main():
                                         args.check_interval_seconds ) )
   handlers.app.install( HmacPlugin( hmac_secret ) )
   CloseStdin()
-  waitress.serve( handlers.app,
-                  host = args.host,
-                  port = args.port,
-                  threads = 30 )
+  handlers.wsgi_server = StoppableWSGIServer( handlers.app,
+                                              host = args.host,
+                                              port = args.port,
+                                              threads = 30 )
+  handlers.wsgi_server.Run()
 
 
 if __name__ == "__main__":
--- a/ycmd/__main__.py
+++ b/ycmd/__main__.py
@@ -104,6 +104,9 @@ def ParseArguments():
                               '[debug|info|warning|error|critical]' )
   parser.add_argument( '--idle_suicide_seconds', type = int, default = 0,
                        help = 'num idle seconds before server shuts down')
+  parser.add_argument( '--check_interval_seconds', type = int, default = 600,
+                       help = 'interval in seconds to check server '
+                              'inactivity' )
   parser.add_argument( '--options_file', type = str, required = True,
                        help = 'file with user options, in JSON format' )
   parser.add_argument( '--stdout', type = str, default = None,
@@ -170,7 +173,8 @@ def Main():
   handlers.UpdateUserOptions( options )
   handlers.SetHmacSecret( hmac_secret )
   SetUpSignalHandler( args.stdout, args.stderr, args.keep_logfiles )
-  handlers.app.install( WatchdogPlugin( args.idle_suicide_seconds ) )
+  handlers.app.install( WatchdogPlugin( args.idle_suicide_seconds,
+                                        args.check_interval_seconds ) )
   handlers.app.install( HmacPlugin( hmac_secret ) )
   CloseStdin()
   waitress.serve( handlers.app,
--- a/ycmd/handlers.py
+++ b/ycmd/handlers.py
@@ -29,6 +29,7 @@ import json
 import logging
 import traceback
 from bottle import request
+from threading import Thread
 
 import ycm_core
 from ycmd import extra_conf_store, hmac_plugin, server_state, user_options_store
@@ -46,6 +47,7 @@ _server_state = None
 _hmac_secret = bytes()
 _logger = logging.getLogger( __name__ )
 app = bottle.Bottle()
+wsgi_server = None
 
 
 @app.post( '/event_notification' )
@@ -223,6 +225,14 @@ def DebugInfo():
   return _JsonResponse( '\n'.join( output ) )
 
 
+@app.post( '/shutdown' )
+def Shutdown():
+  _logger.info( 'Received shutdown request' )
+  ServerShutdown()
+
+  return _JsonResponse( True )
+
+
 # The type of the param is Bottle.HTTPError
 def ErrorHandler( httperror ):
   body = _JsonResponse( BuildExceptionResponse( httperror.exception,
@@ -259,9 +269,20 @@ def _GetCompleterForRequestData( request_data ):
     return _server_state.GetFiletypeCompleter( [ completer_target ] )
 
 
-@atexit.register
 def ServerShutdown():
-  _logger.info( 'Server shutting down' )
+  def Terminator():
+    if wsgi_server:
+      wsgi_server.Shutdown()
+
+  # Use a separate thread to let the server send the response before shutting
+  # down.
+  terminator = Thread( target = Terminator )
+  terminator.daemon = True
+  terminator.start()
+
+
+@atexit.register
+def ServerCleanup():
   if _server_state:
     _server_state.Shutdown()
     extra_conf_store.Shutdown()
--- a/ycmd/tests/client_test.py
+++ b/ycmd/tests/client_test.py
@@ -43,7 +43,7 @@ from ycmd.tests.test_utils import BuildRequest
 from ycmd.user_options_store import DefaultOptions
 from ycmd.utils import ( GetUnusedLocalhostPort, OpenForStdHandle,
                          PathToCreatedTempDir, ReadFile, RemoveIfExists,
-                         SafePopen, ToBytes, ToUnicode )
+                         SafePopen, SetEnviron, ToBytes, ToUnicode )
 
 HEADERS = { 'content-type': 'application/json' }
 HMAC_HEADER = 'x-ycm-hmac'
@@ -90,6 +90,11 @@ class Client_test( object ):
       self._port = GetUnusedLocalhostPort()
       self._location = 'http://127.0.0.1:' + str( self._port )
 
+      # Define environment variable to enable subprocesses coverage. See:
+      # http://coverage.readthedocs.org/en/coverage-4.0.3/subprocess.html
+      env = os.environ.copy()
+      SetEnviron( env, 'COVERAGE_PROCESS_START', '.coveragerc' )
+
       ycmd_args = [
         sys.executable,
         PATH_TO_YCMD,
@@ -105,7 +110,8 @@ class Client_test( object ):
         _popen_handle = SafePopen( ycmd_args,
                                    stdin_windows = subprocess.PIPE,
                                    stdout = stdout,
-                                   stderr = subprocess.STDOUT )
+                                   stderr = subprocess.STDOUT,
+                                   env = env )
         self.server = psutil.Process( _popen_handle.pid )
 
 
--- a/ycmd/tests/client_test.py
+++ b/ycmd/tests/client_test.py
@@ -1,0 +1,231 @@
+# Copyright (C) 2016 ycmd contributors
+#
+# This file is part of ycmd.
+#
+# ycmd is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ycmd is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from future import standard_library
+standard_library.install_aliases()
+from builtins import *  # noqa
+from future.utils import native
+
+from base64 import b64decode, b64encode
+from hamcrest import assert_that, equal_to, has_length, is_in
+from tempfile import NamedTemporaryFile
+import functools
+import json
+import os
+import psutil
+import re
+import requests
+import subprocess
+import sys
+import time
+import urllib.parse
+
+from ycmd.hmac_utils import CreateHmac, CreateRequestHmac, SecureBytesEqual
+from ycmd.tests.test_utils import BuildRequest
+from ycmd.user_options_store import DefaultOptions
+from ycmd.utils import ( GetUnusedLocalhostPort, OpenForStdHandle,
+                         PathToCreatedTempDir, ReadFile, RemoveIfExists,
+                         SafePopen, ToBytes, ToUnicode )
+
+HEADERS = { 'content-type': 'application/json' }
+HMAC_HEADER = 'x-ycm-hmac'
+HMAC_SECRET_LENGTH = 16
+DIR_OF_THIS_SCRIPT = os.path.dirname( os.path.abspath( __file__ ) )
+PATH_TO_YCMD = os.path.join( DIR_OF_THIS_SCRIPT, '..' )
+
+
+class Client_test( object ):
+
+  def __init__( self ):
+    self._location = None
+    self._port = None
+    self._hmac_secret = None
+    self._stdout = None
+    self.server = None
+    self.subservers = []
+    self._options_dict = DefaultOptions()
+
+
+  def setUp( self ):
+    self._hmac_secret = os.urandom( HMAC_SECRET_LENGTH )
+    self._options_dict[ 'hmac_secret' ] = ToUnicode(
+      b64encode( self._hmac_secret ) )
+
+
+  def tearDown( self ):
+    if self.server.is_running():
+      self.server.terminate()
+    if self._stdout:
+      RemoveIfExists( self._stdout )
+    if self.subservers:
+      for subserver in self.subservers:
+        if subserver.is_running():
+          subserver.terminate()
+
+
+  def Start( self, idle_suicide_seconds = 60,
+             check_interval_seconds = 60 * 10 ):
+    # The temp options file is deleted by ycmd during startup
+    with NamedTemporaryFile( mode = 'w+', delete = False ) as options_file:
+      json.dump( self._options_dict, options_file )
+      options_file.flush()
+      self._port = GetUnusedLocalhostPort()
+      self._location = 'http://127.0.0.1:' + str( self._port )
+
+      ycmd_args = [
+        sys.executable,
+        PATH_TO_YCMD,
+        '--port={0}'.format( self._port ),
+        '--options_file={0}'.format( options_file.name ),
+        '--log=debug',
+        '--idle_suicide_seconds={0}'.format( idle_suicide_seconds ),
+        '--check_interval_seconds={0}'.format( check_interval_seconds ),
+      ]
+
+      self._stdout = os.path.join( PathToCreatedTempDir(), 'test.log' )
+      with OpenForStdHandle( self._stdout ) as stdout:
+        _popen_handle = SafePopen( ycmd_args,
+                                   stdin_windows = subprocess.PIPE,
+                                   stdout = stdout,
+                                   stderr = subprocess.STDOUT )
+        self.server = psutil.Process( _popen_handle.pid )
+
+
+  def _IsReady( self, filetype = None ):
+    params = { 'subserver': filetype } if filetype else None
+    response = self.GetRequest( 'ready', params )
+    response.raise_for_status()
+    return response.json()
+
+
+  def WaitUntilReady( self, filetype = None, timeout = 5 ):
+    total_slept = 0
+    while True:
+      try:
+        if total_slept > timeout:
+          raise RuntimeError( 'Waited for the server to be ready '
+                              'for {0} seconds, aborting.'.format(
+                                timeout ) )
+
+        if self._IsReady( filetype ):
+          return
+      except requests.exceptions.ConnectionError:
+        pass
+      finally:
+        time.sleep( 0.1 )
+        total_slept += 0.1
+
+
+  def StartSubserverForFiletype( self, filetype ):
+    # This automatically starts the subserver if not already started.
+    self.WaitUntilReady( filetype )
+
+    response = self.PostRequest(
+      'debug_info',
+      BuildRequest( filetype = filetype )
+    )
+    pid_match = re.search( 'process ID: (\d+)', response.json() )
+    if not pid_match:
+      raise RuntimeError( 'Cannot find PID in debug informations for {0} '
+                          'filetype.'.format( filetype ) )
+    subserver_pid = int( pid_match.group( 1 ) )
+    self.subservers.append( psutil.Process( subserver_pid ) )
+
+
+  def AssertServerAndSubserversAreRunning( self ):
+    for server in [ self.server ] + self.subservers:
+      assert_that( server.is_running(), equal_to( True ) )
+
+
+  def AssertServerAndSubserversShutDown( self, timeout = 5 ):
+    _, alive = psutil.wait_procs( [ self.server ] + self.subservers,
+                                  timeout = timeout )
+    assert_that( alive, has_length( equal_to( 0 ) ) )
+
+
+  def GetRequest( self, handler, params = None ):
+    return self._Request( 'GET', handler, params = params )
+
+
+  def PostRequest( self, handler, data = None ):
+    return self._Request( 'POST', handler, data = data )
+
+
+  def _ToUtf8Json( self, data ):
+    return ToBytes( json.dumps( data ) if data else None )
+
+
+  def _Request( self, method, handler, data = None, params = None ):
+    request_uri = self._BuildUri( handler )
+    data = self._ToUtf8Json( data )
+    headers = self._ExtraHeaders( method,
+                                  request_uri,
+                                  data )
+    response = requests.request( method,
+                                 request_uri,
+                                 headers = headers,
+                                 data = data,
+                                 params = params )
+    return response
+
+
+  def _BuildUri( self, handler ):
+    return native( ToBytes( urllib.parse.urljoin( self._location, handler ) ) )
+
+
+  def _ExtraHeaders( self, method, request_uri, request_body = None ):
+    if not request_body:
+      request_body = bytes( b'' )
+    headers = dict( HEADERS )
+    headers[ HMAC_HEADER ] = b64encode(
+        CreateRequestHmac( ToBytes( method ),
+                           ToBytes( urllib.parse.urlparse( request_uri ).path ),
+                           request_body,
+                           self._hmac_secret ) )
+    return headers
+
+
+  def AssertResponse( self, response ):
+    assert_that( response.status_code, equal_to( requests.codes.ok ) )
+    assert_that( HMAC_HEADER, is_in( response.headers ) )
+    assert_that(
+      self._ContentHmacValid( response ),
+      equal_to( True )
+    )
+
+
+  def _ContentHmacValid( self, response ):
+    our_hmac = CreateHmac( response.content, self._hmac_secret )
+    their_hmac = ToBytes( b64decode( response.headers[ HMAC_HEADER ] ) )
+    return SecureBytesEqual( our_hmac, their_hmac )
+
+
+  @staticmethod
+  def CaptureOutputFromServer( test ):
+    @functools.wraps( test )
+    def Wrapper( self, *args ):
+      try:
+        test( self, *args )
+      finally:
+        if self._stdout:
+          sys.stdout.write( ReadFile( self._stdout ) )
+
+    return Wrapper
--- a/ycmd/tests/shutdown_test.py
+++ b/ycmd/tests/shutdown_test.py
@@ -1,0 +1,92 @@
+# Copyright (C) 2016 ycmd contributors
+#
+# This file is part of ycmd.
+#
+# ycmd is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ycmd is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from future import standard_library
+standard_library.install_aliases()
+from builtins import *  # noqa
+
+from hamcrest import assert_that, equal_to
+
+from ycmd.tests.client_test import Client_test
+
+
+class Shutdown_test( Client_test ):
+
+  @Client_test.CaptureOutputFromServer
+  def FromHandlerWithoutSubserver_test( self ):
+    self.Start()
+    self.WaitUntilReady()
+    self.AssertServerAndSubserversAreRunning()
+
+    response = self.PostRequest( 'shutdown' )
+    self.AssertResponse( response )
+    assert_that( response.json(), equal_to( True ) )
+    self.AssertServerAndSubserversShutDown( timeout = 5 )
+
+
+  @Client_test.CaptureOutputFromServer
+  def FromHandlerWithSubservers_test( self ):
+    self.Start()
+    self.WaitUntilReady()
+
+    # C# subserver cannot be started without a solution file so we don't test
+    # it.
+    filetypes = [ 'go',
+                  'javascript',
+                  'python',
+                  'typescript',
+                  'rust' ]
+    for filetype in filetypes:
+      self.StartSubserverForFiletype( filetype )
+    self.AssertServerAndSubserversAreRunning()
+
+    response = self.PostRequest( 'shutdown' )
+    self.AssertResponse( response )
+    assert_that( response.json(), equal_to( True ) )
+    self.AssertServerAndSubserversShutDown( timeout = 5 )
+
+
+  @Client_test.CaptureOutputFromServer
+  def FromWatchdogWithoutSubserver_test( self ):
+    self.Start( idle_suicide_seconds = 2, check_interval_seconds = 1 )
+    self.WaitUntilReady()
+    self.AssertServerAndSubserversAreRunning()
+
+    self.AssertServerAndSubserversShutDown( timeout = 5 )
+
+
+  @Client_test.CaptureOutputFromServer
+  def FromWatchdogWithSubservers_test( self ):
+    self.Start( idle_suicide_seconds = 2, check_interval_seconds = 1 )
+    self.WaitUntilReady()
+
+    # C# subserver cannot be started without a solution file so we don't test
+    # it.
+    filetypes = [ 'go',
+                  'javascript',
+                  'python',
+                  'typescript',
+                  'rust' ]
+    for filetype in filetypes:
+      self.StartSubserverForFiletype( filetype )
+    self.AssertServerAndSubserversAreRunning()
+
+    self.AssertServerAndSubserversShutDown( timeout = 5 )
--- a/ycmd/watchdog_plugin.py
+++ b/ycmd/watchdog_plugin.py
@@ -47,7 +47,7 @@ class WatchdogPlugin( object ):
 
   def __init__( self,
                 idle_suicide_seconds,
-                check_interval_seconds = 60 * 10 ):
+                check_interval_seconds ):
     self._check_interval_seconds = check_interval_seconds
     self._idle_suicide_seconds = idle_suicide_seconds
 
--- a/ycmd/watchdog_plugin.py
+++ b/ycmd/watchdog_plugin.py
@@ -24,10 +24,12 @@ standard_library.install_aliases()
 from builtins import *  # noqa
 
 import time
-import os
 import copy
-from ycmd import utils
+import logging
 from threading import Thread, Lock
+from ycmd.handlers import ServerShutdown
+
+_logger = logging.getLogger( __name__ )
 
 
 # This class implements the Bottle plugin API:
@@ -95,7 +97,8 @@ class WatchdogPlugin( object ):
       # wait interval to contact us before we die.
       if (self._TimeSinceLastRequest() > self._idle_suicide_seconds and
           self._TimeSinceLastWakeup() < 2 * self._check_interval_seconds):
-        utils.TerminateProcess( os.getpid() )
+        _logger.info( 'Shutting down server due to inactivity' )
+        ServerShutdown()
 
       self._UpdateLastWakeupTime()
 
--- a/ycmd/wsgi_server.py
+++ b/ycmd/wsgi_server.py
@@ -1,0 +1,60 @@
+# Copyright (C) 2016 ycmd contributors
+#
+# This file is part of ycmd.
+#
+# ycmd is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ycmd is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from future import standard_library
+standard_library.install_aliases()
+from builtins import *  # noqa
+
+from future.utils import listvalues
+from waitress.server import TcpWSGIServer
+import select
+
+
+class StoppableWSGIServer( TcpWSGIServer ):
+  """StoppableWSGIServer is a subclass of the TcpWSGIServer Waitress server
+  with a shutdown method. It is based on StopableWSGIServer class from webtest:
+  https://github.com/Pylons/webtest/blob/master/webtest/http.py"""
+
+  shutdown_requested = False
+
+  def Run( self ):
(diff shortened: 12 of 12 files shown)


Valloric:
This is important work; thanks for taking it on! I didn't know we left sub-servers running after ycmd exits on Windows. That's a pretty bad bug.

vheon:
This PR is really awesome and needed! :+1: Is also pretty straightforward to understand since @micbou explained well what was needed. Anyhow I left a comment on a file since I don't understand why the  sleep is needed, and if is needed how do we know that `0.1` is enough? I generally don't like `sleeps`. The only file that is left for me to review is the `server_test`; the reason is that I see in that file lots of function which are similar or used also in the `example_client` or in `ycmd` itself, such the `HMAC` signing and checking, the passing of the options through a temporary file, some `utf8` stuff like `_RecursiveEncodeUnicodeToUtf8` and some other things. So to the question: should we refactor this right now? Or should we do a second pass to do more analysis and refactor later?

---

Reviewed 9 of 10 files at r1.
Review status: 9 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[ycmd/handlers.py, line 269 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K68okxsve6A-L8PBaEn:-K68okxsve6A-L8PBaEo:-1185340488) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L269)):*
Why is this sleep needed?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-K68pHGRxs9qvf6xzfVN:-192503108)*

micbou:
AppVeyor failures are not related to this PR. See PR #283.

> So to the question: should we refactor this right now? Or should we do a second pass to do more analysis and refactor later?

I think this is better to refactor now than later. I already did some refactoring by reusing functions from the `hmac_utils` module. I just found out that `_ToUtf8Json` and `_RecursiveEncodeUnicodeToUtf8` are already defined in the `utils` module so I will use them. Tell me if there are other functions.

---

Review status: 9 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[ycmd/handlers.py, line 269 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K68okxsve6A-L8PBaEn:-K69O5Rd4RGv7NAxaZJX:962476888) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L269)):*
Good question. It is needed to let the time for the server to send the response before being killed by the thread. I should add a comment to explain this.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-K69V8uCRGdRF0noiA0H:688743883)*

vheon:
Before this will land into master we should ping @abingham (emacs-ycmd), @Qusic (atom-youcompleteme), @mawww (kak-ycmd) since this will change how to shutdown the ycmd server.

---

Review status: 9 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-K69_VN5iGQOzfkNuoPA:589664)*

puremourning:
Another thing to consider refactoring is the class which performs full integration tests. A large part of the class in `server_test.py` is just boilerplate for setting up and tearing down the server. Maybe this could be a base to be implemented by something actually containing the `_test()` methods. That sort of separates infrastructure from actual tests.

Would potentially also allow more separation, such as testing Tern completer's "shutdown" method within the `tests/javascript` directory and other `Shutdown` methods for the respective completers.

That might also give us a canonical minimal ycmd client, which could be useful or educational in its own right.

---

Reviewed 10 of 10 files at r1.

---

*[examples/example_client.py, line 110 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69TisWswDShWZeBY1S:-K69TisXoq5MoSzlC-3E:688982945) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/examples/example_client.py#L110)):*
I know this is only an example, but we should probably clean up the server PID handle here. We're expecting the postcondition of this method to be that the server is no longer running. Is there a way to block until the pid exits? Like `thread.join()` or equivalent? Maybe with a more forcefull kill if it times out? May be scope creep for this example client, but perhaps at least 

```python
if self.IsAlive():
  # the server didn't terminate for some reason. kill it.
  self._popen_handle.terminate()
```


it looks like `psutil` has a method for waiting: http://pythonhosted.org/psutil/#psutil.Process.wait
or maybe http://pythonhosted.org/psutil/#psutil.wait_procs

---

*[travis/travis_install.sh, line 39 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69W5jSTB5POBb2m-Se:-K69W5jSTB5POBb2m-Sf:-600368013) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/travis/travis_install.sh#L39)):*
can we break this into 2 lines?

```bash
echo -e "..." > \
  ${YCMD_VENV_DIR}/....
```

I think that works...

---

*[ycmd/__main__.py, line 94 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69XzMxjKboMwV8Kike:-K69XzMxjKboMwV8Kikf:479969223) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/__main__.py#L94)):*
The difference between `--idle_suicide_seconds` and `--check_interval_seconds` seems quite subtle. The help and comments don't really make it obvious which one I care about in which situation.

---

*[ycmd/__main__.py, line 155 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69WNl7CNQYH85Yxghn:-K69WNl8MVdEnQhKPcai:-535936828) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/__main__.py#L155)):*
"These" is the plural form of "This", "Those" is the plural form of "That". Pedantry, I know :)

---

*[ycmd/handlers.py, line 219 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69cq1u7StnjkFzg985:-K69cq1u7StnjkFzg986:2056136094) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L219)):*
We should add a comment here that while this method is a handler for `/shutdown` it is also called by the watchdog plugin which will not have a response context to write to (nor a request to read from). Either that, or we should make it explicit by having this "handler" call a non-handler method. This might be better in fact - it feels... _wrong_ somehow (to me) to call a request handler from anywhere but bottle.

---

*[ycmd/handlers.py, line 269 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K68okxsve6A-L8PBaEn:-K69Ybt06DRoc-027iCJ:769081283) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L269)):*
Presumably, the pause is so that the `/shutdown` command has time to return HTTP 200 to the client before actually shutting down the server. I'm not a fan of this arbitrary delay. As there is no synchronisation, it is still possible for this tread to execute prior to the response being sent.

I feel we should either try to synchronise it (I'm not sure how), or just exit synchronously and state in the API that no response is received to a shutdown request and the communication session is terminated (TCP reset). I'm guessing that you probably tried this and had some sort of problem with the client and/or tests?

---

*[ycmd/tests/server_test.py, line 180 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69eKLP2rZY7rq5S2pH:-K69eKLQtzqYNdVsjMYZ:2127694397) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/tests/server_test.py#L180)):*
I think this proc scans every process in the system. This might be quite slow.

Could we use `psutil`'s `Process` class? It seems to have a [`children`](http://pythonhosted.org/psutil/#psutil.Process.children) method:

> children(recursive=False)[source]
> Return the children of this process as a list of Process objects, pre-emptively checking whether PID has been reused. If recursive is True return all the parent descendants. Example assuming A == this process:

---

*[ycmd/watchdog_plugin.py, line 45 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69cYQivlXeWJFWNeh7:-K69cYQjNNyJQ_TDm-Po:-1648292150) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/watchdog_plugin.py#L45)):*
Hmm. I see both of these options existed before. I still can't claim to understand what they do, or how we use them in the test. Probably just me being dumb.

Do we document the ycmd options anywhere else, like in the README.md ? 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-K69dNYj9KwG1d60MNQE:-2032099476)*

vheon:
Review status: 5 of 10 files reviewed at latest revision, 8 unresolved discussions.

---

*[ycmd/watchdog_plugin.py, line 45 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69cYQivlXeWJFWNeh7:-K6ARxSpZcHiOi6Wee2m:989451010) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/watchdog_plugin.py#L45)):*
The Watchdog plugin check every `check_interval_seconds` how much time is passed since the last request the `ycmd` server have served. If this time is over the `idle_suicide_seconds` then the Watchdog plugin will kill the `ycmd` server.

The description is in the watchdog plugin https://github.com/Valloric/ycmd/blob/8d6cf88518a4cbd90d242abbf13d25b39fee1552/ycmd/watchdog_plugin.py#L29

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282)*

puremourning:
Reviewed 5 of 5 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282)*

micbou:

---

*[examples/example_client.py, line 110 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69TisWswDShWZeBY1S:-K6AS6YQgyCBPkktiF2k:1421596888) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/examples/example_client.py#L110)):*
We can use `self._popen_handle.wait()` but there is no timeout so it may block the script indefinitely.

---

*[travis/travis_install.sh, line 39 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69W5jSTB5POBb2m-Se:-K6AS_GkFfA9Fv-5ppQw:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/travis/travis_install.sh#L39)):*
Done.

---

*[ycmd/__main__.py, line 155 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69WNl7CNQYH85Yxghn:-K6AT1CzH_EJ4ssUG_ee:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/__main__.py#L155)):*
Done.

---

*[ycmd/handlers.py, line 219 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69cq1u7StnjkFzg985:-K6AT2QVWpCYirKhkxsR:1213188744) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L219)):*
I chose the "handler calls a non-handler" path. I added specific logging for the watchdog shutdown to differentiate it from the shutdown one.

---

*[ycmd/handlers.py, line 269 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K68okxsve6A-L8PBaEn:-K6ATiNpF3FftmFO5HYH:568826655) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L269)):*
Using [Bottle `after_request` hook](http://bottlepy.org/docs/dev/api.html#bottle.Bottle.add_hook) could be the solution. I'll test it.

---

*[ycmd/tests/server_test.py, line 180 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69eKLP2rZY7rq5S2pH:-K6AUPUNVRbEzwF1BGud:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/tests/server_test.py#L180)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282)*

micbou:
*[ycmd/handlers.py, line 219 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69cq1u7StnjkFzg985:-K6AVTh6K_dYLEQz60Qa:1457877896) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L219)):*
I mean the "handler one".

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282)*

vheon:
Reviewed 4 of 5 files at r2.

---

*[ycmd/handlers.py, line 269 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K68okxsve6A-L8PBaEn:-K6AWEXpsg3lmMLg_71K:1794731161) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L269)):*
Great!

---

*[ycmd/tests/server_test.py, line 93 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K6AW_xTHnpdjjMRKTHg:-K6AW_xTHnpdjjMRKTHh:-977955353) ([raw file](https://github.com/valloric/ycmd/blob/a4aaf80245a60f0230e4c3363f01e1cf52c66b04/ycmd/tests/server_test.py#L93)):*
can we wrap this in a function like `StartCompleterServer`?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282)*

vheon:

---

*[ycmd/__main__.py, line 94 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69XzMxjKboMwV8Kike:-K6AXOYbP7j7kUAIK_x1:-1628876016) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/__main__.py#L94)):*
Which "I" are we talking about? Are we talking about "I" the user of `ycmd`, the "I" user of YCM or "I" developing `ycmd`??

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282)*

puremourning:

---

*[ycmd/__main__.py, line 94 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K69XzMxjKboMwV8Kike:-K6AhDP3gxcyny82by3A:-1452754292) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/__main__.py#L94)):*
Any/all of the above (except, perhaps the user of YCM). It's fine - I think your explanation below is clear.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282)*

mawww:
Does the SIGTERM signal still works correctly ? or do we need to go through the shutdown command now ?

kak-ycmd is not expected to run directly on windows, but always on cygwin, so I assume SIGTERM will still work there, which means its it not urgent to tweak kak-ycmd to support the shutdown command if SIGTERM is still a supported way to close ycmd.

abingham:
Thanks for looping me in on this. It sounds like the only substantial change from the client point of view is that we'll call `/shutdown` rather than killing the process. If that's the case, this should be a very simple transition for `emacs-ycmd`. (And if not, please enlighten me!)

micbou:
> Another thing to consider refactoring is the class which performs full integration tests. A large part of the class in server_test.py is just boilerplate for setting up and tearing down the server. Maybe this could be a base to be implemented by something actually containing the _test() methods. That sort of separates infrastructure from actual tests.

I moved the infrastructure to a new class `Client_test`. Tests are now in `Shutdown_test`, subclass of `Client_test`.

> Does the SIGTERM signal still works correctly ? or do we need to go through the shutdown command now ?

In PR current state, you can still use the SIGTERM signal to stop the server and it should stay that way. Our goal is to provide a portable (and therefore recommended) way to shutdown the server.

---

Review status: 9 of 11 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/handlers.py, line 269 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K68okxsve6A-L8PBaEn:-K6FRxqgyHlaI0t4M_db:213340911) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L269)):*
Unfortunately, this doesn't work. So, I see three ways to move forward:
 - keep it that way: kill the server from a thread started in the handler function after a fixed delay. Advantages: easy to implement and a 200 response should be returned. Inconvenient: client cannot completely rely on the 200 response because server may be killed before sending the response;
 - directly kill the server in the handler function. Advantages: easy to implement. Inconvenient: no response (connection aborted);
 - implement a subclass of  `TcpWSGIServer` from waitress and add a `shutdown` method like the one in [`StopableWSGIServer` from webtest](https://github.com/Pylons/webtest/blob/master/webtest/http.py#L45). Advantages: a 200 response should be returned (need to implement it to be sure). Inconvenient: not straighforward to implement, depend on waitress internals.
 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-K6FQ2WcFB3iu00xaf_L:-709705176)*

puremourning:
Reviewed 3 of 3 files at r4.

---

*[ycmd/handlers.py, line 269 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K68okxsve6A-L8PBaEn:-K6y4ByNyB1yMcZalmBW:342608997) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L269)):*
My gut tells me the best thing to do is the second option, as it is the safest and simplest. If it is a pain for clients, how much of a pain, taking, say YCM client as example?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282)*

Valloric:

---

*[ycmd/handlers.py, line 269 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K68okxsve6A-L8PBaEn:-K6yTBQT5Jc0MvdAeiQk:-116836390) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L269)):*
Agreed with @puremourning, second option seems the most sensible. I think it's perfectly fine for a `/shutdown` handler to not return anything. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282)*

micbou:

---

*[ycmd/handlers.py, line 269 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K68okxsve6A-L8PBaEn:-K6yaYVNOx68VuBwRFUV:-2143805287) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L269)):*
I tried third approach and it looks good. It does the job better than other solutions by reaching the end of the program and not  killing the process. In particular, `@atexit` function is called on Windows. It is also relatively simple. I'll push the changes this week-end and we'll discuss them.

> If it is a pain for clients, how much of a pain, taking, say YCM client as example?

Just a matter of a `try-except` block. Even if the server returns a 200 response, such block is needed for timeouts.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282)*

vheon:

---

*[ycmd/handlers.py, line 269 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K68okxsve6A-L8PBaEn:-K6zRj4zDeE7rFPa2RJW:-1103892913) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L269)):*
I'm actually worried about the cons part of the third approach

> Inconvenient: not straighforward to implement, depend on waitress internals.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282)*

Valloric:

---

*[ycmd/handlers.py, line 269 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K68okxsve6A-L8PBaEn:-K704IuMewY5APslt2Hd:1710540231) ([raw file](https://github.com/valloric/ycmd/blob/827cf7978406eff77b45ac9676622f0c407d114c/ycmd/handlers.py#L269)):*
Depending on waitress internals is a deal breaker though; we only want to rely on the WSGI interface so that we can swap out waitress for something else if necessary. This isn't merely a theoretical concern; when I was writing ycmd, initially it used a different server (can't remember which) and I switched it to waitress for performance reasons days before it went live. The switch was something like 3 lines of code and that's it.

If we were ever decide to lock ourselves into a particular server framework, it's going to be for a _very_ good reason. This case doesn't strike me as that. Not returning from a `/shutdown` handler isn't pretty, but it's not a big deal.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/282)*

Valloric:
Holy crap, if the coverage bot is correct in that this PR bumps our test coverage from 85% to 90%, _that's goddamn fantastic!_ :D

---

Review status: 2 of 13 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-KEiNSWFDlmEWo0YpMcW:566864321)*

Valloric:

micbou:
In this PR, I want to test that all the subservers are properly shutdown when ycmd is terminated. However, I can't do it properly because:
 - there are different ways to start the subservers: `Go` and `JavaScript` have the `StartServer` subcommand, `Python` and `Rust` got the `RestartServer` subcommand, `C♯` has both subcommands, and `TypeScript` has none;
 - as a client, there is no way to get the PID of the subservers so we can't reliably check if the subservers are down;
 - due to how we interact with the `Go` subserver, we have no control over its process (its handle is not available) and thus we can't get its PID;

I am planning to send PRs to fix each of these points:
- define the `RestartServer` and `StopServer` subcommands for all completers with a server; remove the `StartServer` subcommand;
- display the PID in the debug informations for all completers with a server;
- run the `gocode` executable in server mode.

I'll probably start with the last point.

Valloric:
@micbou Thanks for the update. All the problems you've identified sound like they're worth fixing. At the end, might be worth putting together a doc to list all the methods/behavior/functionality we expect from a semantic completer class so new ones don't forget to add the necessary pieces.

Valloric:
I keep drooling at that coverage bump... :)

vheon:
> I keep drooling at that coverage bump... :)

@Valloric I think it will be even higher since to enable this correctly @micbou said we should refactor the `DebugInfo` to return a json instead of a string and I think we will add more test to cover that :P

micbou:
> @Valloric I think it will be even higher since to enable this correctly @micbou said we should refactor the DebugInfo to return a json instead of a string and I think we will add more test to cover that :P

Returning the debug information as a dictionary will be done (if the idea is approved) after this PR is merged because I don't want to delay it further.

vheon:
> Returning the debug information as a dictionary will be done (if the idea is approved) after this PR is merged because I don't want to delay it further.
 
It was my understanding that displaying the PID of the subservers was required for the tests of this PR. If the DebugInfo to dictionary refactoring is delayed after this PR how are we going to test it? Are we going to parse the DebugInfo string?

micbou:
> Are we going to parse the DebugInfo string?

Yes, that's the plan. You can already see how it's done in the `StartSubserverForFiletype` function.

vheon:
> Yes, that's the plan. You can already see how it's done in the StartSubserverForFiletype function.

Well good that I asked because it is not uniform right now as a string through the completers 😒

micbou:
> Well good that I asked because it is not uniform right now as a string through the completers

I have a PR in preparation for this.

micbou:
Thanks to PRs #505, #506, #533, we are now testing the shutdown of all subservers (except the C# one because it requires a solution file for starting it).

Currently, this PR implements the following approach to shut down the server: it extends the `TcpWSGIServer` class from waitress by adding a `Shutdown` method. This method is called in the new `shutdown` handler through a separate thread to let the server send the response before being shut down. I know there are concerns about relying on waitress internals but all of this is encapsulated in the `StoppableWSGIServer` class so if one day we decide to replace waitress with something else, we only need to update this class. Another concern is the race condition between sending a 200 response and stopping the server. In practice, it never happens as shown by the tests.

I'll send a PR to YCM repository that updates the client to use the new `/shutdown` handler so that you can test it yourself.

I am changing the PR status to RFC.

---

Reviewed 2 of 11 files at r1, 1 of 3 files at r4, 3 of 12 files at r5, 3 of 5 files at r7, 1 of 2 files at r8, 1 of 2 files at r9, 1 of 1 files at r10, 2 of 2 files at r12, 3 of 3 files at r13.

---

*[ycmd/tests/client_test.py, line 93 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-K6AW_xTHnpdjjMRKTHg:-KMebDiO0VIVeOr3NLQd:1632234423) ([raw file](https://github.com/valloric/ycmd/blob/a4aaf80245a60f0230e4c3363f01e1cf52c66b04/ycmd/tests/client_test.py#L93)):*
> can we wrap this in a function like `StartCompleterServer`?
</details>
See the `StartSubserverForFiletype` function.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-KMebQFLC0PiHuEXF6mH:263672256)*

Valloric:
Review status: 12 of 13 files reviewed at latest revision, 6 unresolved discussions, some commit checks broke.

---

*[ycmd/handlers.py, line 273 \[r14\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMjij63OirI0Eb7iMsj:-KMjij63OirI0Eb7iMsk:-1722268997) ([raw file](https://github.com/valloric/ycmd/blob/3d9b124b71ec1faf73c5c0642e346395a7bddbca/ycmd/handlers.py#L273)):*
> ```Python
>       _wsgi_server.Shutdown()
> 
>   terminator = Thread( target = Terminator )
> ```

Probably needs a comment explaining why the separate thread is necessary.

---

*[ycmd/wsgi_server.py, line 31 \[r14\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMjj-b5_WYHTbHJ133G:-KMjj-b5_WYHTbHJ133H:-83189233) ([raw file](https://github.com/valloric/ycmd/blob/3d9b124b71ec1faf73c5c0642e346395a7bddbca/ycmd/wsgi_server.py#L31)):*
> ```Python
> 
> 
> class StoppableWSGIServer( TcpWSGIServer ):
> ```

This really isn't a lot of code. As such, I'm fine with depending on waitress internals here.

Good job!

---

*[ycmd/wsgi_server.py, line 54 \[r14\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMjkQfhLzg7hehV5F8g:-KMjkQfhLzg7hehV5F8h:-218432045) ([raw file](https://github.com/valloric/ycmd/blob/3d9b124b71ec1faf73c5c0642e346395a7bddbca/ycmd/wsgi_server.py#L54)):*
> ```Python
>     self.logger.setLevel( logging.FATAL )
>     self.task_dispatcher.shutdown()
>     while self._map:
> ```

Isn't this accessing private state from the super class? Won't this blow up in our face when we update waitress and they change the class internals?

---

*[ycmd/wsgi_server.py, line 55 \[r14\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMjkAuLp2AXYe-MOkIV:-KMjkAuM3FMXICs_HhSG:45216878) ([raw file](https://github.com/valloric/ycmd/blob/3d9b124b71ec1faf73c5c0642e346395a7bddbca/ycmd/wsgi_server.py#L55)):*
> ```Python
>     self.task_dispatcher.shutdown()
>     while self._map:
>         channels = list( self._map.values() )
> ```

Why the need for a intermediate list object? Why not just loop over `self._map.values()`?

If there's a specific reason, it probably needs a comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282)*

micbou:
I looked more thoroughly at the `waitress` and `asyncore` modules and was able to simplify the `Shutdown` method of `StoppableWSGIServer`. I also added comments to explain what's happening.

---

Reviewed 2 of 2 files at r15.

---

*[ycmd/handlers.py, line 273 \[r14\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMjij63OirI0Eb7iMsj:-KMo3unwRRreX5Lt3VFy:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/3d9b124b71ec1faf73c5c0642e346395a7bddbca/ycmd/handlers.py#L273)):*
> Probably needs a comment explaining why the separate thread is necessary.
</details>
Done.

---

*[ycmd/wsgi_server.py, line 31 \[r14\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMjj-b5_WYHTbHJ133G:-KMo7bKgkxF-sUFZZjju:-298842179) ([raw file](https://github.com/valloric/ycmd/blob/3d9b124b71ec1faf73c5c0642e346395a7bddbca/ycmd/wsgi_server.py#L31)):*
> This really isn't a lot of code. As such, I'm fine with depending on waitress internals here.
> 
> Good job!
</details>
All thanks to webtest (although they made a typo in their class name).

---

*[ycmd/wsgi_server.py, line 54 \[r14\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMjkQfhLzg7hehV5F8g:-KMo42WtxdoCiN82BjOq:1287668155) ([raw file](https://github.com/valloric/ycmd/blob/3d9b124b71ec1faf73c5c0642e346395a7bddbca/ycmd/wsgi_server.py#L54)):*
> Isn't this accessing private state from the super class? Won't this blow up in our face when we update waitress and they change the class internals?
</details>
Unfortunately, I could not find any "public" method to retrieve the `_map` attribute. This attribute is defined in the `dispatcher` class from `asyncore`, which is a base class of `TcpWSGIServer`. I doubt they will change this in waitress anytime soon but I may be wrong so we should be careful here. Do you want me to add a comment about this?

---

*[ycmd/wsgi_server.py, line 55 \[r14\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMjkAuLp2AXYe-MOkIV:-KMo4_gLqQDnymA7GWa1:351281104) ([raw file](https://github.com/valloric/ycmd/blob/3d9b124b71ec1faf73c5c0642e346395a7bddbca/ycmd/wsgi_server.py#L55)):*
> Why the need for a intermediate list object? Why not just loop over `self._map.values()`?
> 
> If there's a specific reason, it probably needs a comment.
</details>
Because the `_map` attribute is modified in the loop. I added a comment to clarify this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-KMo7vjgFnxsqyDfLq9E:-1166788176)*

puremourning:
:lgtm: 

Just one thing to check on a possible race condition.

---

Reviewed 1 of 3 files at r4, 2 of 12 files at r5, 3 of 5 files at r7, 1 of 2 files at r8, 1 of 2 files at r12, 2 of 3 files at r13, 2 of 2 files at r15.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[ycmd/__main__.py, line 179 \[r15\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMoQhOwCYsJgcglQdUt:-KMoQhOwCYsJgcglQdUu:-1928152882) ([raw file](https://github.com/valloric/ycmd/blob/753f7d2f6d191899e8243c498ab972a9c046337d/ycmd/__main__.py#L179)):*
> ```Python
>   handlers.app.install( HmacPlugin( hmac_secret ) )
>   CloseStdin()
>   handlers._wsgi_server = StoppableWSGIServer( handlers.app,
> ```

Isn't it a little naughty to directly interact with `handlers._wsgi_server`? Should we instead wrap this, or call it `handlers.WSGI_SERVER` or something?

---

*[ycmd/handlers.py, line 277 \[r15\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMoROv14XWhGS6DwI0_:-KMoROv14XWhGS6DwI0a:882373443) ([raw file](https://github.com/valloric/ycmd/blob/753f7d2f6d191899e8243c498ab972a9c046337d/ycmd/handlers.py#L277)):*
> ```Python
>   terminator = Thread( target = Terminator )
>   terminator.daemon = True
>   terminator.start()
> ```

I don't know Python's threading model that well, but is there actually any guarantee that the `Shutdown` method actually executes _after_ we send the response? I don't see any synchronisation around it, so I think it is still possible to for us to suicide before the response is actually despatched.

Maybe a 1s delay would be enough (or something), though that's obviously imperfect. TBH i don't really mind as long at the (potential/theoretical) connection-aborted message doesn't cause Vim to other clients to blow up. Most important thing is ycmd exiting and taking its toys with it.

---

*[ycmd/watchdog_plugin.py, line 100 \[r15\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMoS2eMXAqjaeAIxAQ_:-KMoS2eMXAqjaeAIxAQa:-128867111) ([raw file](https://github.com/valloric/ycmd/blob/753f7d2f6d191899e8243c498ab972a9c046337d/ycmd/watchdog_plugin.py#L100)):*
> ```Python
>       if (self._TimeSinceLastRequest() > self._idle_suicide_seconds and
>           self._TimeSinceLastWakeup() < 2 * self._check_interval_seconds):
>         _logger.info( 'Shutting down server for inactivity' )
> ```

for -> "due to" ? Should we also maybe say "due to <x> seconds of inactivity" ? shrug.

---

*[ycmd/wsgi_server.py, line 36 \[r15\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMoSZjhbATbIcpSM481:-KMoSZjhbATbIcpSM482:1826437304) ([raw file](https://github.com/valloric/ycmd/blob/753f7d2f6d191899e8243c498ab972a9c046337d/ycmd/wsgi_server.py#L36)):*
> ```Python
>   https://github.com/Pylons/webtest/blob/master/webtest/http.py"""
> 
>   is_shutdown = False
> ```

`is_shutdown` feels a little awkward. how about `shutdown_requested` ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-KMoT_qeQuqIGMvMBCQ3:-406871577)*

micbou:
Review status: 9 of 13 files reviewed at latest revision, 9 unresolved discussions.

---

*[ycmd/__main__.py, line 179 \[r15\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMoQhOwCYsJgcglQdUt:-KMou5reC_s5zg-cR938:641444917) ([raw file](https://github.com/valloric/ycmd/blob/753f7d2f6d191899e8243c498ab972a9c046337d/ycmd/__main__.py#L179)):*
> Isn't it a little naughty to directly interact with `handlers._wsgi_server`? Should we instead wrap this, or call it `handlers.WSGI_SERVER` or something?
</details>
I removed the leading underscore to the variable name to follow our public variables convention (and to be consistent with the `app` variable).

---

*[ycmd/handlers.py, line 277 \[r15\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMoROv14XWhGS6DwI0_:-KMovc2oImcwrnizn9eM:433975974) ([raw file](https://github.com/valloric/ycmd/blob/753f7d2f6d191899e8243c498ab972a9c046337d/ycmd/handlers.py#L277)):*
> I don't know Python's threading model that well, but is there actually any guarantee that the `Shutdown` method actually executes _after_ we send the response? I don't see any synchronisation around it, so I think it is still possible to for us to suicide before the response is actually despatched.
> 
> Maybe a 1s delay would be enough (or something), though that's obviously imperfect. TBH i don't really mind as long at the (potential/theoretical) connection-aborted message doesn't cause Vim to other clients to blow up. Most important thing is ycmd exiting and taking its toys with it.
</details>
It's not obvious from the `StoppableWSGIServer` class but there is [a timeout of 5s when shutting down the waitress threads](https://github.com/Pylons/waitress/blob/151cb1e33c08ad940b7d766ba383616f7e445d9e/waitress/task.py#L122) so we are giving 5 seconds to the server to send the response, which is way more than enough in practice.

In fact, I don't think there is a perfect solution: either we indefinitely wait for the response to be sent and we take the risk that the server is never shut down because the server is stuck sending the response or we only wait for a predefined time and there is a possibility that the server shut down before sending the response. I prefer the solution where the server is always shut down.

---

*[ycmd/watchdog_plugin.py, line 100 \[r15\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMoS2eMXAqjaeAIxAQ_:-KMoucGAQuXekMiBfCX_:-992112183) ([raw file](https://github.com/valloric/ycmd/blob/753f7d2f6d191899e8243c498ab972a9c046337d/ycmd/watchdog_plugin.py#L100)):*
> for -> "due to" ? Should we also maybe say "due to <x> seconds of inactivity" ? shrug.
</details>
I replaced `for` by `due to`. `due to seconds of inactivity` seems strange without mentioning the number of seconds.

---

*[ycmd/wsgi_server.py, line 36 \[r15\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMoSZjhbATbIcpSM481:-KMovXBOWPFXwt-eoYWM:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/753f7d2f6d191899e8243c498ab972a9c046337d/ycmd/wsgi_server.py#L36)):*
> `is_shutdown` feels a little awkward. how about `shutdown_requested` ?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282)*

Valloric:
:lgtm: I'm pretty happy with this.  :)

---

Reviewed 2 of 11 files at r1, 1 of 3 files at r4, 1 of 12 files at r5, 2 of 5 files at r7, 1 of 2 files at r8, 1 of 2 files at r12, 2 of 3 files at r13, 4 of 4 files at r16.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks broke.

---

*[ycmd/wsgi_server.py, line 54 \[r14\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMjkQfhLzg7hehV5F8g:-KMpUB83K3MBR660uwqR:-709513056) ([raw file](https://github.com/valloric/ycmd/blob/3d9b124b71ec1faf73c5c0642e346395a7bddbca/ycmd/wsgi_server.py#L54)):*
> Unfortunately, I could not find any "public" method to retrieve the `_map` attribute. This attribute is defined in the `dispatcher` class from `asyncore`, which is a base class of `TcpWSGIServer`. I doubt they will change this in waitress anytime soon but I may be wrong so we should be careful here. Do you want me to add a comment about this?
</details>
Yeah, let's at least add a comment here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-KMpUYr51-2WE1oYKBCI:1770495375)*

puremourning:
:lgtm: too :)

---

Reviewed 4 of 4 files at r16.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks broke.

---

*[ycmd/handlers.py, line 277 \[r15\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMoROv14XWhGS6DwI0_:-KMpvaRfb0ilR23-5K34:1691898992) ([raw file](https://github.com/valloric/ycmd/blob/753f7d2f6d191899e8243c498ab972a9c046337d/ycmd/handlers.py#L277)):*
> It's not obvious from the `StoppableWSGIServer` class but there is [a timeout of 5s when shutting down the waitress threads](https://github.com/Pylons/waitress/blob/151cb1e33c08ad940b7d766ba383616f7e445d9e/waitress/task.py#L122) so we are giving 5 seconds to the server to send the response, which is way more than enough in practice.
> 
> In fact, I don't think there is a perfect solution: either we indefinitely wait for the response to be sent and we take the risk that the server is never shut down because the server is stuck sending the response or we only wait for a predefined time and there is a possibility that the server shut down before sending the response. I prefer the solution where the server is always shut down.
</details>
Agree that the most important thing is to be sure we exit eventually.

---

*[ycmd/watchdog_plugin.py, line 100 \[r15\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMoS2eMXAqjaeAIxAQ_:-KMpvh5yyWS-_wSrn_eA:1168660855) ([raw file](https://github.com/valloric/ycmd/blob/753f7d2f6d191899e8243c498ab972a9c046337d/ycmd/watchdog_plugin.py#L100)):*
> I replaced `for` by `due to`. `due to seconds of inactivity` seems strange without mentioning the number of seconds.
</details>
Hah. I was suggesting "due to X seconds", but i (heinously) wrapped X in angled parentheses, which of course is illegal on the internets. This is fine.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-KMpwCb8Vg548pnU1tIy:-1879340630)*

vheon:
:lgtm: aside the inline comment.

---

Reviewed 1 of 3 files at r4, 1 of 12 files at r5, 3 of 5 files at r7, 1 of 2 files at r8, 1 of 2 files at r12, 2 of 3 files at r13, 4 of 4 files at r16.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[ycmd/wsgi_server.py, line 54 \[r14\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMjkQfhLzg7hehV5F8g:-KMpfgv4c7AExF8JE_8S:1396415455) ([raw file](https://github.com/valloric/ycmd/blob/3d9b124b71ec1faf73c5c0642e346395a7bddbca/ycmd/wsgi_server.py#L54)):*
> Won't this blow up in our face when we update waitress and they change the class internals?

Anyway now that we actually spawn a ycmd server in the tests we would know if something stops working before putting in in production :+1: 

---

*[ycmd/tests/shutdown_test.py, line 92 \[r16\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMsQDlK78Byc-iRxnZ7:-KMsQDlK78Byc-iRxnZ8:-885760854) ([raw file](https://github.com/valloric/ycmd/blob/15a99219704c74b92f845a8807321f0fd3d102fb/ycmd/tests/shutdown_test.py#L92)):*
> ```Python
>     self.AssertServerAndSubserversAreRunning()
> 
>     self.AssertServerAndSubserversShutDown()
> ```

Can we explicitly set the `timeout`? is not clear that it will wait for them to be down :confused: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-KMsQXhxfImslCxlnDdD:1177474608)*

micbou:
Waiting for PR #550 to be merged. I am highly suspecting the related issue is causing some flakiness in the tests.

---

Reviewed 4 of 4 files at r16, 2 of 2 files at r17.

---

*[ycmd/wsgi_server.py, line 54 \[r14\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMjkQfhLzg7hehV5F8g:-KMtDaeq_GvORBfKmuSe:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/3d9b124b71ec1faf73c5c0642e346395a7bddbca/ycmd/wsgi_server.py#L54)):*
> > Won't this blow up in our face when we update waitress and they change the class internals?
> 
> Anyway now that we actually spawn a ycmd server in the tests we would know if something stops working before putting in in production :+1: 
</details>
Done.

---

*[ycmd/tests/shutdown_test.py, line 92 \[r16\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMsQDlK78Byc-iRxnZ7:-KMtEGDorhhxovOjYKE5:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/15a99219704c74b92f845a8807321f0fd3d102fb/ycmd/tests/shutdown_test.py#L92)):*
> Can we explicitly set the `timeout`? is not clear that it will wait for them to be down :confused: 
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-KMtEmOhVeDlqk4-3J1m:-1724345799)*

vheon:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[ycmd/tests/shutdown_test.py, line 92 \[r16\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMsQDlK78Byc-iRxnZ7:-KMtStC7V2hPjsC5qRt2:-1938870209) ([raw file](https://github.com/valloric/ycmd/blob/15a99219704c74b92f845a8807321f0fd3d102fb/ycmd/tests/shutdown_test.py#L92)):*
> Done.
</details>
I know that you are going to kill me for this but: can we pass it with its name? so like `self.AssertServerAndSubserversShutDown( timeout = 5 )`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282)*

micbou:
Reviewed 1 of 1 files at r18.

---

*[ycmd/tests/shutdown_test.py, line 92 \[r16\]](https://reviewable.io:443/reviews/valloric/ycmd/282#-KMsQDlK78Byc-iRxnZ7:-KMtUIcloMs6DQWEnK9u:402412329) ([raw file](https://github.com/valloric/ycmd/blob/15a99219704c74b92f845a8807321f0fd3d102fb/ycmd/tests/shutdown_test.py#L92)):*
> I know that you are going to kill me for this but: can we pass it with its name? so like `self.AssertServerAndSubserversShutDown( timeout = 5 )`
</details>
Me? I would never do such thing :smiling_imp:

Anyway, you are right; it is clearer with the named parameter.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282)*

vheon:
:lgtm: 

---

Reviewed 1 of 2 files at r17, 1 of 1 files at r18.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-KMtYBlYPCV6mmLTTKPU:980077423)*

vheon:
@micbou is something still missing? Can we call `homu` on this?

micbou:
Rebased and squashed. This is ready now.

---

Reviewed 1 of 1 files at r19, 1 of 1 files at r20.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/282#-:-KNJ758C9uxqcFYRnyCu:-1301204953)*

Valloric:
Awesome, let's go! 

@homu r=puremourning

ptrv:
The server startup behavior changed also a bit, didn't? After start there is no print `serving on serving on http://127.0.0.1:<port>` anymore. Emacs ycmd used to parse the port number from there. Since we are not specifying the port number in Emacs ycmd. We used the automatically assigned one from the server.

puremourning:
@ptrv does that mean this commit breaks emacs-ycmd ?

How is the ycmd version tied to emacs-ycmd? I seem to recall that some folk install it independently? If so then I think we may have a problem that we need to fix.

ptrv:
Yes this commit breaks emacs-ycmd. Unfortunately we do not tie a specific ycmd version to it. Probably we should, but that's different story.

The quick fix in `emacs-ycmd` would be to specify the port for the server on `emacs-ycmd` side.

puremourning:
The difference is because we are no longer calling `waitress.serve` rather doing that job ourselves. We can add an equivalent print.

ptrv:
That would help a lot for now. and then emacs-ycmd needs to implement a way to tie the server version.

puremourning:
@ptrv any chance you could kick the tyres on the latest commit ? I don't have an emacs at hand...

ptrv:
@puremourning The patch works. Thanks!

puremourning:
Cool. I did try and make an emacs that worked with ycmd, but i'm too much of an emacs noob to make it work. i tried spacemacs and, well, none of the ycmd- commands seems to do anything. So thanks for checking, i'll stick to vim ^_^

PR #286, opened 2015-12-29
Remove unused method

vheon:
I was looking at the coverage and I noticed that this method is not tested, a quick grep on the codebase showed why: we apparently don't use it.


Valloric:
Thanks!

On a side note, Omnisharp startup is making our tests flaky again...

@homu r+

PR #287, opened 2016-01-01
Omnisharp Roslyn support

alistair:
This pull request contains initial support for using omnisharp-rosyln instead of the now legacy OmnisharpServer repository.

While most functionality seems to have been retained there is currently a regression in the YcmCompleter FixIt command.   This is expecting a response but the omnisharp-rosyln server isn't returning one.    I have yet to figure out whether this is expected or not.

There are 2 new commands which should be implemented /getcodeactions and /runcodeaction

examples of these requests and their response can be seen here https://gist.github.com/alistair/73bbd908b6464066e93d#file-getcodeaction

Also not that while I have attempted to make this work on windows,  I haven't tested whether it actually does.

Feedback on this pull request would be welcome.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/287)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,3 +25,6 @@
 [submodule "third_party/tern"]
 	path = third_party/tern
 	url = https://github.com/ternjs/tern.git
+[submodule "third_party/omnisharp-roslyn"]
+	path = third_party/omnisharp-roslyn
+	url = git@github.com:OmniSharp/omnisharp-roslyn.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -27,4 +27,4 @@
 	url = https://github.com/ternjs/tern.git
 [submodule "third_party/omnisharp-roslyn"]
 	path = third_party/omnisharp-roslyn
-	url = git@github.com:OmniSharp/omnisharp-roslyn.git
+	url = https://github.com/OmniSharp/omnisharp-roslyn.git
--- a/build.py
+++ b/build.py
@@ -266,13 +266,11 @@ def BuildYcmdLibs( args ):
 
 
 def BuildOmniSharp():
-  build_command = PathToFirstExistingExecutable(
-    [ 'msbuild', 'msbuild.exe', 'xbuild' ] )
-  if not build_command:
-    sys.exit( 'msbuild or xbuild is required to build Omnisharp' )
-
-  os.chdir( p.join( DIR_OF_THIS_SCRIPT, 'third_party', 'OmniSharpServer' ) )
-  subprocess.check_call( [ build_command, '/property:Configuration=Release' ] )
+  os.chdir( p.join( DIR_OF_THIS_SCRIPT, 'third_party', 'omnisharp-roslyn' ) )
+  if OnWindows():
+    subprocess.check_call( [ 'build.bat' ] )
+  else:
+    subprocess.check_call( [ './build.sh' ] )
 
 
 def BuildGoCode():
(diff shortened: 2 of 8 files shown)


alistair:
Ok,  i'm hoping this will actually go green now (at least on linux).  There are a number of things which need to be looked at and evaluated in terms of impact.

alistair:
i'm suspecting that omnisharp-roslyn is taking too long to startup on travis.   but will have to investigate later.   far too late now.

vheon:
We already had a PR for OmnisharpServer Roslyn #213. Did you look at it?

Valloric:
Thanks for the PR! :)

Like @vheon said, we've had a PR that tried to update to Omnisharp-roslyn so it's worth checking that out.

For the tests, I see that some are being marked as skipped. That's fine while you're developing, but of course those tests can't be skipped in the final code that will be merged. We'd all be happier with the roslyn version of omnisharp, but it can't introduce regressions.

For new subcommands, I'd suggest focusing this PR on replacing the functionality we have with the new omnisharp and leave exposing new features to a future PR.

---

Review status: 0 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[.gitmodules, line 28 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/287#-K6yQsWkcKfVIRCWqKKt:-K6yQsWlDVMMcl0Wde3c:1966314521) ([raw file](https://github.com/valloric/ycmd/blob/00a773bff4e20720d71ff868b97414234f9ec134/.gitmodules#L28)):*
Shouldn't the other omnisharp submodule be removed if this is being added?

---

*[ycmd/completers/cs/cs_completer.py, line 599 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/287#-K6yRMi_IaQn1zQb9nF7:-K6yRMi_IaQn1zQb9nF8:1116800717) ([raw file](https://github.com/valloric/ycmd/blob/00a773bff4e20720d71ff868b97414234f9ec134/ycmd/completers/cs/cs_completer.py#L599)):*
A better name might be `PostToServer`

---

*[ycmd/tests/cs/get_completions_test.py, line 114 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/287#-K6yRXMABe48gwokst8U:-K6yRXMABe48gwokst8V:651920153) ([raw file](https://github.com/valloric/ycmd/blob/00a773bff4e20720d71ff868b97414234f9ec134/ycmd/tests/cs/get_completions_test.py#L114)):*
Why are these tests being skipped?

I'm assuming this is temporary and the idea is to resolve this before the PR is declared ready to merge, correct?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/287#-:-K6yRn70Kc4VwoRmAOjs:-192056382)*

mispencer:
As far as I know, none of the issues with Omnisharp-rosyln mentioned in my earlier PR have been addressed, . It still lacks support for import types, hence those skip tests in this PR as well as in my original PR. This is a crucial feature to me, so I still can't support switching to Omnisharp-rosyln permanently. It also still has the [Recommender issue](https://github.com/OmniSharp/omnisharp-roslyn/issues/78) open, so I'm assuming that is also an issue.

Last time, the performance of the HTTP interface was lacking, so I had to implement the STDIO interface. This quickly became a whirlpool of pain, so if the performance of HTTP has improved, I would prefer to use that for MUCH simpler interface.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/287#-:-K6yXnFLWArJszYnSKsN:88445917)*

alistair:
I really thought I looked whether someone had started work on rosyln before I started.... 
Oh well,  I will chalk it up to experience learning ycmd internals.    But yes the skipped tests were temporary :)
The other pull request looks further along so im going to close this.   If you need any help on the other one,  just ping me.  more than interested to help :)

PR #288, opened 2016-01-01
Use 127.0.0.1 instead of localhost in C♯ completer

micbou:
Use `127.0.0.1` because `localhost` may not be defined by the OS.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/288)

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -584,7 +584,7 @@ class CsharpSolutionCompleter:
 
 
   def _ServerLocation( self ):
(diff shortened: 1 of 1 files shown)


micbou:
Should have tested it before sending a PR. [OmniSharp Server is listening on localhost](https://github.com/OmniSharp/omnisharp-server/blob/e1902915c6790bcec00b8d551199c8a3537d33c9/OmniSharp/Program.cs#L127). Closing.

Valloric:
@micbou You might want to add a comment there with the information you've discovered.

PR #297, opened 2016-01-08
[NOT READY] Bump Clang version to 3.8.0

micbou:
This is strange but [no binaries are available for Mac OS X](http://llvm.org/releases/download.html#3.7.1). I suppose this is an oversight and it should be added soon.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/297)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -44,7 +44,13 @@ if ( USE_CLANG_COMPLETER AND
      NOT PATH_TO_LLVM_ROOT AND
      NOT EXTERNAL_LIBCLANG_PATH )
 
-  set( CLANG_VERSION "3.7.0" )
+  # We are still using Clang 3.7.0 for Mac OS X because no official 3.7.1
+  # binaries were released on this platform.
+  if ( APPLE )
+    set( CLANG_VERSION "3.7.0" )
+  else()
+    set( CLANG_VERSION "3.7.1" )
+  endif()
 
   if ( APPLE )
     set( CLANG_DIRNAME "clang+llvm-${CLANG_VERSION}-x86_64-apple-darwin" )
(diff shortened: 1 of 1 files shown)


Valloric:
That's... really weird. It has happened before if I recall correctly, but it was for a different platform. Unlikely to get resolved by itself.

Valloric:
Oh, LGTM too.

puremourning:
I'm not keen on supporting different libclang versions for different platforms. It makes supporting the product harder. "Oh that's fixed if you're not using a mac... go complain to cfe-dev if you care...". It also restricts what we can do in the test suite.

What do you guys think? This change itself is fine with that caveat, though I think we should be finding out where the missing binaries are for OS X.

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/297#-:-K7kjNsursdJKFQbfLad:-41159235)*

micbou:
I agree with you and I was thinking about asking on the llvm-dev list but I'm afraid no one will answer. This kind of issue really makes me want to build our own LLVM binaries.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/297#-:-K7lLdm65QYQIXJYZOL-:1842169838)*

oblitum:
I agree with not adopting different versions except in case it starts to get usual to happen or versions by platform start to lag considerably. I have a feeling OS X binaries may come only in the next version.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/297#-:-K7lqmP9i4VwC9s8Npo_:1205398402)*

vheon:
> I'm not keen on supporting different libclang versions for different platforms. It makes supporting the product harder. "Oh that's fixed if you're not using a mac... go complain to cfe-dev if you care...". It also restricts what we can do in the test suite.
>
What do you guys think? This change itself is fine with that caveat, though I think we should be finding out where the missing binaries are for OS X.

I agree.

>  This kind of issue really makes me want to build our own LLVM binaries.

Would it be worth it?

micbou:
> Would it be worth it?

Probably not but we might consider it if no OS X binaries are available for the 3.8.0 release.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/297#-:-K7nOnHkJm4URUhSKGmo:-881172078)*

Valloric:
Solid points on not diverging with versions. My understanding was that 3.7.1 is a small bugfix release (so going out of sync wouldn't be too bad) and that the Mac download missing was a temporary oversight (so the out of sync was temporary).

Do we have a reason to believe either of the above points are false?

I'm also fine with use just sticking to 3.7.0 on all platforms until 3.8.0. Hopefully they wouldn't skip Mac then. I'd rather not be in a place where we distribute libclang, although it wouldn't be the end of the world.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/297#-:-K7ncQUAKJBbTD2-8hRb:2099175510)*

micbou:
Let's stay on 3.7.0 then. I'll update this PR for the 3.8.0.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/297#-:-K81x_cJapxJWygtPkNL:-290352273)*

micbou:
Closed in favor of PR #353.

PR #299, opened 2016-01-09
Error message when rust_src_path unset

Valloric:
If a completion request to racerd fails AND the user has not set the
rust_src_path config option, then we return a nice error message
mentioning to the user that they likely reason of the failure is the
empty value for that option. This should provide the user with something
to search for in the YCM docs.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/299)

### DIFF
--- a/ycmd/completers/rust/rust_completer.py
+++ b/ycmd/completers/rust/rust_completer.py
@@ -91,6 +91,11 @@ class RustCompleter( Completer ):
     self._server_state_lock = threading.RLock()
     self._keep_logfiles = user_options[ 'server_keep_logfiles' ]
     self._hmac_secret = ''
+    self._rust_source_path = self._GetRustSrcPath()
+
+    if not self._rust_source_path:
+      _logger.warn( 'No path provided for the rustc source. Please set the '
+                    'rust_src_path option' )
 
     if not self._racerd:
       _logger.error( BINARY_NOT_FOUND_MESSAGE )
@@ -205,7 +210,7 @@ class RustCompleter( Completer ):
     try:
       completions = self._FetchCompletions( request_data )
     except requests.HTTPError:
-      if not self._GetRustSrcPath():
+      if not self._rust_source_path:
         raise RuntimeError( ERROR_FROM_RACERD_MESSAGE )
       raise
 
@@ -263,12 +268,8 @@ class RustCompleter( Completer ):
       env = os.environ.copy()
       env[ 'RUST_BACKTRACE' ] = '1'
 
-      rust_src_path = self._GetRustSrcPath()
-      if rust_src_path:
-        args.extend( [ '--rust-src-path', rust_src_path ] )
-      else:
-        _logger.warn( 'No path provided for the rustc source. Please set the '
-                      'ycm_rust_src_path option' )
+      if self._rust_source_path:
+        args.extend( [ '--rust-src-path', self._rust_source_path ] )
 
       filename_format = p.join( utils.PathToTempDir(),
                                 'racerd_{port}_{std}.log' )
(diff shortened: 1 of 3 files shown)


Valloric:
@jwilm @vheon @puremourning @micbou

micbou:
Should we raise the error only one time like we are doing in the Javascript completer with the missing `.tern-project` configuration file?

---

---

*[ycmd/completers/rust/rust_completer.py, line 271 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/299#-K7ZpxNwdDw1CxNl-c0b:-K7ZpxNxL3r8oDgt_KKR:-570933750) ([raw file](https://github.com/valloric/ycmd/blob/269135da6149ae78009cd31df5aeef9ad3ddb98d/ycmd/completers/rust/rust_completer.py#L271)):*
Maybe it would be better to put this warning in `_GetRustSrcPath` method.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/299#-:-K7ZoW4vjZbQXvGCnCjN:-1910880473)*

Valloric:
Well, the error is being raised no matter what. racerd served us a 500, and we never want to swallow that. So we're never  _adding_ an error message, merely returning one message instead of another. So I can't see how we can do this only once without also swallowing the 500 from racerd.

---

---

*[ycmd/completers/rust/rust_completer.py, line 271 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/299#-K7ZpxNwdDw1CxNl-c0b:-K7ZrUzqz4SNA_Rr_uyG:1445501817) ([raw file](https://github.com/valloric/ycmd/blob/269135da6149ae78009cd31df5aeef9ad3ddb98d/ycmd/completers/rust/rust_completer.py#L271)):*
It was there, I just moved it. The reason for the move is that I want to avoid the message being printed to the log every time someone make a completion request that fails.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/299#-:-K7Zqlp96EfWRPVShtoe:1211000929)*

micbou:
Right. Didn't think that we were raising an error anyway.

---

Reviewed 1 of 3 files at r1.
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/completers/rust/rust_completer.py, line 271 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/299#-K7ZpxNwdDw1CxNl-c0b:-K7ZsXUJdN-IhhoRuAVZ:915020627) ([raw file](https://github.com/valloric/ycmd/blob/269135da6149ae78009cd31df5aeef9ad3ddb98d/ycmd/completers/rust/rust_completer.py#L271)):*
Oops, didn't see that. Then we should create a property `_rust_src_path` instead of calling `_GetRustSrcPath` each time. If I am not mistaken, this method always returns the same value once the server is started.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/299#-:-K7Zt3Wj0eoEpfcseIBL:1365003636)*

Valloric:

---

*[ycmd/completers/rust/rust_completer.py, line 271 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/299#-K7ZpxNwdDw1CxNl-c0b:-K7Zzg0O4OPHvOwiETdG:1260437658) ([raw file](https://github.com/valloric/ycmd/blob/269135da6149ae78009cd31df5aeef9ad3ddb98d/ycmd/completers/rust/rust_completer.py#L271)):*
I was thinking the same thing. Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/299)*

micbou:
:lgtm:

---

Reviewed 1 of 3 files at r1, 2 of 2 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/299#-:-K7_1RJHAqGn_D7O6M6e:1417088721)*

Valloric:
@homu r+

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/299#-:-K7_52rN1TxnhlQ3VfXY:-1142673777)*

Valloric:
@homu retry

PR #311, opened 2016-01-16
[READY] Support for rename in tern completer

puremourning:
## Tern rename

Tern server supports renames of variables only. It can do this across multiple files, only when it knows about those files. This feature is supported by tern_for_vim, so with the addition of this feature we should have feature-parity on that front.

## Refactoring infrastructure

This re-uses the existing FixIt infrastructure, with no changes to the JSON API. The only change is that this is the first time the command will return changes in multiple files.

The largest change is moving the Cs* classes that are equivalent of the C++ classes representing diagnostics/FixIts to `responses.py`, so they can be shared by other completers.

A Vim client change is required to recognise FixIt responses by the contents of the response, rather than the command name, and to apply the chunks across files.

## FileReadyToParse

Previously, the server would only learn about files if we sent a request (such as a completion request), or by appearing in the loadEagerly list.  Now we pass the file data to the server in the parse request.  This ensures that the server knows about files as soon as possible, so that they are available for completion suggestions and for rename/etc. commands.

## Tests

Simple tests for the single- and multi-file rename cases, both eagerly loaded and picked up from parse request.

## JSON API Changes

None.

## Client changes required

* FixIt commands are now identified by the presence of `'fixit'` entry in the  response
* FixIt commands can now apply across multiple files. The client is responsible for applying the chunks in an appropriate fashion.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/311)

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -41,41 +41,6 @@ PATH_TO_OMNISHARP_BINARY = os.path.join(
   'OmniSharp', 'bin', 'Release', 'OmniSharp.exe' )
 
 
-# TODO: Handle this better than dummy classes
-class CsharpDiagnostic:
-  def __init__ ( self, ranges, location, location_extent, text, kind ):
-    self.ranges_ = ranges
-    self.location_ = location
-    self.location_extent_ = location_extent
-    self.text_ = text
-    self.kind_ = kind
-
-
-class CsharpFixIt:
-  def __init__ ( self, location, chunks ):
-    self.location = location
-    self.chunks = chunks
-
-
-class CsharpFixItChunk:
-  def __init__ ( self, replacement_text, range ):
-    self.replacement_text = replacement_text
-    self.range = range
-
-
-class CsharpDiagnosticRange:
-  def __init__ ( self, start, end ):
-    self.start_ = start
-    self.end_ = end
-
-
-class CsharpDiagnosticLocation:
-  def __init__ ( self, line, column, filename ):
-    self.line_number_ = line
-    self.column_number_ = column
-    self.filename_ = filename
-
-
 class CsharpCompleter( Completer ):
   """
   A Completer that uses the Omnisharp server as completion engine.
@@ -246,14 +211,15 @@ class CsharpCompleter( Completer ):
   def _QuickFixToDiagnostic( self, quick_fix ):
     filename = quick_fix[ "FileName" ]
 
-    location = CsharpDiagnosticLocation( quick_fix[ "Line" ],
-                                         quick_fix[ "Column" ], filename )
-    location_range = CsharpDiagnosticRange( location, location )
-    return CsharpDiagnostic( list(),
-                             location,
-                             location_range,
-                             quick_fix[ "Text" ],
-                             quick_fix[ "LogLevel" ].upper() )
+    location = responses.Location( quick_fix[ "Line" ],
+                                   quick_fix[ "Column" ],
+                                   filename )
+    location_range = responses.Range( location, location )
+    return responses.Diagnostic( list(),
+                                 location,
+                                 location_range,
+                                 quick_fix[ "Text" ],
+                                 quick_fix[ "LogLevel" ].upper() )
 
 
   def GetDetailedDiagnostic( self, request_data ):
@@ -513,11 +479,12 @@ class CsharpSolutionCompleter:
 
     result = self._GetResponse( '/fixcodeissue', request )
     replacement_text = result[ "Text" ]
-    location = CsharpDiagnosticLocation( request_data['line_num'],
-                                         request_data['column_num'],
-                                         request_data['filepath'] )
-    fixits = [ CsharpFixIt( location,
-                            _BuildChunks( request_data, replacement_text ) ) ]
+    location = responses.Location( request_data['line_num'],
+                                   request_data['column_num'],
+                                   request_data['filepath'] )
+    fixits = [ responses.FixIt( location,
+                                _BuildChunks( request_data,
+                                              replacement_text ) ) ]
 
     return responses.BuildFixItResponse( fixits )
 
@@ -654,10 +621,10 @@ def _BuildChunks( request_data, new_buffer ):
   ( start_line, start_column ) = _IndexToLineColumn( old_buffer, start_index )
   ( end_line, end_column ) = _IndexToLineColumn( old_buffer,
                                                  old_length - end_index )
-  start = CsharpDiagnosticLocation( start_line, start_column, filepath )
-  end = CsharpDiagnosticLocation( end_line, end_column, filepath )
-  return [ CsharpFixItChunk( replacement_text,
-                             CsharpDiagnosticRange( start, end ) ) ]
+  start = responses.Location( start_line, start_column, filepath )
+  end = responses.Location( end_line, end_column, filepath )
+  return [ responses.FixItChunk( replacement_text,
+                                 responses.Range( start, end ) ) ]
 
 
 def _FixLineEndings( old_buffer, new_buffer ):
--- a/ycmd/completers/javascript/tern_completer.py
+++ b/ycmd/completers/javascript/tern_completer.py
@@ -188,6 +188,16 @@ class TernCompleter( Completer ):
   def OnFileReadyToParse( self, request_data ):
     self._WarnIfMissingTernProject()
 
+    # Keep tern server up to date with the file data. We do this by sending an
+    # empty request just containing the file data
+    try:
+      self._PostRequest( {}, request_data )
+    except:
+      # The server might not be ready yet or the server might not be running.
+      # in any case, just ignore this we'll hopefully get another parse request
+      # soon.
+      pass
+
 
   def GetSubcommandsMap( self ):
     return {
@@ -568,7 +578,6 @@ class TernCompleter( Completer ):
                             filename ) )
 
 
-
     def BuildFixItChunk( change ):
       return responses.FixItChunk(
         change[ 'text' ],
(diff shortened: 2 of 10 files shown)


puremourning:
I think this is basically complete, but marked as `[RFC]` for now as it is a new feature for ycmd and requires client changes (which are on their way).

puremourning:
Obligatory Vim client demo:

Multi-file rename:
ycm-tern-rename-multi-file-1

And another one, opening the files:
ycm-tern-rename-multi-file-2

mispencer:
I would really like the various `BuildWhateverData` to be refactored to take dictionaries instead of the classes with the weird `whatever_` properties.  That's what my `TODO: Handle this better than dummy classes` note was referring to in `CsCompleter`. As I recall, this was an implementation detail for the clang completer, which was the first completer with diagnostics, and I just hacked it when I implemented that in the C# completer (which was the second). I would like to hide this (before it has even more usages), and if we're refactoring it, this would be a good time.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/311#-:-K8B5gnH3lCFuFxUWwAM:-1332284909)*

Valloric:
_Stellar_ work as usual, :lgtm: on all of it (code & general approach)!

---

Reviewed 10 of 10 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/311#-:-K8CVb63dukNRACGQDWF:2061543664)*

vheon:
> I would really like the various BuildWhateverData to be refactored to take dictionaries instead of the classes with the weird whatever_ properties. That's what my TODO: Handle this better than dummy classes note was referring to in CsCompleter. As I recall, this was an implementation detail for the clang completer, which was the first completer with diagnostics, and I just hacked it when I implemented that in the C# completer (which was the second). I would like to hide this (before it has even more usages), and if we're refactoring it, this would be a good time.

@mispencer Why do you think bare dictionary are better than named field in classes?

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/311#-:-K8EwHDpbXGewXO4W1bJ:-526273857)*

mispencer:
Because the only usage of those classes, as far as I remember, is to pass them to those Build methods to be converted into bare dictionaries. It would be simpler to just build the dictionaries directly.

The reason we aren't doing so now is an implementation detail with the Clang C++ completer - those classes are what the Clang C++ bridge returns, and the conversion was done above the completer layer, so when I implemented the C# diagnostics, I elected to mock those classes instead of refactoring. 

Though honestly, I am mostly just bothered by the trailing underscores in the properties' names, which I assume is because they are conversions of a C++ classes.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/311#-:-K8FPzRObNKAbdeg6e_o:1856656244)*

puremourning:
The other reason for the `Build*` methods is that it confines the JSON API to `responses.py`, which is a legit separation of concerns.

The usage of them at least is quite intuitive:

```python
    def BuildRange( filename, start, end ):
      return responses.Range(
        responses.Location( start[ 'line' ] + 1,
                            start[ 'ch' ] + 1,
                            filename ),
        responses.Location( end[ 'line' ] + 1,
                            end[ 'ch' ] + 1,
                            filename ) )


    def BuildFixItChunk( change ):
      return responses.FixItChunk(
        change[ 'text' ],
        BuildRange( os.path.abspath( change[ 'file'] ),
                    change[ 'start' ],
                    change[ 'end' ] ) )
```

I think a bare dictionary would be somewhat more verbose (see the comment above) in `tern_compelter.py`.

I agree that the magic trailing `_` is less than idea, but all the more reason to bundle in `requests` module.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/311#-:-K8B9P0HW1yDO2WL7TK1:1188750901)*

puremourning:
I think i'm happy with this now. I'd still appreciate it if reviewers could take it for a spin with https://github.com/Valloric/YouCompleteMe/pull/1917.

micbou:
Reviewed 10 of 10 files at r1.

---

*[ycmd/completers/javascript/tern_completer.py, line 520 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/311#-K8LvnIfxkiUrM5G45xK:-K8LvnIgHnkXDxxmcvVb:2043701117) ([raw file](https://github.com/valloric/ycmd/blob/e7058f1693991eaeb391357c68f54853e1045563/ycmd/completers/javascript/tern_completer.py#L520)):*
Need to raise a `RuntimeError` with an appropriate message if index does not exist. Currently, User get `IndexError: list index out of range` in Vim if no argument is given to `RefactorRename`.

---

*[ycmd/completers/javascript/tern_completer.py, line 591 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/311#-K8LwejxutlSjt-7B2-N:-K8LwejxutlSjt-7B2-O:72964020) ([raw file](https://github.com/valloric/ycmd/blob/e7058f1693991eaeb391357c68f54853e1045563/ycmd/completers/javascript/tern_completer.py#L591)):*
Should we rename `FixIt` to `Refactor`?

---

*[ycmd/tests/javascript/subcommands_test.py, line 23 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/311#-K8Lxinr8NGbNz7GdRnn:-K8Lxinr8NGbNz7GdRno:-1397982027) ([raw file](https://github.com/valloric/ycmd/blob/e7058f1693991eaeb391357c68f54853e1045563/ycmd/tests/javascript/subcommands_test.py#L23)):*
Spaces after `(` and before `)`.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/311)*

puremourning:
Review status: 7 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[ycmd/completers/javascript/tern_completer.py, line 520 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/311#-K8LvnIfxkiUrM5G45xK:-K8M0QrAhJF7oeBG-WfP:-1625871788) ([raw file](https://github.com/valloric/ycmd/blob/e7058f1693991eaeb391357c68f54853e1045563/ycmd/completers/javascript/tern_completer.py#L520)):*
Good point! Thanks.

---

*[ycmd/completers/javascript/tern_completer.py, line 591 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/311#-K8LwejxutlSjt-7B2-N:-K8M0SRQE3pCS_P9g4Hg:1399379111) ([raw file](https://github.com/valloric/ycmd/blob/e7058f1693991eaeb391357c68f54853e1045563/ycmd/completers/javascript/tern_completer.py#L591)):*
Hmmm. Tricky one. The JSON api still says 'fixit' so it still feels like a FixIt response, but you're right - it's now a more general command.

I've updated the doctoring for the `FixIt` class and `BuildFixItResponse` to signpost that they are useful for refactor commands too.

---

*[ycmd/tests/javascript/subcommands_test.py, line 23 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/311#-K8Lxinr8NGbNz7GdRnn:-K8M0gtZFBsbdCYt8yGj:-1745992165) ([raw file](https://github.com/valloric/ycmd/blob/e7058f1693991eaeb391357c68f54853e1045563/ycmd/tests/javascript/subcommands_test.py#L23)):*
yup. fixed.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/311)*

Valloric:
Review status: 7 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[ycmd/completers/javascript/tern_completer.py, line 591 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/311#-K8LwejxutlSjt-7B2-N:-K8Q1WFNoauI-yO3jDB7:-738009497) ([raw file](https://github.com/valloric/ycmd/blob/e7058f1693991eaeb391357c68f54853e1045563/ycmd/completers/javascript/tern_completer.py#L591)):*
Well, FixIt comes from libclang which has fixits for diagnostic errors. Those aren't really refactors since they change the semantics (from broken to working :)).

I think "refactor" has a more narrow meaning than "fixit," and I think we need the more general one.

The new docs/comments are nice though.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/311)*

Valloric:
I prefer having the Build functions. As @puremourning surmised correctly, they're there to confine the JSON API to `responses.py`. We have many completers with more to come, and I don't like having our JSON API (which is our contract to clients) spread all over the place. I think tests also use the Build functions to prepare expected responses, so having the same code in prod and in tests is good.

Adding more constructs to the JSON API should not in fact be too easy. 

---

Reviewed 3 of 3 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/311#-:-K8Q2rKkOaONLWZDfGv9:649551888)*

vheon:
Reviewed 7 of 10 files at r1, 3 of 3 files at r2.

---

*[ycmd/completers/javascript/tern_completer.py, line 591 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/311#-K8LwejxutlSjt-7B2-N:-K8Qhd-8jzofTHThJ1xo:1864463781) ([raw file](https://github.com/valloric/ycmd/blob/e7058f1693991eaeb391357c68f54853e1045563/ycmd/completers/javascript/tern_completer.py#L591)):*
On top of all @Valloric said, I kinda like `FixIt` :stuck_out_tongue_closed_eyes: 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/311)*

puremourning:
I *think* the appveyor build is spurious. I certainly don't understand the reason for failure (test directory not empty at the end?), but there are so many warnings from the compiler it is hard to see the wood for the trees.

Valloric:
@puremourning That appveyor failure is almost certainly flakiness; only one run out of 6 failed, and all travis runs succeeded as well. So let's go with it.

@homu r+

PR #295, opened 2016-01-07
[READY] Make python binary specifiable by the user

vheon:
I'm starting this PR as WIP because it lacks tests and I wanted to ask for suggestion about that. What do you think would be the best approach? I was thinking about configuring the `user_option` and check how the `utils.SafePopen` was called but in general I'm not a fan of this kind of tests, so I'm open to suggestion :smiley:

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/295)

PR #313, opened 2016-01-18
support py3

ocehugo:
See YCM push request. All tests working (didn't finished the GO ones).

YCM still missing completion in vim with --python3 but I'm leaving this to someone that can push harder than me now (maybe it's simple).

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/313)

### DIFF
--- a/build.py
+++ b/build.py
@@ -6,9 +6,15 @@ import os.path as p
 import sys
 
 major, minor = sys.version_info[ 0 : 2 ]
-if major != 2 or minor < 6:
-  sys.exit( 'The build script requires Python version >= 2.6 and < 3.0; '
-            'your version of Python is ' + sys.version )
+# support python3
+if major >2:
+    PY3 = True
+else:
+    PY3 = False
+
+if major == 2 and minor < 6:
+     sys.exit( 'The build script requires Python version >= 2.6 and < 3.0; '
+              'your version of Python is ' + sys.version )
 
 DIR_OF_THIS_SCRIPT = p.dirname( p.abspath( __file__ ) )
 DIR_OF_THIRD_PARTY = p.join( DIR_OF_THIS_SCRIPT, 'third_party' )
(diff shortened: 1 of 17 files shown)


vheon:
I'm sorry but, why do we want this?

ocehugo:
@vheon,
sorry but why you don't want this?

If u had a look at the code I take the try to write a plausible port that would run in both py2 and py3 versions. The only thing i did change that is not backward compatible (hopefully) was the vim scripts ( and I did that for testing/personal purposes since i don't have vim with py2).

So answering you question:

You need this for the people out there that are using python3. Several of my vimscripts already support python3. I use vim to develop python3 code and since my code is not in the python2 folder i can't get completions to work (ycm use jedi, that loads the packages in the python interpreter directory prefix).

So I was thinking if I port the ycm completer I could use the whole vim with python3 and have completions within my package and other py3 packages and functions. 

Today you cannot use python3 and ycm in vim.

From this push I found kinda difficult since new push keeps happening with python2 style (just fixed some of them), so another reason to port and keep using py3 instead is that in the future will be harder and harder to port since more code will be add with old syntax/calls.

If the devs want they can take it from here or burry this into oblivion.  I just wanna show that it was not too hard and it would not be too difficult to maintain support, as long the contributors update some of the "way of coding in py2".

puremourning:
> I use vim to develop python3 code and since my code is not in the python2 folder i can't get completions to work (ycm use jedi, that loads the packages in the python interpreter directory prefix).

We're currently waiting for the regression tests to complete on a PR which adds [python 3 completion to ycm](https://github.com/Valloric/ycmd/pull/295). So very soon (like today) you should be able to complete python 3 code by setting 1 Vim variable.

ocehugo:
That is really nice. Well, I will not put anymore work on this (unless people are interested). Feel free to import whatever changes you may find good for the project

puremourning:
@ocehugo thanks for going to the effort! porting the code to python 3 has come up before and this will certainly help. right now there's not a huge appetite for changing the python version and everything that goes with it because the various complexities of various distribution's python installations is probably the most common source of installation problems for YCM :/

micbou:
Thanks for the work. We'll need to do this one day.

I was wondering: did you consider using a tool like [`futurize` to add python3 support](http://python-future.org/automatic_conversion.html)?

ocehugo:
No. I didn't really look at. Never ported py2->py3 before. I was so annoyed that I couldn't use the completer that I decide to give a try... hopefully there was a lot of tests !

puremourning:
There's a whole book on porting, which i started reading in order to review this PR http://python3porting.com

jwilm:
> There's a whole book on porting

:sob:

Valloric:
Thanks @ocehugo for doing this work and sending a PR! :)

We don't _currently_ have porting to python3 on our roadmap, but as @micbou said, we're probably going to have to do this at some point. And as others have pointed out, we'll be able to complete python3 very soon even with the architecture we have right now.

But this work will be a useful reference when we do decide to go down this path. 

I'll close this now since we're not going to be doing this just yet. Thanks again!

PR #318, opened 2016-01-21
[READY] Properly split EXTRA_CMAKE_ARGS environment variable in build.py script

micbou:
While I was playing with `EXTRA_CMAKE_ARGS` environment variable to answer issue Valloric/YouCompleteMe#1925, I found out it was not possible to pass CMake arguments with spaces in it because the variable is split on spaces. For example:
```sh
EXTRA_CMAKE_ARGS="-DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_CXX_FLAGS='-march=native -O3'"
```
gives
```python
['-DCMAKE_VERBOSE_MAKEFILE=ON', "-DCMAKE_CXX_FLAGS='-march=native", "-O3'"]
```
which obviously fail.

This is fixed by using the [`shlex` module and its split feature](https://docs.python.org/2/library/shlex.html). With this, example above results in:
```python
['-DCMAKE_VERBOSE_MAKEFILE=ON', '-DCMAKE_CXX_FLAGS=-march=native -O3']
```
which make us happy.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/318)

### DIFF
--- a/build.py
+++ b/build.py
@@ -4,6 +4,7 @@ import os
 import subprocess
 import os.path as p
 import sys
+import shlex
 
 major, minor = sys.version_info[ 0 : 2 ]
 if major != 2 or minor < 6:
@@ -219,7 +220,8 @@ def GetCmakeArgs( parsed_args ):
     cmake_args.append( '-DUSE_SYSTEM_BOOST=ON' )
 
   extra_cmake_args = os.environ.get( 'EXTRA_CMAKE_ARGS', '' )
-  cmake_args.extend( extra_cmake_args.split() )
+  # We use shlex split to properly parse quoted CMake arguments.
+  cmake_args.extend( shlex.split( extra_cmake_args ) )
   return cmake_args
 
 


puremourning:
Tbh i think we should have `-- <extra cmake arts>` for install. The env. is a bit annoying. 

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/318#-:-K8Zh94ji6dgHv7nHNCt:788268407)*

vheon:
> Tbh i think we should have -- <extra cmake arts> for install. The env. is a bit annoying.

It would be easier to use for sure but maybe less explicit?

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/318#-:-K8_EswrqUph_qiYF3hZ:-1042603532)*

micbou:
> Tbh i think we should have -- <extra cmake arts> for install. The env. is a bit annoying. 

This would be in addition to the environment variable then because it is used in `run_test.py` to pass the `-DUSE_DEV_FLAGS=ON` and `-DUSE_CLANG_COMPLETER=ON` flags.

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/318#-:-K8aTXspvtxCJFlduVdO:-665946667)*

puremourning:
:lgtm:

@micbou you're right. `-- <args>` might be useful for humans, but we wouldn't dispense with`EXTRA_CMAKE_ARGS`, so having 2 ways to do the same thing is probably worse.

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/318#-:-K8novKIyFUoElF27bBh:221050390)*

Valloric:
Nice catch! :)

@homu r+

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/318#-:-K8psx_KwTNNQKKI_nZN:367226421)*

PR #319, opened 2016-01-23
[READY] Relax semantic trigger matching to match between column start and current cursor column

micbou:
This PR fixes issues Valloric/YouCompleteMe#1602 and Valloric/YouCompleteMe#1930.

I updated the tests and added some cases.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/319)

### DIFF
--- a/ycmd/completers/completer.py
+++ b/ycmd/completers/completer.py
@@ -147,10 +147,11 @@ class Completer( object ):
       return False
     current_line = request_data[ 'line_value' ]
     start_column = request_data[ 'start_column' ] - 1
+    column_num = request_data[ 'column_num' ] - 1
     filetype = self._CurrentFiletype( request_data[ 'filetypes' ] )
 
     return self.prepared_triggers.MatchesForFiletype(
-        current_line, start_column, filetype )
+        current_line, start_column, column_num, filetype )
 
 
   def QueryLengthAboveMinThreshold( self, request_data ):
--- a/ycmd/completers/completer_utils.py
+++ b/ycmd/completers/completer_utils.py
@@ -23,7 +23,7 @@ import re
 
 
 class PreparedTriggers( object ):
-  def __init__( self,  user_trigger_map = None, filetype_set = None ):
+  def __init__( self, user_trigger_map = None, filetype_set = None ):
     user_prepared_triggers = ( _FiletypeTriggerDictFromSpec(
         dict( user_trigger_map ) ) if user_trigger_map else
         defaultdict( set ) )
@@ -36,17 +36,23 @@ class PreparedTriggers( object ):
     self._filetype_to_prepared_triggers = final_triggers
 
 
-  def MatchingTriggerForFiletype( self, current_line, start_column, filetype ):
+  def MatchingTriggerForFiletype( self, current_line, start_column, column_num,
+                                  filetype ):
     try:
       triggers = self._filetype_to_prepared_triggers[ filetype ]
     except KeyError:
       return None
-    return _MatchingSemanticTrigger( current_line, start_column, triggers )
+    return _MatchingSemanticTrigger( current_line,
+                                     start_column,
+                                     column_num,
+                                     triggers )
 
 
-  def MatchesForFiletype( self, current_line, start_column, filetype ):
+  def MatchesForFiletype( self, current_line, start_column, column_num,
+                          filetype ):
     return self.MatchingTriggerForFiletype( current_line,
                                             start_column,
+                                            column_num,
                                             filetype ) is not None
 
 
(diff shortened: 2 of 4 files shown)


puremourning:
Generally looks good. 

Some people thoughts:
- does this change potentially break any existing custom triggers? I know there are a number of hacks out there to (heinously?) trigger semantic on each keystroke. 
- should we add tests for the cases where this came up like triggers including identifier chars? 
- how does this change the interaction with the cache (if at all)? I confess I can't remember what goes in as the key in the cache, but it seems now that the 'prefix' is more fluid which might interfere with existing assumptions. 

---

Reviewed 3 of 3 files at r1.

---

*[ycmd/completers/completer_utils.py, line 87 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/319#-K8nP9YfyNDDPvEY46JL:-K8nP9YfyNDDPvEY46JM:156435841) ([raw file](https://github.com/valloric/ycmd/blob/a254497802fe59f23e2f439baab95c6696b2ed11/ycmd/completers/completer_utils.py#L87)):*
I guess this could benefit from a comment. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/319#-:-K8nNfrMfKlYwvUr4gHi:842769768)*

micbou:
> does this change potentially break any existing custom triggers? I know there are a number of hacks out there to (heinously?) trigger semantic on each keystroke. 

Lookahead regex hack will work again (it was broken by PR #292) but will not be needed anymore. I can't think of any other hack.

> should we add tests for the cases where this came up like triggers including identifier chars? 

Probably. We could use our dummy completer for these tests.

> how does this change the interaction with the cache (if at all)? I confess I can't remember what goes in as the key in the cache, but it seems now that the 'prefix' is more fluid which might interfere with existing assumptions. 

This PR only affects the value returned by `MatchesForFiletype` and this value is not cached.

---

---

*[ycmd/completers/completer_utils.py, line 87 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/319#-K8nP9YfyNDDPvEY46JL:-K8nUKUHJEwCH_ZemjPm:-449979228) ([raw file](https://github.com/valloric/ycmd/blob/a254497802fe59f23e2f439baab95c6696b2ed11/ycmd/completers/completer_utils.py#L87)):*
I'll add one.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/319#-:-K8nULhN4elP5LVb7gD3:1136869299)*

puremourning:
OK with that minor comment this :lgtm:. Great work.

I wonder if there is anywhere that this behaviour is currently documented, and if we need to update that? The previous behaviour was fairly subtle, and now it is subtly different. Not really sure. Maybe an update to YCM's README for the option, or the comments around the default trigger map. 

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/319#-:-K8nqbZWtzwSrCIHfRSR:1114144401)*

vheon:
I'm a little puzzled by some changes (I have to admit that I don't remember how this whole piece of code works):

- Why we need both `start_column` and `column_num`? isn't `column_num` enough to determine if we have to trigger a semantic completion?
- Why are we checking for `match.end() <=` while before we were checking if `match.end() ==`?

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/319#-:-K8o4rTz87ESD1JzzbOt:1694585651)*

oblitum:
Once remaining comments are added, it's :LGTM: for me too.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/319#-:-K8oZ3nuTRBDchm0AIcc:-833622400)*

micbou:
I added the comment (hope it is understandable) and an integration test.

> Why we need both start_column and column_num? isn't column_num enough to determine if we have to trigger a semantic completion?

> Why are we checking for match.end() <= while before we were checking if match.end() ==?

Current behavior is to check at `start_column` because this is where we expect to find a semantic (non-identifier) trigger. However, this doesn't work when the trigger is an identifier itself (more precisely, when it ends with an identifier character). For this reason, we need to match between `start_column` and `column_num`. An example:
```cpp
       
foo.bar_qux
    ^      ^
    |      column_num
    start_column
```
If our trigger is `.`, its end will match at `4 = start_column` (0-based index). If our trigger is `_`, its end will match at `9 <= column_num`.  

---

Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/completers/completer_utils.py, line 87 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/319#-K8nP9YfyNDDPvEY46JL:-K8pcM-zmaLhAmGnzhpF:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/a254497802fe59f23e2f439baab95c6696b2ed11/ycmd/completers/completer_utils.py#L87)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/319#-:-K8o74b-VEmyookdBl0K:-284255505)*

oblitum:
@homu +r. Good explanation, I'm not sure the act of triggering needed to be so tricky!

---

Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/319#-:-K8qI36_hP3jmenm0AvE:1091592284)*

oblitum:
@homu r+

---

Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/319#-:-K8qJhmyCzrDLumpLaWh:-1142673777)*

PR #320, opened 2016-01-24
improve perl6 identifier completions

atweiden:
supports kebab case

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/320)

### DIFF
--- a/ycmd/identifier_utils.py
+++ b/ycmd/identifier_utils.py
@@ -75,9 +75,7 @@ FILETYPE_TO_IDENTIFIER_REGEX = {
     'haskell': re.compile( r"[_a-zA-Z][\w']*", re.UNICODE ),
 
     # Spec: http://doc.perl6.org/language/syntax
-    'perl6': re.compile(
-        r"(?:[A-Za-z_\xC0-\xFF](?:[A-Za-z_\xC0-\xFF0-9]|[-'](?=[A-Za-z_\xC0-\xFF]))*)",
-        re.UNICODE ),
+    'perl6': re.compile( r"[_a-zA-Z](?:\w|[-'](?=[_a-zA-Z]))*", re.UNICODE ),
 }
 
 FILETYPE_TO_IDENTIFIER_REGEX[ 'scss' ] = FILETYPE_TO_IDENTIFIER_REGEX[ 'css' ]
--- a/ycmd/identifier_utils.py
+++ b/ycmd/identifier_utils.py
@@ -73,6 +73,11 @@ FILETYPE_TO_IDENTIFIER_REGEX = {
     # Spec: http://www.haskell.org/onlinereport/lexemes.html
     # Section 2.4
     'haskell': re.compile( r"[_a-zA-Z][\w']*", re.UNICODE ),
+
+    # Spec: http://doc.perl6.org/language/syntax
+    'perl6': re.compile(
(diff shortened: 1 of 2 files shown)


puremourning:
Thanks for the PR!

The regex is quite... opaque, so it is quite hard to review what it is doing. Could you update the PR description with an overview of the change and how you tested it?

We totally need auto tests for this too :)

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/320#-:-K8pHPeraFNfd0A1AIlq:1496713213)*

atweiden:
This regex is from [vim-perl](https://github.com/vim-perl/vim-perl/blob/2cef2074b423e88d981bad60496dd0f1a1d5ecf9/syntax/perl6.vim#L66).

A [complete regex](https://github.com/perl6/std/blob/28329a77382cda87f1bba12a8522005766a1054a/STD.pm6#L711) would require recursion (absent in python re module):

```perl6
token ident
{
    <.alpha> \w*
}

token apostrophe
{
    <[ ' \- ]>
}

token identifier
{
    <.ident> [ <.apostrophe> <.ident> ]*
}
```

Note that the current YCM completer will not complete such vars as:

```perl6
my $kebab-case-variable;
my $variable-with-an-apostrophe'here;
```

Also note neither the current YCM completer nor my modified completer will complete vars containing unicode chars.

```perl6
my $vxyarי;
my $vxyaræ;
my $vxyarь;
my $vxyarѳ;
my $vxya۰r;
```

No functionality is lost.

Valloric:
Thanks so much for adding tests! :)

That regex looks very complicated while the tests seem fairly light; are you confident you're covering enough (useful) edge-cases? If so, I'm happy with this if @puremourning is also happy.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/320#-:-K9-E3-SId3U5RhWFwtE:-1974980111)*

atweiden:
I've been using this regex in [vim-perl](https://github.com/vim-perl/vim-perl) since April 2015. The regex lacks support for variables containing unicode, but otherwise works well. ycmd's default perl6 completion does not support completing variables containing unicode either, so no functionality is lost.

As well, I've been locally testing this PR's [patch](https://github.com/atweiden/pkgbuilds-plugged/commit/f34c4f73fb77f9df8f9ebb482acbbcb6e3a7467d) over the past two days without issue. I've been converting all my p6 code from snake_case to kebab-case, and it's the same experience as before (except with completion of kebab).

Valloric:
@atweiden I believe the regex works well _right now_, :) but having good test coverage is critical to ensure it stays working in the future. Someone will want to change that regex later on, and we don't want regressions. Enabling future refactoring and maintenance is the point of tests, not confirming that what we have right now works (though that's a nice benefit).

Valloric:
@micbou We keep seeing this transient "directory is not empty" error in appveyor runs. You have the best idea how our appveyor setup works; could you take a look at what's going on there when you find the time?

micbou:
It's probably better to remove the `\xC0-\xFF` characters range in the regex because YCM does not support extended ASCII.

@Valloric See PR #325.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/320#-:-K99PjF88AaYgAocQilj:-362801217)*

Valloric:
This looks good! Thanks for the PR! :)

@homu r+

@atweiden In the future, add a comment to a PR when you add more commits. GitHub doesn't post notifications to others when new commits are added, so we're likely to miss your changes.

PR #322, opened 2016-01-26
Add groovy semantic trigger

jreiter:
[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/322)

### DIFF
--- a/ycmd/completers/completer_utils.py
+++ b/ycmd/completers/completer_utils.py
@@ -156,7 +156,7 @@ DEFAULT_FILETYPE_TRIGGERS = {
   'cpp,objcpp' : ['->', '.', '::'],
   'perl' : ['->'],
   'php' : ['->', '::'],
-  'cs,java,javascript,typescript,d,python,perl6,scala,vb,elixir,go' : ['.'],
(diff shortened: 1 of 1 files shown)


puremourning:
This looks fine,  though we should add some tests for it.

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/322#-:-K8ykRpkNALZCa0Si5K-:470855543)*

jreiter:
I don't mind doing that if you can give me a little guidance.  The existing language-specific tests (clang, cs, go, etc.) seem to all be based on external completion engines.  For groovy, it would all be identifier and omnifunc-based (in my case, through eclim) completion.

I don't have any python dev experience, though, so it's possible I'm missing something.

vheon:
@puremourning I don't think we ever tested this kind of stuff.

Valloric:
Thanks for the PR! :)

If we don't have a setup to test this kind of change already, I don't want to ask @jreiter to add it. Let's merge it and then we can follow up on this with tests for semantic triggers.

@homu r+

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/322#-:-K9-Csex9rVM24KfEn4X:1792428312)*

PR #325, opened 2016-01-28
[READY] Fix error when removing temporary folder on AppVeyor

micbou:
So there is this new error that occurs occasionally on AppVeyor:
```sh
WindowsError: [Error 145] The directory is not empty: 'c:\\users\\appveyor\\appdata\\local\\temp\\1\\ycm_build.***\\ycm\\tests'
```
It probably happens because the `ycm_core_tests` executable is still used by some process while the build folder is removed.

I see three ways to fix this:
 - ignore the error;
 - ignore it but print a warning;
 - retry several times (`3` should be enough).

I implemented the first approach because it is the simplest but I don't really like the idea of leaving (in rare cases) some files in the temporary folder (Windows does not automatically clean this folder). This is why the PR is marked as RFC. Ideally, we should implement the third approach.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/325)

### DIFF
--- a/build.py
+++ b/build.py
@@ -38,6 +38,10 @@ def OnWindows():
   return platform.system() == 'Windows'
 
 
+def OnTravisOrAppVeyor():
+  return 'CI' in os.environ
+
+
 # On Windows, distutils.spawn.find_executable only works for .exe files
 # but .bat and .cmd files are also executables, so we use our own
 # implementation.
@@ -266,7 +270,7 @@ def BuildYcmdLibs( args ):
       RunYcmdTests( build_dir )
   finally:
     os.chdir( DIR_OF_THIS_SCRIPT )
-    rmtree( build_dir )
+    rmtree( build_dir, ignore_errors = OnTravisOrAppVeyor() )
 
 
 def BuildOmniSharp():


Valloric:
Can we make `rmtree` ignore errors only on AppVeyor and Travis? I'd be fine with that. I know Travis exposes some env vars we can detect to know we're running inside Travis; AppVeyor probably has something similar.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/325#-:-K99YYTHr025uwuY3ozA:-846291770)*

micbou:
They both define the `CI` environment variable so we could use it or we only do this for AppVeyor and use `APPVEYOR` in this case.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/325#-:-K99aKxzmh1njSjPBhV-:-678292652)*

micbou:
So I decided to use the `CI` environment variable. I think this is ready.

---

Reviewed 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/325#-:-K9C9H-D1Q5WFVFnF270:1604167634)*

vheon:
:lgtm: 

---

Reviewed 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/325#-:-K9ECdQc7XEa_N1ngfj_:980077423)*

Valloric:
Thanks! :)

@homu r+

PR #327, opened 2016-01-29
Rename completion properties

Qusic:
This PR is part of https://github.com/Valloric/ycmd/pull/137
including (c) (d) (f) (g)

    (a) kind            ==> kind (unchanged) // such as "class", "variable"
    (b) extra_data      ==> extra_data (unchanged) // completer-specific data
    (c) extra_menu_info ==> result_type (renamed) // type of the completion expression, such as "int", "void"
    (d) menu_text       ==> display_string (renamed) // intended to be showed in UI for that completion
    (e) insertion_text  ==> completion_chunks (converted)
    (f) detailed_info   ==> doc_string (merged)
    (g)                 ==> typed_string (added) // type in this string to get this completion. used for filtering, sorting and grouping.

Names such as `menu_text` and `extra_menu_info` are too specific to the vim client and confusing to people who make clients for other editors or new completers for other languages.
Actually `detailed_info` and `doc_string` are both displayed in the same preview window, concatenated. So merging them will also make it clearer. Actually `detailed_info` is removed and I will explain it later.

Another notable change is that duplicate completions are no longer merged in this PR. For example, a function and its overloads are now separate completions. To elaborate it:

1. `detailed_info` used to be lines of function signatures of different overloads and now it is not needed any more. the function signature(or anything that helps to use a completion) can be obtained in `display_string` of each separate completion.
2. `doc_string` used to be the value returned by `clang_getCompletionBriefComment` for the first overload of that function. however, doc strings(if any) for other overloads were lost.
3. a function and its overloads will have the same value for `typed_string` and clients are still able to group completions if necessary.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/327)


Valloric:
Thanks for sending a PR! :)

This is pretty extensive. @micbou @vheon @puremourning Thoughts?

---

Review status: 0 of 25 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/327#-:-K9Ed3JYrYYI72kPYHIB:-761378422)*

vheon:
I absolutely agree that the api needs to rename some field and change other, but I'm still not 100% convinced that those are the right names, (I'm probably overthinking this but, oh well...); 

- `doc_string` is this a "two word" field because is a doc in a form of a string? I mean why just not `doc` or `documentation` or make it `docstring` directly?
- `display_string` should be used only for overloaded function? Or do you see other purposes for this? is not clear so maybe the name is not optimal (?)

---

Review status: 0 of 25 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/327#-:-K9IdlptH8anXUZyB0d9:-1525378048)*

puremourning:
We totally need to include the authors of the other clients in this, and update the example client. I think this would be a great opportunity to document this part of the API too ^_^

I haven't fully digested this change, but went quickly through it. I think there are a couple of changes in behaviour, which I'm not totally sure about - could be that i'm missing something.

---

Review status: 0 of 25 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/go/tests/gocode_completer_test.py, line 115 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/327#-K9Ijk39utJwu4-tJdqf:-K9Ijk39utJwu4-tJdqg:1977981966) ([raw file](https://github.com/valloric/ycmd/blob/6f7fe3b9d7083a74e3f6a75aa5bdc62ae8c1fa5e/ycmd/completers/go/tests/gocode_completer_test.py#L115)):*
here it looks like some useful information was lost.

---

*[ycmd/completers/javascript/tern_completer.py, line 180 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/327#-K9IjOknumOE3Z-LLurO:-K9IjOknumOE3Z-LLurP:-2106000959) ([raw file](https://github.com/valloric/ycmd/blob/6f7fe3b9d7083a74e3f6a75aa5bdc62ae8c1fa5e/ycmd/completers/javascript/tern_completer.py#L180)):*
why change this to not use BuildDoc? In any case, if BuildDoc is no longer required it should be removed (else it is just dead code)

---

*[ycmd/tests/javascript/get_completions_test.py, line 275 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/327#-K9Ike_AIcD_fAiaW-0a:-K9Ike_AIcD_fAiaW-0b:1681387432) ([raw file](https://github.com/valloric/ycmd/blob/6f7fe3b9d7083a74e3f6a75aa5bdc62ae8c1fa5e/ycmd/tests/javascript/get_completions_test.py#L275)):*
why this change in behaviour? inclusion of the signature was deliberate.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/327#-:-K9IktS3XJlvJG1b5l_T:-1197503862)*

vheon:
Let's ping client's authors:

- Vim: done :stuck_out_tongue_closed_eyes: 
- Emacs: @abingham, @ptrv
- Atom: is the author of the PR
- kakuone: @mawww

---

Review status: 0 of 25 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/go/tests/gocode_completer_test.py, line 115 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/327#-K9Ijk39utJwu4-tJdqf:-K9Inadrxj1kQlqREoYh:-1906942592) ([raw file](https://github.com/valloric/ycmd/blob/6f7fe3b9d7083a74e3f6a75aa5bdc62ae8c1fa5e/ycmd/completers/go/tests/gocode_completer_test.py#L115)):*
No information was lost here. Actually there were redundant information before: you can recreate `detailed_info` using the information in other fields. Is just up to the client to choose what and how should the information be displayed.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/327#-:-K9IoN28lVAwqMgEZlKI:536174893)*

Qusic:
@puremourning
* I just kept the old name. https://github.com/Valloric/ycmd/blob/master/cpp/ycm/ClangCompleter/CompletionData.h#L114
* `display_string` is a required field. it will show in the UI for that completion. For example, when a completer provides a snippet completion for a class definition, this field will be helpful because it is not a good idea to display the full code in the completion list view.

---

Review status: 0 of 25 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/go/tests/gocode_completer_test.py, line 115 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/327#-K9Ijk39utJwu4-tJdqf:-K9KoWOCY6qEu9W4NfIK:-1702536500) ([raw file](https://github.com/valloric/ycmd/blob/6f7fe3b9d7083a74e3f6a75aa5bdc62ae8c1fa5e/ycmd/completers/go/tests/gocode_completer_test.py#L115)):*
`detailed_info` is for reference after completion because vim client cannot insert placeholders for parameters. however such information, such as function signatures, is very likely to be in `display_string`.

---

*[ycmd/completers/javascript/tern_completer.py, line 180 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/327#-K9IjOknumOE3Z-LLurO:-K9KqL0OmZFlzGPu3EO0:-1596326720) ([raw file](https://github.com/valloric/ycmd/blob/6f7fe3b9d7083a74e3f6a75aa5bdc62ae8c1fa5e/ycmd/completers/javascript/tern_completer.py#L180)):*
`doc_string` should be purely what it should be. this change make it consistent across all completers.

---

*[ycmd/tests/javascript/get_completions_test.py, line 275 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/327#-K9Ike_AIcD_fAiaW-0a:-K9KqhX3adZD39syCii9:-231769322) ([raw file](https://github.com/valloric/ycmd/blob/6f7fe3b9d7083a74e3f6a75aa5bdc62ae8c1fa5e/ycmd/tests/javascript/get_completions_test.py#L275)):*
This is a test case for tern_completer.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/327#-:-K9Kr-M6zdKb-6Q8quQx:-1246257443)*

ptrv:
I think the API change is really good and makes the completion keywords less vim specific and therefore more clear, what they are for.

With providing the overloaded functions as separate completion entries makes the Emacs client much simpler because there we added code for splitting up the `detailed_info` to get the overloads.

I have already a branch for `emacs-ycmd` where I adapted the new API and it was pretty straight forward.

About the keyword names, I was a bit confused with `typed_string` and `display_string` but I guess when it's documented then it's fine.

I have only one concern, which might not be an issue at all but now we have a lot more data to transfer with the overloaded functions. Could it have impact on performance with large projects?

puremourning:
Review status: 0 of 25 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/javascript/tern_completer.py, line 180 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/327#-K9IjOknumOE3Z-LLurO:-K9Mn7jWPZg0F9V3dZgo:-61679329) ([raw file](https://github.com/valloric/ycmd/blob/6f7fe3b9d7083a74e3f6a75aa5bdc62ae8c1fa5e/ycmd/completers/javascript/tern_completer.py#L180)):*
> `doc_string` should be purely what it should be

I don't understand this response, sorry. I think you have to define what 'should be' means here. The doc string generated here was intentionally including both the signature and the comment, because that is more useful to users that what Tern actually returns, and matches what many other completion engines produce. I still see this as a regression, unless we're saying that the client is responsible for language-specific documentation concatenation. I'm not too sure that's a good thing.

---

*[ycmd/tests/javascript/get_completions_test.py, line 275 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/327#-K9Ike_AIcD_fAiaW-0a:-K9MnhqHzuFSBmwUhH-B:-980351911) ([raw file](https://github.com/valloric/ycmd/blob/6f7fe3b9d7083a74e3f6a75aa5bdc62ae8c1fa5e/ycmd/tests/javascript/get_completions_test.py#L275)):*
I know that. But as mentioned, i see it as a regression, unless there is more to this which i'm missing. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/327)*

vheon:
> About the keyword names, I was a bit confused with typed_string and display_string but I guess when it's documented then it's fine.

If you have suggestion for those we can discuss them :smiley: 

> I have only one concern, which might not be an issue at all but now we have a lot more data to transfer with the overloaded functions. Could it have impact on performance with large projects?

My gut says no, but we don't have benchmark for this in place to measure.

---

Review status: 0 of 25 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/327#-:-K9MoEV6j2HiRAbxBoJu:-1377912819)*

vheon:
Review status: 0 of 25 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/tests/javascript/get_completions_test.py, line 275 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/327#-K9Ike_AIcD_fAiaW-0a:-K9Mpv9qPszX5JuSJbiQ:-32160595) ([raw file](https://github.com/valloric/ycmd/blob/6f7fe3b9d7083a74e3f6a75aa5bdc62ae8c1fa5e/ycmd/tests/javascript/get_completions_test.py#L275)):*
@puremourning what do you mean "a regression"? what have we lost?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/327)*

Qusic:
Review status: 0 of 25 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/javascript/tern_completer.py, line 180 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/327#-K9IjOknumOE3Z-LLurO:-K9NFER2m283KBTa9uiI:2044013538) ([raw file](https://github.com/valloric/ycmd/blob/6f7fe3b9d7083a74e3f6a75aa5bdc62ae8c1fa5e/ycmd/completers/javascript/tern_completer.py#L180)):*
We don't need to define these fields strictly. I think their names should be suggestions on how they should be displayed in the UI. (Of course we should document it later.)
For example, we would have a table list for completions like this:

```
---------------------------------------
| result_type | display_string | kind |
---------------------------------------
| result_type | display_string | kind |
---------------------------------------
| result_type | display_string | kind |
---------------------------------------
| (list footer)                       |
---------------------------------------
```

the footer of the list shows `doc_string` of the highlighted completion item.

Vim client uses a preview window, so an option is to show all of `result_type`, `display_string` and `doc_string` in the window.

In other words, signatures, or something else for the same purpose, are already in other fields and we don't need to repeat them.

BTW, the preview window is mainly for reference of parameters because YCM doesn't support inserting parameter placeholders when completing, which is the final issue I am trying to resolve.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/327)*

Valloric:
This PR seems abandoned, closing.

PR #328, opened 2016-01-29
Allow skipping tests for specific completers

mispencer:
Currently, you can skip tests for the clang completion engine. This pull requests extends this feature for the other completion engines.

The existing filter for clang was based on the test name. I attempted to use this method for the other completions, but not all tests were consistently named with the language or completer engine. So instead of refactoring all those name, I moved the test files so that all the tests for each completion engine were in their own directory, and, using a new nose plugin, excluded tests in said directory. As an additional benefit, directory filtering is less brittle to future changes.

For consistence, I also moved the other tests outside the test folder into the test folder.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/328)


Valloric:
I was just thinking the same thing yesterday! :)

Coveralls is saying that coverage went down; is this expected? 

@micbou You reorganized the tests recently so you probably have some opinions on how this should be done. 

I'm LGTM on this if @micbou is.

---

Review status: 0 of 25 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9ERhQ-Ywewm8I4x7pG:1648274161)*

mispencer:
Coveralls had 100% coverage of ycmd/completers/completer_utils_test.py. After this change, it's not longer listed. This was the only change in coverage. I assume that completer_utils_test.py never should have been listed in the first place, since it's a test file, so I think this is fine.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9F-Na77PxBrlPhTmLt:1116779178)*

puremourning:
I haven't looked at this in too much detail, but a couple things spring to mind:

- I normally just do `./run_tests.py --skip-build -- ycmd.tests.javascript` or something
- we kept the Integration and Module tests separate before. I think they are now merged. Is that right?
- both @vheon  and I have working vagrant files for ycmd tests  presumably the issue is that there is quite a lot of setup required to run everything ?

---

Review status: 0 of 25 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9GrhEo6ytFVSijLByN:1305306511)*

vheon:
> I normally just do ./run_tests.py --skip-build -- ycmd.tests.javascript or something

That is basically what I usually do too.

On the other hand it would be tedious without this change to to run all the tests except for the omnisharp one for example. The question here is that this is a real use case :confused:

micbou:
I am not sure about moving unit tests with integration tests. It makes sense for unit tests to be near the module they are testing and integration tests to be at the root module. I would just move the `ycmd/completers/completer_utils_test.py` to `ycmd/completers/tests/completer_utils_test.py` to be consistent with other tests. It would also remove it from coverage.

Concerning the new options to exclude tests from completers, I don't think they are really useful. With them, we can do (for example):
```sh
./run_tests.py --skip-build --no-javascript-completer
```
instead of
```sh
./run_tests.py --skip-build --exclude-dir=ycmd/tests/javascript
```
Not a big gain.

---

Review status: 0 of 25 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9HvsZtYqvY2eo47OAm:-2135242793)*

mispencer:
My use case is I want to run `./run_tests.py` locally with appropriate arguments to test everything I can without getting superficial errors from tests for completers I don't have installed. Outside CI, this utility is only used by people working on the code. I assumed enough of the people doing this don't have every completer installed that this use cause would be worth while.

`./run_tests.py --skip-build --no-javascript-completer` and `./run_tests.py --skip-build --exclude-dir=ycmd/tests/javascript` are equivalent. But`./run_tests.py --no-omnisharp-completer` and `./run_tests.py  --exclude-dir=ycmd/tests/cs` aren't, since `--no-omnisharp-completer` affects the build. I put in ones that don't affect the build mostly for consistence.

In general, if there's a use case for `--no-clang-completer`, there is a use case of `--no-omnisharp-completer` and all the others. So we should have them all, or remove `--no-clang-completer`.

Regarding the location, the pattern I am used to is having both unit tests and integration tests together, so that is the pattern I have implemented here. Indeed, I didn't realizes there were tests outside the tests directory until I put this PR together. If desired, I guess I can move them back. If so, at least `FindGoCodeBinary_test` should be moved, since that's a integration test. I am unsure whether the other tests in `gocode_completer_test.py` and `filename_completer_test.py` count as integration tests, since they use the file system. I am also unsure what to do about completer-specific unit tests in this case - should they be excluded under `--no-whatever-completer`?

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9IHKcftj4QrhUV-EON:2069385051)*

micbou:
I changed my view on this. I think it is fine to move all the tests in the `tests` root folder. For the new options, I didn't think about the build so they make sense. Just one thing, could they all be named after the language they are targeting (`--no-cs-completer` instead of `--no-omnisharp-completer`, etc.)?

---

Reviewed 37 of 39 files at r1.
Review status: 23 of 25 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[ycmd/tests/go/get_completions_test.py, line 24 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9OF6o6ZQEMHc2rBZEM:-K9OF6o7IPlt6pr-pCj7:-1178104524) ([raw file](https://github.com/valloric/ycmd/blob/9ffd8b0af34ccddde186d3a0353a242fe5a7c337/ycmd/tests/go/get_completions_test.py#L24)):*
I am currently working on a overhaul of the Go completer and one of the change is to rename it `GoCompleter` so let's keep the current name of this class. Other tests will probably be moved here.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9IPVBKH7iPVlMptBz-:-758373575)*

vheon:
What I'm afraid is that more languages we support more option we're going to have here which IMHO could go out of hand. If we do like @micbou suggested:

> Just one thing, could they all be named after the language they are targeting (--no-cs-completer instead of --no-omnisharp-completer, etc.)

then we could make a single option like `--no-completer=<language> <language> ...` and be done with it. What do you think?

---

Review status: 23 of 25 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9OWANFxP_Boei7w4mj:-1931883646)*

micbou:
I think this is an excellent idea. I am just wondering how it will interact with the remaining arguments. I assume adding `--` is mandatory in this case.

---

Review status: 23 of 25 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9OZFvw4JtYIU6XdmCS:-669353606)*

Valloric:
Comma separated might be easier. `--no-completer=cpp,cs,python`

---

Review status: 23 of 25 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9Ob-nfP3HyXuBLnN66:644244663)*

micbou:
`argparse` can handle spaces separated arguments: `--no-completers cpp cs python` and using `--` to pass arguments to nosetests is working. I just tested it.

If we decide to add this parameter, it makes sense to do the same thing for the `build.py` arguments (we would probably need to mark the old parameters as deprecated instead of removing them to not break users).

---

Review status: 23 of 25 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9OzCQCYI3yN96Ycya1:-674838036)*

Valloric:
@micbou I understand that argparse can parse that easily, but for _humans_ I think it's easier to grasp how `--foo=bar,goo` works than `--foo bar goo`. The latter isn't as obvious.  Also, I think the space separated args might interfere with possible future positional arguments (not that I foresee us having any).

I'm not a fan of CLI interfaces taking several params to a flag with space separation. It feels a bit too magic. KISS and all that.

---

Review status: 23 of 25 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9P2scEUWi7NYBHVVow:-336479185)*

Valloric:
Review status: 23 of 25 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[ycmd/tests/go/get_completions_test.py, line 24 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9OF6o6ZQEMHc2rBZEM:-K9P3lJ-zS8B57YoUeJK:2119263656) ([raw file](https://github.com/valloric/ycmd/blob/9ffd8b0af34ccddde186d3a0353a242fe5a7c337/ycmd/tests/go/get_completions_test.py#L24)):*
Along those lines, I'd love it if users never had to see semantic engine names. It should be `--cfamily-completer`, '--cs-completer` etc (old flags should still work, but they would be undocumented aliases). For internal class names I don't care much.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328)*

mispencer:
Ok, I have changed this to take a single parameter of  --no-completers. I also added a --with-completers for specifying only certain completers and deprecated the existing --no-clang-completer.

* Both currently take space delimited argument because that's the only thing that argparser provides by default, so it would take custom code to make it comma delimited and I'm lazy. Let me know if you all feel strongly about this, and I'll try to implement it.

* Both will prefer the language name, but will also taken a list of aliases. Let me know if there are missing aliases. These aliases are not listed in the help output.

* Both have a custom type to do validation. I tried argparse's choices, but it was too long in the help, and I couldn't hide alias (e.g. clang).

I elected to implement the "with" option since locally I only have a few completers dependencies installed, and it's much easier to lists the ones I have installed than not installed. It's also slightly more backward compatible, since it wouldn't fail should more completers be added.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9SxeSDyaV0odRbFCUL:328264577)*

Valloric:
This looks fine to me. You should also change `TESTS.md` since some of the flags it now reccommends are obsolete. With that, :lgtm:.

---

Review status: 22 of 24 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[run_tests.py, line 34 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9TGSx90QMqAaUwEzJ7:-K9TGSx90QMqAaUwEzJ8:-151249294) ([raw file](https://github.com/valloric/ycmd/blob/f3d6e9b9569e6fbbba948b0c76dc252887d7f096/run_tests.py#L34)):*
Maybe make this all caps?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9TGse2C4ZKrEdkIyJj:-2055849839)*

mispencer:
Test document has been updated.

---

Review status: 22 of 25 files reviewed at latest revision, 2 unresolved discussions.

---

*[run_tests.py, line 34 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9TGSx90QMqAaUwEzJ7:-K9TLqZxWihvrtXp4DPk:2136771930) ([raw file](https://github.com/valloric/ycmd/blob/f3d6e9b9569e6fbbba948b0c76dc252887d7f096/run_tests.py#L34)):*
Changed to all caps.

---

*[ycmd/tests/go/get_completions_test.py, line 24 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9OF6o6ZQEMHc2rBZEM:-K9SxYSvXBOAu4FqMQ3H:-1763544199) ([raw file](https://github.com/valloric/ycmd/blob/9ffd8b0af34ccddde186d3a0353a242fe5a7c337/ycmd/tests/go/get_completions_test.py#L24)):*
it's been reverted to back to  Go_GetCompletions_test.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9TLuUWDuSK_Qmrymt2:1120424965)*

Valloric:
Reviewed 1 of 39 files at r1, 1 of 2 files at r2, 2 of 2 files at r3.

---

*[TESTS.md, line 38 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9TYCTr4oO_qvfj_T2K:-K9TYCTr4oO_qvfj_T2L:828342697) ([raw file](https://github.com/valloric/ycmd/blob/17cd43df7845a573272cabd65335081d35fb9a2d/TESTS.md#L38)):*
"can able" typo

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328)*

vheon:
Reviewed 37 of 39 files at r1, 1 of 2 files at r2, 2 of 2 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328)*

vheon:
except the typo :lgtm: too

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9TlrezzNbt028PGg7a:-232293205)*

mispencer:
I have corrected the grammar issues.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9Tox-OFesAjbaQClX5:-1704080263)*

micbou:
I find aliases a little overkill. Do we really need them?

---

Reviewed 1 of 39 files at r1, 1 of 2 files at r2, 2 of 2 files at r4.

---

*[run_tests.py, line 38 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9TmujVRcFxZtuB-8_q:-K9TmujVRcFxZtuB-8_r:1296422492) ([raw file](https://github.com/valloric/ycmd/blob/17cd43df7845a573272cabd65335081d35fb9a2d/run_tests.py#L38)):*
Last comma not needed.

---

*[run_tests.py, line 48 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9Tmk1x1RnORjGGzJ4u:-K9Tmk1yUIIIwN5ditZ2:-161860489) ([raw file](https://github.com/valloric/ycmd/blob/17cd43df7845a573272cabd65335081d35fb9a2d/run_tests.py#L48)):*
`tern` instead of `term`

---

*[run_tests.py, line 97 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9Tm8YFUq1cfeXSScqz:-K9Tm8YGY3hAIPujKIqY:-582735495) ([raw file](https://github.com/valloric/ycmd/blob/17cd43df7845a573272cabd65335081d35fb9a2d/run_tests.py#L97)):*
I would have named this option simply `--completers` (or `--no-completers` should be named `--without-completers`).

---

*[run_tests.py, line 120 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9TmdPIpr6-b5TZCNH4:-K9TmdPIpr6-b5TZCNH5:1971039587) ([raw file](https://github.com/valloric/ycmd/blob/17cd43df7845a573272cabd65335081d35fb9a2d/run_tests.py#L120)):*
Over 80 characters.

---

*[TESTS.md, line 55 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9TnJ-w3FWzHpo00RwD:-K9TnJ-yn0pxMeXndB9w:1478678277) ([raw file](https://github.com/valloric/ycmd/blob/17cd43df7845a573272cabd65335081d35fb9a2d/TESTS.md#L55)):*
Double `the`

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9TnrrwRsCXtG22Sow_:-612431121)*

vheon:
> I find aliases a little overkill. Do we really need them?

How can we keep the old options working without the aliases?

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9Tt9SUT0pHEHLiecuy:-1276234601)*

micbou:
Which old options? There was no `--no-*-completer` options except `--no-clang-completer` for `run_tests.py`. And the new options are in a different tormat anyway: `--no-completers <completer> <...> <completer>`

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9Tu-49EBGLCKj8lGZk:-723549622)*

mispencer:
Should I remove the aliases or keep them?

---

Review status: 23 of 25 files reviewed at latest revision, 7 unresolved discussions.

---

*[run_tests.py, line 48 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9Tmk1x1RnORjGGzJ4u:-K9UM0WiL8XuB6mQ4un-:-1888472227) ([raw file](https://github.com/valloric/ycmd/blob/17cd43df7845a573272cabd65335081d35fb9a2d/run_tests.py#L48)):*
term => tern

---

*[run_tests.py, line 97 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9Tm8YFUq1cfeXSScqz:-K9ULuwbwvz72Q-baqXE:-2116502727) ([raw file](https://github.com/valloric/ycmd/blob/17cd43df7845a573272cabd65335081d35fb9a2d/run_tests.py#L97)):*
Changed `--with-completers` to `--completers`

---

*[run_tests.py, line 120 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9TmdPIpr6-b5TZCNH4:-K9ULnMZy2jPasW9HLnx:-782799887) ([raw file](https://github.com/valloric/ycmd/blob/17cd43df7845a573272cabd65335081d35fb9a2d/run_tests.py#L120)):*
Broke up this line.

---

*[TESTS.md, line 38 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9TYCTr4oO_qvfj_T2K:-K9ULqFuQgK2XXNn43ZE:-2127133785) ([raw file](https://github.com/valloric/ycmd/blob/17cd43df7845a573272cabd65335081d35fb9a2d/TESTS.md#L38)):*
Resolved typo.

---

*[TESTS.md, line 55 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9TnJ-w3FWzHpo00RwD:-K9ULsrUX2ytPjMYVCez:-2127133785) ([raw file](https://github.com/valloric/ycmd/blob/17cd43df7845a573272cabd65335081d35fb9a2d/TESTS.md#L55)):*
Resolved typo.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9UM34GHH57uqF_ddvI:-1434582537)*

Valloric:
Let's keep the aliases. `run_tests.py` is only for dev, so I don't care about keeping the old flags supported. If this were changing `build.py`, that would be a different thing altogether.

---

Reviewed 2 of 2 files at r5.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9UfEoJsXjS2UntPlHC:-1615426909)*

Valloric:

---

*[run_tests.py, line 114 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9ePc-2sjT_QZP500QT:-K9ePc-2sjT_QZP500QU:883524157) ([raw file](https://github.com/valloric/ycmd/blob/8e8b0c8a00e3e06e86740aa7f49486e99464d564/run_tests.py#L114)):*
Looking at this a bit more, could you pull out the logic that builds `completers` into a separate function?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328)*

mispencer:
Merge conflicts resolved.

---

Review status: 17 of 25 files reviewed at latest revision, 2 unresolved discussions.

---

*[run_tests.py, line 114 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/328#-K9ePc-2sjT_QZP500QT:-K9mk373FkARdMbj3o4-:-35541377) ([raw file](https://github.com/valloric/ycmd/blob/8e8b0c8a00e3e06e86740aa7f49486e99464d564/run_tests.py#L114)):*
`completer` building has been refactored out.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9mkElRwrdn36ZJojwM:-1428521142)*

vheon:
Reviewed 8 of 8 files at r6.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328)*

Valloric:
Awesome work, thanks for the PR! :)

@homu r+

---

Reviewed 24 of 39 files at r1, 8 of 8 files at r6.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/328#-:-K9nJd9K97JV14sWnLqF:-481847120)*

PR #332, opened 2016-01-31
[READY] Adding new python-future dependency

Valloric:
We'll use this for Python 3 compatibility. Also changed responses.py
using futurize so that it uses a module from python-future to ensure the
sys.path module pickup is working.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/332)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,3 +25,6 @@
 [submodule "third_party/racerd"]
 	path = third_party/racerd
 	url = https://github.com/jwilm/racerd
+[submodule "third_party/python-future"]
+	path = third_party/python-future
+	url = https://github.com/PythonCharmers/python-future
--- a/run_tests.py
+++ b/run_tests.py
@@ -10,6 +10,8 @@ DIR_OF_THIRD_PARTY = p.join( DIR_OF_THIS_SCRIPT, 'third_party' )
 
 python_path = []
 for folder in os.listdir( DIR_OF_THIRD_PARTY ):
+  if folder == 'python-future':
+    folder = os.path.join( folder, 'src' )
   python_path.append( p.abspath( p.join( DIR_OF_THIRD_PARTY, folder ) ) )
 if os.environ.get( 'PYTHONPATH' ) is not None:
   python_path.append( os.environ['PYTHONPATH'] )
--- a/ycmd/responses.py
+++ b/ycmd/responses.py
@@ -17,6 +17,10 @@
 # You should have received a copy of the GNU General Public License
 # along with YouCompleteMe.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import ( absolute_import, division,
+                         print_function, unicode_literals )
+from builtins import *  # flake8: noqa
+
 import os
 
 YCM_EXTRA_CONF_FILENAME = '.ycm_extra_conf.py'
@@ -135,7 +139,7 @@ def BuildRangeData( source_range ):
   }
 
 
-class Diagnostic:
+class Diagnostic( object ):
   def __init__ ( self, ranges, location, location_extent, text, kind ):
     self.ranges_ = ranges
     self.location_ = location
@@ -144,7 +148,7 @@ class Diagnostic:
     self.kind_ = kind
 
 
-class FixIt:
+class FixIt( object ):
   """A set of replacements (of type FixItChunk) to be applied to fix a single
   diagnostic. This can be used for any type of refactoring command, not just
   quick fixes. The individual chunks may span multiple files."""
@@ -155,7 +159,7 @@ class FixIt:
     self.chunks = chunks
 
 
-class FixItChunk:
+class FixItChunk( object ):
   """An individual replacement within a FixIt (aka Refactor)"""
 
   def __init__ ( self, replacement_text, range ):
@@ -164,7 +168,7 @@ class FixItChunk:
     self.range = range
 
 
-class Range:
+class Range( object ):
   """Source code range relating to a diagnostic or FixIt (aka Refactor)."""
 
   def __init__ ( self, start, end ):
@@ -173,7 +177,7 @@ class Range:
     self.end_ = end
 
 
-class Location:
+class Location( object ):
   """Source code location for a diagnostic or FixIt (aka Refactor)."""
 
   def __init__ ( self, line, column, filename ):
@@ -185,7 +189,6 @@ class Location:
 
 
 def BuildDiagnosticData( diagnostic ):
-
   kind = ( diagnostic.kind_.name if hasattr( diagnostic.kind_, 'name' )
            else diagnostic.kind_ )
 
--- a/ycmd/utils.py
+++ b/ycmd/utils.py
@@ -213,7 +213,11 @@ def AddNearestThirdPartyFoldersToSysPath( filepath ):
     raise RuntimeError(
         'No third_party folder found for: {0}'.format( filepath ) )
 
+  # NOTE: Any hacks for loading modules that can't be imported without custom
+  # logic need to be reproduced in run_tests.py as well.
   for folder in os.listdir( path_to_third_party ):
+    if folder == 'python-future':
+      folder = os.path.join( folder, 'src' )
     sys.path.insert( 0, os.path.realpath( os.path.join( path_to_third_party,
                                                         folder ) ) )
 


Valloric:
Grr, another flaky failure. This time with omnisharp returning a 500.

micbou:
Reviewed 5 of 5 files at r1.

---

*[ycmd/responses.py, line 21 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/332#-K9RUfonNcOBp5d2ePcI:-K9RUfosABOBUfffdgwi:-1156397311) ([raw file](https://github.com/valloric/ycmd/blob/76f33275f2fa140908c015fe56fc8026b8c7ca9a/ycmd/responses.py#L21)):*
I don't understand why this is needed. Only the `os` module is imported. There is no division, no print, and the strings does not contain unicode characters.

---

*[ycmd/responses.py, line 22 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/332#-K9RUhWDFJVXu-mHZxvR:-K9RUhWDFJVXu-mHZxvS:790153352) ([raw file](https://github.com/valloric/ycmd/blob/76f33275f2fa140908c015fe56fc8026b8c7ca9a/ycmd/responses.py#L22)):*
We are only using the `str` and `object` builtins so we should explicitly import them.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/332)*

Valloric:

---

*[ycmd/responses.py, line 21 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/332#-K9RUfonNcOBp5d2ePcI:-K9StKMr1_cVdH3BXYtD:-114121550) ([raw file](https://github.com/valloric/ycmd/blob/76f33275f2fa140908c015fe56fc8026b8c7ca9a/ycmd/responses.py#L21)):*
Ha, I knew this will be controversial. :)

I was thinking we might want to always have the `__future__` imports up top along with all the builtins imported as well. This way we can just always write Python 3 without concern.

I wouldn't do this for anything but the Python 3 compatibility layer.

I'm worried that if we only import the parts of the shim layer that are used in the file, someone's going to change something in it which would work on one version of Python but break on another, and if we're lucky Travis would catch it. It's one thing to remember "I have to import the `os` module to use `os.foo`" and quite another to realize one must import `division` from `__future__` before dividing any numbers. And if you forget, it's going to work on _one_ version of Python, the one you're developing on.

So I think the shim layer might be a special case; what do the rest of you think? We need to decide how we're going to be doing this codebase-wide.

@puremourning @vheon 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/332)*

vheon:
Reviewed 5 of 5 files at r1.

---

*[ycmd/responses.py, line 21 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/332#-K9RUfonNcOBp5d2ePcI:-K9ThOUeiAHL1pmI189f:-1018150732) ([raw file](https://github.com/valloric/ycmd/blob/76f33275f2fa140908c015fe56fc8026b8c7ca9a/ycmd/responses.py#L21)):*
I was thinking about this today when I read the page for writing Python 3 from scratch using `python-future`. I see two option:

- put this on top of the file that needs Python 3 compatibility
- use only the needed imports and put in place `pylint` to note when imports are needed.

---

*[ycmd/responses.py, line 22 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/332#-K9RUhWDFJVXu-mHZxvR:-K9Tk44igc-oeAYDAeLt:1539202633) ([raw file](https://github.com/valloric/ycmd/blob/76f33275f2fa140908c015fe56fc8026b8c7ca9a/ycmd/responses.py#L22)):*
is the same as above

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/332)*

Valloric:

---

*[ycmd/responses.py, line 21 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/332#-K9RUfonNcOBp5d2ePcI:-K9TtXlnI7O8We8Pwjh1:-54446610) ([raw file](https://github.com/valloric/ycmd/blob/76f33275f2fa140908c015fe56fc8026b8c7ca9a/ycmd/responses.py#L21)):*
I'm not sure would pylint catch the issue. For instance, you can use `dict` in both Python 2 & 3 and call `items()` on both, but on py2 it's going to create copies. So it will work, just slower. `from builtins import *` will import their `dict` on py2 which is a dict subclass that calls `iteritems` on py2 (among other things) thus making behavior consistent across versions.

And no test would catch a missing dict import since everything would still work (just slower on py2). I really think we need to import everything in all files.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/332)*

vheon:

---

*[ycmd/responses.py, line 21 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/332#-K9RUfonNcOBp5d2ePcI:-K9TvIcFlpDLz0lmAqp4:79069409) ([raw file](https://github.com/valloric/ycmd/blob/76f33275f2fa140908c015fe56fc8026b8c7ca9a/ycmd/responses.py#L21)):*
I'm ok with using all imports from `python-future` regardless of the uses.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/332)*

micbou:

---

*[ycmd/responses.py, line 21 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/332#-K9RUfonNcOBp5d2ePcI:-K9Tvp1YE-ZVXAnoBR64:-318522694) ([raw file](https://github.com/valloric/ycmd/blob/76f33275f2fa140908c015fe56fc8026b8c7ca9a/ycmd/responses.py#L21)):*
It looks like this is [the recommended way](http://python-future.org/quickstart.html#if-you-are-writing-code-from-scratch). I am fine with it then. Maybe we should add a comment? 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/332)*

vheon:

---

*[ycmd/responses.py, line 21 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/332#-K9RUfonNcOBp5d2ePcI:-K9TyvsatqgHdB2qZn8F:-1136332556) ([raw file](https://github.com/valloric/ycmd/blob/76f33275f2fa140908c015fe56fc8026b8c7ca9a/ycmd/responses.py#L21)):*
Yeah that was the page I was referring to. Do you want to put a comment on every file?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/332)*

micbou:
*[ycmd/responses.py, line 21 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/332#-K9RUfonNcOBp5d2ePcI:-K9TzjBD0JPfD56wz1_g:-1593489991) ([raw file](https://github.com/valloric/ycmd/blob/76f33275f2fa140908c015fe56fc8026b8c7ca9a/ycmd/responses.py#L21)):*
At least, separate it from other imports.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/332)*

Valloric:
Review status: 4 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/responses.py, line 21 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/332#-K9RUfonNcOBp5d2ePcI:-K9U-e1POQERGLL8HLS8:-449730372) ([raw file](https://github.com/valloric/ycmd/blob/76f33275f2fa140908c015fe56fc8026b8c7ca9a/ycmd/responses.py#L21)):*
Separating from other imports with an empty line is fine (and done); adding a comment would be a good idea in general, but probably not in this instance since we'd then have to have it in every single file.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/332)*

micbou:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/332#-:-K9U1Za7Z-RveF-lm5Vi:1417088721)*

vheon:
:lgtm: @homu r+

---

Reviewed 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/332#-:-K9U3e26c_5kxcwKYEk-:-141632775)*

puremourning:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[ycmd/responses.py, line 21 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/332#-K9RUfonNcOBp5d2ePcI:-K9U4S_IcRnLjluDOmRq:1917020107) ([raw file](https://github.com/valloric/ycmd/blob/76f33275f2fa140908c015fe56fc8026b8c7ca9a/ycmd/responses.py#L21)):*
For the record, I think the always import so you don't have to remember is fine. I believe that is what one of the many magic port scripts does too 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/332)*

Valloric:
Travis is clogged a bit and enough rows have finished successfully for me to consider it safe to merge this sans-homu.

PR #334, opened 2016-02-01
[READY] Adding vagrant setup for ycmd development

### LINKED ISSUE #273: Create Vagrantfile to ease ramp up and full testing
Testing all ycmd is not an easy to do due to the amount of dependency we have:

- python (with headers and all)
- compiler
- cmake
- Omnisharp
  - mono
- Tern.js
 - node
 - npm
- gocode
- typescript

And I'm sure I'm forgetting something... So for a new contributor who want to add a feature and want to run all the tests a Vagrantfile with all the dependency baked in would be nice.

Valloric:
Should make our lives easier, especially since we'll want to easily switch between Python 2 and 3 for local dev and test running.

Fixes #273.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/334)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,6 +94,10 @@ Creating good pull requests
     sometimes what you want can be done in a different way if the reason for the
     change is known. _What goal is your change trying to accomplish?_
 
+You might also want to easily set up [an environment for ycmd
+development][dev-setup].
+
 [build-bots]: https://travis-ci.org/Valloric/ycmd
 [ycmd-users]: https://groups.google.com/forum/?hl=en#!forum/ycmd-users
 [ycmd-tests]: https://github.com/Valloric/ycmd/blob/master/TESTS.md
+[dev-setup]: https://github.com/Valloric/ycmd/blob/master/DEV_SETUP.md
--- a/DEV_SETUP.md
+++ b/DEV_SETUP.md
@@ -1,0 +1,20 @@
+# Setting up for ycmd development
+
+1. Install [Vagrant][].
+2. `cd` into the folder where you checked out ycmd.
+3. `$ vagrant up && vagrant ssh`. This will take a while because the VM is being
+	 built and set up. Only needs to happen once though.
+4. You are now in the VM. Run the tests with `$ ./run_tests.sh`.
+5. Hack away. When done, exit the ssh connection with `exit`.
+6. `$ vagrant suspend` so that you can quickly get back to hacking later.
+7. Later on: `$ vagrant resume && vagrant ssh`. This will be _much_ faster.
+
+That's it!
+
+You can switch between Python versions with `pyenv global 2.6.6` and `pyenv
+global 3.3.0`.
+
+If you ever feel like you've screwed up the VM, just kill it with
+`vagrant destroy` and then run `vagrant up` again to get to a clean state.
+
+[vagrant]: https://www.vagrantup.com/
--- a/TESTS.md
+++ b/TESTS.md
@@ -2,6 +2,11 @@
 
 This readme documents instructions on running the test suite.
 
+The easiest way to run the ycmd test suite locally is to [set up your ycmd dev
+environment using Vagrant][dev-setup] which will have everything installed and
+ready to go. Reading this file is still useful because you'll learn how to skip
+the build or parts of the test suite.
+
 An alternative (canonical) reference is the scripts used for running the tests
 on Travis CI. These can be found in `.travis.yml` and `./travis` directory.
 
@@ -105,3 +110,5 @@ Likely to be a problem with the OmniSharpServer.
     FAIL: ycmd.completers.general.tests.filename_completer_test.FilenameCompleter_test.SystemPathCompletion_test
 
 Ensure that you have UTF-8 support in your environment (see above)
+
+[dev-setup]: https://github.com/Valloric/ycmd/blob/master/DEV_SETUP.md
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,22 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
+  # TODO: update to xenial64 when that comes out
+  config.vm.box = "ubuntu/trusty64"
+
+  # On startup, run our bootstrap script to setup the VM
+  config.vm.provision :shell, :path => "vagrant_bootstrap.sh"
+
+  config.vm.provider "virtualbox" do |v|
+    # MAGIC for faster guest networking
+    v.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
+    v.customize ["modifyvm", :id, "--natdnsproxy1", "on"]
+    v.customize ["modifyvm", :id, "--nictype1", "virtio"]
+
+    # We need quite a bit of memory to compile more than one ycmd C++ file at a
+    # time.
+    v.memory = 2048
+    v.cpus = 2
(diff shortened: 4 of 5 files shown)


Valloric:
And _another_ flaky failure, this time JediHTTP timing out. Hm, might be something to do with Travis.

vheon:
We should add that if fixes https://github.com/Valloric/ycmd/issues/273

> And another flaky failure, this time JediHTTP timing out. Hm, might be something to do with Travis.

I really doubt that JediHTTP would have problem to be spawned and to be ready to serve so I suspect that Travis have problem spawning other servers? :confused: 

---

Reviewed 4 of 5 files at r1, 1 of 1 files at r2.

---

*[Vagrantfile, line 7 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/334#-K9TayfxeaCT_1kpV_ca:-K9TayfxeaCT_1kpV_cb:-1738917435) ([raw file](https://github.com/valloric/ycmd/blob/e514ab01476eb892b084124bc5df43bc77e97dc2/Vagrantfile#L7)):*
I assume that the default method for synchronize is to use the shared folder of Vagrant, am I correct? Then one downside of this is that if for example I'm on OSX then after running the tests I would have the compiled library for ubuntu. How would I overcome this? (Sorry for the comment on a blank like)

---

*[Vagrantfile, line 12 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/334#-K9Ta8VSrlASgLmnYJn7:-K9Ta8VSrlASgLmnYJn8:167437215) ([raw file](https://github.com/valloric/ycmd/blob/e514ab01476eb892b084124bc5df43bc77e97dc2/Vagrantfile#L12)):*
Question: where does this magic come from? :stuck_out_tongue_closed_eyes: 

---

*[Vagrantfile, line 20 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/334#-K9TaHIqiHTFm5qCt9E8:-K9TaHIqiHTFm5qCt9E9:1772923896) ([raw file](https://github.com/valloric/ycmd/blob/e514ab01476eb892b084124bc5df43bc77e97dc2/Vagrantfile#L20)):*
In my local Vagrantfile I used only `1024` for `memory` and a swap file but I think is reasonable for today standards to ask for 2Gb of ram and 2 cpu...

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/334#-:-K9T_TSJ4ZcWyj6OcLYC:1929296915)*

Valloric:
Yeah I don't think the issue was with JediHTTP itself, but either Travis or the way we've written our tests.

Added that this fixes #273.

---

---

*[Vagrantfile, line 7 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/334#-K9TayfxeaCT_1kpV_ca:-K9TrYVo9was5T_5WTgj:-579405885) ([raw file](https://github.com/valloric/ycmd/blob/e514ab01476eb892b084124bc5df43bc77e97dc2/Vagrantfile#L7)):*
I had the exact problem today and I realize the easiest fix is to develop in folder that I'm not loading by default in vim. That's it really.

---

*[Vagrantfile, line 12 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/334#-K9Ta8VSrlASgLmnYJn7:-K9Trl4k-0-5ToRDPIWw:-63938843) ([raw file](https://github.com/valloric/ycmd/blob/e514ab01476eb892b084124bc5df43bc77e97dc2/Vagrantfile#L12)):*
Random stackoverflow posts. They didn't go into details on why it worked otherwise I would have linked them here. All I know is that these lines made the VM have the same internet connection speed as the host instead of 10x slower. Apparently it's a common issue.

---

*[Vagrantfile, line 20 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/334#-K9TaHIqiHTFm5qCt9E8:-K9Tt7Vn_NDkVXZZfSuy:1775589054) ([raw file](https://github.com/valloric/ycmd/blob/e514ab01476eb892b084124bc5df43bc77e97dc2/Vagrantfile#L20)):*
I agree.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/334#-:-K9TrMbHWvDsFo1CXt3-:1103501944)*

vheon:

---

*[Vagrantfile, line 7 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/334#-K9TayfxeaCT_1kpV_ca:-K9TtuKLrzpsrVj6t4AG:-295890209) ([raw file](https://github.com/valloric/ycmd/blob/e514ab01476eb892b084124bc5df43bc77e97dc2/Vagrantfile#L7)):*
I used the `rsync` method on mine, this way the sync can be configured to be only one way and exclude certain file. The downside is that on windows you have to download rsync or use cygwin to install it. But honestly I don't know which is better :confused: 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/334)*

Valloric:

---

*[Vagrantfile, line 7 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/334#-K9TayfxeaCT_1kpV_ca:-K9Uo2OhVutRkFTO2OHx:1467675240) ([raw file](https://github.com/valloric/ycmd/blob/e514ab01476eb892b084124bc5df43bc77e97dc2/Vagrantfile#L7)):*
I'm not sure how you would use rsync here. Either way, seems like overkill. Let's keep things simple, at least for now.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/334)*

Valloric:
Note that the Travis failure is caused by flakiness (and this doesn't change any code, so can't really break anything).

---

Review status: 4 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/334#-:-K9VdGdcexbavftu93xE:-999208388)*

vheon:
:lgtm: 

---

Reviewed 1 of 1 files at r3.

---

*[Vagrantfile, line 7 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/334#-K9TayfxeaCT_1kpV_ca:-K9YWYE_qPnl_8uQN-uR:813065473) ([raw file](https://github.com/valloric/ycmd/blob/e514ab01476eb892b084124bc5df43bc77e97dc2/Vagrantfile#L7)):*
Vagrant has native support for rsync so I had this in my Vagrantfile

```ruby
  config.vm.synced_folder ".", "/home/vagrant/ycmd", type: "rsync",
    # default value but without the "--delete" argument
    # so the guest can keep the libclang archive and builded binaries
    rsync__args: [ "--verbose", "--archive", "-z", "--copy-links" ],
    rsync__exclude: [ ".git/", "ycm_core.so", "ycm_client_support.so", "libclang.dylib" ]
```

Naturally it comes with a downside which is that you have to `vagrant rsync` before re-running the test to sync the code. But I agree that the default sync folder is straightforward and less complex.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/334#-:-K9YXHM3NXGXeOozvk_l:980077423)*

Valloric:
Thanks for the review! 

@homu r+

puremourning:
flake. @homu retry

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/334#-:-K9Yuz0vsDG76i_4EEdU:-1288271755)*

PR #335, opened 2016-02-02
[READY] Pruning the size of the Travis test matrix

Valloric:
We used to have 9 (5 Linux, 4 Mac) rows in our test matrix; these
changes should bring it down to 6 (3 Linux, 3 Mac).

First, we cut out clang on/off for Python 2.7. There's already
two rows for clang on/off on Python 2.6, and testing the same
on/off setup with 2.7 seems wasteful. We now just test 2.7 with clang
on. This is two rows down (Linux & Mac each).

Another row killed is Linux-clang-py2.7, because we also have
Linux-clang-py2.7-coverage. That one runs all the tests and also runs
coverage, so contains the row we're killing.

All together, this should speed up our Travis runs.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/335)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,14 +23,17 @@ env:
     - YCM_CORES=1
   matrix:
     - USE_CLANG_COMPLETER=true YCMD_PYTHON_VERSION=2.7
-    - USE_CLANG_COMPLETER=false YCMD_PYTHON_VERSION=2.7
-    - USE_CLANG_COMPLETER=true YCMD_PYTHON_VERSION=2.6
-    - USE_CLANG_COMPLETER=false YCMD_PYTHON_VERSION=2.6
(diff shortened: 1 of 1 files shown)


micbou:
We'll probably need to reduce the number of runs on AppVeyor too. There will be too much tests with Python 3.

:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/335#-:-K9UKEv9cGqSdy3zK831:1799724130)*

Valloric:
That's the reason why I'm cutting this down now. :) I have further ideas for speeding up Travis. Stay tuned.

@homu r+

Valloric:
Travis is a bit clogged up with our runs right now so everything is taking forever; this is a safe change so I'm just going to merge it sans-homu.

PR #346, opened 2016-02-05
[READY] Remove interpreter line from Python sources

micbou:
We only need the shebang line for Python scripts (`build.py`, `run_tests.py`, and `example_client.py`).

Also, [use `coding: utf-8` for Python encoding instead of `-*- coding: utf-8 -*-`](https://www.python.org/dev/peps/pep-0263/), which is specific to Emacs.

I'll update the copyrights next.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/346)

### DIFF
--- a/examples/.ycm_extra_conf.py
+++ b/examples/.ycm_extra_conf.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
-#
 # Copyright (C) 2014  Google Inc.
 #
 # This file is part of YouCompleteMe.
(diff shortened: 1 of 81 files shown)


Valloric:
Nice, thanks! 

@homu r+

Valloric:
@micbou I just noticed that this PR also removed the shebang line in `example_client.py`, which breaks it. :)

Valloric:
We may also want to leave the shebang in `__main__.py` as well.

micbou:
Yes, I messed up. I'll send a PR.

> We may also want to leave the shebang in __main__.py as well.

I am not sure why. `ycmd` is started by doing `python ycmd/`, not `./ycmd/__main__.py`.

PR #352, opened 2016-02-06
[READY] Add --all option to build.py

puremourning:
I find this useful when developing, as I always enable all completers.

```
./install.py --all
```

rather than

```
./install.py --clang-completer --omnisharp-completer --gocode-completer --racer-completer
```

If you guys think this is also useful, we can merge it and I will update YCM's README. If not, then I'm not precious about this - i'll just maintain it in my fork :)

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/352)

### DIFF
--- a/build.py
+++ b/build.py
@@ -204,6 +204,10 @@ def ParseArguments():
   parser.add_argument( '--tern-completer',
                        action = 'store_true',
                        help   = 'Enable tern javascript completer' ),
+  parser.add_argument( '--all',
+                       action = 'store_true',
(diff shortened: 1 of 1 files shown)


Valloric:
Makes complete sense! :lgtm: 

@homu r+

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/352#-:-K9rz_WOPG61Hemv4gvn:-856299176)*

PR #353, opened 2016-02-07
[READY] Bump Clang version to 3.8.0

micbou:
Follow up to PR #297. Currently updated to rc2.

Good news is that Mac OS X binaries are available.

Fix Clang bug with `auto`. Tests updated to reflect this.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/353)


Valloric:
If rc2 is out, that means full 3.8.0 should be out soon as well. Any specific reason why we should use rc2 now instead of wait for full release? That being said, :lgtm:.

Happy to see the auto bug fixed!

---

Review status: 0 of 68 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353#-:-K9wz3BwHWOURWULQyT3:654931892)*

micbou:
It's in preparation of the full release. This is why I put the `NOT READY` tag.

---

Review status: 0 of 68 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353#-:-K9x4Wry1NTgHV-hSs6w:1669457773)*

Valloric:
> It's in preparation of the full release. This is why I put the NOT READY tag.

Ah, that makes complete sense. :)

---

Review status: 0 of 68 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353#-:-K9x9lRcsCPFWPS87u5v:2077243106)*

micbou:
Clang 3.8.0 is out. Updated the PR.

---

Review status: 0 of 68 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353#-:-KCMlC1z2-D5rKNHanCU:1340966841)*

vheon:
:lgtm: 

---

Reviewed 66 of 68 files at r1, 2 of 2 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353#-:-KCMvxaXD4tV2Z1LW1Us:980077423)*

Valloric:
:lgtm: with minor comment.

---

---

*[cpp/llvm/include/clang-c/Index.h, line 1211 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCNd9TTaPJdIatD38H5:-KCNd9TTaPJdIatD38H6:-95689428) ([raw file](https://github.com/valloric/ycmd/blob/c66213da2951173f5882f0155700d84901fc5708/cpp/llvm/include/clang-c/Index.h#L1211)):*
We should now use the enum name in the C++ code instead of the raw constant.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353#-:-KCNdJVTZE-pGxoi08SA:-1120555899)*

micbou:

---

*[cpp/llvm/include/clang-c/Index.h, line 1211 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCNd9TTaPJdIatD38H5:-KCNmrk4n3McqHfQIAXU:7553213) ([raw file](https://github.com/valloric/ycmd/blob/c66213da2951173f5882f0155700d84901fc5708/cpp/llvm/include/clang-c/Index.h#L1211)):*
Are we dropping support for Clang 3.7? If so, there are two other places to update.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353)*

Valloric:

---

*[cpp/llvm/include/clang-c/Index.h, line 1211 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCNd9TTaPJdIatD38H5:-KCNxMFC_vHuL0UhhYMJ:698356246) ([raw file](https://github.com/valloric/ycmd/blob/c66213da2951173f5882f0155700d84901fc5708/cpp/llvm/include/clang-c/Index.h#L1211)):*
Yes. As soon as we switch to a new Clang in CMakeLists.txt, we stop supporting the old version. That has always been the case.

We might want to bump the core version number as well, if necessary.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353)*

micbou:
Updated to drop Clang 3.7 support.

---

Reviewed 1 of 68 files at r1, 1 of 1 files at r4, 2 of 3 files at r5, 1 of 1 files at r6, 1 of 1 files at r7.

---

*[cpp/llvm/include/clang-c/Index.h, line 1211 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCNd9TTaPJdIatD38H5:-KCbtTELeBXnTxo_2BYp:1454474153) ([raw file](https://github.com/valloric/ycmd/blob/c66213da2951173f5882f0155700d84901fc5708/cpp/llvm/include/clang-c/Index.h#L1211)):*
Ok. We'll need to update the README after merging this PR to say that we only support Clang 3.8+.

> We might want to bump the core version number as well, if necessary.

No changes to the API so it should not be needed.

---

*[cpp/ycm/ClangCompleter/ClangCompleter.cpp, line 92 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCbsynws1Z006Wgc-A8:-KCbsynxvLBCqswnOo0N:-1973283545) ([raw file](https://github.com/valloric/ycmd/blob/1521e468db74936e23a418c026605439df5267b5/cpp/ycm/ClangCompleter/ClangCompleter.cpp#L92)):*
Since we don't reparse in the TU constructor anymore, we now always reparse here. This was needed to fix the diagnostic tests. `LatestDiagnostics` method was only used here so I completely removed it.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353#-:-KCbuUH8Tl7D8Zh2NHc7:1344550984)*

Valloric:

---

*[cpp/llvm/include/clang-c/Index.h, line 1211 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCNd9TTaPJdIatD38H5:-KCc47rLfDqgGUfqUYmx:505874899) ([raw file](https://github.com/valloric/ycmd/blob/c66213da2951173f5882f0155700d84901fc5708/cpp/llvm/include/clang-c/Index.h#L1211)):*
> No changes to the API so it should not be needed.

Ah, but API changes aren't the only reason why we bump the version number. We often do it when significant features are added to the C++ layer. In general, the logic for bumping the version number is "do we want users to re-run install.py?" And I think the answer is yes here, because we want people to pick up a newer libclang which is harder/better/faster/stronger.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353)*

Valloric:

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 114 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCc5lDCb4ExaGkKWNDw:-KCc5lDDgMkedAxjF_uE:1554832057) ([raw file](https://github.com/valloric/ycmd/blob/650e4864a11900f0b2a7083aec5b5da75559a2b3/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L114)):*
Maybe we should have some sort of check for libclang version. If somebody uses 3.7 or earlier with this line deleted, it's going to slow.

Possibly not this PR though.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353)*

micbou:
Reviewed 1 of 1 files at r8.

---

*[cpp/llvm/include/clang-c/Index.h, line 1211 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCNd9TTaPJdIatD38H5:-KCc71cI7bP5XHKPOJrt:-1134169408) ([raw file](https://github.com/valloric/ycmd/blob/c66213da2951173f5882f0155700d84901fc5708/cpp/llvm/include/clang-c/Index.h#L1211)):*
That makes sense. Version bumped.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353)*

vheon:
Reviewed 2 of 3 files at r5, 1 of 1 files at r6, 1 of 1 files at r8.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 114 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCc5lDCb4ExaGkKWNDw:-KCcAtjLNaIBjfvwGDr5:2060000481) ([raw file](https://github.com/valloric/ycmd/blob/650e4864a11900f0b2a7083aec5b5da75559a2b3/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L114)):*
maybe keep a todo so we don't forget? or just open an issue to track it

---

*[ycmd/tests/clang/diagnostics_test.py, line 30 \[r8\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCcBJhT3nJT1mUgg7SP:-KCcBJhT3nJT1mUgg7SQ:-393807392) ([raw file](https://github.com/valloric/ycmd/blob/04b433ecb1ef3768babc393861b5d49d6cfc1cd2/ycmd/tests/clang/diagnostics_test.py#L30)):*
I'm missing the reason of this change from Shared to Isolated :confused: 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353)*

micbou:

---

*[ycmd/tests/clang/diagnostics_test.py, line 30 \[r8\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCcBJhT3nJT1mUgg7SP:-KCcBmAxlzfkF7sWy8aY:-1557283808) ([raw file](https://github.com/valloric/ycmd/blob/04b433ecb1ef3768babc393861b5d49d6cfc1cd2/ycmd/tests/clang/diagnostics_test.py#L30)):*
When removing the reparse in the TU constructor, some diagnostic tests were still successful (but they should have failed) because the reparse was done by a previous test. We don't want this kind of behavior so we isolate these tests. In fact, I wonder if we should isolate all the Clang tests.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353)*

micbou:

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 114 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCc5lDCb4ExaGkKWNDw:-KCcJJ8p-tnWM1_9nEln:-762106437) ([raw file](https://github.com/valloric/ycmd/blob/650e4864a11900f0b2a7083aec5b5da75559a2b3/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L114)):*
With the changes introduced by this PR, ycmd cannot be compiled anymore with Clang 3.7 (or earlier).

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353)*

Valloric:

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 114 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCc5lDCb4ExaGkKWNDw:-KCcVTp9UV63CHNcq49D:1291526927) ([raw file](https://github.com/valloric/ycmd/blob/650e4864a11900f0b2a7083aec5b5da75559a2b3/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L114)):*
Ok, that solves that problem. :)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353)*

puremourning:
:lgtm:

I might be missing something, but I seem to recall that @d0k suggested that this change to 3.8 would remove the need for all the hacks we do to get system include paths working (in particular on OS X).

However, I tested this and it does not remove the need for hacking in additional system paths (at least not on my macs when working with YCM cpp code).

---

Reviewed 66 of 68 files at r1, 2 of 2 files at r3, 2 of 3 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 1 of 1 files at r8.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 114 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCc5lDCb4ExaGkKWNDw:-KCfryJ_ReSa6RfjA_yV:-1842333546) ([raw file](https://github.com/valloric/ycmd/blob/650e4864a11900f0b2a7083aec5b5da75559a2b3/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L114)):*
I suspect that will lead to a number of issues like "YCM doesn't link anymore with `--system-libclang`". Perhaps with a few !!!11one.

Should we consider removing `--system-libclang` if we're going to be using the latest `libclang` releases? Folk like me can still use `-DPATH_TO_LLVM_ROOT` or equivalent... The reason I suggest this is that there aren't likely to be vary many system lib clangs which are compatible with 3.8 (for quite a while at least).

---

*[ycmd/tests/clang/subcommands_test.py, line 342 \[r8\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCftVB4bfBwQ8b8ruM8:-KCftVB4bfBwQ8b8ruM9:-442444587) ([raw file](https://github.com/valloric/ycmd/blob/04b433ecb1ef3768babc393861b5d49d6cfc1cd2/ycmd/tests/clang/subcommands_test.py#L342)):*
I have a plan to fix this. I think there are a couple of options but the simplest is to just compare the strings. Then `GetType` works great for `auto`. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353#-:-KCg2i4hykEDc9VWuAPr:-866253576)*

d0k:
Thanks for cleaning up for me :)

> I might be missing something, but I seem to recall that @d0k suggested that this change to 3.8 would remove the need for all the hacks we do to get system include paths working (in particular on OS X).

It's not that easy. To get standard library headers it needs the full path in the compilation database, just custom flags won't work. Then libclang will find libc++ relative to that path. CMake seems to generate absolute paths for me but I've seen that failing before. Also xcode 7 seems to have changed again which makes compiles file if the SDKROOT env var isn't set, I haven't fully figured out the root cause there. It'll probably take more tweaks until we finally have a hack-free environment on OS X :(

Valloric:

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 114 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/353#-KCc5lDCb4ExaGkKWNDw:-KCgixGTjBfoX3gQ99QU:-916263683) ([raw file](https://github.com/valloric/ycmd/blob/650e4864a11900f0b2a7083aec5b5da75559a2b3/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L114)):*
`--system-libclang` has always been strongly suggested-against. Like, _extremely_ strongly (see the wording we use in the README).

It was only added because there were so many people screaming that they want that option and how they know what they're doing, so whatever, fine, here you go. It's largely undocumented (intentionally). It's only mentioned in the FreeBSD install instructions, and that's not an officially supported platform.

I say we leave it. If anyone uses it and it explodes for them, there's a large helping of "I told you so" waiting for them.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353)*

micbou:
For me, this PR is ready except the question about the Clang tests: should we isolate all of them because of the TU parsing from previous tests? Diagnostic tests were isolated for this reason. Time to run these tests is not an issue: they take about 3 seconds when isolated.

---

Reviewed 65 of 68 files at r1, 2 of 2 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353#-:-KD03bG2P6EDMK3wGWGk:707483797)*

Valloric:
@micbou Do we have a problem with them that needs fixing? If not, maybe best to just leave it as-is.

You have three LGTMs here so feel free to trigger homu at your convenience.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353#-:-KDB8psqfGelVl_L1glp:1653424436)*

micbou:
> Do we have a problem with them that needs fixing? If not, maybe best to just leave it as-is.

Not really. Let's keep them this way.

@homu r+

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/353#-:-KDBCa_QguNUyWGf947x:-1510441747)*

PR #356, opened 2016-02-09
[READY] Ignore vagrant directory

vheon:
[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/356)


Valloric:
Thanks!

This is trivial enough that I'm fine with it landing while #358 is still pending.

@homu r+

PR #361, opened 2016-02-11
Find system-libclang on Debian x386 systems

stefansonski:
Search not only for clang, but also for libclang.so.1.
This way the system-wide installed libclang is found on Debian-based systems.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/361)


Valloric:
Thanks for the PR, but we actively recommend against using the system libclang in all of our docs. We've had no end of problems with system libclangs having weird patches applied that brake in weird ways. The upstream libclangs from llvm.org are a known quantity and we test against them in CI so we know they work.

So while I do appreciate the patch, we will not be merging it. Thanks for understanding.

Valloric:
Ah crap, this is changing the code behind the SYSTEM_LIBCLANG flag. Forgot we had that. :D

The it's fine. Sorry for the confusion.

Valloric:
@homu r+

PR #364, opened 2016-02-13
Modify C triggers for completer_utils.py

ailisp:
For C programmers, there are lots of long function and macro names, such as a_foo_bar, which is similar to a->foo->bar in c++, ycmd may be better for c programmers to add `DEFAULT_FILETYPE_TRIGGERS` for C. Before changing this, ycmd can only complete for struct members in C, which is not very useful. Since it's only a minor change, I didn't write a test, but only test it in my emacs. As for completion speed, it runs with no extra time when adding GLib and all standard librarys for C.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/364)

### DIFF
--- a/ycmd/completers/completer_utils.py
+++ b/ycmd/completers/completer_utils.py
@@ -143,7 +143,7 @@ def FiletypeCompleterExistsForFiletype( filetype ):
 TRIGGER_REGEX_PREFIX = 're!'
 
 DEFAULT_FILETYPE_TRIGGERS = {
-  'c' : ['->', '.'],
+  'c' : ['->', '.', r're![_a-zA-Z][_a-zA-Z0-9]*'],
   'objc' : ['->',
             '.',
             r're!\[[_a-zA-Z]+\w*\s',    # bracketed calls


vheon:
I'm not convinced about this since you have basically disabled the identifier completer for all C users. IMHO if a user really want this behaviour he/she can easily put the settings in their configuration. In any case the tests are failing and I'm not sure that is a bug on our code :confused:

ailisp:
@vheon Excuse me, but how to turn on this feature in ycm config file?

vheon:
@ailisp you've mentioned that you use emacs, so I don't know in detail how to do it from the editor but if you look the ycmd config file you can see that there is a `semantic_triggers` key which gets merged with the default one.

ailisp:
@vheon I see. Before this modifying, ycmd completion of this kind is only include identifiers in the current file, but not a clang-based candidates from all include files, for example, if I typed the following before modifying this:
``` c
#include <glib.h>
g_list```
I will get no completions since `g_list_*` doesn't appear in current file but in glib.h.
After this modifying, completions will be:
```
g_list_next
g_list_free1
g_datalist_id_set_data
...
```
Sorry, I can't insert an image here, but this feature seems important to any C Programmer. Ycmd completions after `.` or `->` and identifier based on current file is enough for C++ since C++ is an OOP and doesn't have long function names, but  in larger C programs, functions have to be very long to make a difference between operations on different type of structs.

vheon:
you can always `<C-Space>` if you know you need a semantic completion.

vheon:
You can also make the triggers more specific to what you're working with. Look at https://github.com/Valloric/YouCompleteMe/issues/1930, is a discussion on the vim client but you'll get what I want to say (like `re!g_`).

Valloric:
@ailisp Thanks for the PR, but I've always been against auto-triggering semantic completion on every keystroke (instead of just after triggers) for performance reasons. Libclang just isn't fast enough to make that a good experience for most users, and in most cases the identifier completer is good enough. We have the `<c-space>` shortcut you can use to trigger semantic completion wherever if you need it though.

Thanks again! :)

PR #365, opened 2016-02-14
[READY] Fix Python 2 support and add Python 3 support on Windows

micbou:
### Python 2 support 

We cannot pass an environment variable containing unicode literals to `Popen` on Windows and Python2. Instead of converting all environment variables to native strings, we only convert the ones we use.

### Python 3 support

We update the requests submodule to fix a compatibility error between requests and Python 3.5.

Fix from PR #211 is only needed for Python 2.

In `build.py`, we prepend the folder of the `ycm_core_tests` executable to the `PATH` instead of replacing it so that the executable is able to find the `python35.dll` library.

We don't reuse Omnisharp instances in C♯ tests on Windows and Python 3 because of the following random error occuring when running those tests:
```python
OSError: [WinError 6] The handle is invalid
```

### AppVeyor configuration

We append the `-DUSE_DEV_FLAGS=ON` option to `EXTRA_CMAKE_ARGS` environment variable in `run_tests.py` script in order to use `EXTRA_CMAKE_ARGS` when running tests. We need it on AppVeyor to tell CMake the correct path to the Python 3 library according to architecture. Otherwise, it always pick the 64 bits one. I can't reproduce this issue on my machine so I consider it is specific to AppVeyor.

We only do one run for MSVC 11, MSVC 12, and Python 2.7. This reduces the number of runs from 6 to 5.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/365)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,18 +6,23 @@ branches:
 environment:
   USE_CLANG_COMPLETER: true
   matrix:
-  - msvc: 11
-    arch: 32
+  # We only test MSVC 11 and 12 with Python 3.5 on 64 bits.
   - msvc: 11
     arch: 64
-  - msvc: 12
-    arch: 32
+    python: 35
   - msvc: 12
     arch: 64
+    python: 35
   - msvc: 14
     arch: 32
+    python: 35
+  # We only test Python 2.7 with MSVC 14 on 64 bits.
   - msvc: 14
     arch: 64
+    python: 27
+  - msvc: 14
+    arch: 64
+    python: 35
 install:
   - ci\appveyor\appveyor_install.bat
 build_script:
--- a/build.py
+++ b/build.py
@@ -1,9 +1,12 @@
 #!/usr/bin/env python
 
-from __future__ import unicode_literals
+# Passing an environment variable containing unicode literals to a subprocess
+# on Windows and Python2 raises a TypeError. Since there is no unicode
+# string in this script, we don't import unicode_literals to avoid the issue.
 from __future__ import print_function
 from __future__ import division
 from __future__ import absolute_import
+
 import os
 import subprocess
 import os.path as p
--- a/build.py
+++ b/build.py
@@ -283,7 +283,10 @@ def RunYcmdTests( build_dir ):
   new_env = os.environ.copy()
 
   if OnWindows():
-    new_env[ 'PATH' ] = DIR_OF_THIS_SCRIPT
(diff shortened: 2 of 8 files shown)


Valloric:
Thanks for making sure we continue to work well on Windows! :) Without you, it would probably bitrot.

:lgtm: with minor comments.

---

Review status: 0 of 9 files reviewed at latest revision, 2 unresolved discussions.

---

*[build.py, line 3 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/365#-KAY0KrZamUciq4pkGNp:-KAY0Kr_offc0kp4l0rE:1572773173) ([raw file](https://github.com/valloric/ycmd/blob/eeefabe407167cef0707c87865ed019a73e0b28f/build.py#L3)):*
Not having this needs a comment explaining why it's missing otherwise someone will add it again in the future.

---

*[build.py, line 283 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/365#-KAY0ZIR8RgM9mncRkjr:-KAY0ZIR8RgM9mncRkjs:5572663) ([raw file](https://github.com/valloric/ycmd/blob/a7dbfa8078fa748ff06330a42044cf3eabe7a277/build.py#L283)):*
Needs comment.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/365#-:-KAY16Xft4CQWZEYbhOA:672295503)*

micbou:
Reviewed 8 of 9 files at r1, 1 of 1 files at r2.

---

*[build.py, line 3 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/365#-KAY0KrZamUciq4pkGNp:-KA_dUI4Q5_Xw-Xy52Gk:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/eeefabe407167cef0707c87865ed019a73e0b28f/build.py#L3)):*
Done.

---

*[build.py, line 283 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/365#-KAY0ZIR8RgM9mncRkjr:-KA_dUnjSrHTOrMQ_9Lo:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/a7dbfa8078fa748ff06330a42044cf3eabe7a277/build.py#L283)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/365)*

Valloric:
Thanks!

PR #372, opened 2016-02-17
[READY] Running ./style_format.sh to reformat C++

Valloric:
People forget to run it in PRs so we've accumulated some things that
don't follow the style guide.

Since astyle isn't perfect, this also makes some code worse (like
the #define indents), but we don't want to make manual changes to astyle
output. (If clang-format were more flexible, we could use that, but
currently it produces worse output that astyle for our codebase.)

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/372)

### DIFF
--- a/cpp/ycm/ClangCompleter/ClangCompleter.cpp
+++ b/cpp/ycm/ClangCompleter/ClangCompleter.cpp
@@ -213,7 +213,7 @@ ClangCompleter::GetFixItsForLocationInFile(
   int column,
   const std::vector< UnsavedFile > &unsaved_files,
   const std::vector< std::string > &flags,
-  bool reparse) {
+  bool reparse ) {
 
   ReleaseGil unlock;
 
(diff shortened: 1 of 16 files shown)


micbou:
My poor `#define`s.

:lgtm:

---

Reviewed 16 of 16 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/372#-:-KAgbwTJV_yJaU-0sGcV:80688890)*

vheon:
I don't think there's much to review since is an automatic tool, but it is good to add it to the Vagrantfile as a dependency :+1: 

:lgtm: @homu r+

---

Reviewed 16 of 16 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/372#-:-KAgcg2ezo9oXL9cYRSX:98847075)*

PR #373, opened 2016-02-17
Update JediHTTP

vheon:
Fixes https://github.com/Valloric/YouCompleteMe/issues/1924

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/373)


Valloric:
Thanks!

I edited your PR description to contain "Fixes <issue>" so that the YCM issue is autoclosed when this is merged.

@homu r+

vheon:
> I edited your PR description to contain "Fixes " so that the YCM issue is autoclosed when this is merged.

But it will not be until we update the ycmd submodule in YCM

Valloric:
> But it will not be until we update the ycmd submodule in YCM

That's a release detail. The problem is fixed even if the fix hasn't yet been pushed out. Doesn't change how we use the issue tracker.

PR #380, opened 2016-02-20
[READY] Fix python-future import

micbou:
We cannot import `future` before appending `python-future` to the `sys.path`. This means that we need to do this in `SetUpPythonPath` before importing `utils` (because `utils` is using `future`).

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/380)


vheon:
I was starting to look into this but apparently you beat me to the punch :+1: 

:lgtm: 

---

Reviewed 3 of 3 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KAzaIm5Rtgy3TtDRiz4:82046437)*

Valloric:
How will this work for YCM? It uses `AddNearestThirdPartyFoldersToSysPath` to import third_party libs both in its own repo and in the ycmd repo. AFAIR it doesn't call `SetUpPythonPath`. 

I think we have to have `AddNearestThirdPartyFoldersToSysPath` do everything in some way.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KAzdxZzmcIp-48jUo0h:1281110830)*

micbou:
We just add this line in `SetUpPython` in `autoload/youcompleteme.vim`:
```python
sys.path.append( os.path.join( script_folder, '..', 'third_party', 'ycmd',
                               'third_party', 'python-future', 'src' ) )
```
before importing `utils`.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KAzehnY37hLbs8yTIST:186636736)*

Valloric:
@micbou Sure, but that's a hack, no? What happens the next time we have a third-party lib that needs special logic? The idea of `AddNearestThirdPartyFoldersToSysPath` was that it handles all that logic for the caller. Now it doesn't. :(

Could we refactor the code somehow so we still have one function that can be called from both ycmd and YCM that handles all of this? It doesn't have to be `AddNearestThirdPartyFoldersToSysPath`, it could be something else. I'd prefer to encapsulate our hacks inside functions that Just Work.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-A6cjllimA3Q-_0MC:71346883)*

micbou:
We can move `AddNearestThirdPartyFoldersToSysPath` to `server_utils`. It is only used here in `ycmd`.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-G_SXWapztz3rF2CF:183731470)*

Valloric:
> We can move AddNearestThirdPartyFoldersToSysPath to server_utils. It is only used here in ycmd.

I'm not sure I understand; I might be confused. `AddNearestThirdPartyFoldersToSysPath` is also used in YCM in [`autoload/youcompleteme.vim`](https://github.com/Valloric/YouCompleteMe/blob/master/autoload/youcompleteme.vim#L151).

What I'd like us to have is one function (its location & name I don't care much about) that accepts a path, searches for the nearest third_party folder and adds everything there to sys.path with correct handling of weird libs like python-future encapsulated within it so that the caller doesn't care.

What do you propose is the best way to achieve that? I'm open to all sorts of refactoring to make it happen.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-Se_x-h2i47SbOPgE:-936578294)*

micbou:
Here's my proposal: we move `AddNearestThirdPartyFoldersToSysPath` and its subfunctions (`PathToNearestThirdPartyFolder` and `AncestorFolders`) to `server_utils`. In YCM, instead of importing `utils.AddNearestThirdPartyFoldersToSysPath`, we import `server_utils.AddNearestThirdPartyFoldersToSysPath`.

I pushed the changes. Since `AncestorFolders` is used in the Javascript completer, we replace it with `PathToAllParentFolders` which is moved to `utils.py` from `extra_conf_store.py`. It is simplified by following the same logic as `AncestorFolders` and it manages the case where the path is a folder. I added a test for this case. `PathToAllParentFolders` tests are moved to `utils_test` module. Finally, we hardcode the `third_party` path in the Rust completer instead of using `PathToNearestThirdPartyFolder` function.

---

Review status: 1 of 9 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-Tvgd_6MMV7ogFAc5:1733139818)*

Valloric:
All of that sounds reasonable. :) 

:lgtm: with comment.

@vheon Since this is now almost entirely different than it was when you LGTM'd it, you probably want to take another look.

---

Reviewed 1 of 3 files at r1, 8 of 8 files at r2.

---

*[ycmd/server_utils.py, line 34 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/380#-KB-Zb9E6l6Payf1NsaI:-KB-Zb9E6l6Payf1NsaJ:-1114703857) ([raw file](https://github.com/valloric/ycmd/blob/471e956e0c6f86216ee377842a99d0b62b10a1a3/ycmd/server_utils.py#L34)):*
This is one of those "what was I _thinking_" moments; why was this ever a function? :)

---

*[ycmd/server_utils.py, line 48 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/380#-KB-_9XsQBt1ztkECeLC:-KB-_9XsQBt1ztkECeLD:-1742178053) ([raw file](https://github.com/valloric/ycmd/blob/471e956e0c6f86216ee377842a99d0b62b10a1a3/ycmd/server_utils.py#L48)):*
Does this still need to be here? `AddNearestThirdPartyFoldersToSysPath` is doing it as well.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-Z5FSdNJ7HkWJhbZz:1107888059)*

Valloric:
@micbou Sure, run it only on windows. We have a ton of conditional tests now that only run on py2 for instance. Worth knowing why it fails though, but I don't care too much (like you said, those paths don't matter on linux/mac).

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-e3GKj27axNaRQeut:-951036726)*

micbou:
It fails because on Linux/OS X, the parent of `C:/` is `C:` (`C:` is a valid folder name). Old behavior was incorrect.

I think we should use the [`Skip` nosetests plugin](http://nose.readthedocs.org/en/latest/plugins/skip.html) for the conditional tests. This way, we have always the same number of tests on all platforms/configurations.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-ePmmP2QnF2mJAQNM:-1244483826)*

micbou:
~~the parent of C:/ is C:~~

the parent of `C:` is the empty path.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-glR1sumnO2yT2O_I:140237992)*

Valloric:
Looking at the skip docs you linked to, I'm not sure how that's configured.

In any case, don't worry about it in this PR. Go with the approach we currently use; we can refactor all the conditional tests in a separate PR.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-iBIoS_7qt2DjHkmb:939018147)*

Valloric:
Feel free to merge this with homu when ready.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-iTo_zobKyG2QqFRO:-1769298074)*

micbou:
Updated. I am waiting for the tests.

---

Reviewed 1 of 3 files at r1, 5 of 8 files at r2, 2 of 2 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-jmG0mA87H3Ri2XEv:-727894309)*

micbou:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[ycmd/server_utils.py, line 34 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/380#-KB-Zb9E6l6Payf1NsaI:-KB-pAphH1RhAOkA6jrH:1842649476) ([raw file](https://github.com/valloric/ycmd/blob/471e956e0c6f86216ee377842a99d0b62b10a1a3/ycmd/server_utils.py#L34)):*
Same thing in `extra_conf_store.py`.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380)*

mispencer:
SkipTest is a builtin plugin, so:

```
from nose import SkipTest
# Then later
raise SkipTest( 'Some reason why skipping' )
```

It really is that simple.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-rzpzfS9D83Zx7bhe:1163052096)*

vheon:
:lgtm: 

for skipping tests I prefer `unittest.skipIf` to use as a decorator.

---

Reviewed 6 of 8 files at r2, 2 of 2 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-w-M3EE4FYvDBS3BQ:177710919)*

vheon:
I don't know what happened but for some reason I apparently merged this :scream: I will rever this and call `homu`

vheon:
Apparently reverting means create another PR to back out this commit. What should I do?

micbou:
Let's keep it merged. It was ready anyway. PR #381 just needs to be updated.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/380#-:-KB-xGOLKuDv8GLHIZ8f:-1906467519)*

vheon:
And to make things worse I screwed up #381 

I'm really sorry guys :sob:

vheon:
@micbou `homu` didn't run though, so we didn't run the tests.

Valloric:
@vheon Eh, don't worry about it. :) "Only people who do no work make no mistakes."

vheon:
> "Only people who do no work make no mistakes."

You're not the first one who told me that, a coworker told me that the second week at work :)

Anyway I believe that I clicked the merge button on reviewable right after the "Publish" button since you already mark it as LGTM. @Valloric is configurable? since we almost only merge things through `homu` can we tell reviewable to not show the merge button (which is huge btw)?

A simpler solution would be to be careful of course :stuck_out_tongue_closed_eyes:

Valloric:
Reviewable [has a repo](https://github.com/Reviewable/Reviewable) on github where issues can be reported. Feel free to open one there. :)

PR #382, opened 2016-02-21
[READY] Use decorator to skip conditional tests

micbou:
This way, we have the same list of tests on all configurations.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/382)

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -5,5 +5,6 @@ PyHamcrest>=1.8.5
 WebTest>=2.0.20
 ordereddict>=1.1
 nose-exclude>=0.4.1
+unittest2>=1.1.0
 # This needs to be kept in sync with submodule checkout in third_party
 future==0.15.2
--- a/ycmd/tests/bottle_utils_test.py
+++ b/ycmd/tests/bottle_utils_test.py
@@ -23,33 +23,36 @@ from future import standard_library
 standard_library.install_aliases()
 from builtins import *  # noqa
 
-from future.utils import PY2
 from nose.tools import eq_
 from mock import patch, call
 from ycmd import bottle_utils
+from ycmd.tests.test_utils import Py2Only, Py3Only
 import bottle
 
 
-if PY2:
-  @patch( 'bottle.response' )
-  def SetResponseHeader_Py2CorrectTypesWithStr_test( *args ):
-    bottle_utils.SetResponseHeader( 'foo', 'bar' )
-    eq_( bottle.response.set_header.call_args, call( 'foo', u'bar' ) )
+@Py2Only
+@patch( 'bottle.response' )
+def SetResponseHeader_Py2CorrectTypesWithStr_test( *args ):
+  bottle_utils.SetResponseHeader( 'foo', 'bar' )
+  eq_( bottle.response.set_header.call_args, call( 'foo', u'bar' ) )
 
 
-  @patch( 'bottle.response' )
-  def SetResponseHeader_Py2CorrectTypesWithUnicode_test( *args ):
-    bottle_utils.SetResponseHeader( u'foo', u'bar' )
-    eq_( bottle.response.set_header.call_args, call( 'foo', u'bar' ) )
-
-else:
-  @patch( 'bottle.response' )
-  def SetResponseHeader_Py3CorrectTypesWithBytes_test( *args ):
-    bottle_utils.SetResponseHeader( b'foo', b'bar' )
-    eq_( bottle.response.set_header.call_args, call( u'foo', u'bar' ) )
+@Py2Only
+@patch( 'bottle.response' )
+def SetResponseHeader_Py2CorrectTypesWithUnicode_test( *args ):
+  bottle_utils.SetResponseHeader( u'foo', u'bar' )
+  eq_( bottle.response.set_header.call_args, call( 'foo', u'bar' ) )
 
 
-  @patch( 'bottle.response' )
-  def SetResponseHeader_Py3CorrectTypesWithUnicode_test( *args ):
-    bottle_utils.SetResponseHeader( u'foo', u'bar' )
-    eq_( bottle.response.set_header.call_args, call( u'foo', u'bar' ) )
+@Py3Only
+@patch( 'bottle.response' )
+def SetResponseHeader_Py3CorrectTypesWithBytes_test( *args ):
+  bottle_utils.SetResponseHeader( b'foo', b'bar' )
+  eq_( bottle.response.set_header.call_args, call( u'foo', u'bar' ) )
+
+
+@Py3Only
+@patch( 'bottle.response' )
+def SetResponseHeader_Py3CorrectTypesWithUnicode_test( *args ):
+  bottle_utils.SetResponseHeader( u'foo', u'bar' )
+  eq_( bottle.response.set_header.call_args, call( u'foo', u'bar' ) )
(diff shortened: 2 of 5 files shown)


Valloric:
Am I the only one who thinks this is substantially uglier than what we currently have? Instead of having the IF once, we now have it repeated in every single test. 

I don't see having the same list of tests in all configurations being worth the loss of readability & maintainability. Having the same list is nice, but isn't it mostly cosmetic?

Willing to be convinced otherwise, but currently I'm against this change.

micbou:
I think this is useful for keeping track of all tests. It also remove the indentation. Would you prefer with a decorator on each test? I can write one.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/382#-:-KB2ACQCUq0SNnmN8pKJ:-1222476971)*

Valloric:
A decorator would be better, yes. If it were something like `@skipIf(<bool expression causing skipping>)` so that it's used like `@skipIf(not PY2)` in our case, then that would be great. The current approach with an IF inside the test is IMO not an improvement.

TBH I actually expected the skip infrastructure in nosetests to have such a decorator already. The `SkipTest` thing they have sucks.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/382#-:-KB481xtslfqqq-Otfly:-1256107542)*

puremourning:
I thought @vheon mentioned a decorator that he was already using in JediHTTP ?

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/382#-:-KB49678k1XMUocSlFSw:-1548498793)*

vheon:
> I thought @vheon mentioned a decorator that he was already using in JediHTTP ?

Yes I use them like: https://github.com/vheon/JediHTTP/blob/master/jedihttp/tests/end_to_end_test.py#L132 and are as simple as https://github.com/vheon/JediHTTP/blob/master/jedihttp/tests/utils.py#L34-L35.

Valloric:
@vheon Those are _very_ nice! We should use it. We probably need a `@windows_only` tag as well.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/382#-:-KB4slljT8RNIOA8kQuU:-579121085)*

micbou:
Thanks @vheon. I'll update the PR with those kinds of decorator.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/382#-:-KB4tIfk0UWVBqwdoF7L:-1444398702)*

micbou:
Updated.

---

Reviewed 4 of 4 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/382#-:-KBB00pUeyiia9PSg-bB:-719267080)*

Valloric:
Review status: 3 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[ycmd/tests/bottle_utils_test.py, line 47 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/382#-KBBCcfArXtzZAQIA7cK:-KBBCcfB7vIvm0df5zrU:1543507829) ([raw file](https://github.com/valloric/ycmd/blob/1f3b881afb86baac129c8aae5552141acd5d0525/ycmd/tests/bottle_utils_test.py#L47)):*
Er, this is a Py3 test (as is the one below). There should be a `@Py3Test`.

I think there's more than one instance of py3-only tests, though I do try to make py3 tests run on py2 as well.

---

*[ycmd/tests/test_utils.py, line 41 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/382#-KBBC7nMuPtrebFxP7ci:-KBBC7nMuPtrebFxP7cj:-1067773225) ([raw file](https://github.com/valloric/ycmd/blob/1f3b881afb86baac129c8aae5552141acd5d0525/ycmd/tests/test_utils.py#L41)):*
I'm not against these names in particular, but why use the `Test` suffix? Why not `Py2Only` and `WindowsOnly`? 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/382)*

micbou:
Reviewed 2 of 2 files at r4, 4 of 4 files at r5.

---

*[ycmd/tests/bottle_utils_test.py, line 47 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/382#-KBBCcfArXtzZAQIA7cK:-KBBGL478w6kTWSQsp0x:-1815223848) ([raw file](https://github.com/valloric/ycmd/blob/1f3b881afb86baac129c8aae5552141acd5d0525/ycmd/tests/bottle_utils_test.py#L47)):*
Oops, you didn't see anything.

---

*[ycmd/tests/test_utils.py, line 41 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/382#-KBBC7nMuPtrebFxP7ci:-KBBFLbmcqB2sCoBAocB:1049617194) ([raw file](https://github.com/valloric/ycmd/blob/1f3b881afb86baac129c8aae5552141acd5d0525/ycmd/tests/test_utils.py#L41)):*
I thought that `Test` was more informative ("it is a Python 2 test" versus "it is only Python 2") but I don't mind using `Only`.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/382)*

Valloric:
:lgtm: Thanks for the refactoring! :) I loooove refactoring-only PRs; second only to new-tests-only PRs.

---

---

*[ycmd/tests/test_utils.py, line 41 \[r4\]](https://reviewable.io:443/reviews/valloric/ycmd/382#-KBBC7nMuPtrebFxP7ci:-KBBL4ZK1cDIyQZ4Mq5v:1075765008) ([raw file](https://github.com/valloric/ycmd/blob/1f3b881afb86baac129c8aae5552141acd5d0525/ycmd/tests/test_utils.py#L41)):*
Don't get me wrong, I'm not convinced my version is better. It just seems redundant to have a decorator on a function with the`Test` substring when the function itself certainly ends with `test`.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/382#-:-KBBLj0IAHpCTZaV7yEr:-1872037050)*

Valloric:
Reviewed 1 of 2 files at r4, 4 of 4 files at r5.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/382)*

vheon:
:lgtm: @homu r+

---

Reviewed 1 of 2 files at r4, 3 of 4 files at r5, 1 of 1 files at r6.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/382#-:-KBEo3gvHm3jjpAyyRn0:-141632775)*

PR #384, opened 2016-02-21
[WIP] Pass filename as a C++ compatible string to FlagsForFile

### LINKED ISSUE #378: Error with GetCompilationInfoForFile and python 2
Hi,

I'm getting the following error when I try to use a `compile_commands.json` and ycmd with python 2:

```python
  File "/home/peter/src/ycmd/third_party/bottle/bottle.py", line 861, in _handle
    return route.call(**args)
  File "/home/peter/src/ycmd/third_party/bottle/bottle.py", line 1734, in wrapper
    rv = callback(*a, **ka)
  File "/home/peter/src/ycmd/ycmd/../ycmd/watchdog_plugin.py", line 105, in wrapper
    return callback( *args, **kwargs )
  File "/home/peter/src/ycmd/ycmd/../ycmd/hmac_plugin.py", line 70, in wrapper
    body = callback( *args, **kwargs )
  File "/home/peter/src/ycmd/ycmd/../ycmd/handlers.py", line 78, in EventNotification
    event_handler )( request_data )
  File "/home/peter/src/ycmd/ycmd/../ycmd/completers/cpp/clang_completer.py", line 318, in OnFileReadyToParse
    flags = self._FlagsForRequest( request_data )
  File "/home/peter/src/ycmd/ycmd/../ycmd/completers/cpp/clang_completer.py", line 381, in _FlagsForRequest
    return self._flags.FlagsForFile( filename, client_data = client_data )
  File "/home/peter/src/ycmd/ycmd/../ycmd/completers/cpp/flags.py", line 95, in FlagsForFile
    client_data )
  File "/home/peter/src/ycmd/ycmd/../ycmd/completers/cpp/flags.py", line 157, in _CallExtraConfFlagsForFile
    return module.FlagsForFile( filename, client_data = client_data )
  File "/home/peter/src/ycmd/cpp/ycm/.ycm_extra_conf.py", line 166, in FlagsForFile
    compilation_info = GetCompilationInfoForFile( filename )
  File "/home/peter/src/ycmd/cpp/ycm/.ycm_extra_conf.py", line 159, in GetCompilationInfoForFile
    return database.GetCompilationInfoForFile( filename )
ArgumentError: Python argument types in
    CompilationDatabase.GetCompilationInfoForFile(CompilationDatabase, unicode)
did not match C++ signature:
    GetCompilationInfoForFile(YouCompleteMe::CompilationDatabase {lvalue}, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)
```

micbou:
Fixes #378.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/384)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -154,9 +154,9 @@ def _CallExtraConfFlagsForFile( module, filename, client_data ):
   # For the sake of backwards compatibility, we need to first check whether the
   # FlagsForFile function in the extra conf module even allows keyword args.
   if inspect.getargspec( module.FlagsForFile ).keywords:
-    return module.FlagsForFile( filename, client_data = client_data )
-  else:
-    return module.FlagsForFile( filename )
+    return module.FlagsForFile( ToCppStringCompatible( filename ),
+                                client_data = client_data )
+  return module.FlagsForFile( ToCppStringCompatible( filename ) )
 
 
 def PrepareFlagsForClang( flags, filename ):
--- a/ycmd/tests/clang/flags_test.py
+++ b/ycmd/tests/clang/flags_test.py
@@ -15,7 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
 
-from __future__ import unicode_literals
+# Intentionally not importing unicode_literals!
 from __future__ import print_function
 from __future__ import division
 from __future__ import absolute_import
@@ -23,9 +23,11 @@ from future import standard_library
 standard_library.install_aliases()
 from builtins import *  # noqa
 
+from future.utils import PY2
 from nose.tools import eq_
 from nose.tools import ok_
 from ycmd.completers.cpp import flags
+import imp
 
 
 def SanitizeFlags_Passthrough_test():
@@ -234,3 +236,43 @@ def ExtraClangFlags_test():
       num_found += 1
 
   eq_( 1, num_found )
+
+
+def CreateModule( name, code ):
+  module = imp.new_module( name )
+  exec( code, module.__dict__ )
+  return module
+
+
+def FlagsForFile_OldSignature_PassCppCompatibleString_test():
+  code = """
+def FlagsForFile( filename ):
+  return filename
+"""
+  module = CreateModule( 'extra_conf', code )
+
+  filename = flags._CallExtraConfFlagsForFile( module, 'some_filename', None )
+
+  if PY2:
+    eq_( filename, 'some_filename' )
+    eq_( type( filename ), type( '' ) )
+  else:
+    eq_( filename, bytes( b'some_filename' ) )
+    ok_( isinstance( filename, bytes ) )
+
+
+def FlagsForFile_NewSignature_PassCppCompatibleString_test():
+  code = """
+def FlagsForFile( filename, **kwargs ):
+  return filename
+"""
+  module = CreateModule( 'extra_conf', code )
+
+  filename = flags._CallExtraConfFlagsForFile( module, 'some_filename', None )
+
+  if PY2:
+    eq_( filename, 'some_filename' )
+    eq_( type( filename ), type( '' ) )
+  else:
+    eq_( filename, bytes( b'some_filename' ) )
+    ok_( isinstance( filename, bytes ) )
--- a/ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py
+++ b/ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py
@@ -2,7 +2,7 @@ import os.path
 
 
 def FlagsForFile( filename, **kwargs ):
-  d = os.path.dirname( filename )
+  d = os.path.dirname( filename.decode( 'utf8' ) )
   return {
     'flags': [ '-iquote', os.path.join( d, 'quote' ),
                '-I', os.path.join( d, 'system' ) ],


Valloric:
Good catch!

---

---

*[ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py, line 5 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/384#-KB449dqzsNIQtvgUljl:-KB449dqzsNIQtvgUljm:-841845574) ([raw file](https://github.com/valloric/ycmd/blob/d7764082f63fa8da1b61b61ad82f7942cba17ef6/ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py#L5)):*
Why is this necessary if the change in flags.py works? This function should only receive py2 `str` or  py3 `bytes`.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/384#-:-KB44LUyEo6K-OZ-XS6e:422879044)*

Valloric:

---

*[ycmd/completers/cpp/flags.py, line 157 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/384#-KB44Tg6ZiX1OokH0iFn:-KB44Tg6ZiX1OokH0iFo:1443006329) ([raw file](https://github.com/valloric/ycmd/blob/d7764082f63fa8da1b61b61ad82f7942cba17ef6/ycmd/completers/cpp/flags.py#L157)):*
It would be good to have a test for this to ensure we don't pass unicode objects to FlagsForFile functions.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/384)*

micbou:
Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---

*[ycmd/completers/cpp/flags.py, line 157 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/384#-KB44Tg6ZiX1OokH0iFn:-KB4TFJFO0hlm4EmgshZ:-450724552) ([raw file](https://github.com/valloric/ycmd/blob/d7764082f63fa8da1b61b61ad82f7942cba17ef6/ycmd/completers/cpp/flags.py#L157)):*
Added a test.

---

*[ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py, line 5 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/384#-KB449dqzsNIQtvgUljl:-KB4Hx3lqc9UgSC_oLKc:-1096746826) ([raw file](https://github.com/valloric/ycmd/blob/d7764082f63fa8da1b61b61ad82f7942cba17ef6/ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py#L5)):*
Like you said, `filename` is a `bytes` on py3 and joining a `str` and a `bytes` raises the following error:
```python
TypeError: Can't mix strings and bytes in path components 
```
so we need to convert it to a string.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/384)*

puremourning:
Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---

*[ycmd/tests/clang/flags_test.py, line 250 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/384#-KB4f_73roi7YFmOIfmv:-KB4f_73roi7YFmOIfmw:-1215237025) ([raw file](https://github.com/valloric/ycmd/blob/d10f6a34319bbe98099ca1b060be1df356466c6a/ycmd/tests/clang/flags_test.py#L250)):*
I may be missing key information, but are we using `type()` here because `isinstance()` would return `True` due to `future` ?

---

*[ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py, line 5 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/384#-KB449dqzsNIQtvgUljl:-KB4duvzgVmuvxy5TajJ:-92737987) ([raw file](https://github.com/valloric/ycmd/blob/d7764082f63fa8da1b61b61ad82f7942cba17ef6/ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py#L5)):*
This is an interesting problem. If i', understanding correctly, it means that we can't isolate (or perhaps insulate) users from the PY2/PY3 non-interoperable nastiness. If they want their `.ycm_extra_conf.py` to work regardless of which Python ycmd is running in (like I would for the one I maintain at work), they must go through all of the same pain as everyone else to support multiple python versions. I don't think there is a way to solve this, because Python.

I just fear that we're gonna have a lot of reports of people going "YAY PYTHONz 3szzz!!." then "OMG my YCM broke with some obscure python error; you guys suck.". Ideas on a postcard as to how/if we can insulate users from this pain. I doubt we can.

We can update the FAQ, maybe, but that just reduces typing for us. "See FAQ" is easier than the alternative responses.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/384)*

Valloric:

---

*[ycmd/tests/clang/flags_test.py, line 250 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/384#-KB4f_73roi7YFmOIfmv:-KB4frEpDH5ujOyD56gq:900528012) ([raw file](https://github.com/valloric/ycmd/blob/d10f6a34319bbe98099ca1b060be1df356466c6a/ycmd/tests/clang/flags_test.py#L250)):*
I think so. See comments on `ToBytes` in `utils.py`. (I wrote it all down with comments and tests so I wouldn't have to remember it.)

---

*[ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py, line 5 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/384#-KB449dqzsNIQtvgUljl:-KB4gFpSg-4q_1J8i4A7:-370583924) ([raw file](https://github.com/valloric/ycmd/blob/d7764082f63fa8da1b61b61ad82f7942cba17ef6/ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py#L5)):*
> they must go through all of the same pain as everyone else to support multiple python versions.

Entirely correct. I'm aware of this, but like you said, "because Python."

> I just fear that we're gonna have a lot of reports

Your fears are not merely justified, but prescient. Yes, we will get these reports without a doubt.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/384)*

Valloric:
:lgtm: 

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/384#-:-KB4h05ziD0p7G7Dfnk1:980077423)*

micbou:
Test is failing on Python 2. I am trying a new approach.

---

---

*[ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py, line 5 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/384#-KB449dqzsNIQtvgUljl:-KB4h9JeBHBhAeG4yPKJ:2016671416) ([raw file](https://github.com/valloric/ycmd/blob/d7764082f63fa8da1b61b61ad82f7942cba17ef6/ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py#L5)):*
Ideally, `GetCompilationInfoForFile` should accept unicode strings but I assume it is not possible.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/384#-:-KB4hJGr6Dv9U-g6a5E_:573634995)*

micbou:
Ok, it should pass the tests now (I added a test for the old `FlagsForFile` signature).

---

Review status: 2 of 4 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/384#-:-KB4kqCuwbBTlNRAu6L1:-1007248440)*

Valloric:
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py, line 5 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/384#-KB449dqzsNIQtvgUljl:-KB4t6ht2u5_XwbvgbrL:2062535286) ([raw file](https://github.com/valloric/ycmd/blob/d7764082f63fa8da1b61b61ad82f7942cba17ef6/ycmd/tests/clang/testdata/test-include/.ycm_extra_conf.py#L5)):*
@micbou It's likely possible. We just need to overload that function in C++ with a version that accepts a Python unicode object (probably `boost::python::unicode`, though I'm not sure how it interacts with py3). 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/384)*

Valloric:
Review status: 2 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[ycmd/utils.py, line 335 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/384#-KB51M-alboyZCxYLMaG:-KB51M-alboyZCxYLMaH:-252092737) ([raw file](https://github.com/valloric/ycmd/blob/59db1412c8ee8405ff2322c46f62400b222918e5/ycmd/utils.py#L335)):*
If this is used only in one test file, then it should probably just be a utility method in that file.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/384)*

Valloric:
Reviewed 1 of 2 files at r4, 1 of 1 files at r5.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/384)*

micbou:
Closing in favor of #394.

PR #385, opened 2016-02-21
Added logic to find a standard library path that was recently added o…

indianajohn:
…n MacOSX, and add that path to the list of flags that is sent to libclang.

As discussed in https://github.com/Valloric/YouCompleteMe/issues/1992#issuecomment-185876944 , the include paths for MacOSX's version of Clang have recently changed. Not only have they changed, but they include a version string (7.0.2) that could change in the near future, or be different between different systems.  This pull request uses some simple logic to find that directory despite any version changes that might occur, and add it to the list of flags.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/385)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -49,6 +49,8 @@ FILE_FLAGS_TO_SKIP = set(['-MD', '-MMD', '-MF', '-MT', '-MQ', '-o'])
 # See Valloric/ycmd#266
 CPP_COMPILER_REGEX = re.compile( r'\+\+(-\d+(\.\d+){0,2})?$' )
 
+CLANG_TOOLCHAIN_PARENT_PATH = '/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang'
+
 class Flags( object ):
   """Keeps track of the flags necessary to compile a file.
   The flags are loaded from user-created python files (hereafter referred to as
@@ -279,12 +281,7 @@ def _RemoveUnusedFlags( flags, filename ):
   return new_flags
 
 
-# These are the standard header search paths that clang will use on Mac BUT
-# libclang won't, for unknown reasons. We add these paths when the user is on a
-# Mac because if we don't, libclang would fail to find <vector> etc.
-# This should be fixed upstream in libclang, but until it does, we need to help
-# users out.
-# See Valloric/YouCompleteMe#303 for details.
+# Include paths specific to MacOSX
 # It is possible that, later on, further such exceptions will need to be
 # added. The complete list of paths that need to be added can be retrieved
 # on a machine running MacOSX with the command:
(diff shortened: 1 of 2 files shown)


Valloric:
Thanks for the PR! :)

This needs tests though (probably in the `flags_test.py` file). You'll probably want to import and use `patch` and `call` from the `mock` module to avoid talking to the filesystem. (See `utils_test.py` for examples of usage. [More docs on `mock`.](http://www.voidspace.org.uk/python/mock/)) You might even end up using `assert_that` and the `contains` matcher from PyHamcrest ([docs](https://pypi.python.org/pypi/PyHamcrest)) to ensure the expected generated path is included. For examples of usage of PyHamcrest, see our `clang/diagnostics_test.py` file.

---

Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cpp/flags.py, line 296 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/385#-KB46qfnLeqBaKPWoW4G:-KB46qfnLeqBaKPWoW4H:2114673855) ([raw file](https://github.com/valloric/ycmd/blob/22251f9d7285c49a243735b3092698f5fd448fbc/ycmd/completers/cpp/flags.py#L296)):*
I'm fine with making this a function, but why not also have:

```python
MAC_INCLUDE_PATHS = _MacIncludePaths()
```

This way we cache the path calculations. It's not as if the paths are going to change across calls, right?

---

*[ycmd/completers/cpp/flags.py, line 297 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/385#-KB44zEkqJ_M3i2_1DxR:-KB44zElwpd5GlGq5nIx:1016725527) ([raw file](https://github.com/valloric/ycmd/blob/22251f9d7285c49a243735b3092698f5fd448fbc/ycmd/completers/cpp/flags.py#L297)):*
Two space indents please. :)

---

*[ycmd/completers/cpp/flags.py, line 309 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/385#-KB47ai7xqWrexQRuXY4:-KB47ai7xqWrexQRuXY5:397289504) ([raw file](https://github.com/valloric/ycmd/blob/22251f9d7285c49a243735b3092698f5fd448fbc/ycmd/completers/cpp/flags.py#L309)):*
It might be easier to test this if you took this path and made it a constant in the module. This way you could use `patch` to replace the value with a path to a test directory in our `testdata` dir.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/385#-:-KB457STAU1bEbu7gY4_:-1867520928)*

indianajohn:
Alright, so I added a test that mocks the same process in the testdata dir, made it so that the flags are cached in the module instead of evaluated each time. I used the way you suggested - by declaring the parent path as a variable in the module and using patch to substitute it for the test. I also fixed the indents.

Excuse the 3 commits - as you can see, it's really just one, but I made some last-minute changes and force-pushed.

Would you like me to squash the two commits into one? Any other things that need changed?

Thanks for your patience and work on the project.

Valloric:
Thanks for addressing the comments I made! :) I added a couple more.

---

Reviewed 1 of 1 files at r1, 3 of 3 files at r2.

---

*[ycmd/completers/cpp/flags.py, line 308 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/385#-KB92Ktl0mzdpLwvECr1:-KB92Ktl0mzdpLwvECr2:1249388496) ([raw file](https://github.com/valloric/ycmd/blob/ca106825747a2d559451de0b07bb04ad9bea2583/ycmd/completers/cpp/flags.py#L308)):*
Space after `(`.

Also, you seem to re-use the `CLANG_TOOLCHAIN_PARENT_PATH + '/' + leaf` expression two lines later, so maybe pull this into a variable above the IF?

---

*[ycmd/completers/cpp/flags.py, line 309 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/385#-KB92as_BPsxd4tzHV5F:-KB92as_BPsxd4tzHV5G:293507544) ([raw file](https://github.com/valloric/ycmd/blob/ca106825747a2d559451de0b07bb04ad9bea2583/ycmd/completers/cpp/flags.py#L309)):*
What happens when there's several directories here? Say, 7.0.2 and 7.0.3? Wouldn't we want the code to select the latest one instead of the first one?

---

*[ycmd/tests/clang/flags_test.py, line 231 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/385#-KB922TKMlc8ufbPKN5h:-KB922TKMlc8ufbPKN5i:1723753638) ([raw file](https://github.com/valloric/ycmd/blob/ca106825747a2d559451de0b07bb04ad9bea2583/ycmd/tests/clang/flags_test.py#L231)):*
Is this within 80 columns?

---

*[ycmd/tests/clang/flags_test.py, line 234 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/385#-KB93JKSmLXs8bSdu2tg:-KB93JKSmLXs8bSdu2th:799149813) ([raw file](https://github.com/valloric/ycmd/blob/ca106825747a2d559451de0b07bb04ad9bea2583/ycmd/tests/clang/flags_test.py#L234)):*
Space after `(`

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/385#-:-KB92y1_KKUQ0281hyAb:1901297569)*

puremourning:
FWIW I've tried this out on my mac (after fixing trivial merge conflict) and it works (i.e. automatically finds the system headers again). Thanks!

indianajohn:
Yup, sorry for the delay, I'm in the middle of an international move. I'll get to it tomorrow or the next day.

Valloric:
Closed in favor of #415.

Thanks for sending a PR anyway! :)

PR #389, opened 2016-02-24
Fix Boost.Python error in ycm_extra_conf.py

skogler:
Python loads the compilation database using unicode strings.
However, Boost.Python cannot handle this and throws an error.
This patch converts the paths to utf-8 strings before passing them on to C++.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/389)


skogler:
See also this link for other people having the same problem:
http://programfan.github.io/contents/2015/05/01/fix-boost-python-argument-error-lvalue/

Valloric:
The issue in question is #378.

The "correct" fix is to provide an overload to `GetCompilationInfoForFile` that accepts an Python unicode object, thus fixing this issue without any changes to extra conf files. I'll try to do that later today. (Closing this PR.)

skogler:
You are right, that would be better.

PR #396, opened 2016-02-25
[READY] Fix Python support tests on Windows

micbou:
Fix tests by using a specific `compile_commands.json` for Windows. Refactor them to reduce code duplication.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/396)

### DIFF
--- a/ycmd/tests/python_support_test.py
+++ b/ycmd/tests/python_support_test.py
@@ -23,12 +23,24 @@ from future import standard_library
 standard_library.install_aliases()
 # Intentionally not importing all builtins!
 
+import os
+
 from nose.tools import eq_
 from future.utils import native
 
 import ycm_core
-from ycmd.tests.test_utils import ClangOnly, Py2Only, Py3Only, PathToTestFile
-from ycmd.utils import ToBytes, ToUnicode
+from ycmd.tests.test_utils import ClangOnly, Py2Only, Py3Only
+from ycmd.utils import ToBytes, ToUnicode, OnWindows
+
+
+# We don't use PathToTestFile from test_utils module because this module
+# imports future modules that may change the path type.
+PATH_TO_TESTDATA = os.path.abspath( os.path.join( os.path.dirname( __file__ ),
+                                                  'testdata' ) )
+PATH_TO_COMPILE_COMMANDS = (
+  os.path.join( PATH_TO_TESTDATA, 'windows' ) if OnWindows() else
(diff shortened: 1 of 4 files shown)


Valloric:
I agree that we should have a separate `compile_commands.json` file for Windows support, but I'm not a fan of the test refactoring. Tests should simple be above all else; the new approach reduces some code duplication but it costs too much in code complexity. 

I don't think duplicated assertion code (which are just data declarations) is a big deal. IMO it's the non-assertion boilerplate in tests that needs to be factored out, not the assertions themselves.

micbou:
It looks like `CompilationDatabase_Py2Str_test` and `CompilationDatabase_Py3Bytes_test` are identicals (both use `ToBytes`). Should we merge them? 

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/396#-:-KBPBxavU7hA8b1I8poE:-1122705810)*

Valloric:
What would be a good name for both? We actually want to test two different things, native `str` on py2 and `bytes` on py3 that _just so happen_ to be the test code (if done carefully; the `b` prefix is needed). Does this mean we should actually merge them, I'm not sure. I'd lean towards keeping them separate because that one test would probably have too much nuance in it.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/396#-:-KBPNWgcI0n8zNsrfYkE:-1061336030)*

Valloric:
Reviewed 5 of 5 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/396)*

micbou:
Updated with a simpler approach.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/396#-:-KBPWgd35_XjJTppfodv:1722068727)*

Valloric:
I'm sold. :) Thanks for this!

@homu r+

---

Reviewed 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/396#-:-KBPdtAkTMLOWnMZvMJc:1911183092)*

PR #406, opened 2016-02-29
[READY] Prevent annoying on OS X El Capitan when using system python

puremourning:
# Problem

In OS X El Capitan, Apple introduced System Integrity Protection.  Amongst other things, this introduces features to the dynamic loader (dyld) which cause it to "sanitise" (and complain about) embedded LC_RPATH entries which contain @executable_path when then are loaded into "restricted" binaries.  For our purposes, "restricted" here means "supplied by Apple" and includes the system versions of python.  For unknown reasons, the libclang.dylib that comes from llvm.org includes an LC_RPATH entry '@executable_path/../lib' which causes the OS X dynamic loader to print a cryptic warning to stderr of the form:

> dyld: warning, LC_RPATH @executable_path/../lib in /path/to/ycmd/libclang.dylib being ignored in restricted program because of @executable_path

Note: This does not happen when using homebrew, pyenv or any other non-Apple versions of python.

# Mysteries

As far as I can tell, this warning/behaviour should have been present (at least) since OS X 10.11 (El Capitan), though it only started printing to the terminal (and interfering with terminal vim's default display) after commit 599de71575d542d4fc5c33344c65b3931e4d7ed2.

It is not clear precisely why this is now visible, but the change to remove `ycm_client_support.so` was the change when this started to *become* visible.

I'm almost certain that previously the warning (which is printed to stderr I believe) was getting eaten by the specific series of initialisations that was going on, but I have literally no evidence to this other than provably checking out `599de71575d542d4fc5c33344c65b3931e4d7ed2^` and building/running does not have the problem, whereas anything at `599de71575d542d4fc5c33344c65b3931e4d7ed2` or after it does.

# Solution

In order to prevent this harmless and annoying message appearing, we simply strip the rpath entry from the dylib.  There's no way any @executable_path that python might have could be in any way useful to libclang.dylib, and we always take a _copy_ of any system or externally supplied libclang, so this seems perfectly safe.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/406)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -326,6 +326,35 @@ if( LIBCLANG_TARGET )
       POST_BUILD
       COMMAND ${CMAKE_COMMAND} -E copy "${LIBCLANG_TARGET}" "$<TARGET_FILE_DIR:${PROJECT_NAME}>"
     )
+
+    if( APPLE )
+      # In OS X El Capitan, Apple introduced System Integrity Protection.
+      # Amongst other things, this introduces features to the dynamic loader
+      # (dyld) which cause it to "sanitise" (and complain about) embedded
+      # LC_RPATH entries which contain @executable_path when then are loaded
+      # into "restricted" binaries.  For our purposes, "restricted" here means
+      # "supplied by Apple" and includes the system versions of python.  For
+      # unknown reasons, the libclang.dylib that comes from llvm.org includes an
+      # LC_RPATH entry '@executable_path/../lib' which causes the OS X dynamic
+      # loader to print a cryptic warning to stderr of the form:
+      #
+      #    dyld: warning, LC_RPATH @executable_path/../lib in
+      #    /path/to/ycmd/libclang.dylib being ignored in restricted program
+      #    because of @executable_path
+      #
+      # In order to prevent this harmless and annoying message appearing, we
+      # simply strip the rpath entry from the dylib.  There's no way any
+      # @executable_path that python might have could be in any way useful to
+      # libclang.dylib, so this seems perfectly safe.
+      get_filename_component( LIBCLANG_TAIL ${LIBCLANG_TARGET} NAME )
+      add_custom_command( TARGET ${PROJECT_NAME}
+                          POST_BUILD
+                          COMMAND install_name_tool
+                          "-delete_rpath"
+                          "@executable_path/../lib"
+                          "$<TARGET_FILE_DIR:${PROJECT_NAME}>/${LIBCLANG_TAIL}"
+                        )
+    endif()
   endif()
 endif()
 


puremourning:
FWIW, this is the closest to 'documentation' of this feature I can find: https://developer.apple.com/library/mac/documentation/Security/Conceptual/System_Integrity_Protection_Guide/RuntimeProtections/RuntimeProtections.html

>  Any dynamic linker (dyld) environment variables, such as DYLD_LIBRARY_PATH, are purged when launching protected processes.

It's possible that this is not related to System Integrity Protection at all, and just some other security feature, but in any case the same logic (and solution) apply.

puremourning:
OK, well dyld is a bit open source, so [its source code](http://www.opensource.apple.com/source/dyld/dyld-360.18/src/dyld.cpp) seems to be the best type of documentation. I don't think this is the exact error I'm seeing, but the code below looks like it might be a good explanation:

```c++
		// executable_path cannot be in used in any binary in a setuid process rdar://problem/4589305
		if ( sProcessIsRestricted && !sProcessRequiresLibraryValidation )
			throwf("unsafe use of @executable_path in %s with restricted binary", context.origin);
```

puremourning:
OK the lack of precision was annoying me. The [actual code](http://www.opensource.apple.com/source/dyld/dyld-360.18/src/ImageLoaderMachO.cpp) looks like this:

```c++
                                else if ( (strncmp(path, "@executable_path", 16) == 0) && ((path[16] == '/') || (path[16] == '\0')) ) {
                                        if ( context.processIsRestricted && !context.processRequiresLibraryValidation ) {
                                                dyld::warn("LC_RPATH %s in %s being ignored in restricted program because of @executable_path\n", path, this->getPath());
                                                break;
                                        }
```

Valloric:
_Sigh_, yet another hack we have to carry forward. :( 

(Not saying there's a better way, just bitching.) 

Thanks for looking into this and fixing it! Must not have been fun.

:lgtm: 

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/406#-:-KBeoe-7rLhslvhkHYUO:-1963697926)*

puremourning:
It was a head scratcher. And, yes I may have ranted a bit on gitter ;)

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/406#-:-KBerRzc9rF8G1FVTtOR:-1941417541)*

puremourning:
I had a eureka moment about why this has just started happening. Or at least a guess.

Previously the chain of `dlopen` calls was:
 - `python` (protected)
 - `dlopen`ed `ycm_core.so` (not protected)
 - which `dlopen`ed `ycm_client_suppor.so`
 - which opened `libclang.dylib`

This meant that when the check was run, the "binary" was `ycm_core.so`, not `python`. Now:
 - `python` (protected)
 - `dlopen`s `ycm_core.so`
 - which opens `libclang.dylib`

I think that means that the `dlopen` check happens against `python` instead of `ycm_client_support.so`.

OK having written that out, I'm not 100% confident, but it might well be in the right ballpark of explanations.

micbou:
:lgtm: with minor comment.

---

Reviewed 1 of 1 files at r1.

---

*[cpp/ycm/CMakeLists.txt, line 328 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/406#-KBh6_K13vWvGdFJuOYh:-KBh6_K2eNu3-9lJ0yCi:-370334077) ([raw file](https://github.com/valloric/ycmd/blob/88f348c2cf20a05755c7aab9dc6b1f8f812892f6/cpp/ycm/CMakeLists.txt#L328)):*
I would add a blank line here...

---

*[cpp/ycm/CMakeLists.txt, line 348 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/406#-KBh6bLdsaBva3rgy-op:-KBh6bLet6JqeCKV98Ky:344509902) ([raw file](https://github.com/valloric/ycmd/blob/88f348c2cf20a05755c7aab9dc6b1f8f812892f6/cpp/ycm/CMakeLists.txt#L348)):*
... and remove this one.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/406#-:-KBh6ejtUdQRWeBkoUEF:-1120555899)*

puremourning:
Updated with new whitespacing :)

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[cpp/ycm/CMakeLists.txt, line 328 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/406#-KBh6_K13vWvGdFJuOYh:-KBoqzpeYJyX7JlNiWsu:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/88f348c2cf20a05755c7aab9dc6b1f8f812892f6/cpp/ycm/CMakeLists.txt#L328)):*
Done.

---

*[cpp/ycm/CMakeLists.txt, line 348 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/406#-KBh6bLdsaBva3rgy-op:-KBor-3Kfj-LXr8FNLu1:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/88f348c2cf20a05755c7aab9dc6b1f8f812892f6/cpp/ycm/CMakeLists.txt#L348)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/406#-:-KBor-zNophm9gQ3no8c:1578810661)*

micbou:
@homu r+

---

Reviewed 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/406#-:-KBosEd7XsVoXa-nPnV8:-1142673777)*

puremourning:
@homu retry

puremourning:
There seems to be some transient problem with travis

PR #339, opened 2016-02-02
Persistent Omnisharp in tests

mispencer:
Running all the tests in ycmd takes a long time. The reason for this is
repeatedly starting and stopping Omnisharp servers. To mitigated this, whenever
we start an instance of Omnisharp during testing, we cache it, and reuses it
from the cache whenever a test requires a instance of Omnisharp on the same
solution.

As part of this, add Set/GetOmnisharpPort commands to the C# completer. These
commands allow setting the port it uses to communicate with the Omnisharp
server. This allows sharing of the Omnisharp server by retrieving the port and
using it in other clients, or get the port another client is using and using it
here.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/339)

PR #409, opened 2016-02-29
[RFC] Rewrite tests to reuse server state

### LINKED ISSUE #338: Refactor tests to reuse semantic servers
We currently bring up a new instance of omnisharp, gocode, racerd etc for every single test that needs them. It makes tests take _forever_, especially the omnisharp tests. 

We need to figure out a way to reuse the server instances once they're up without making tests clobber each other. If we only manage to get C# tests to reuse omnisharp, I'll call that a win.

@puremourning @micbou @vheon Any takers? :)

micbou:
This PR rewrites tests from completers using a server by separating them into two classes:
 - isolated tests: each test use its own server instance;
 - persistent tests: use the same server instance.

We use classes because it provides a clean way to do this (no global variables needed).

This PR also simplifies the C♯ tests by adding filepaths used to start the server to a list and stopping the server for each filepath in the list at the end of the tests. With this approach, we don't need to expose additional functions in the C♯ completer (`GetOmnisharpPort` and `SetOmnisharpPort` are not needed). Furthermore, we don't have the issue with persistent tests (introduced by PR #339) on Windows and Python 3. See PR #365 for details.

And now, the interesting part; the execution times on different CI configurations before and after the changes:

Configuration | Before (s) | After (s)
------------- | -----------| ---------
AppVeyor Python 2.7 | 58.4 | 28.8
AppVeyor Python 3.5 | 161.1 | 32.8
Travis Linux | 49.0 | 32.6
Travis OS X | 59.4 | 42.8 

Closes #338.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/409)


Valloric:
Holy crap this is AWESOME! :D Thanks so much!

I have some minor comments, but otherwise :lgtm:; showing me test speed improvements makes me instantly like a PR. :)

@puremourning You probably have some opinions on this PR. I'd love to hear them.

---

Review status: 0 of 38 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/tests/get_completions_test.py, line 46 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KBjkzY1LZD8-nSKm60Y:-KBjkzY2yohCneONKKQh:-1172023737) ([raw file](https://github.com/valloric/ycmd/blob/63bab42be2dcdf7c77870081efe38480cf08ab32/ycmd/tests/get_completions_test.py#L46)):*
Seems like these three lines could be put into one free function that takes `handlers.app` and returns `TestApp`. It would reduce code duplication.

---

*[ycmd/tests/cs/cs_handlers_test.py, line 179 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KBjlHW3ZjlC9d50_e_w:-KBjlHW3ZjlC9d50_e_x:-1846173182) ([raw file](https://github.com/valloric/ycmd/blob/bd569e9d0696f8de540b185465277cebb3589574/ycmd/tests/cs/cs_handlers_test.py#L179)):*
I loooove to see code getting deleted! It's the best thing one can do with code.

---

*[ycmd/tests/cs/cs_persistent_test.py, line 1 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KBjlzyqv8rbki5L_Jdi:-KBjlzyqv8rbki5L_Jdj:681625430) ([raw file](https://github.com/valloric/ycmd/blob/63bab42be2dcdf7c77870081efe38480cf08ab32/ycmd/tests/cs/cs_persistent_test.py#L1)):*
Not sure about the isolated/persistent nomenclature. Isolated is fine, but perhaps "shared" instead of "persistent"?  

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/409#-:-KBjkQZWv9Ajtgrwadw8:916470929)*

Valloric:
Reviewed 41 of 41 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/409)*

vheon:
Reviewed 41 of 41 files at r1.

---

*[ycmd/tests/cs/cs_persistent_test.py, line 62 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KBo58vxHzTEVcKfkcSx:-KBo58vxHzTEVcKfkcSy:1382854873) ([raw file](https://github.com/valloric/ycmd/blob/63bab42be2dcdf7c77870081efe38480cf08ab32/ycmd/tests/cs/cs_persistent_test.py#L62)):*
Aren't we starting an OmniSharpServer for each file? Shouldn't we start a OmniSharpServer for each solution?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/409)*

micbou:
Reviewed 3 of 47 files at r1, 19 of 24 files at r2.
Review status: 33 of 38 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/tests/get_completions_test.py, line 46 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KBjkzY1LZD8-nSKm60Y:-KBo_rjyYErlUyT5qjh-:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/63bab42be2dcdf7c77870081efe38480cf08ab32/ycmd/tests/get_completions_test.py#L46)):*
Done.

---

*[ycmd/tests/cs/cs_shared_test.py, line 1 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KBjlzyqv8rbki5L_Jdi:-KBo_zlm1OkDLUucp3Y0:-1890050697) ([raw file](https://github.com/valloric/ycmd/blob/63bab42be2dcdf7c77870081efe38480cf08ab32/ycmd/tests/cs/cs_shared_test.py#L1)):*
Yes, "shared" is better. I changed it.

---

*[ycmd/tests/cs/cs_persistent_test.py, line 62 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KBo58vxHzTEVcKfkcSx:-KBoBC2_dEE3bXow6KwQ:1610781336) ([raw file](https://github.com/valloric/ycmd/blob/63bab42be2dcdf7c77870081efe38480cf08ab32/ycmd/tests/cs/cs_persistent_test.py#L62)):*
No, the completer automatically find the solution for `filepath` and start a server for this solution only if not already started.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/409)*

puremourning:
@valloric I took a quick scan through. The only concern is just about maintainability. Having practically all the tests in a single class/file just makes a single large file where all the tests are, which is a bit of a god class pattern. I don't think this is the end of the world, but there was something quite nice about having the separation of subcommand, completion, diagnostics, etc. where that makes sense. 

It also wasn't completely obvious to me exactly what made a test fit in the "isolated" vs. "shared" buckets. Do we have a rule of thumb about when we should put tests in each?

We just have to decide if 20s per test run is worth this particular architecture. I don't really have a strong opinion on it, though I am in favour of obvious organisation within test suites as this leads to less duplication and (hopefully) more accessibility to newcomers.

---

Reviewed 1 of 47 files at r1.
Review status: 33 of 38 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/409#-:-KBooXp1xw5Z1DihFe9r:1149528925)*

Valloric:
@puremourning I must admit I didn't notice that this merges various test files into one. I too very much liked our separation of subcommand, diags etc test files. 

@micbou Any way we could keep that organization? I'd like to have our cake and eat it too if possible.

---

Review status: 33 of 38 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/409#-:-KBpA6WJFROeCMSJgIZ3:-416529917)*

micbou:
So, I was able to keep the current organization by doing this at a package level. This means that we are using global variables but it doesn't look too bad.

I added two decorators `@Isolated` and `@Shared` (defined for the tests in `ycmd/tests` and each completer tests) to make it easy to write tests. This way, we just need to attach the decorator and add the parameter `app` to the test. Important: if this is a [test generator](http://nose.readthedocs.org/en/latest/writing_tests.html#test-generators), it must be attached to the yielded test. See the code for examples.

I removed the classes because they don't bring any value to this approach.

> It also wasn't completely obvious to me exactly what made a test fit in the "isolated" vs. "shared" buckets. Do we have a rule of thumb about when we should put tests in each?

A test should be isolated if it changes the server state in a irreversible way (example: a semantic server is stopped or restarted) or if it expects a clean state (example: no semantic server started, no `.ycm_extra_conf.py` loaded, etc).

---

Reviewed 2 of 47 files at r1, 3 of 24 files at r2, 46 of 52 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/409#-:-KC7CnnTgNSr_Lafw2T9:843104259)*

Valloric:
@micbou Thanks for taking another stab at this! It's a tremendous amount of work that few would enjoy taking on so it's very much appreciated.

I love the new decorators, though I think it might be clearer to name them `@SharedYcmd` and `@IsolatedYcmd` because it's not clear at-a-glance what exactly is being shared (for isolated, it's clearer).

I love that the test classes have gone away; I was never really a huge fan of them.

Some inline comments, but :lgtm: from me.

---

---

*[ycmd/tests/__init__.py, line 39 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KC7jt3fe9f3pVnHA0Wu:-KC7jt3fe9f3pVnHA0Wv:-1021327512) ([raw file](https://github.com/valloric/ycmd/blob/1ad7c688de0da2e9aaf460af8d727fe41f1de345/ycmd/tests/__init__.py#L39)):*
Seems like this might benefit from some docs. Similar with `Shared`.

---

*[ycmd/tests/clang/__init__.py, line 63 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KC7l1Qf254CjyOVsPjn:-KC7l1Qf254CjyOVsPjo:865691415) ([raw file](https://github.com/valloric/ycmd/blob/1ad7c688de0da2e9aaf460af8d727fe41f1de345/ycmd/tests/clang/__init__.py#L63)):*
Are these empty lines in `Wrapper` needed? They don't seem to improve readability; it's 4 lines. :)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/409#-:-KC7k6iM471fCcp7xw1S:797227692)*

puremourning:
Holy crap, I *love* this change. Thanks *so* much. :lgtm: with couple of comments - i think we need to check that we're properly clearing down after an exception is thrown by our wrapper.

---

Reviewed 1 of 47 files at r1, 6 of 24 files at r2, 52 of 52 files at r3.

---

*[ycmd/tests/clang/__init__.py, line 58 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KC7o3Pf0j-k_kmTa7_D:-KC7o3Pf0j-k_kmTa7_E:263248621) ([raw file](https://github.com/valloric/ycmd/blob/1ad7c688de0da2e9aaf460af8d727fe41f1de345/ycmd/tests/clang/__init__.py#L58)):*
if SetUpApp() of the function raise an exception i don't think this gets reset. Is that a problem? Should we be doing a `try: finally:` here?

---

*[ycmd/tests/javascript/__init__.py, line 116 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KC7p4Wi-09Zl2vX2ntb:-KC7p4Wi-09Zl2vX2ntc:-1478635826) ([raw file](https://github.com/valloric/ycmd/blob/1ad7c688de0da2e9aaf460af8d727fe41f1de345/ycmd/tests/javascript/__init__.py#L116)):*
Same here, I think we should either be using a context manager or `try/finally` in here (to reset `cwd`)

---

*[ycmd/tests/cs/cs_isolated_test.py, line 0 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KC7pds7khO8iMN9T1k6:-KC7pds7khO8iMN9T1k7:-818203167) ([raw file](https://github.com/valloric/ycmd/blob/a489ae73f41d174fff60cde2c042c25ff2a32729/ycmd/tests/cs/cs_isolated_test.py#L0)):*
Could be a reviewable.io bug/arefact, but I think we now have an empty test file which we probably don't need.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/409#-:-KC7qAjGqzG2D9BtqUq5:437505109)*

micbou:
> I love the new decorators, though I think it might be clearer to name them @SharedYcmd and @IsolatedYcmd because it's not clear at-a-glance what exactly is being shared (for isolated, it's clearer).

I renamed them.

---

Reviewed 4 of 52 files at r3, 22 of 30 files at r4, 13 of 13 files at r5.

---

*[ycmd/tests/__init__.py, line 39 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KC7jt3fe9f3pVnHA0Wu:-KCCmzuWZpKqZQS-0xKL:1060069661) ([raw file](https://github.com/valloric/ycmd/blob/1ad7c688de0da2e9aaf460af8d727fe41f1de345/ycmd/tests/__init__.py#L39)):*
Added documentation to `setUpPackage`, `tearDownPackage`, `SharedYcmd`, and `IsolatedYcmd` functions (when they exist).

---

*[ycmd/tests/clang/__init__.py, line 58 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KC7o3Pf0j-k_kmTa7_D:-KCCnHSz6NjoQdaq5ir4:-1787246838) ([raw file](https://github.com/valloric/ycmd/blob/1ad7c688de0da2e9aaf460af8d727fe41f1de345/ycmd/tests/clang/__init__.py#L58)):*
You are absolutely right. I added a `try/finally` block.

---

*[ycmd/tests/clang/__init__.py, line 63 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KC7l1Qf254CjyOVsPjn:-KCCnQor0-MeXNtcNGKs:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/1ad7c688de0da2e9aaf460af8d727fe41f1de345/ycmd/tests/clang/__init__.py#L63)):*
Done.

---

*[ycmd/tests/javascript/__init__.py, line 116 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KC7p4Wi-09Zl2vX2ntb:-KCCpCAMA8wfeEigjst0:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/1ad7c688de0da2e9aaf460af8d727fe41f1de345/ycmd/tests/javascript/__init__.py#L116)):*
Done.

---

*[ycmd/tests/cs/cs_isolated_test.py, line  \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/409#-KC7pds7khO8iMN9T1k6:-KCCpF43qc9bncvWhJcU:779067302) ([raw file](https://github.com/valloric/ycmd/blob/a489ae73f41d174fff60cde2c042c25ff2a32729/ycmd/tests/cs/cs_isolated_test.py#L)):*
No idea why it is here but it says `File not in pull request at selected revision` and I am sure it is not in the PR.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/409#-:-KCCp2BDKKdqyoCRTyPk:-1825650025)*

Valloric:
This looks like it's good to go. Thanks again!

@homu r+

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/409#-:-KCCq8L3wJonCNs3iFlK:284336125)*

PR #412, opened 2016-03-02
Closes #411 Fixed python2 ivalid encoding argument for builtin open

### LINKED ISSUE #411: ycmd crashes after update
After last update ycmd crashes with following traceback.
Reason is python2 open has no encoding keyword argument.
I suggest use io.open in PY2/PY3
```python
  File "<string>", line 19, in <module>                                                                                                                
  File "/home/nes/config-files/.vim/bundle/YouCompleteMe/autoload/../python/ycm/setup.py", line 38, in SetUpYCM                                        
    base.LoadJsonDefaultsIntoVim()                                                                                                                     
  File "/home/nes/config-files/.vim/bundle/YouCompleteMe/autoload/../python/ycm/base.py", line 46, in LoadJsonDefaultsIntoVim                          
    defaults = user_options_store.DefaultOptions()                                                                                                     
  File "/home/nes/config-files/.vim/bundle/YouCompleteMe/python/ycm/../../third_party/ycmd/ycmd/user_options_store.py", line 55, in DefaultOptions     
    options = json.loads( ReadFile( settings_path ) )                                                                                                  
  File "/home/nes/config-files/.vim/bundle/YouCompleteMe/python/ycm/../../third_party/ycmd/ycmd/utils.py", line 55, in ReadFile                        
    with open( filepath, encoding = 'utf8' ) as f:                                                                                                     
TypeError: 'encoding' is an invalid keyword argument for this function    
```

m-novikov:
[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/412)

### DIFF
--- a/ycmd/utils.py
+++ b/ycmd/utils.py
@@ -24,6 +24,7 @@ standard_library.install_aliases()
 from builtins import *  # noqa
 from future.utils import PY2, native
 
+import io
 import tempfile
 import os
 import sys
@@ -51,7 +52,7 @@ ACCESSIBLE_TO_ALL_MASK = ( stat.S_IROTH | stat.S_IWOTH | stat.S_IXOTH |
 # doesn't get closed. So, a helper func.
 # Also, all files we read are UTF-8.
 def ReadFile( filepath ):
-  with open( filepath, encoding = 'utf8' ) as f:
+  with io.open( filepath, encoding = 'utf8' ) as f:
     return f.read()
 
 


micbou:
This should not be needed because this line
```python
from builtins import *
```
in `ycmd/utils.py` takes care of that. It imports a [custom open from `python-future`](https://github.com/PythonCharmers/python-future/blob/6b8341c0412ddf8081151692e9034986a6f867bc/src/future/builtins/misc.py#L43).

m-novikov:
Then probably it should be
```python
from future.builtins import *
```
builtins is a python module

micbou:
> Then probably it should be
> from future.builtins import *

Not the recommended way [according to the documentation](http://python-future.org/quickstart.html).

> builtins is a python module

Not on a default python 2 installation. In the python interpreter:
```python
Python 2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 20:40:30) [MSC v.1500 64 bit (AMD64)] on win32
>>> import builtins
  File "<stdin>", line 1, in <module>
ImportError: No module named builtins
```

m-novikov:
My bad some package from pip added file builtins.py in dist-packages which caused python-futures monkey patching to fail.

PR #423, opened 2016-03-09
[RFC] TypeScript: get detailed completion after filtered candidates < 100

icholy:
I haven't really tested this ... 

https://github.com/Valloric/YouCompleteMe/issues/2040
https://github.com/Valloric/YouCompleteMe/issues/1878

---
This change is [Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/423)

### DIFF
--- a/ycmd/completers/completer.py
+++ b/ycmd/completers/completer.py
@@ -168,8 +168,7 @@ class Completer( with_metaclass( abc.ABCMeta, object ) ):
 
     candidates = self._GetCandidatesFromSubclass( request_data )
     if request_data[ 'query' ]:
-      candidates = self.FilterAndSortCandidates( candidates,
-                                                 request_data[ 'query' ] )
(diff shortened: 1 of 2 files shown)


icholy:
I'm going to dog-food this for a few days and after that I'll clean it up. For now I just want to know if there's anything fundamentally wrong with my approach.

icholy:
It works, but there's an introduced lag when it fetches the additional details. It's very noticeable (for me at least). I'm thinking that grabbing the details in a background thread would make it snappier, but that's going to get really complicated. I'm thinking it's just not worth it.

Valloric:
@icholy You have the most experience with this, so I'll defer to your opinion.

icholy:
I'm giving up on this approach for now. IMO it's too much complexity for a marginal improvement in UX.

PR #424, opened 2016-03-09
Use current python from environment when launch jedihttp

allangarret:
Here we always use the vim python interpreter when we launch the
jedihttp unless python_binary_path is defined. Now, we use the python
interpreter from the environment, so if we use virtualenv, the
jedihttp is launched with the right python version and the completion
works out of the box for python2/3 with virtualenv

---
This change is [Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/424)

### DIFF
--- a/ycmd/completers/python/jedi_completer.py
+++ b/ycmd/completers/python/jedi_completer.py
@@ -37,7 +37,6 @@ import logging
 import urllib.parse
 import requests
 import threading
-import sys
 import os
 from distutils.spawn import find_executable
 
--- a/ycmd/completers/python/jedi_completer.py
+++ b/ycmd/completers/python/jedi_completer.py
@@ -39,6 +39,7 @@ import requests
 import threading
 import sys
 import os
+from distutils.spawn import find_executable
 
 
 HMAC_SECRET_LENGTH = 16
@@ -69,7 +70,7 @@ class JediCompleter( Completer ):
     self._logfile_stderr = None
     self._keep_logfiles = user_options[ 'server_keep_logfiles' ]
     self._hmac_secret = ''
-    self._python_binary_path = sys.executable
+    self._python_binary_path = find_executable('python')
 
     self._UpdatePythonBinary( user_options.get( 'python_binary_path' ) )
     self._StartServer()
--- a/ycmd/tests/python/user_defined_python_test.py
+++ b/ycmd/tests/python/user_defined_python_test.py
@@ -26,7 +26,7 @@ from builtins import *  # noqa
 from hamcrest.core.base_matcher import BaseMatcher
 from hamcrest import assert_that, has_item, contains, equal_to, is_not  # noqa
 from mock import patch
-import sys
+from distutils.spawn import find_executable
 
 from ycmd import utils
 from ycmd.completers.python.jedi_completer import BINARY_NOT_FOUND_MESSAGE
@@ -81,7 +81,7 @@ def was_called_with_python( python ):
 @patch( 'ycmd.utils.SafePopen' )
 def UserDefinedPython_WithoutAnyOption_DefaultToYcmdPython_test( app, *args ):
   app.get( '/ready', { 'subserver': 'python' } )
-  assert_that( utils.SafePopen, was_called_with_python( sys.executable ) )
+  assert_that( utils.SafePopen, was_called_with_python( find_executable('python') ) )
 
 
 @IsolatedYcmd
@@ -121,7 +121,7 @@ def UserDefinedPython_RestartServerWithoutArguments_WillReuseTheLastPython_test(
   request = BuildRequest( filetype = 'python',
                           command_arguments = [ 'RestartServer' ] )
   app.post_json( '/run_completer_command', request )
-  assert_that( utils.SafePopen, was_called_with_python( sys.executable ) )
+  assert_that( utils.SafePopen, was_called_with_python( find_executable('python') ) )
 
 
 @IsolatedYcmd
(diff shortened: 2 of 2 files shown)


Valloric:
Thanks for the PR!

This seems reasonable to me; @vheon what do you think?

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/424#-:-KCRUtMoxWspkfwE2jQg:1243534355)*

vheon:
I'm ok with using the PATH for searching the `python` executable but I'm not sure with the current implementation and using it as a default. What I would do is leave the default as `sys.executable`; this is because the Python completer is the only one not configurable inside `ycmd` so everyone has it but not everyone use it so to mitigate possible problems I don't want the default to be dependable on the user's environment, but using the user `PATH` when checking the `python_binary_path` executable is what I wanted to do (but didn't find the time to do it :sob: ) , so if the user what to spawn JediHTTP using whatever `python` executable is first find into the `PATH` they have to say it explicitely using `python_binary_path = 'python'`. What do you think?

@allangarret To make the current implementation more "Windows friendly" use the `utils.FindExecutable` or `utils.PathToFirstExistingExecutable` function.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/424#-:-KCS8pbOt9iRxSXOqxeK:-1336798128)*

allangarret:
I can leave the default as `sys.executable` and search in `PATH` if the user explicitly use `python_binary_path = 'python'` as you want, but I'm still thinking that if the user runs vim inside a virtualenv is because she/he wants to perform the autocompletion against this virtualenv. Maybe, I can check if the environment variable `VIRTUAL_ENV`exists and use the python version from `PATH` in such case. At the end, I think we could check the python version like this:
```python
def __init__(self, user_options):
    self._python_binary_path = sys.executable
    self._UpdatePythonBinary( user_options.get( 'python_binary_path' ) )

def _UpdatePythonBinaryPath(self, binary):
    if binary:
        if binary == 'python':
            # resolve python using path
        else:
            # check if user supplied python exists
    elif 'VIRTUAL_ENV' in os.environ:
        # resolve python using path
```

@vheon what do you think?

Also, I will use `utils.FindExecutable`

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/424#-:-KCWdhcaBmmh1JQxsrlw:-1609473639)*

vheon:
Honestly it seems a little too magic, that is why I am in favour of just saying:

- default to `sys.executable`
- `python_binary_path` is used to spawn a JediHTTP respecting the user's PATH

It seems simpler to understand and not that hard to configure (e.g. if you want your behaviour just put `python_binary_path` in your config, or if you know you don't work on multiple project at the same time just put the full path to the one you want). Anyway the fact that I don't like that it doesn't mean that I'm right, so other feedback is more than welcome, after all I'm not a python programmer so I might have a distorted vision on how a dev would work. If I'm not mistaken I had a prototype of the approach I wanted to test, but the first free time I have is this weekend, so If you're patient for a couple of days we can compare the results :)

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/424#-:-KCXSMs2FUU_UPs_-lav:-980432762)*

allangarret:
So, something like this is ok to you?

```python
def __init__(self, user_options):
    self._python_binary_path = sys.executable
    self._UpdatePythonBinary( user_options.get( 'python_binary_path' ) )

def _UpdatePythonBinaryPath(self, binary):
    if binary:
        self._python_binary_path = utils.FindExecutable(binary)
```

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/424#-:-KCaj_mm5jEw_b1HUG0T:992766838)*

vheon:
That would not work if you specify the full path. But thinking about it more to make it work like I wanted the only thing to do is to change the `CheckBinaryExists` method. I'm working on a prototype right now, I'll let you know.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/424#-:-KCanzmyXUBpUCryqn7D:1722969597)*

vheon:
I've just pushed a commit into one of my branches https://github.com/vheon/ycmd/tree/check-python-interpreter with the approach I was talking about. I've taken the `IsPythonVersionCorrect` from YouCompleteMe, so if we think this is a valid and functional approach I will grab the relative tests and change YouCompleteMe to use the one from `ycmd` but we'll see :)

vheon:
Maybe I have another solution which is similar to what @allangarret is doing, I'll let you know :+1:

allangarret:
It's nice, but we could add a little magic to make it better:

```python
def _UpdatePythonBinaryPath(self, binary):
    if binary:
        if os.path.isabs( binary ):
            binary_path = binary
        else:
            binary_path = utils.FindExecutable( binary )
        if not self._CheckBinaryExists( binary_path ):
            msg = BINARY_NOT_FOUND_MESSAGE.format( binary_path )
            self._logger.error( msg )
            raise RuntimeError( msg )
        self._python_binary_path = binary_path
```
This way, works with `python_binary_path = /some/random/python` and `python_binary_path = python`. Also, it will work on Windows too

allangarret:
In the previous code, you can replace the `_CheckBinaryExists` function with your new implementation

vheon:
> This way, works with python_binary_path = /some/random/python and python_binary_path = python. Also, it will work on Windows too

That is basically the other approach I was talking about and that I tried to implement. The results is in #429. We can discuss which one we can consider to merge in :+1:

allangarret:
I think we should close this and merge #429 instead

Valloric:

micbou:
Closed by PR #429.

PR #336, opened 2016-02-02
[READY] More travis run speedups

Valloric:
- Using clang instead of gcc; compiles faster and with less memory,
meaning we can build in parallel.
- Caching more stuff so startup is faster. In particular, the Rust
toolchain.
- Using the ninja build tool which tends to be faster than Make.

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/336)

PR #431, opened 2016-03-14
always use ninja to compile if on the path

stanionascu:
allows to use ninja for building ycmd with gcc in msys64 environment and use ninja with MS compiler:

to build using ninja on windows for msvc 2015:
```
"%VS140COMNTOOLS%\..\..\VC\vcvarsall.bat" amd64
set PATH=%PATH%;c:\path\to\cmake-and-ninja;c:\path\to\python
python build.py --all
```

to build in msys64 simply:
```
./build.py --all
```

---
This change is [Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/431)

### DIFF
--- a/build.py
+++ b/build.py
@@ -203,7 +203,9 @@ def CustomPythonCmakeArgs():
 
 
 def GetGenerator( args ):
-  if OnWindows():
+  if PathToFirstExistingExecutable( ['ninja'] ):
+    return 'Ninja'
+  elif OnWindows():
     if args.msvc == 14:
       generator = 'Visual Studio 14'
     elif args.msvc == 12:
@@ -216,8 +218,6 @@ def GetGenerator( args ):
       generator = generator + ' Win64'
     return generator
 
-  if PathToFirstExistingExecutable( ['ninja'] ):
-    return 'Ninja'
   return 'Unix Makefiles'
 
 


Valloric:
This is up to @micbou whether he wants to support ninja on Windows.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/431#-:-KCqU4qXlHlUXvkko4Do:1793641181)*

vheon:
I remember that @micbou tried Ninja on Windows and said that we were talking about 10s faster build, which isn't that much, but I'm guessing this is more for msys64 than for MSVC :confused: 

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/431#-:-KCrLG4bqiptxlTsDOwH:-389089924)*

micbou:
I don't really want to support it for the following reasons:
 - Ninja was added to speed up runs on Travis, not for a normal use. See PR #336;
 - additional configuration is needed from the user to make it work with MSVC (`vcvarsall.bat` script);
 - building speedup is small with MSVC (9s on my machine for a compilation run of 1m10);
 - building on MSYS is irrelevant because we don't support it;
 - not officially supported on Windows by CMake. From the Ninja docs:

> A widely used meta-build system that can generate Ninja files on Linux as of CMake version 2.8.8. (There is some Mac and Windows support — ReactOS uses Ninja on Windows for their buildbots, but those platforms are not yet officially supported by CMake as the full test suite doesn’t pass.) 

 - can't make it work on AppVeyor.

Thanks for the PR though.
 

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/431#-:-KCqie5SJp7MWO5pti4D:-1612701515)*

stanionascu:
Indeed it was to get ycmd to build in msys+gcc w/o patching build.py and as a side effect making msvc build faster.
according to CMake docs, since 3.3 Ninja generator is no longer marked as experimental.

Thanks for checking.

PR #291, opened 2016-01-02
[READY] C/C++ semantic highlighting.

davits:
Hi Guys and Happy New Year,
I've decided to turn on Christmas lights in the Vim for the occasion :)

This is a proof of  a concept and lot more polishing needs to be done.
I just wanted to share the initial results, and see your concerns and remarks about the current implementation and the direction it is going. The process of extraction and usage of semantic tokens is pretty much similar to the one with diagnostics, so I implemented semantics by changing return value of file reparse function to be pair of diagnostics and semantics.
For now it supports/returns tokens of the following types: namespace, class, struct, union, typedef, enum, enum constant, macro, function, function argument. And all these token types can be linked to the different highlighting groups.

PS
I will create another pull request for YouCompleteMe, so you can test this.

PPS
Here are some screenshots to please your eyes. (Note that bold version of Type used to highlight user defined types, and bold version of Normal to highlight function arguments).
Image
Image
Image

[Review on Reviewable](https://reviewable.io/reviews/valloric/ycmd/291)

PR #433, opened 2016-03-16
[READY] Include headers from external projects last

micbou:
Currently, headers are included by CMake in this order (when compiling with Clang support):
 - BoostParts
 - Python
 - Clang
 - `cpp\ycm`
 - `cpp\ycm\ClangCompleter`

The issue with this order is that if one of our headers has the same name as one of BoostParts, Python, or Clang headers, the wrong header may be included (depending on how it is included in the code). It can't happen with GCC and Clang compilers because the `-isystem` flag is used for these includes (`-isystem` directories are searched after `-I` directories). However, it is possible on Windows since there is no `-isystem` equivalent for MSVC. See issue in PR #291 where Python `token.h` is included instead of ycmd `Token.h`.

We prevent this by including the Boost, Python, and Clang directories after the ycmd ones.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/433)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -199,17 +199,6 @@ else()
   set( Boost_LIBRARIES BoostParts )
 endif()
 
-# The SYSTEM flag makes sure that -isystem[header path] is passed to the
-# compiler instead of the standard -I[header path]. Headers included with
-# -isystem do not generate warnings (and they shouldn't; e.g. boost warnings are
-# just noise for us since we won't be changing them).
-include_directories(
-  SYSTEM
-  ${Boost_INCLUDE_DIR}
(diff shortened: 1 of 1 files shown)


Valloric:
:lgtm:

Thanks!

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/433#-:-KD-PJOKVboKxz6Tcfe5:563354983)*

vheon:
:lgtm: @homu r=Valloric

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/433#-:-KD04XZ1t4mcGITDRtQF:583508563)*

PR #434, opened 2016-03-17
Removed comment about clients building ycmd

akirakyle:
It seems that all the listed known ycmd clients do in fact require the user to build and set up ycmd on their systems (except for YouCompleteMe, but even that still requires occasional recompilation). Removing this comment reduces potential confusion for people seeking to set up their ycmd client as most client's install instructions link to this section.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/434)

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,6 @@ built one.
 
 Building
 --------
-
-[Clients commonly build and set up ycmd for you; you are unlikely to need to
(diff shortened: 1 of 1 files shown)


Valloric:
Thanks for the PR!

I don't like it, but you have a point.

PR #439, opened 2016-03-23
[WIP] Assume column number in request is codepoint indexed

micbou:
See PR Valloric/YouCompleteMe#2073.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/439)

### DIFF
--- a/ycmd/request_wrap.py
+++ b/ycmd/request_wrap.py
@@ -23,7 +23,6 @@ from future import standard_library
 standard_library.install_aliases()
 from builtins import *  # noqa
 
-from ycmd.utils import ToUnicode, ToBytes
 from ycmd.identifier_utils import StartOfLongestIdentifierEndingAtIndex
 from ycmd.request_validation import EnsureRequestValid
 
@@ -101,18 +100,8 @@ def CompletionStartColumn( line_value, column_num, filetype ):
     foo.bar^
   with the cursor being at the location of the caret (so the character *AFTER*
   'r'), then the starting column would be the index of the letter 'b'."""
-  # NOTE: column_num and other numbers on the wire are byte indices, but we need
-  # to walk codepoints for identifier checks.
-
-  utf8_line_value = ToBytes( line_value )
-  unicode_line_value = ToUnicode( line_value )
-  codepoint_column_num = len(
-      str( utf8_line_value[ : column_num - 1 ], 'utf8' ) ) + 1
 
   # -1 and then +1 to account for difference betwen 0-based and 1-based
   # indices/columns
-  codepoint_start_column = StartOfLongestIdentifierEndingAtIndex(
-      unicode_line_value, codepoint_column_num - 1, filetype ) + 1
-
-  return len(
-      unicode_line_value[ : codepoint_start_column - 1 ].encode( 'utf8' ) ) + 1
+  return StartOfLongestIdentifierEndingAtIndex(
+      line_value, column_num - 1, filetype ) + 1
--- a/ycmd/tests/request_wrap_test.py
+++ b/ycmd/tests/request_wrap_test.py
@@ -96,8 +96,8 @@ def StartColumn_Dot_test():
 
 
 def StartColumn_DotWithUnicode_test():
-  eq_( 7,
-       RequestWrap( PrepareJson( column_num = 11,
+  eq_( 5,
+       RequestWrap( PrepareJson( column_num = 8,
                                  contents = 'fäö.bär') )[ 'start_column' ] )
 
 
@@ -122,7 +122,7 @@ def StartColumn_AfterWholeWord_Html_test():
 
 def StartColumn_AfterWholeUnicodeWord_test():
   eq_( 1,
-       RequestWrap( PrepareJson( column_num = 6,
+       RequestWrap( PrepareJson( column_num = 4,
                                  contents = u'fäö') )[ 'start_column' ] )
 
 


vheon:
This way we would break all the clients, right?

micbou:
All clients are already broken for unicode characters. For clients sending byte-indexed column like YCM, it is broken as shown in issue Valloric/YouCompleteMe#1378. For clients sending codepoint-indexed column, it is broken by `CompletionStartColumn` returning the wrong start column.

---

Reviewed 2 of 2 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/439#-:-KD_D8s13KyVF_Xnr9-A:1562568237)*

puremourning:
The README says:

> All line and column numbers are 1-based, not 0-based. They are also byte offsets, not Unicode codepoint offsets.

And I know this has been said explicitly before. I'm not saying don't change it, I'm just wondering why it was *explicitly* done this way in the first place. @Valloric ?

puremourning:
Looking at the code, I like the fact that we are removing complexity, but I'm a little sceptical that we're safe to make this assumption elsewhere. Like, I haven't checked what the assumption made by tern or libclang are when it comes to offsets. Currently we just pass through the offsets from the client unmolested. If the client starts sending unicode code points, then these offsets would surely be wrong either now, or before ?

---

Reviewed 2 of 2 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/439#-:-KD_hdu6nQxQj9XePjyU:1971988383)*

micbou:
They are wrong now because we are sending the buffer contents as unicode to these engines but `column_num` and `start_column` as byte offsets. It can't work for unicode characters.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/439#-:-KD_gLwzutYwJMz_BF8P:-1993612582)*

Valloric:
[Everyone in this debate should read [Utf8 Everywhere](http://utf8everywhere.org/) in case you're rusty on the Unicode parts.]

I'm not sure this is the right approach (it might be). I did explicitly choose utf8 strings in the API along with byte offsets. Codepoints might be better, but it's just slightly less wrong because of surrogate pairs in Unicode. People like to think that "one codepoint is one letter" but that's wrong; several codepoints together might render as one letter (a `u` codepoint plus a codepoint for "accent over `u`" rendering as "accented u"), or one codepoint might render as two letters (`nj` is two letters in English but one letter in Croatian, for instance).

So this might move us to a slightly less broken world, but not the ideal one. The reason I picked byte offsets is because this tends to force people to think about this a bit more carefully because _obviously_ the column number is not the number of characters. That may not be the right call though.

Also, what the various semantic engines do is orthogonal to what our API should be. We can accept byte offsets and do whatever computation necessary to make the engines happy (much like we do with 0 vs 1 for indexing).

This needs careful consideration. Ideas & opinions welcome.

puremourning:
I have spent literally 2 days looking at this (including many hex dumps and much use of [PyClewn](http://pyclewn.sourceforge.net). And now my head hurts. 

Anyway, I am now quite certain that the problem we have is that while the API asks for UTF-8 and byte offsets, the *code* assumes (all over the place) that `start_column` (as returned by `request_wrap.CompletionStartColumn`) is a 'character' offset, but `line_value` is a unicode string.

This goes wrong in situations like this:

```python
    start_column = request_data[ 'start_column' ] - 1
    current_line = request_data[ 'line_value' ]
    return ( start_column and
             ( current_line[ start_column - 1 ] in self._triggers or
               self.ShouldCompleteIncludeStatement( request_data ) ) )
```

This is `ShouldUseNowInner` and it is trying to inspect the character *before* the start column to see if it is a `/` (or equivalent).

However, `current_line` in s a unicode string `str` object:

```
(Pdb) pp current_line
u"var x = '\u2020'.this_is_some_text"
```

But `start_column` is a byte offset into the UTF-8 encoded version of current_line:

```
(Pdb) pp start_column
14

UTF-8:
1  2  3  4  5  6  7  8  9  10 11 12 13 14
76 61 72 20 79 20 3d 20 27 e2 80 a0 27 2e
v  a  r     x     =     '  †        '  .
```

And thus the expression `current_line[ start_column - 1 ]` does not yield the correct character (it should be the `'` in this case):

```
(Pdb) pp current_line[ start_column - 1]
u'h'
```

It is (as expected), exactly 2 bytes out - the number of additional bytes in the † character.

I think there are other places where we do this sort of thing; we don't consider that `start_column` is a byte offset, and thus `start_colum - 1` is not a valid operation. 

In fact, all of these operations where we need to look at the previous character need to convert the byte offset to a codepoint offset (in a similar way to `CompletionStartColumn`:

```python
  utf8_line_value = ToBytes( line_value )
  unicode_line_value = ToUnicode( line_value )
  codepoint_column_num = len(
      str( utf8_line_value[ : column_num - 1 ], 'utf8' ) ) + 1

  # -1 and then +1 to account for difference betwen 0-based and 1-based
  # indices/columns
  codepoint_start_column = StartOfLongestIdentifierEndingAtIndex(
      unicode_line_value, codepoint_column_num - 1, filetype ) + 1
```

Does this make sense/seem reasonable?

If so, then I think we should

- [ ] Create `ycmd.utils.ByteOffsetToCodePointOffset( line, byte_offset )`
- [ ] survey all the uses of `start_column` (and possibly `column_num` from the `request_data`), and ensure that where they are making assumptions about 'characters' in the line, they use the above.

Thoughts? I'm fairly confident this is the root of the issue, as I have tried a number of hypotheses and they all seem to point to this "interpreting it  as a character offset" problem.

Valloric:
@puremourning Excellent analysis!

> Does this make sense/seem reasonable?

Entirely. Does `-1` expressions (and in fact, any goddamn math expressions on column numbers) are painfully broken. Ugh.

> Create ycmd.utils.ByteOffsetToCodePointOffset

That may not be enough. See my previous comment; you can have two codepoints that make up _one_ character (see [Combining Characters](https://en.wikipedia.org/wiki/Combining_character)). What we probably need to do is enforce that all text passed to ycmd is [NFC normalized](https://en.wikipedia.org/wiki/Unicode_equivalence#Normalization).

What you propose would probably fix 99% of the issues though.

I'd also want to encapsulate all the math logic behind carefully reviewed functions. So something like `MoveNCharacters(string, byte_offset, num_chars_to_move) -> byte_offset`. Similar for any other operations we have. All the unicode functions should probably be in one `unicode_helpers.py` file so that it's easy to audit this shit in the future, instead of having it in various parts of the codebase.

puremourning:
Cool.

This approach is working. I've successfully (trivially) fixed a number of issues by adding `start_codepoint` as well as `start_column` as "magic" properties of `request_wrap`.

This allows us to just (for now) change places where we're doing:

```python
  start_column = request[ 'start_column' ] - 1
```

To

```python
  start_codepoint = request[ 'start_codepoint' ] - 1
```

That way the semantics are clear, and so does the code work. Fixing the tern completer was as simple as this:

```diff
diff --git a/ycmd/completers/javascript/tern_completer.py b/ycmd/completers/javascript/tern_completer.py
index 85e1277..ab53c22 100644
--- a/ycmd/completers/javascript/tern_completer.py
+++ b/ycmd/completers/javascript/tern_completer.py
@@ -337,7 +337,7 @@ class TernCompleter( Completer ):
     def MakeTernLocation( request_data ):
       return {
         'line': request_data[ 'line_num' ] - 1,
-        'ch':   request_data[ 'start_column' ] - 1
+        'ch':   request_data[ 'start_codepoint' ] - 1
       }

     full_query = {
```

I think you're right about abstracting any "move X characters" operations out, which I will also do.

micbou:
> In fact, all of these operations where we need to look at the previous character need to convert the byte offset to a codepoint offset (in a similar way to CompletionStartColumn:

Not only when we need to look at the previous character but each time we try to access a unicode string with a byte offset. Look at your example:
```
1  2  3  4  5  6  7  8  9  10 11 12 13 14
1  2  3  4  5  6  7  8  9  10       11 12   
v  a  r     x     =     '  †        '  .
```
The correct index is 12, not 14 so even `current_line[ start_column ]` is wrong.

> That may not be enough. See my previous comment; you can have two codepoints that make up one character (see Combining Characters). What we probably need to do is enforce that all text passed to ycmd is NFC normalized.

[NFC will not be enough](http://utf8everywhere.org/#myth.strlen). Ideally, we should use grapheme clusters but [Python does not support them](http://utf8everywhere.org/#faq.python).

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/439#-:-KDfFB4ANlVPbOFivp7V:-1459793719)*

Valloric:
I'm aware that grapheme clusters is the way to go, but codepoints gets us 99% there and NFC gets us 99.99%. Grapheme clusters is indeed something we should do; we should explore what Python libraries we can use for that though. Out-of-the-box, Python is clueless, as you pointed out.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/439#-:-KDfVYhRZgT2LKm4aQi3:2087583930)*

Valloric:
For grapheme clusters, `uniseg` [seems to have what we want](http://uniseg-python.readthedocs.org/en/latest/graphemecluster.html). Sadly doesn't run on 2.6, but does support 2.7 and 3.3+.

I think NFC might be the best we can do, short of dropping support for Python 2.6.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/439#-:-KDfWRLSVtzY8FBm-Pex:820319400)*

puremourning:
> Not only when we need to look at the previous character but each time we try to access a unicode string with a byte offset. 

Quite. Basically, the fact that `line_value` returns a unicode string and `start_column` returns a byte offset into a completely different string is just going to confuse everyone. I'm adding `line_bytes` and `start_codepoint` (and judicious comments!) to sort that out for now. Once that it done (it's about roughly working now, with tests), I will look to compress to hide complexity.

TBH it is a bit of a minefield (like C++ and javascript, and Python 3 all support unicode chars in identifiers, but YCM doesn't). I'm trying to work out where to draw the line :)

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/439#-:-KDfscu062jYxjjNmGfh:1947971251)*

micbou:
Closed in favor of #455.

PR #440, opened 2016-03-24
[READY] Don't send non-javascript files to tern

puremourning:
This fixes a _rare_ tern server crash noticed when editing a Python file and a JavaScript file simultaneously. It's not completely fair to expect tern server to parse Python, so it can occasionally crash doing so. As a result, only include files who have the `javascript` filetype.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/440)


micbou:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/440#-:-KDa6DaAN1tbt0-QTdqo:1417088721)*

Valloric:
Needs a test! :)

puremourning:
I wrote a whole spiel about how it wasn't possible to test that this actually *works*, only to *exercise* it, but as I was doing so I came up with a scheme! Typical :)

The scheme is to write some real javascript that would normally offer suggestions, but tell ycmd that it is in fact Python code. Ensure that tern *doesn't* offer those suggestions.

Done!

puremourning:
Ping ?

Sorry for the chaser, but this is causing me pain whilst writing ycmd tests. And nobody would want me to stop writing tests for ycmd now :)

Valloric:
Oops, missed this one. Thanks for the ping!

@homu r=micbou

Issue #430, opened 2016-03-13
Strange caching behavior on completions requests


vincentwoo:
Hi there,

In testing YCMD as a remote completion API, I appear to be hitting a weird caching bug. I'm just testing against Jedi autocompletion currently.

For the buffer

```python
import requests
import datetime

requests.
```

I (on the remote server where ycmd is running), write out the buffer to the filesystem, send a `FileReadyToParse` event to YCMD, and then ask for completions. This works okay, I get the appropriate results for `requests`.

However, If I change the buffer to:

```python
import requests
import datetime

datetime.
```

and perform the same actions, I still get returned the completion results for `requests` and not `datetime`. I believe this to be because of YCMD's caching on line and character offsets. Is there an API for saying "this buffer needs to be reparsed"? I couldn't find one in the sample client.

vincentwoo:
On a bit of further digging I realized that YCMD provides [hooks](https://github.com/Valloric/ycmd/blob/30b04ebd1458d8a09e14587b57e81aabc15e2a53/ycmd/completers/completer.py#L253) for completion engines to listen to editor events. However, I could not find evidence that these had been overridden in the Jedi completer. This problem also seems to happen with JavaScript, even though the tern completer has an [OnFileReadyToParse](https://github.com/Valloric/ycmd/blob/30b04ebd1458d8a09e14587b57e81aabc15e2a53/ycmd/completers/javascript/tern_completer.py#L194) hook.

vincentwoo:
Furthermore, the completer engine appears to to ignore `filepath` for caching purposes. This suggests that YCMD is built to only complete one buffer at a time. Is that correct, and if so, are there any ways to work around the issue?

Valloric:
@vincentwoo Well ycmd is designed for editor completion on a local box with one user. Zero thought has been put into multi-user support since it's not a feature localhost editors need. There's probably plenty of issues in ycmd that would prevent multi-user support.

We might be willing to accept patches improving multi-user support, but only if they're _very_ low-maintenance.

The caching issue you've hit is indeed a corner case that can happen when two same-length names of the two different objects are used on the same line & column. It's pretty rare though, but we should fix it somehow.

You may want to read the comments (and the code) in `completer.py` to understand how the caching works.

vincentwoo:
Cool, yeah it's not a huge deal to go multi-user, might be able to hack around it from the outside.

[...]

PR #443, opened 2016-03-25
Skip completion cache if force_semantic is set

vincentwoo:
This is a simple way to deal with the issues described in #430 as suggested by @puremourning in https://github.com/Valloric/ycmd/issues/430#issuecomment-201540248.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/443)

### DIFF
--- a/ycmd/completers/completer.py
+++ b/ycmd/completers/completer.py
@@ -174,21 +174,22 @@ class Completer( with_metaclass( abc.ABCMeta, object ) ):
 
 
   def _GetCandidatesFromSubclass( self, request_data ):
-    cache_completions = self._completions_cache.GetCompletionsIfCacheValid(
-          request_data[ 'line_num' ],
-          request_data[ 'start_column' ],
-          self.CompletionType( request_data ) )
+    if not ForceSemanticCompletion( request_data ):
+      cache_completions = self._completions_cache.GetCompletionsIfCacheValid(
+            request_data[ 'line_num' ],
+            request_data[ 'start_column' ],
+            self.CompletionType( request_data ) )
 
-    if cache_completions:
-      return cache_completions
-    else:
-      raw_completions = self.ComputeCandidatesInner( request_data )
-      self._completions_cache.Update(
-          request_data[ 'line_num' ],
-          request_data[ 'start_column' ],
-          self.CompletionType( request_data ),
-          raw_completions )
-      return raw_completions
+      if cache_completions:
+        return cache_completions
+
+    raw_completions = self.ComputeCandidatesInner( request_data )
+    self._completions_cache.Update(
+        request_data[ 'line_num' ],
+        request_data[ 'start_column' ],
+        self.CompletionType( request_data ),
+        raw_completions )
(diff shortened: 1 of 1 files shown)


puremourning:
Needs tests :/

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/443#-:-KDk0ataPcswG8jvTmYp:-537945060)*

mispencer:
It's common for me to force completions, have it time out, force it again, and get completions from the  cache. I just tested this, and with this change, I can't get any completions at all, since it always times out.

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/443#-:-KDk5EarDDKQWSV60iE-:-870972918)*

micbou:
Another reason to add a `cache` parameter instead of using `force_semantic` in the request.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/443#-:-KDk8FsJz303GyvIQJIb:-1321736620)*

Valloric:
Thanks for the PR! :)

I'm leaning towards not merging this because of the use-case @mispencer brought up. I too have used that in the past.

Thoughts?

@vincentwoo In the future, please make sure you're adding tests for any new changes. We don't accept PRs without tests. 

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/443#-:-KDkIBWfv9gcLG3qbwBn:-1199278700)*

vincentwoo:
No prob. I'll redo this with an explicit new parameter (cache: false seems fine) as well as a test.

---

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/valloric/ycmd/443#-:-KDkWa3VvBZQxlDwoaXj:-148185615)*

Valloric:
[See [my comment](https://github.com/Valloric/ycmd/issues/430#issuecomment-201656058) on #430 first]

Again, thank you for the PR! 

@micbou, @puremourning and I can't imagine a case where YCM, emacs-ycmd or the other clients in ycmd's designed-for category of clients would _ever_ set the `cache: false` param, so supporting it _forever_ (ycmd tries hard to not break backwards compatibility) does not seem worth it, sorry. We have a general rule that features need to be widely useful to be worth future maintenance.

oblitum:
The suggested `cache` flag is very useful/important for https://github.com/Valloric/YouCompleteMe/pull/1300.

oblitum:
Reason. It suggest multiplexing normal completion and hints from the same start_column, but conditioned to the presence of characters following the hint trigger or not. If the user forces completion after opening function parenthesis, for example, it should bring hints, if the user has typed any identifier character after the parenthesis and forces semantic completion, the expected is not to have hints anymore, so it should not bring the cache from start_column, it should force complete the standard way without hints.

oblitum:
Just for reference, I've personally solved it this way: https://github.com/Valloric/ycmd/issues/491#issuecomment-219296474.

PR #453, opened 2016-04-09
[READY] Fix IndexError exception from C++

### LINKED ISSUE #177: Error with non-ascii character in file name
I have a file with non-ascii cahracter I want to run completion on and I get the following error:

The issue appeared because we did some test for `emacs-ycmd` regarding encoding. We had some problems with current Emacs HEAD.

```
2015-07-16 23:51:40,136 - INFO - Adding buffer identifiers for file: /home/peter/src/test/testå.cpp
2015-07-16 23:51:44,529 - INFO - Received completion request
2015-07-16 23:51:44,530 - DEBUG - Using filetype completion: False
  File "/home/peter/src/ycmd/third_party/bottle/bottle.py", line 861, in _handle
    return route.call(**args)
  File "/home/peter/src/ycmd/third_party/bottle/bottle.py", line 1734, in wrapper
    rv = callback(*a, **ka)
  File "/home/peter/src/ycmd/ycmd/../ycmd/watchdog_plugin.py", line 100, in wrapper
    return callback( *args, **kwargs )
  File "/home/peter/src/ycmd/ycmd/../ycmd/hmac_plugin.py", line 62, in wrapper
    body = callback( *args, **kwargs )
  File "/home/peter/src/ycmd/ycmd/../ycmd/handlers.py", line 101, in GetCompletions
    completer.ComputeCandidates( request_data ),
  File "/home/peter/src/ycmd/ycmd/../ycmd/completers/general/general_completer_store.py", line 83, in ComputeCandidates
    candidates += completer.ComputeCandidates( request_data )
  File "/home/peter/src/ycmd/ycmd/../ycmd/completers/all/identifier_completer.py", line 52, in ComputeCandidates
    ToUtf8IfNeeded( request_data[ 'filetypes' ][ 0 ] ) )
IndexError: bitset::set: __position (which is 18446744073709551555) >= _Nb (which is 128)
```

micbou:
This PR fixes the `IndexError` exception returned by the `FilterAndCandidatesWrap` function when a candidate or the query contains non-ascii characters. Its error message depends on the OS:
 - on Windows: `invalid bitset<N> position`
 - on Linux: `bitset::set: __position (which is xxx) >= _Nb (which is 128)`
 - on OS X: `bitset set argument out of range`

It is caused by the `LetterBitsetFromString` function trying to set a bit to the `Bitset` object with a character index that is out of range. This is solved by only setting bits for character indices satisfying `0 ≤ i < 128`.

We add two tests for the `LetterBitsetFromString` function: one test to check the lower and upper character bounds and another to check that non-ascii characters are ignored. These tests are failing without this change. For example, on Windows:
```
[ RUN      ] LetterBitsetFromStringTest.Boundaries
unknown file: error: C++ exception with description "invalid bitset<N> position" thrown in the test body.
[  FAILED  ] LetterBitsetFromStringTest.Boundaries (1 ms)
[ RUN      ] LetterBitsetFromStringTest.IgnoreNonAsciiCharacters
unknown file: error: C++ exception with description "invalid bitset<N> position" thrown in the test body.
[  FAILED  ] LetterBitsetFromStringTest.IgnoreNonAsciiCharacters (0 ms)
```

Fixes #177 and Valloric/YouCompleteMe#2103.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/453)


puremourning:
Reviewing the code a bit I find a number of other places where the code assumes that each `char letter` can be used to index into `LetterNodeListMap::letters_`. I expect that we will have to fix all of them too.

For example in `cpp/ycm/LetterNodeListMap.cpp`:

```cpp
std::list< LetterNode * > &LetterNodeListMap::operator[] ( char letter ) {
  int letter_index = IndexForChar( letter );
  std::list< LetterNode * > *list = letters_[ letter_index ];
  ...
}
...
std::list< LetterNode * > *LetterNodeListMap::ListPointerAt( char letter ) {
  return letters_[ IndexForChar( letter ) ];
}
...
bool LetterNodeListMap::HasLetter( char letter ) const {
  return letters_[ IndexForChar( letter ) ] != NULL;
}
```

according to [the docs](http://www.boost.org/doc/libs/1_60_0/doc/html/boost/array.html#idm45507162975808-bb) this could breach the requirement that `letter` here be < `NUM_LETTERS`.

I suspect that what is happening is that this is always reading or writing past the end of the array, which could lead to ~~undefined behaviour~~ nuclear war. I could be wrong. Running under `valgrind` would be a cheap way to test this.

---

---

*[cpp/ycm/Candidate.cpp, line 69 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KEv8u6mHcTIp5lNqpmI:-KEv8u6mHcTIp5lNqpmJ:699431402) ([raw file](https://github.com/valloric/ycmd/blob/dc3ecd1f9df5aaf3ea38fcb4bd3153249dc281d2/cpp/ycm/Candidate.cpp#L69)):*
`char_index` should be declared where it is initialised (i.e. on line 71). In general, variables should be initialised when they are declared (e.g. `int char_index = 0`) but redundant assignments should be avoided too (so `int char_index = IndexForChar( letter )`). Good style also to declare close to where it is used (within the smallest reasonable scope).

---

*[cpp/ycm/Candidate.cpp, line 72 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KEv9qp-kZFVr-hYwfVf:-KEv9qp-kZFVr-hYwfVg:-1857201606) ([raw file](https://github.com/valloric/ycmd/blob/dc3ecd1f9df5aaf3ea38fcb4bd3153249dc281d2/cpp/ycm/Candidate.cpp#L72)):*
I personally think that it is more natural to read (and write): `if ( char_index >= 0 && char_index < NUM_LETTERS )`

There is a sort of old-school fallacy that in order to prevent accidental use of the assignment operator, programmers would (heinously IMO) obfuscate their code with `if ( 0 == myInt )`, rather than `if ( myInt == 0 )`. 

Anyway, I always feel like the variable (lvalue) should be of the left, because we normally say "is this apple red" not "is red this apple". 

</soapbox>

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KEvC3d6_O9OqI2z-F5c:-680321309)*

micbou:
Good catch. I'll try to write tests for these functions.

---

Reviewed 1 of 2 files at r1.

---

*[cpp/ycm/Candidate.cpp, line 69 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KEv8u6mHcTIp5lNqpmI:-KEvIkS9td0PqU8l9SUH:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/dc3ecd1f9df5aaf3ea38fcb4bd3153249dc281d2/cpp/ycm/Candidate.cpp#L69)):*
Done.

---

*[cpp/ycm/Candidate.cpp, line 72 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KEv9qp-kZFVr-hYwfVf:-KEvGjgKDKPLZU11-v6q:19558356) ([raw file](https://github.com/valloric/ycmd/blob/dc3ecd1f9df5aaf3ea38fcb4bd3153249dc281d2/cpp/ycm/Candidate.cpp#L72)):*
I wrote it this way to be closer to the mathematical form: `0  ≤ i < NUM_LETTERS`. But let's be consistent and always write the variable to the left.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KEvKU0eopNVQ-0KUROE:-961767404)*

micbou:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[cpp/ycm/Candidate.cpp, line 70 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KEvLbciG65jN-ddBOPT:-KEvLbciG65jN-ddBOPU:764777606) ([raw file](https://github.com/valloric/ycmd/blob/3ca5e3cd76b2fe76162c5dcc2b8f86712c28a963/cpp/ycm/Candidate.cpp#L70)):*
I changed `char_index` to `letter_index` because this name is used in other places.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453)*

puremourning:
I think this fixes https://github.com/Valloric/YouCompleteMe/issues/278 too (along with my upcoming PR for unicode)

micbou:
This fixes the `bitset` exception, not the filename unicode support.

---

Reviewed 1 of 2 files at r1, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KEwyZyRXZ3LfXDMUNhP:-1639804952)*

puremourning:
Right, the combination of the 2 should fix it, though I think :)

---

Reviewed 2 of 2 files at r1, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KEwz_nZT-gphwfGrGcC:871354717)*

micbou:
I mean, even with our fixes, completing filenames with non-ascii characters will not be possible and I think this is what issue #278 is about.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KEx-HQvP3Gp4Wg7wc0T:-1171813140)*

puremourning:
Well the report it self talks about fixing the `bitset` exception. If we fixed the filter to support unicode, then the filename completer would also continue to work. The filename completer itself handles unicode fine (at least after my PR) :)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KEx3qs821xfqRq2b11q:1881378191)*

Valloric:
:lgtm: from me if @puremourning is fine with it too. I think he's now officially the unicode expert as it pertains to ycmd. :)

---

---

*[cpp/ycm/Candidate.cpp, line 70 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KEvLbciG65jN-ddBOPT:-KEy-yecJXbnqNrmnkNJ:-500113922) ([raw file](https://github.com/valloric/ycmd/blob/3ca5e3cd76b2fe76162c5dcc2b8f86712c28a963/cpp/ycm/Candidate.cpp#L70)):*
Fine with this, though it looks wierd that we now have `letter_index = IndexForChar`. Maybe update the function name too?

I'm fine with leaving it the way you have it too.

---

*[cpp/ycm/tests/LetterBitsetFromString_test.cpp, line 60 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KEy0KID1gpHWaNkiXXI:-KEy0KIETQpMLh8BcWdt:-41957037) ([raw file](https://github.com/valloric/ycmd/blob/3ca5e3cd76b2fe76162c5dcc2b8f86712c28a963/cpp/ycm/tests/LetterBitsetFromString_test.cpp#L60)):*
Maybe jam a chinese letter in there too. They tend to use more that two bytes in UTF-8.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KEy0aVhVlI4G4QU6iNc:-1311416538)*

micbou:
So I added tests for the `LetterNode` class: one with ascii characters and another with non-ascii characters.

As expected, the non-ascii test was failing by returning out of bounds pointers instead of a NULL pointer. Fixed by checking that the letter index is in range in `LetterNodeListMap` methods.

---

Reviewed 2 of 4 files at r5, 4 of 4 files at r6.
Review status: 6 of 7 files reviewed at latest revision, 6 unresolved discussions.

---

*[cpp/ycm/Candidate.cpp, line 70 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KEvLbciG65jN-ddBOPT:-KF-KZqPBLtks2fivjBi:-1113579270) ([raw file](https://github.com/valloric/ycmd/blob/3ca5e3cd76b2fe76162c5dcc2b8f86712c28a963/cpp/ycm/Candidate.cpp#L70)):*
You want to rename it to `IndexForLetter`? I don't mind.

---

*[cpp/ycm/LetterNodeListMap.cpp, line 30 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-JtQKZqIBN6_SZJoY:-KF-JtQKZqIBN6_SZJoZ:-1905098456) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.cpp#L30)):*
@puremourning I wrote the condition as `0 <= i && i < n` instead of `i >= 0 && i < n` because it is done this way in the `IsUppercase` function, just above.

---

*[cpp/ycm/LetterNodeListMap.h, line 35 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-KqqjyDfIysIet1V1:-KF-KqqjyDfIysIet1V2:-1541735650) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.h#L35)):*
Should we inline these functions?

---

*[cpp/ycm/LetterNodeListMap.h, line 42 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-Q74jarg6j15Pz9ef:-KF-Q74k0G8KkpGEEGu2:-1572153668) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.h#L42)):*
@Valloric This method is never used. Should we remove it?

---

*[cpp/ycm/LetterNodeListMap.h, line 48 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-Q9nWj8aW7CrYPriM:-KF-Q9nWj8aW7CrYPriN:-1948804644) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.h#L48)):*
This method too.

---

*[cpp/ycm/tests/LetterBitsetFromString_test.cpp, line 60 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KEy0KID1gpHWaNkiXXI:-KF-L4B5wgcB51QzPYZV:1040065394) ([raw file](https://github.com/valloric/ycmd/blob/3ca5e3cd76b2fe76162c5dcc2b8f86712c28a963/cpp/ycm/tests/LetterBitsetFromString_test.cpp#L60)):*
`€` is a 3-bytes character and all chinese characters seem to be 3-bytes (`ø` is 2-bytes). We could use the `𐍈` character: it is 4-bytes. You know what: I'll update the string to `uni¢𐍈d€` (`¢` is 2-bytes) so that we use 1, 2, 3, and 4-bytes characters in the same string (there are no characters with more than 4-bytes in UTF-8).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KF-N-5XnaQ_Uo6FWU4z:702064765)*

puremourning:
Reviewed 2 of 4 files at r5, 2 of 4 files at r6, 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[cpp/ycm/LetterNodeListMap.cpp, line 71 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF0042aqRK8XRC_Lhp0:-KF0042aqRK8XRC_Lhp1:1926928658) ([raw file](https://github.com/valloric/ycmd/blob/db27875ffe9ce3449f3eaaeebd31d69a9deb6b0c/cpp/ycm/LetterNodeListMap.cpp#L71)):*
I think this leaks memory (who calls the `delete` for this `new`?), and it is difficult to come up with a scheme where the return value from this call is non-const and doesn't leak. (one option would be something like `static std::list< LetterNode *>() out_of_range_result; return out_of_range_result`, but this would lead to strange behaviour if someone used that result and modified it)

As it happens, I think it is reasonable to have a precondition that `operator[]` takes a value which strictly must be `InAsciiRange`. We can add an `assert( IsInAsciiRange( letter_index ) )` or just throw an exception (as we typically don't test with debug builds).

---

*[cpp/ycm/LetterNodeListMap.h, line 35 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-KqqjyDfIysIet1V1:-KF01lurrX97qB6LWykP:813459498) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.h#L35)):*
I'm not sure how well inlining exported symbols works, TBH. But inlining `IsInAsciiRange` could be done. 

---

*[cpp/ycm/LetterNodeListMap.h, line 42 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-Q74jarg6j15Pz9ef:-KF026o8HYJrdvK9POOi:1541987304) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.h#L42)):*
If we have dead code, I think we should remove it (personally).

---

*[cpp/ycm/LetterNodeListMap.h, line 48 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-Q9nWj8aW7CrYPriM:-KF02AhPgel4NO1HczYD:-469765787) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.h#L48)):*
ditto

---

*[cpp/ycm/tests/LetterNode_test.cpp, line 30 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF02LI6TENWweAMYOLT:-KF02LI6TENWweAMYOLU:-1488771367) ([raw file](https://github.com/valloric/ycmd/blob/db27875ffe9ce3449f3eaaeebd31d69a9deb6b0c/cpp/ycm/tests/LetterNode_test.cpp#L30)):*
I don't think this needs to be on the heap, does it? Even if it does, who `delete`s it?

OK a small memory leak in our test isn't a big deal, but it means the destructor code (which is still code we should test) isn't run.

---

*[cpp/ycm/tests/LetterNode_test.cpp, line 65 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF02ccphQ0q8Sh09bdS:-KF02ccphQ0q8Sh09bdT:1075660548) ([raw file](https://github.com/valloric/ycmd/blob/db27875ffe9ce3449f3eaaeebd31d69a9deb6b0c/cpp/ycm/tests/LetterNode_test.cpp#L65)):*
also probably leaks. Can probably just be `LetterNode node( "un...." );`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453)*

puremourning:
Reviewed 1 of 4 files at r6.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453)*

Valloric:
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[cpp/ycm/Candidate.cpp, line 70 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KEvLbciG65jN-ddBOPT:-KF0Y4sGWR6l04djHUxh:1173564609) ([raw file](https://github.com/valloric/ycmd/blob/3ca5e3cd76b2fe76162c5dcc2b8f86712c28a963/cpp/ycm/Candidate.cpp#L70)):*
Sure, why not.

---

*[cpp/ycm/LetterNode.cpp, line 30 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF0YAUTLIW9XbjwDgEC:-KF0YAUTLIW9XbjwDgED:970228635) ([raw file](https://github.com/valloric/ycmd/blob/09f216441f99c02d51f9460e006683c5fe83669e/cpp/ycm/LetterNode.cpp#L30)):*
Ha, at least I was honest about it. :D

---

*[cpp/ycm/LetterNodeListMap.cpp, line 30 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-JtQKZqIBN6_SZJoY:-KF0YNLd1z9Z4JAjE1dX:-188564544) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.cpp#L30)):*
Nice. The formatting in IsUppercase was intentional; it more clearly shows that letter needs to be between A and Z.

The formatting works here just as well.

---

*[cpp/ycm/LetterNodeListMap.cpp, line 71 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF0042aqRK8XRC_Lhp0:-KF0YxD8-ANw1IKNHTi0:159811067) ([raw file](https://github.com/valloric/ycmd/blob/db27875ffe9ce3449f3eaaeebd31d69a9deb6b0c/cpp/ycm/LetterNodeListMap.cpp#L71)):*
This is tricky, I agree. And yes, as written, it leaks.

I'm not against having an assert.

---

*[cpp/ycm/LetterNodeListMap.h, line 35 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-KqqjyDfIysIet1V1:-KF0ZKYU7oPJEQPCh26D:-177438094) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.h#L35)):*
Let's leave it up to the compiler.

---

*[cpp/ycm/LetterNodeListMap.h, line 42 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-Q74jarg6j15Pz9ef:-KF0ZOl0FnVzKU7NUL1d:-1977107377) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.h#L42)):*
Sure, kill it.

---

*[cpp/ycm/LetterNodeListMap.h, line 48 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-Q9nWj8aW7CrYPriM:-KF0ZRe5RfBNTgGxHWAr:261621654) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.h#L48)):*
Kill it with fire.

---

*[cpp/ycm/tests/LetterBitsetFromString_test.cpp, line 60 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KEy0KID1gpHWaNkiXXI:-KF0gZF22ZJUe70wLmot:1689193178) ([raw file](https://github.com/valloric/ycmd/blob/3ca5e3cd76b2fe76162c5dcc2b8f86712c28a963/cpp/ycm/tests/LetterBitsetFromString_test.cpp#L60)):*
Sounds good.

---

*[cpp/ycm/tests/LetterNode_test.cpp, line 30 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF02LI6TENWweAMYOLT:-KF0gelGFpmEuI7BWkTP:-1211221615) ([raw file](https://github.com/valloric/ycmd/blob/db27875ffe9ce3449f3eaaeebd31d69a9deb6b0c/cpp/ycm/tests/LetterNode_test.cpp#L30)):*
Leaks are bad in tests too because it means you can't run leak detectors on your test suite.

And I agree, this shouldn't be on the heap.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453)*

micbou:
Reviewed 1 of 1 files at r7, 6 of 7 files at r8.
Review status: 7 of 8 files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/ycm/Candidate.cpp, line 70 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KEvLbciG65jN-ddBOPT:-KF19tM6E5sz1GppZXr9:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/3ca5e3cd76b2fe76162c5dcc2b8f86712c28a963/cpp/ycm/Candidate.cpp#L70)):*
Done.

---

*[cpp/ycm/LetterNodeListMap.cpp, line 71 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF0042aqRK8XRC_Lhp0:-KF19y_wJfy1Y7gFJ_SD:896348891) ([raw file](https://github.com/valloric/ycmd/blob/db27875ffe9ce3449f3eaaeebd31d69a9deb6b0c/cpp/ycm/LetterNodeListMap.cpp#L71)):*
I decided to use the [`at` method from Boost](http://www.boost.org/doc/libs/1_60_0/doc/html/boost/array.html#idm45507162967984-bb) for the `[]` operator and the `ListPointerAt` method.

---

*[cpp/ycm/LetterNodeListMap.h, line 42 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-Q74jarg6j15Pz9ef:-KF1Ae4Mo2U9cToBO-c2:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.h#L42)):*
Done.

---

*[cpp/ycm/LetterNodeListMap.h, line 48 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-Q9nWj8aW7CrYPriM:-KF1AebPeEJ_yaj_g7G2:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.h#L48)):*
Done.

---

*[cpp/ycm/tests/LetterNode_test.cpp, line 30 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF02LI6TENWweAMYOLT:-KF1Aik90H6ROQt9Sgx_:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/db27875ffe9ce3449f3eaaeebd31d69a9deb6b0c/cpp/ycm/tests/LetterNode_test.cpp#L30)):*
Done.

---

*[cpp/ycm/tests/LetterNode_test.cpp, line 65 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF02ccphQ0q8Sh09bdS:-KF1AjfNqIGFHRZdk9J0:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/db27875ffe9ce3449f3eaaeebd31d69a9deb6b0c/cpp/ycm/tests/LetterNode_test.cpp#L65)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453)*

micbou:
Should we bump the core version for these changes?

---

Review status: 7 of 8 files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KF1C9StBYDefjG18TFf:1601697514)*

Valloric:
Yes, core version should be bumped.

---

Review status: 7 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[cpp/ycm/LetterNodeListMap.cpp, line 71 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF0042aqRK8XRC_Lhp0:-KF1w34Sr26EgHZIFR_8:621081265) ([raw file](https://github.com/valloric/ycmd/blob/db27875ffe9ce3449f3eaaeebd31d69a9deb6b0c/cpp/ycm/LetterNodeListMap.cpp#L71)):*
What do we do in higher levels of the stack when the C++ layer throws an exception? We should make sure throwing here works.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KF1wHM3Oym_hmXb0StS:-609577034)*

micbou:
Version bumped.

---

Reviewed 1 of 1 files at r9, 1 of 1 files at r10.

---

*[cpp/ycm/LetterNodeListMap.cpp, line 71 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF0042aqRK8XRC_Lhp0:-KF3JXgkPtQvc3ScvFFs:-1689124233) ([raw file](https://github.com/valloric/ycmd/blob/db27875ffe9ce3449f3eaaeebd31d69a9deb6b0c/cpp/ycm/LetterNodeListMap.cpp#L71)):*
This exception will never get to the Python layer because [candidates containing non-ascii (nonprintable) characters are discarded](https://github.com/Valloric/ycmd/blob/master/cpp/ycm/CandidateRepository.cpp#L75). I checked the behavior without discarding such candidates and we get a `IndexError: array<>: index out of range` exception in Python so this is working as expected.

If we want to be extra cautious, we could add the `FilterAndSortCandidates` tests from PR #455 to this PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KF3PsvnUyeCkUnotZiy:328850284)*

Valloric:
Awesome stuff!

---

Reviewed 1 of 4 files at r5, 1 of 1 files at r7, 6 of 7 files at r8, 1 of 1 files at r9, 1 of 1 files at r10.

---

*[cpp/ycm/LetterNodeListMap.cpp, line 71 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF0042aqRK8XRC_Lhp0:-KF5LdnRMDcK55N4kf86:-1959025863) ([raw file](https://github.com/valloric/ycmd/blob/db27875ffe9ce3449f3eaaeebd31d69a9deb6b0c/cpp/ycm/LetterNodeListMap.cpp#L71)):*
Nah, we're good.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KF5Lo3t-shwPMAsC-aO:1434142575)*

vheon:
Reviewed 1 of 4 files at r5, 1 of 1 files at r7, 6 of 7 files at r8, 1 of 1 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453)*

puremourning:
:lgtm:

---

Reviewed 6 of 7 files at r8, 1 of 1 files at r9, 1 of 1 files at r10.

---

*[cpp/ycm/LetterNodeListMap.h, line 35 \[r6\]](https://reviewable.io:443/reviews/valloric/ycmd/453#-KF-KqqjyDfIysIet1V1:-KF_EvdQsIcENnDz2ZKX:-443343619) ([raw file](https://github.com/valloric/ycmd/blob/c07d4c5b69a412105c5036fbc4fc60668848c7ca/cpp/ycm/LetterNodeListMap.h#L35)):*
Not sure the compiler can inline it if the definition is in the TU rather than the header. (Well, it can inline it in the TU, but not elsewhere, which I think is the point of the question). In any case, it is early optimisation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KF_FwSujfsAQ5eg-exW:1417088721)*

Valloric:
Thanks for the PR! :)

@homu r+

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/453#-:-KFbQAjUzcAgqUU2LTo-:-1087463726)*

PR #456, opened 2016-04-11
Add netrw to the default filetype blacklist

yous:
As Netrw is a browser for either local or remote files, it doesn't need completion. When YCM is compiled with `--clang-completer` and user doesn't have set `g:ycm_global_ycm_extra_conf`, browsing any directory in which doesn't have `.ycm_extra_conf.py` using Netrw makes YCM print an error message.

This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/456)

### DIFF
--- a/ycmd/default_settings.json
+++ b/ycmd/default_settings.json
@@ -25,6 +25,7 @@
     "qf": 1,
     "notes": 1,
     "markdown": 1,
+    "netrw": 1,
     "unite": 1,
     "text": 1,
     "vimwiki": 1,


Valloric:
Thanks for sending a PR.

Please explain why this change is necessary in the PR description.

yous:
Oops, updated. If there is any information that has to be added, please let me know.

Valloric:
Thanks for providing a rationale! :)

:lgtm: 

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/456#-:-KF5W7MXQ3wW0ckN0Qlf:793662393)*

micbou:
I think the blacklisting of filetypes specific to Vim (`tagbar`, `qf`, `unite`, `vimwiki`, `netwr`) should be done in YCM, not ycmd.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/456#-:-KF5ZuSMbtpeApFTHCML:1339968803)*

Valloric:
@micbou Agreed. But are we doing that already? If yes, then this PR should be replaced with a corresponding one on YCM. If we aren't doing this already, let's not start doing it in this PR because we need more groundwork for it.

:lgtm_cancel:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/456#-:-KF6A3rUTy2ptapZ4JdG:1740040229)*

micbou:
No, we are not so I guess we can merge this PR.

:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/456#-:-KF6EXy3qkulwAsOMWnR:107921677)*

Valloric:
OK, then let's think about fixing it in a future PR.

@homu r+

PR #457, opened 2016-04-12
Adding support for .netcore project file (project.json)

teonivalois:
When trying to figure out the solution file, we are sticking to *.sln as if we could only have valid .Net projects when a .sln file is available.

Well, it used to be true, but now with .net core (https://dotnet.github.io) we can have individual .net projects. These project folders are recognized when a project.json file is available.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/457)

### DIFF
--- a/ycmd/completers/cs/solutiondetection.py
+++ b/ycmd/completers/cs/solutiondetection.py
@@ -87,11 +87,12 @@ def GuessFile( filepath ):
   for i in reversed( range( len( tokens ) - 1 ) ):
     path = os.path.join( *tokens[ : i + 1 ] )
     candidates = glob.glob1( path, '*.sln' )
-    
-    """ If there's no .sln file, try to find a project.json file for dotnetcore (https://dotnet.github.io) projects """
+
+    """ If there's no .sln file, try to find a project.json file for
+    dotnetcore (https://dotnet.github.io) projects """
     if len( candidates ) == 0:
       candidates = glob.glob1( path, 'project.json' )
-    
+
     if len( candidates ) > 0:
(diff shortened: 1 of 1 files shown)


Valloric:
Thanks for sending a PR!

You need to add tests for this feature. Also, please update the PR description with links to docs describing this `project.json`.

---

---

*[ycmd/completers/cs/solutiondetection.py, line 93 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/457#-KFAMh7ZTsz83GMbg5ab:-KFAMh7_74tAPdzQi1tQ:1526774622) ([raw file](https://github.com/valloric/ycmd/blob/e368ff269e48954a0a9fa7352dfac228634c6484/ycmd/completers/cs/solutiondetection.py#L93)):*
The pythonic way of writing this check is to do `if not candidates`. The check below yours is also bad and should be `if candidates`. Please fix that too while you're here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/457#-:-KFAMuuVyW4aSn-CXiCE:-133243377)*

Valloric:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/457)*

Valloric:
PR seems stalled so closing. We can reopen in the future if needed.

Wasn't the project.json way deprecated recently?

teonivalois:
I couldn't get the unit tests working based on the specs. we also had the project.json deprecated. so it's fine to close.

PR #458, opened 2016-04-14
TypeScript: Add GoToType command

icholy:
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/458)

### DIFF
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -359,6 +359,7 @@ class TypeScriptCompleter( Completer ):
                description = ref[ 'lineText' ]
              ) for ref in response[ 'refs' ] ]
 
+
   def _GoToType( self, request_data ):
     self._Reload( request_data )
     try:
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -301,6 +301,8 @@ class TypeScriptCompleter( Completer ):
                            self._GoToDefinition( request_data ) ),
       'GoToReferences' : ( lambda self, request_data, args:
                            self._GoToReferences( request_data ) ),
+      'GoToType'       : ( lambda self, request_data, args:
+                           self._GoToType( request_data ) ),
       'GetType'        : ( lambda self, request_data, args:
                            self._GetType( request_data ) ),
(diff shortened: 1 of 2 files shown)


Valloric:
Thanks for the PR!

Looks good in general, but the coverage bot says coverage decreases with this PR; that's not good. Please improve test coverage.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/458#-:-KFL0UMabeQFf1QE80TO:-1756430183)*

vheon:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/458#-:-KFLZmUfqP0_TOhBTL3i:1417088721)*

Valloric:
Nice!

@homu r=vheon

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/458#-:-KFL_VrRhxG9jx6Ro66o:165669495)*

Valloric:
I think we might have pulled the trigger too early on this PR. I thought we had a `GoToType` subcommand already supported in one of the other completers, but we don't. The reason why is that this is just the `GoTo` command. There should be no need for a separate `GoToType` subcommand.

@icholy Could you send a separate PR to rename the new subcommand to `GoTo`? We'll also need a separate PR to YCM that updates the docs (typescript should be listed as supported for `GoTo`).

vheon:
> I think we might have pulled the trigger too early on this PR. I thought we had a GoToType subcommand already supported in one of the other completers, but we don't. The reason why is that this is just the GoTo command. There should be no need for a separate GoToType subcommand.

Well they are not the same thing. A `GoTo` try to go to the definition or the declaration of what is under the cursor while this command moves you to declaration of the type of the thing under the cursor. So in the positive test that we have now a `GoTo` would tell you that you already are in the declaration of what is under thee cursor, while the `GoToType` moves you on the declaration of the `Foo` type.

Valloric:
@vheon You're right. My mind is fully in statically typed languages, where you'd use GoTo on the variable to jump to the declaration and GoTo on the type in the decl to jump to the type definition. But with TS, the type is possibly not there, so a separate GoToType subcomman makes sense.

I take what I said back then. We just need docs for the new GoToType subcommand.

PR #459, opened 2016-04-15
[READY] Fix and improve pyenv setup in Travis script

micbou:
These Git commands:
```sh
git fetch --tags
git checkout v20160202
```
are not executed in the right folder when setting up pyenv on Travis. Execute them in the pyenv folder.

Instead of using `set +e`  and `set -e`, check if the `~/.pyenv` folder exists before cloning the pyenv repository.

Remove the `git fetch --tags` command because it is not needed after cloning a repository.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/459)

PR #460, opened 2016-04-17
[READY] Fix pyenv setup on Travis

micbou:
I made two mistakes in [previous PR](https://github.com/Valloric/ycmd/pull/459) (shame on me).
First mistake is that even if Travis don't cache anything, an empty directory is still created so the condition that checks if the `pyenv` directory doesn't exist is always false, resulting in a failure when nothing is cached. We fix this by checking if the `.git` folder exists instead.
Second mistake is the removal of the tags fetching: it is still useful if we bump the `pyenv` tag without clearing the cache (may not be a good idea though).

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/460)

### DIFF
--- a/ci/travis/travis_install.sh
+++ b/ci/travis/travis_install.sh
@@ -17,10 +17,11 @@ source ci/travis/travis_install.${TRAVIS_OS_NAME}.sh
 
 export PYENV_ROOT="${HOME}/.pyenv"
 
-if [ ! -d "${PYENV_ROOT}" ]; then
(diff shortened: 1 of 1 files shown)


vheon:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/460#-:-KFZAS3_oYkOg-w7WLWF:1417088721)*

Valloric:
Thanks! @homu r=vheon

PR #465, opened 2016-04-25
[READY] Clear completions cache when sharing the application between tests

micbou:
### Problem

When sharing the application in tests (using the `@SharedYcmd` decorator), the completions cache is not invalidated between each test. This can make a test not working like expected. For instance, the completions are detailed in the `GetCompletions_MaxDetailedCompletion_test` TypeScript test because of the previous test. However, the test is still passing because the `has_entries` matcher does not check for exact entries.

### Solution

Improve the `GetCompletions_MaxDetailedCompletion_test` TypeScript test by checking that there are no detailed completions.
Invalidate the completions cache for all completers stored by the server state before running each test when sharing the application.

For now, I am only sending the test improvement so that we can confirm the test is failing because we don't clear the completions cache between tests. Then, I'll send the cache invalidation for shared tests.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/465)

### DIFF
--- a/ycmd/tests/__init__.py
+++ b/ycmd/tests/__init__.py
@@ -26,7 +26,7 @@ from builtins import *  # noqa
 import functools
 import os
 
-from ycmd.tests.test_utils import SetUpApp
+from ycmd.tests.test_utils import ClearCompletionsCache, SetUpApp
 
 shared_app = None
 
@@ -55,5 +55,6 @@ def SharedYcmd( test ):
 
   @functools.wraps( test )
   def Wrapper( *args, **kwargs ):
+    ClearCompletionsCache()
     return test( shared_app, *args, **kwargs )
   return Wrapper
--- a/ycmd/tests/clang/__init__.py
+++ b/ycmd/tests/clang/__init__.py
@@ -27,7 +27,7 @@ import functools
 import os
 
 from ycmd import handlers
-from ycmd.tests.test_utils import SetUpApp
+from ycmd.tests.test_utils import ClearCompletionsCache, SetUpApp
 
 shared_app = None
 
@@ -56,6 +56,7 @@ def SharedYcmd( test ):
 
   @functools.wraps( test )
   def Wrapper( *args, **kwargs ):
+    ClearCompletionsCache()
     return test( shared_app, *args, **kwargs )
   return Wrapper
 
--- a/ycmd/tests/cs/__init__.py
+++ b/ycmd/tests/cs/__init__.py
@@ -29,7 +29,7 @@ import os
 import time
 
 from ycmd import handlers
-from ycmd.tests.test_utils import BuildRequest, SetUpApp
+from ycmd.tests.test_utils import BuildRequest, ClearCompletionsCache, SetUpApp
 
 shared_app = None
 shared_filepaths = []
@@ -123,6 +123,7 @@ def SharedYcmd( test ):
 
   @functools.wraps( test )
   def Wrapper( *args, **kwargs ):
+    ClearCompletionsCache()
     return test( shared_app, *args, **kwargs )
   return Wrapper
 
--- a/ycmd/tests/go/__init__.py
+++ b/ycmd/tests/go/__init__.py
@@ -26,7 +26,7 @@ from builtins import *  # noqa
 import functools
 import os
 
-from ycmd.tests.test_utils import BuildRequest, SetUpApp
+from ycmd.tests.test_utils import BuildRequest, ClearCompletionsCache, SetUpApp
 
 shared_app = None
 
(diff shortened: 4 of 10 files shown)


micbou:

---

*[ycmd/tests/typescript/get_completions_test.py, line 102 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/465#-KGCzB9TyRjjbDzpZhy-:-KGCzB9TyRjjbDzpZhy0:-2108125817) ([raw file](https://github.com/valloric/ycmd/blob/94496e3bfdfb93f93c12f48fde30056c9b11382f/ycmd/tests/typescript/get_completions_test.py#L102)):*
If you are wondering why this is a little convoluted, the reason is that some completions may be detailed but not others so we need to be sure that not a single completion has detailed completions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/465)*

micbou:
As expected, the test failed. I've pushed the fix.

---

Reviewed 1 of 1 files at r1, 9 of 9 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/465#-:-KGDCwoZExZ5XnyQ93ss:1082638197)*

vheon:

---

*[ycmd/tests/test_utils.py, line 174 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/465#-KGEFdhspI6mtPGjTSpD:-KGEFdhspI6mtPGjTSpE:-1405979245) ([raw file](https://github.com/valloric/ycmd/blob/51eb91645bbfaf2ceea71781e1afc509602fd9bf/ycmd/tests/test_utils.py#L174)):*
why don't we use `itervalues` here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/465)*

micbou:
The tests should pass now.

---

Reviewed 1 of 1 files at r3.

---

*[ycmd/tests/test_utils.py, line 174 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/465#-KGEFdhspI6mtPGjTSpD:-KGETCNeM1ywffnVQG48:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/51eb91645bbfaf2ceea71781e1afc509602fd9bf/ycmd/tests/test_utils.py#L174)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/465#-:-KGESp8ky9g672Drw2vJ:-782194131)*

Valloric:
Hm, I'm not a fan of how we have to copy-paste the same `CleareCompletionsCache` line across files. This is unlikely to be the last time we need to have extra logic in `@SharedYcmd`. Any way we could unify these?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/465#-:-KGEwSHP0smLeSZcB86C:-1064339427)*

micbou:
Me neither but I don't see how to do it differently. `@SharedYcmd` must be defined by package because it depends on a package-level variable. We can't move it to `test_utils.py`.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/465#-:-KGHJ8Prd9AAsXdnbGXG:1789522658)*

vheon:
I don't see at the moment another way of doing it :confused: so :lgtm: 

---

Reviewed 1 of 1 files at r1, 8 of 9 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/465#-:-KGOgbyE69Lld7P8lD4R:1189038211)*

puremourning:
:lgtm:

We already have some duplication in this space, and this change does't require us to remember to do anything on _each test_, which means it isn't likely to be forgotten (when creating a new wrapper, we'd probably just copy/paste anyway). Duplication is regrettable, but I don't imagine this particular change makes the situation we have any worse.

I'll leave it to @valloric for the final say, of course :)

> I don't see at the moment another way of doing it :confused: so :lgtm: </details>


---

Reviewed 1 of 1 files at r1, 8 of 9 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/465#-:-KHG556OCHTMfuPtCEvT:1432251098)*

Valloric:
I'll live with it. :)

@homu r=vheon

> :lgtm:
> 
> We already have some duplication in this space, and this change does't require us to remember to do anything on _each test_, which means it isn't likely to be forgotten (when creating a new wrapper, we'd probably just copy/paste anyway). Duplication is regrettable, but I don't imagine this particular change makes the situation we have any worse.
> 
> I'll leave it to @valloric for the final say, of course :)</details>


---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/465#-:-KHGIpJZp2ZOifTJywmA:-619657625)*

PR #467, opened 2016-04-26
[READY] Add checks to ycm_core library when starting ycmd

micbou:
Do the following checks on the ycm_core library when starting ycmd:
 - missing library;
 - compiled with Python 2 but imported with Python 3;
 - compiled with Python 3 but imported with Python 2;
 - outdated library.

When one of these errors is encountered, exit with a non-zero status code specific to the error. These status codes will be used in the YCM client to display an appropriate error message to the user. In addition, starting a separate process to check the core version will not be needed anymore. This should slighty improve the Vim startup time. See issue Valloric/YouCompleteMe#2085.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/467)


Valloric:
This is a pretty fantastic change! I'm not a fan of us parsing import error messages, but I don't see a better way of doing it so we'll live.

@abingham @ptrv @mawww @Qusic Note that this is a breaking change to clients because it deletes `check_core_version.py`. See the comments in `CompatibleWithCurrentCore()` in `server_utils.py` in this PR for a list of ycmd process return codes that are replacing that file.

@micbou This PR should probably change the README as well to document this. Shouldn't be much beyond the comment you have in the code transplanted into a new section in the README.

---

Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[ycmd/server_utils.py, line 40 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGK1LqsEgXuGVdhjX_j:-KGK1LqsEgXuGVdhjX_k:551974073) ([raw file](https://github.com/valloric/ycmd/blob/694e5ae6ea63cf19cdaafe814680d4c92f76dd9a/ycmd/server_utils.py#L40)):*
This message is likely to be encountered by people completely unfamiliar with YCM so it should be more detailed. For instance, it should refer to the installation guide, much like the old message in handlers.py.

In general, I try to avoid emitting error messages that don't indicate a clear way towards solving the problem. "You need to compile it" may be clear to me and you, but the more direct we are, the fewer issue reports we'll get.

---

*[ycmd/tests/server_utils_test.py, line 66 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGK27AG0du3DF75JSe0:-KGK27AG0du3DF75JSe1:1646437207) ([raw file](https://github.com/valloric/ycmd/blob/694e5ae6ea63cf19cdaafe814680d4c92f76dd9a/ycmd/tests/server_utils_test.py#L66)):*
This 0, 1, 0 construction appears too magic. Can we refactor this a bit so that it's more clear to the reader what's going on?

I think we used a different idiom for asserting calls on mocks in other tests, no?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467#-:-KGK2aqwiDEncZJTL-6_:-1317941234)*

puremourning:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/server_utils.py, line 71 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGM-TMHErnDxP9iUurU:-KGM-TMHErnDxP9iUurV:756412690) ([raw file](https://github.com/valloric/ycmd/blob/694e5ae6ea63cf19cdaafe814680d4c92f76dd9a/ycmd/server_utils.py#L71)):*
Drive-by comment: can we define the return codes with symbolic names? They can then also be imported by the client to avoid magic numbers. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467)*

micbou:
I added a section to the README and I assigned a specific exit code for the "unexpected error" case instead of using the general exit code `1` so that we can handle it in the clients.

---

Reviewed 5 of 6 files at r1, 1 of 1 files at r2, 2 of 3 files at r4, 2 of 2 files at r5, 1 of 1 files at r6.

---

*[ycmd/server_utils.py, line 40 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGK1LqsEgXuGVdhjX_j:-KGNOTLgrTpcC8vDJZZo:-164052590) ([raw file](https://github.com/valloric/ycmd/blob/694e5ae6ea63cf19cdaafe814680d4c92f76dd9a/ycmd/server_utils.py#L40)):*
I updated the messages.

---

*[ycmd/server_utils.py, line 71 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGM-TMHErnDxP9iUurU:-KGNP0CTh0PlRhcNrGGV:723852259) ([raw file](https://github.com/valloric/ycmd/blob/694e5ae6ea63cf19cdaafe814680d4c92f76dd9a/ycmd/server_utils.py#L71)):*
Done. Since the symbolic names can only be used by python clients, the README only talks about the hardcoded values. 

---

*[ycmd/tests/server_utils_test.py, line 66 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGK27AG0du3DF75JSe0:-KGN_eQ8bOhSb5brlCeT:1750946275) ([raw file](https://github.com/valloric/ycmd/blob/694e5ae6ea63cf19cdaafe814680d4c92f76dd9a/ycmd/tests/server_utils_test.py#L66)):*
I added the `log_level` parameter to the tests so that we can easily use the `assert_called_once_with` method. I think it is clearer this way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467#-:-KGNbAfRcQpQjKz0oqTt:734694135)*

Valloric:
:lgtm:

---

Reviewed 4 of 6 files at r1, 2 of 2 files at r5, 1 of 1 files at r6.

---

*[ycmd/server_utils.py, line 71 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGM-TMHErnDxP9iUurU:-KGOED0W0OcUBQ1qxqmA:-723350966) ([raw file](https://github.com/valloric/ycmd/blob/694e5ae6ea63cf19cdaafe814680d4c92f76dd9a/ycmd/server_utils.py#L71)):*
The documentation of what the states mean should now probably be on the constants themselves instead of here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467#-:-KGOEQSUi3xH_1_LZYOL:1417088721)*

micbou:
Reviewed 1 of 1 files at r7.

---

*[ycmd/server_utils.py, line 71 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGM-TMHErnDxP9iUurU:-KGONzNOMH-h8OXue8QD:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/694e5ae6ea63cf19cdaafe814680d4c92f76dd9a/ycmd/server_utils.py#L71)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467)*

Valloric:

---

*[ycmd/server_utils.py, line 51 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGOTnRxm0nummx440Kv:-KGOTnRxm0nummx440Kw:1792158097) ([raw file](https://github.com/valloric/ycmd/blob/bb062da537ffcd5177d501dd9313734435b15a66/ycmd/server_utils.py#L51)):*
Why the doc comment? Why not just lines prefixed with `#`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467)*

vheon:
Reviewed 4 of 6 files at r1, 1 of 2 files at r5, 1 of 1 files at r6, 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[ycmd/tests/server_utils_test.py, line 73 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGOVHYbVxDwHC6xMlMA:-KGOVHYc7uKipsrZQHIJ:474180380) ([raw file](https://github.com/valloric/ycmd/blob/bb062da537ffcd5177d501dd9313734435b15a66/ycmd/tests/server_utils_test.py#L73)):*
If we inline the implementation for this case we could simplify the implementation of `RunCompatibleWithCurrentCore` and improve its readability. If we do that maybe we can also rename it like `RunNonCompatibleWithCurrentCore` since we would be using it only for the non-compatible cases. What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467)*

micbou:
Reviewed 1 of 1 files at r8.

---

*[ycmd/server_utils.py, line 51 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGOTnRxm0nummx440Kv:-KGOXAfj9Lal4Ii7oC9w:-311602127) ([raw file](https://github.com/valloric/ycmd/blob/bb062da537ffcd5177d501dd9313734435b15a66/ycmd/server_utils.py#L51)):*
Trying to be smart I guess. I changed it.

---

*[ycmd/tests/server_utils_test.py, line 73 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGOVHYbVxDwHC6xMlMA:-KGOXm9Cg7ZID21dQf5q:-1825180131) ([raw file](https://github.com/valloric/ycmd/blob/bb062da537ffcd5177d501dd9313734435b15a66/ycmd/tests/server_utils_test.py#L73)):*
We can't really simplify the `RunCompatibleWithCurrentCore` function because there are other tests not using the `import_error` key: the two `CompatibleWithCurrentCore_Outdated` tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467)*

vheon:

---

*[ycmd/tests/server_utils_test.py, line 73 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGOVHYbVxDwHC6xMlMA:-KGTKUpOcFOdIjxSjvfJ:-1486964134) ([raw file](https://github.com/valloric/ycmd/blob/bb062da537ffcd5177d501dd9313734435b15a66/ycmd/tests/server_utils_test.py#L73)):*
I've overlooked that apparently. Is that  to me it looks like we're trying to squeeze three kind of test into one only because they share 1 line of code and the other one a similar. But I guess is just me :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467)*

micbou:
I made some changes to the tests (they are now more robust and more readable). I am marking the PR as ready. Tell me if there is something that should be improved before we merge this.

---

Reviewed 1 of 1 files at r9.

---

*[ycmd/tests/server_utils_test.py, line 73 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGOVHYbVxDwHC6xMlMA:-KGh_KMpGWz8XbDGzaT6:-1415831610) ([raw file](https://github.com/valloric/ycmd/blob/bb062da537ffcd5177d501dd9313734435b15a66/ycmd/tests/server_utils_test.py#L73)):*
So I followed your advice of not merging all the tests into one and I think you were right, it is better now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467#-:-KGhaF7CCTjMZf_iXlr1:-1379305924)*

Valloric:
I continue to be satisfied with this. :)

---

Reviewed 1 of 1 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467#-:-KGhqLGkHXW_Ej5FeGr8:-1561552746)*

vheon:
Reviewed 1 of 1 files at r8.

---

*[ycmd/tests/server_utils_test.py, line 73 \[r7\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGOVHYbVxDwHC6xMlMA:-KGi3v3V6ldKG8TQmjkX:-470081708) ([raw file](https://github.com/valloric/ycmd/blob/bb062da537ffcd5177d501dd9313734435b15a66/ycmd/tests/server_utils_test.py#L73)):*
I like it better now :+1: 

---

*[ycmd/tests/server_utils_test.py, line 150 \[r9\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGi4R6xml9t_izXY7l0:-KGi4R6xml9t_izXY7l1:-379649233) ([raw file](https://github.com/valloric/ycmd/blob/85d5f6bbc7bf003fc7a6253a6786ed0f459c6019/ycmd/tests/server_utils_test.py#L150)):*
Should we use the variable names instead of the bare numbers like for the exit status? (I'm saying this here because if it is the case then we should add a comment to the top of this method to explain why there are `10` and `11` as `return_value` for some mocked function.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467)*

micbou:

---

*[ycmd/tests/server_utils_test.py, line 150 \[r9\]](https://reviewable.io:443/reviews/valloric/ycmd/467#-KGi4R6xml9t_izXY7l0:-KGi6my3Ni3tv-5tzgSo:-1069821069) ([raw file](https://github.com/valloric/ycmd/blob/85d5f6bbc7bf003fc7a6253a6786ed0f459c6019/ycmd/tests/server_utils_test.py#L150)):*
It is better to use the numbers in tests so that we also check the correspondance between these numbers and the symbolic names.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467)*

vheon:
Then I believe that this is ready to go. :lgtm: @homu r=Valloric

---

Reviewed 1 of 1 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/467#-:-KGnDfpByVG76eBbN2wK:-1792667422)*

PR #468, opened 2016-04-27
More helping message at opening: "Found ... Load?"

maelvalais:
I found it REALLY frustrating that YouCompleteMe won't provide you more help on error messages. I went to :h YouCompleteMe but no way to find the place where I could eventually turn of that warning.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/468)

### DIFF
--- a/ycmd/responses.py
+++ b/ycmd/responses.py
@@ -28,7 +28,7 @@ import os
 YCM_EXTRA_CONF_FILENAME = '.ycm_extra_conf.py'
 
 CONFIRM_CONF_FILE_MESSAGE = ('Found {0}. Load? \n\n(Question can be turned '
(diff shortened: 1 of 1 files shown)


vheon:
I'm afraid we cannot accept this change since `ycmd` is editor agnostic while you're proposing to look into the vim help doc.

maelvalais:
Hmm I didn't think of that. Still, having those kind of "go to documentation and read 10 pages before you can do anything" is a pain. Bummer :(

Valloric:
Like @vheon said, we can't accept this, ycmd is used in emacs as well. Thanks for feedback about the error message though, we'll take it into consideration.

PR #470, opened 2016-04-28
Use libclang prebuilt binaries even for 32 bit linux

voria:
llvm.org provides prebuilt x86 binaries for Fedora 23, use them instead
of forcing the user to compile clang from source.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/470)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -58,11 +58,11 @@ if ( USE_CLANG_COMPLETER AND
            "3120c3055ea78bbbb6848510a2af70c68538b990cb0545bac8dad01df8ff69d7" )
       set( CLANG_FILENAME "${CLANG_DIRNAME}.tar.xz" )
     else()
-      # Clang 3.3 is the last version with pre-built x86 binaries upstream.
-      # That's too old now.
-      message( FATAL_ERROR "No pre-built Clang ${CLANG_VERSION} binaries for 32 bit linux. "
-        "You'll have to compile Clang ${CLANG_VERSION} from source. "
-        "See the YCM docs for details on how to use a user-compiled libclang." )
+      # Use the prebuilt x86 binaries for Fedora 23.
+      set( CLANG_DIRNAME "clang+llvm-${CLANG_VERSION}-i686-fedora23" )
+      set( CLANG_SHA256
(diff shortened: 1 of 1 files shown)


Valloric:
Thanks for sending a PR!

Hm, unsure about this... It's peculiar that llvm.org does not provide 32bit binaries for Ubuntu but does for Fedora. Makes me think that this might blow up if loaded on a 32bit Ubuntu. That's not necessarily a blocker for merging this, but would be good to know before we do.

@voria Could you check this on a 32 bit Ubuntu LTS and report back?

voria:
I just tested it in a virtual machine, on a ubuntu 16.04 LTS i386 installation, and everything works as expected.

Effectively, it is strange that llvm.org does not provide x86 binaries for ubuntu and debian. Yet they do for Fedora and OpenSUSE. I suppose it's so because they already provide [apt repositories for debian based distributions](http://llvm.org/apt/) (I'm speculating).

vheon:
> I just tested it in a virtual machine, on a ubuntu 16.04 LTS i386 installation, and everything works as expected.

I was afraid that being compiled for Fedora it had a different glib but if you have tested and it returns completions just fine then I would go for it. Should we add a 32 bit box on Travis (if we don't have one already)?

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/470#-:-KGYAB48UbWxNZ1XCo6P:-89381343)*

Valloric:
If you tried it and it worked, :lgtm: from me.

Thanks again for the PR! :)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/470#-:-KGYrj_VATOQ60q-one8:-502638199)*

micbou:
> Should we add a 32 bit box on Travis (if we don't have one already)?

There is no 32-bit Ubuntu available on Travis.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/470#-:-KGaZtR6gd8GIWmYC2xT:7086054)*

vheon:
> There is no 32-bit Ubuntu available on Travis.

So there is nothing else we can do here. 

:lgtm: @homu r=Valloric

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/470#-:-KGbgPIGelyxOHLskz99:-891635696)*

PR #480, opened 2016-05-07
Add tex identifier regex

bijancn:
Commonly used latex identifiers are not covered by the default regex.
The new regex is covered by test cases similar to the ones for perl and haskell.

With this regex it is easy to reference with YCM from all figures, sections, etc. in the document. Using `fig:` and so on is best practice in latex but I haven't found a spec that could serve as reference.

I wasn't able to run the full test suite (see message on ycmd-users) but I ran
```
nosetests ycmd/tests/identifier_utils_test.py
```
successfully.

Ultimatively, I think the user should be able to give a custom identifier regex for different filetypes in their .vimrc (and its analogue in other editors).

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/480)

### DIFF
--- a/ycmd/identifier_utils.py
+++ b/ycmd/identifier_utils.py
@@ -93,6 +93,10 @@ FILETYPE_TO_IDENTIFIER_REGEX = {
     # Section 2.4
     'haskell': re.compile( r"[_a-zA-Z][\w']*", re.UNICODE ),
 
+    # Spec: ?
+    # Labels like \label{fig:foobar} are very common
+    'tex': re.compile( r"[_a-zA-Z:-]*", re.UNICODE ),
+
     # Spec: http://doc.perl6.org/language/syntax
     'perl6': re.compile( r"[_a-zA-Z](?:\w|[-'](?=[_a-zA-Z]))*", re.UNICODE ),
 }
--- a/ycmd/identifier_utils.py
+++ b/ycmd/identifier_utils.py
@@ -91,11 +91,11 @@ FILETYPE_TO_IDENTIFIER_REGEX = {
 
     # Spec: http://www.haskell.org/onlinereport/lexemes.html
     # Section 2.4
-    'haskell': re.compile( r"[_a-zA-Z][\w']*", re.UNICODE ),
+    'haskell': re.compile( r"[_a-zA-Z][\w']+", re.UNICODE ),
 
     # Spec: ?
     # Labels like \label{fig:foobar} are very common
-    'tex': re.compile( r"[_a-zA-Z:-]*", re.UNICODE ),
+    'tex': re.compile( r"[_a-zA-Z:-]+", re.UNICODE ),
 
     # Spec: http://doc.perl6.org/language/syntax
     'perl6': re.compile( r"[_a-zA-Z](?:\w|[-'](?=[_a-zA-Z]))*", re.UNICODE ),
--- a/ycmd/tests/identifier_utils_test.py
+++ b/ycmd/tests/identifier_utils_test.py
@@ -183,6 +183,7 @@ def IsIdentifier_Css_test():
   ok_( not iu.IsIdentifier( '-3' , 'css' ) )
   ok_( not iu.IsIdentifier( '3'  , 'css' ) )
   ok_( not iu.IsIdentifier( 'a'  , 'css' ) )
+  ok_( not iu.IsIdentifier( '' , 'css' ) )
 
 
 def IsIdentifier_R_test():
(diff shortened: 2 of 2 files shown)


Valloric:
Thanks for the PR! 

:lgtm: 

> Ultimatively, I think the user should be able to give a custom identifier regex for different filetypes in their .vimrc (and its analogue in other editors).

This would lead to a similar situation we have today with the `g:ycm_semantic_triggers` option. People who know what a better trigger should be for a language fix it for themselves and never send a PR, so everyone else loses out.

Unfortunately we have to more strongly motivate people to do the right thing.

---

---

*[ycmd/identifier_utils.py, line 98 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/480#-KHBosrg7fgVs4z3nSEm:-KHBosrg7fgVs4z3nSEn:-226277850) ([raw file](https://github.com/valloric/ycmd/blob/f65c7796f7316a1335b3da961d2487fc7c7f7347/ycmd/identifier_utils.py#L98)):*
Shouldn't this use `+` instead of `*`? I doubt an empty string is a valid identifier. Having a test for empty string would be a good idea too.

While you're at it, could you fix the haskell one above too? It has the same issue.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/480#-:-KHBo40syqzA8LnIPC7H:-1445214916)*

bijancn:
I was just concerned that in some languages it might be a matter of convention what should be allowed as identifier. But I guess YCM can always cover the bigger set.

I changed the `*` to `+` as it makes more sense but the empty test also works with `*`.
It is catched by this `if`
```
def IsIdentifier( text, filetype = None ):
  if not text:
    return False
```

BTW shouldn't `frozendict` and `bottle` be in the `test_requirements.txt`? I needed them to run the test.

Valloric:
> BTW shouldn't frozendict and bottle be in the test_requirements.txt? I needed them to run the test.

They're in third_party folder. I recommend using our vagrant environment for ycmd development. See [DEV_SETUP](https://github.com/Valloric/ycmd/blob/master/DEV_SETUP.md).

bijancn:
Well I tried vagrant. On one machine I couldn't use it because in the Ubuntu 14.04 packages there is only an old version that doesn't find your machine and I have no root access there. On the other machine I think it failed due to a lack of RAM and/or disk space. Something a bit more lightweight like docker would be nice.

Anyhow, if I understand this reviewable correctly now this PR only needs two thumbs up?

vheon:
> there is only an old version that doesn't find your machine

I'm sorry, what do you mean? An old version of what? And which machine of ours don't you find?

bijancn:
Sorry for not being clear. The version of vagrant (1.4.3) in the Ubuntu 14.04 packages does not work with your Vagrantfile which demands "ubuntu/trusty64" which is to be expected if I understand https://github.com/Varying-Vagrant-Vagrants/VVV/issues/354 correctly. So there is no trusty inception possible ;). And I didn't want to go through the hassle of building vagrant myself just to be able to build ycmd.

vheon:
> "ubuntu/trusty64" 

That is the image that will be downloaded from the vagrantcloud, not the version of ubuntu required.

bijancn:
Yes I am aware of that. Vagrant 1.4.3 does not find trusty64 in the vagrantcloud as a lot of people are confirming in the issue I mentioned.

vheon:
> Yes I am aware of that

Sorry 😜

puremourning:
If you're not able to get our vagrant setup working, then there are some [manual instructions](https://github.com/Valloric/ycmd/blob/master/TESTS.md).

Alternatively, if you're totally stuck, you can just enable Travis for your fork and Travis will run everything when you push a commit. Painful and slow TBH. I successfully run the tests on multiple OSes outside of our vagrant setup.

bijancn:
Thanks :smiley:. I will try manual installation next time harder.

Would still be happy to see 2 LGTMs as all tests pass.

puremourning:
:lgtm:

Thanks!

@homu r=Valloric

<details><summary><i>Previously, bijancn wrote…</i></summary>
> Thanks :smiley:. I will try manual installation next time harder.
> 
> 
> 
> Would still be happy to see 2 LGTMs as all tests pass.</details>


---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/480#-:-KHG0tPeYMtV8d1RVCWr:-1791670325)*

Issue #476, opened 2016-05-04
YCM runs on large-ish non-source files, spews a backtrace when you enter insert mode


jlebar:
STR:

    # Create a file that's large, but not so large as to cause ycm to disable itself
    $ cd `mktemp -d`
    $ dd if=/dev/urandom of=test bs=500k count=1
    $ vim test

Now enter insert mode.

Result: YCM error spew, included below.

I bisected the test file size and noticed that the error starts occurring somewhere between files of size 100k and 300k.

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/youcompleteme/autoload/../python/ycm/youcompleteme.py",
line 255, in NativeFiletypeCompletionUsable
    self.NativeFiletypeCompletionAvailable() )
  File "/usr/lib/youcompleteme/autoload/../python/ycm/youcompleteme.py",
line 250, in NativeFiletypeCompletionAvailable
    vimsupport.CurrentFiletypes() ] )
  File "/usr/lib/youcompleteme/autoload/../python/ycm/youcompleteme.py",
line 240, in FiletypeCompleterExistsForFiletype
    exists_completer = SendCompleterAvailableRequest( filetype )
  File "/usr/lib/youcompleteme/autoload/../python/ycm/client/completer_available_request.py",
line 57, in SendCompleterAvailableRequest
    request.Start()
  File "/usr/lib/youcompleteme/autoload/../python/ycm/client/completer_available_request.py",
line 45, in Start
    'semantic_completion_available' )
  File "/usr/lib/youcompleteme/autoload/../python/ycm/client/base_request.py",
line 81, in PostDataToHandler
    timeout ) )
  File "/usr/lib/youcompleteme/autoload/../python/ycm/client/base_request.py",
line 177, in JsonFromFuture
    _ValidateResponseObject( response )
  File "/usr/lib/youcompleteme/autoload/../python/ycm/client/base_request.py",
line 206, in _ValidateResponseObject
    their_hmac = ToBytes( b64decode( response.headers[ _HMAC_HEADER ] ) )
  File "/usr/lib/youcompleteme/third_party/ycmd/third_party/requests/requests/structures.py",
line 54, in __getitem__
    return self._store[key.lower()][1]
KeyError: u'x-ycm-hmac'
E858: Eval did not return a valid python object

```
+@r4nt

vheon:
[...]

PR #482, opened 2016-05-07
[READY] Return a proper response when the request is too big.

vheon:
We have a hard limit on the size of the request that ycmd can handle that is set to 1Mb. This is dictated by Bottle and there is no way around it.

#### Why is this a problem:

When a client makes a request greater than 1Mb Bottle returns an error that is not generated by us and thus it is not a json response, but an html one, and it doesn't contain the HMAC header which causes  errors like #476. 

For now I've pushed only the failing test. Once I see it fail on Travis I will push the fix.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/482)

### DIFF
--- a/ycmd/handlers.py
+++ b/ycmd/handlers.py
@@ -25,7 +25,6 @@ from builtins import *  # noqa
 
 import atexit
 import bottle
-import http.client
 import json
 import logging
 import traceback
@@ -221,12 +220,14 @@ def DebugInfo():
 
 
 # The type of the param is Bottle.HTTPError
-@app.error( http.client.INTERNAL_SERVER_ERROR )
 def ErrorHandler( httperror ):
   body = _JsonResponse( BuildExceptionResponse( httperror.exception,
                                                 httperror.traceback ) )
   hmac_plugin.SetHmacHeader( body, _hmac_secret )
   return body
+
+# For every error Bottle encounters it will use this as the default handler
+app.default_error_handler = ErrorHandler
 
 
 def _JsonResponse( data ):
--- a/ycmd/tests/misc_handlers_test.py
+++ b/ycmd/tests/misc_handlers_test.py
@@ -56,7 +56,6 @@ def MiscHandlers_EventNotification_ReturnJsonOnBigFileError_test( app ):
   app.post_json( '/event_notification', event_data, expect_errors = True ).json
 
 
-
 @SharedYcmd
 def MiscHandlers_FilterAndSortCandidates_Basic_test( app ):
   candidate1 = { 'prop1': 'aoo', 'prop2': 'bar' }
--- a/ycmd/tests/misc_handlers_test.py
+++ b/ycmd/tests/misc_handlers_test.py
@@ -47,6 +47,17 @@ def MiscHandlers_EventNotification_AlwaysJsonResponse_test( app ):
 
 
 @SharedYcmd
+def MiscHandlers_EventNotification_ReturnJsonOnBigFileError_test( app ):
+  # We generate a content greater than Bottle.MEMFILE_MAX, which is set to 1Mb.
+  contents = "foo " * 500000
+  event_data = BuildRequest( contents = contents,
+                             event_name = 'FileReadyToParse' )
+
+  app.post_json( '/event_notification', event_data, expect_errors = True ).json
+
+
+
+@SharedYcmd
 def MiscHandlers_FilterAndSortCandidates_Basic_test( app ):
   candidate1 = { 'prop1': 'aoo', 'prop2': 'bar' }
   candidate2 = { 'prop1': 'bfo', 'prop2': 'zoo' }


Valloric:
I think we have something like a "known to fail" annotation that @puremourning introduced; can't recall the exact name. This seems like a good use of it.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482#-:-KHBGbMZo3G7FwHzJOKq:-819905319)*

micbou:
Is it not better to handle this on the client side by [only validating the response after the `raise_for_status` instead of the other way around](https://github.com/Valloric/YouCompleteMe/blob/master/python/ycm/client/base_request.py#L174)? Is it a security hole to do that?

> I think we have something like a "known to fail" annotation that @puremourning introduced; can't recall the exact name. This seems like a good use of it.

[`ExpectedFailure`](https://github.com/Valloric/ycmd/blob/master/ycmd/tests/test_utils.py#L183)

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482#-:-KHBK7qj7l5n-zQ1NjQs:-1694425322)*

vheon:
> I think we have something like a "known to fail" annotation that @puremourning introduced; can't recall the exact name. This seems like a good use of it.

I don't want to test that we fail but that we return the correct response even if we fail in that case. I've renamed the test to better explain what I wanted to test.

> Is it not better to handle this on the client side by only validating the response after the raise_for_status instead of the other way around? Is it a security hole to do that?

I think that since is a hard limit we should return a meaningful error on all client. And why we would return a correct response (json hmac-signed) for a 500 error and not for this when is a known limitation on our side?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482#-:-KHBQTqQX2FW16Aml1zb:-423786076)*

puremourning:
Tidy!

:lgtm:

---

---

*[ycmd/tests/misc_handlers_test.py, line 59 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHBTjryKrYzsJB3vQyq:-KHBTjrzVwYEthEFnexV:-352080927) ([raw file](https://github.com/valloric/ycmd/blob/0e01c0fcefdd5ae7031699a90320324e72e47d3c/ycmd/tests/misc_handlers_test.py#L59)):*
3 blank lines is just opulent. Let's not use up all the worlds blank lines!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482#-:-KHBTysTxDrPXRBj34qj:1266560606)*

vheon:

---

*[ycmd/tests/misc_handlers_test.py, line 59 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHBTjryKrYzsJB3vQyq:-KHBdlYYutf3f24dzA8V:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/0e01c0fcefdd5ae7031699a90320324e72e47d3c/ycmd/tests/misc_handlers_test.py#L59)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482)*

micbou:
> I think that since is a hard limit we should return a meaningful error on all client. And why we would return a correct response (json hmac-signed) for a 500 error and not for this when is a known limitation on our side?

That's a good point.

:lgtm:

---

Reviewed 1 of 1 files at r2, 1 of 1 files at r4.

---

*[ycmd/handlers.py, line 224 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHBfwN-3u8w_wE_T3pk:-KHBfwN-3u8w_wE_T3pl:1522058146) ([raw file](https://github.com/valloric/ycmd/blob/0e01c0fcefdd5ae7031699a90320324e72e47d3c/ycmd/handlers.py#L224)):*
I didn't expected the fix to be as simple as this. That's great.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482#-:-KHBgIloaAKcZ9-MN59x:-1357122042)*

Valloric:

---

*[ycmd/handlers.py, line 224 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHBfwN-3u8w_wE_T3pk:-KHBnIA0FhlbJijY3oQD:1388295617) ([raw file](https://github.com/valloric/ycmd/blob/0e01c0fcefdd5ae7031699a90320324e72e47d3c/ycmd/handlers.py#L224)):*
Wait, this is it? How does this even work?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482)*

vheon:

---

*[ycmd/handlers.py, line 224 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHBfwN-3u8w_wE_T3pk:-KHBq79jOKeps5UuXkDV:305444122) ([raw file](https://github.com/valloric/ycmd/blob/0e01c0fcefdd5ae7031699a90320324e72e47d3c/ycmd/handlers.py#L224)):*
When bottle encounter an error it uses its own error page except when you tell otherwise. Until now we told Bottle that when it encounter a 500 error (like when a completer raises an Exception) to use this function instead, which returns to the user a JSON formatted response signed with the HMAC header. When we send a request greater than 1Mb (which exceed the MEMFILE_MAX) Bottle do not return a 500 error (which would be handled by this function, but return a 413 which is handled by Bottle itself and return an HTML response. With this line we tell Bottle to treat the 413 like the 500.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482)*

Valloric:

---

*[ycmd/handlers.py, line 224 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHBfwN-3u8w_wE_T3pk:-KHBsXD6cK0aiR6vu-T9:1999313977) ([raw file](https://github.com/valloric/ycmd/blob/0e01c0fcefdd5ae7031699a90320324e72e47d3c/ycmd/handlers.py#L224)):*
Makes sense. Is there any other such special error code that Bottle will try to handle itself? Probably best if we didn't let it handle anything internally because whatever it sends back will fail hmac validation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482)*

vheon:

---

*[ycmd/handlers.py, line 224 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHBfwN-3u8w_wE_T3pk:-KHBteTmLN9dhN9TPApm:-718672437) ([raw file](https://github.com/valloric/ycmd/blob/0e01c0fcefdd5ae7031699a90320324e72e47d3c/ycmd/handlers.py#L224)):*
I was asking this myself but from the top of my head I couldn't think of any other error that we could encounter :confused:  I don't know if there is a way to tell Bottle "hey for every error you stumble upon use this function".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482)*

micbou:

---

*[ycmd/handlers.py, line 224 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHBfwN-3u8w_wE_T3pk:-KHBuUTIgZ0xDR6IDl3S:1758943431) ([raw file](https://github.com/valloric/ycmd/blob/0e01c0fcefdd5ae7031699a90320324e72e47d3c/ycmd/handlers.py#L224)):*
Here's the list: https://docs.python.org/3/library/http.html#http-status-codes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482)*

vheon:

---

*[ycmd/handlers.py, line 224 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHBfwN-3u8w_wE_T3pk:-KHBuulFTqTXW030zF1z:-1499462913) ([raw file](https://github.com/valloric/ycmd/blob/0e01c0fcefdd5ae7031699a90320324e72e47d3c/ycmd/handlers.py#L224)):*
> Here's the list: https://docs.python.org/3/library/http.html#http-status-codes

Yes, but they are too many to be handled like:

```python
@app.error( http.client.CONTINUE )
@app.error( http.client.SWITCHING_PROTOCOLS )
...
```

Reading the Bottle docs it seems that there is no documented function to do the "all" error thing. But if we look at the code http://bottlepy.org/docs/dev/_modules/bottle.html#Bottle.error we could do something like:

```python
app.default_error_handler = ErrorHandler
```

And we would be good. I know is not documented but following the python convention that should be a public member, am I correct?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482)*

micbou:

---

*[ycmd/handlers.py, line 224 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHBfwN-3u8w_wE_T3pk:-KHByAIbwNpeQYSGfgNF:-431917424) ([raw file](https://github.com/valloric/ycmd/blob/0e01c0fcefdd5ae7031699a90320324e72e47d3c/ycmd/handlers.py#L224)):*
I tried your suggestion and it seems to work so I am fine with it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482)*

Valloric:

---

*[ycmd/handlers.py, line 224 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHBfwN-3u8w_wE_T3pk:-KHCT-kYsRltZ3dennQe:228365858) ([raw file](https://github.com/valloric/ycmd/blob/0e01c0fcefdd5ae7031699a90320324e72e47d3c/ycmd/handlers.py#L224)):*
+1 on us adding a custom default error handler.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482)*

vheon:

---

*[ycmd/handlers.py, line 224 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHBfwN-3u8w_wE_T3pk:-KHCUt0Otvqz2H5lDKWP:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/0e01c0fcefdd5ae7031699a90320324e72e47d3c/ycmd/handlers.py#L224)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482)*

vheon:
Review status: 1 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/handlers.py, line 231 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHCWbzj6lXc8ytegANt:-KHCWbzj6lXc8ytegANu:1252513132) ([raw file](https://github.com/valloric/ycmd/blob/ba337eae1031361a19fb890f4582049602196bbb/ycmd/handlers.py#L231)):*
I thought of making a little decorator to mark the `ErrorHandler` function as the Bottle default error handler (I don't quite like this line laying around here) but I feared that it was too overengineering a simple thing. Let me know what do you think. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482)*

Valloric:

---

*[ycmd/handlers.py, line 231 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/482#-KHCWbzj6lXc8ytegANt:-KHE3To9AfOMTMfQ7B4p:-2105960664) ([raw file](https://github.com/valloric/ycmd/blob/ba337eae1031361a19fb890f4582049602196bbb/ycmd/handlers.py#L231)):*
> I thought of making a little decorator to mark the `ErrorHandler` function as the Bottle default error handler (I don't quite like this line laying around here) but I feared that it was too overengineering a simple thing. Let me know what do you think. </details>
Let's go with the simple approach for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482)*

micbou:
@homu r+

> 
> [Coverage Status](https://coveralls.io/builds/6089707)
> 
> Coverage remained the same at 84.901% when pulling **6c1bb7f82bdf583b78ec389ab61109d69141aea4 on vheon:fix-no-hmac-header-on-big-request** into **8617de85073fa280c071ae55152b24f909de47ae on Valloric:master**.
> </details>


---

Reviewed 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/482#-:-KHE8tdH6QOzwwqtA6Xr:-1142673777)*

PR #475, opened 2016-05-02
Reduce memory usage by: only allocating arrays when needed, removing …

dgel:
…std::list usage, using short integer indices instead of pointers

The main algorithms used stays the same, including big-O complexity. These changes imply that individual identifiers are restricted to about 30k bytes, which I believe to be reasonable. In total, the 3 changes made here imply a 60-70% reduction in memory usage. I mentioned these changes in #474 , this PR does not include any usage of a more naive algorithm for identifying matches

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/475)

PR #498, opened 2016-05-17
[RFC] Move member variable to function local in LetterNode

vheon:
While I was refreshing my knowledge of the `cpp` code for #475 I noticed that
`letternode_per_text_index` is actually only used during one of the two
constructor (the one for the root node) as a utility data structure, so it doesn't make much sense
to keep it around longer than necessary. Since #475 is proposed I don't know if we really want to merge this, but I wanted to see what do you think, hence the `RFC` mark ;)

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/498)

### DIFF
--- a/cpp/ycm/LetterNode.cpp
+++ b/cpp/ycm/LetterNode.cpp
@@ -18,6 +18,7 @@
 #include "LetterNode.h"
 #include "standard.h"
 
+#include <vector>
 
 namespace YouCompleteMe {
 
@@ -30,22 +31,22 @@ LetterNode::LetterNode( char letter, int index )
 LetterNode::LetterNode( const std::string &text )
   : is_uppercase_( false ),
     index_( -1 ) {
-  letternode_per_text_index_.resize( text.size() );
+  std::vector< LetterNode * > letternode_per_text_index( text.size() );
 
   for ( uint i = 0; i < text.size(); ++i ) {
     char letter = text[ i ];
     LetterNode *node = new LetterNode( letter, i );
     letters_[ letter ].push_back( node );
-    letternode_per_text_index_[ i ] = boost::shared_ptr< LetterNode >( node );
+    letternode_per_text_index[ i ] = node;
   }
 
-  for ( int i = static_cast< int >( letternode_per_text_index_.size() ) - 1;
+  for ( int i = static_cast< int >( letternode_per_text_index.size() ) - 1;
         i >= 0; --i ) {
-    LetterNode *node_to_add = letternode_per_text_index_[ i ].get();
+    LetterNode *node_to_add = letternode_per_text_index[ i ];
 
     for ( int j = i - 1; j >= 0; --j ) {
-      letternode_per_text_index_[ j ]->PrependNodeForLetter( text[ i ],
-                                                             node_to_add );
+      letternode_per_text_index[ j ]->PrependNodeForLetter( text[ i ],
+                                                            node_to_add );
     }
   }
 }
--- a/cpp/ycm/LetterNode.h
+++ b/cpp/ycm/LetterNode.h
@@ -22,9 +22,7 @@
 #include "LetterNodeListMap.h"
 
 #include <boost/utility.hpp>
-#include <boost/shared_ptr.hpp>
 
-#include <vector>
 #include <list>
 #include <string>
 
@@ -59,7 +57,6 @@ public:
 private:
 
   LetterNodeListMap letters_;
-  std::vector< boost::shared_ptr< LetterNode > > letternode_per_text_index_;
   bool is_uppercase_;
   int index_;
 };


puremourning:
I think this could have a neat little performance benefit.  So nice!

I added some comments - I think there is a couple of other efficiencies to be had, and we can't really introduce a memory leak :)

You can use `valgrind --leak-check=yes` to check for leaks. It's a bit fiddly to do this with ycmd and our test harness, but it's probably possible to just run something like `PYTHONPATH=./run_tests.py --printpath valgrind --leak-check=yes nosetests <args>`.

> #### \[RFC\] Move member variable to function local in LetterNode
> 
> While I was refreshing my knowledge of the `cpp` code for #475 I noticed that
> 
> `letternode_per_text_index` is actually only used during one of the two
> 
> constructor (the one for the root node) as a utility data structure, so it doesn't make much sense
> 
> to keep it around longer than necessary. Since #475 is proposed I don't know if we really want to merge this, but I wanted to see what do you think, hence the `RFC` mark ;)
> 
> 
> </details>


---

Reviewed 2 of 2 files at r1.

---

*[cpp/ycm/LetterNode.cpp, line 35 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-qLmGqBWkwdkzJQWC:-KI-qLmHgYD13F-97HUx:-2048171907) ([raw file](https://github.com/valloric/ycmd/blob/045ac9eea026268bba0434c24486959a4041dcd4/cpp/ycm/LetterNode.cpp#L35)):*
> ```cpp
>     index_( -1 ) {
>   std::vector< LetterNode * > letternode_per_text_index;
>   letternode_per_text_index.resize( text.size() );
> ```

This looks a little suspect now. Either we should use the `int` constructor to `vector`, or call `reserve` rather than `resize`. Given that we always write exactly `text.size()` items to `letternode_per_text_index` (hence the name), I suspect that we can skip the initialisation overhead of `resize()` and just do a `reserve()`. 

Moreover, this fixes the -ahem- memory leak that has been introduced (by declaring it `std::vector< LetterNode >`)

---

*[cpp/ycm/LetterNode.cpp, line 39 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-rI7wFuvROLCRq-nt:-KI-rI7x0n7-2hmuv5_3:661053136) ([raw file](https://github.com/valloric/ycmd/blob/18fc8d8fbcb81cd29df53c2d753b9d3cef968d8f/cpp/ycm/LetterNode.cpp#L39)):*
> ```cpp
>     LetterNode *node = new LetterNode( letter, i );
>     letters_[ letter ].push_back( node );
>     letternode_per_text_index_[ i ] = boost::shared_ptr< LetterNode >( node );
> ```

By changing this to `LetterNode *` each element now leaks (allocated by `new LetterNode` above, but never freed). Either: change it to `vector< LetterNode >` as above and avoid the heap allocation for each `LetterNode` (yay, big performance win), or change it to a `std::vector< std::unique_ptr >`...  Wait we don't support c++ 11 right? So I guess you have to keep it as a shared_ptr or write a whole bunch of ugly c++98 code.

I would recommend `std::vector< LetterNode >` and begone `operator new`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/498#-:-KI-sirhO42eis5vOTwV:-1744698280)*

vheon:

---

*[cpp/ycm/LetterNode.cpp, line 35 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-qLmGqBWkwdkzJQWC:-KI1i_5Hfg_t6bDsyxbL:541444460) ([raw file](https://github.com/valloric/ycmd/blob/045ac9eea026268bba0434c24486959a4041dcd4/cpp/ycm/LetterNode.cpp#L35)):*
> ```cpp
>     index_( -1 ) {
>   std::vector< LetterNode * > letternode_per_text_index;
>   letternode_per_text_index.resize( text.size() );
> ```

I agree on changing either the `resize` in `reserve` or using the `int` constructor but I'm not following:

> (by declaring it std::vector< LetterNode >)

where did I declared this? I see `std::vector<LetterNode*>`

---

*[cpp/ycm/LetterNode.cpp, line 39 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-rI7wFuvROLCRq-nt:-KI1itLbiXBYs6iNrG0c:141777388) ([raw file](https://github.com/valloric/ycmd/blob/18fc8d8fbcb81cd29df53c2d753b9d3cef968d8f/cpp/ycm/LetterNode.cpp#L39)):*
> By changing this to `LetterNode *` each element now leaks (allocated by `new LetterNode` above, but never freed). Either: change it to `vector< LetterNode >` as above and avoid the heap allocation for each `LetterNode` (yay, big performance win), or change it to a `std::vector< std::unique_ptr >`...  Wait we don't support c++ 11 right? So I guess you have to keep it as a shared_ptr or write a whole bunch of ugly c++98 code.
> 
> I would recommend `std::vector< LetterNode >` and begone `operator new`.
</details>
actually the memory is owned by `letters_` which when it will go out of scope it will free all the memory, look at the destructor of `LetterNodeListMap`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/498)*

puremourning:

---

*[cpp/ycm/LetterNode.cpp, line 35 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-qLmGqBWkwdkzJQWC:-KI2--m6k89Uv0Vai7fQ:-1127923295) ([raw file](https://github.com/valloric/ycmd/blob/045ac9eea026268bba0434c24486959a4041dcd4/cpp/ycm/LetterNode.cpp#L35)):*
> I agree on changing either the `resize` in `reserve` or using the `int` constructor but I'm not following:
> 
> > (by declaring it std::vector< LetterNode >)
> 
> where did I declared this? I see `std::vector<LetterNode*>`
</details>
It ok. I missed that we also put these in `letters_` so it cannot be changed to a vector of objects rather than pointers. 

---

*[cpp/ycm/LetterNode.cpp, line 39 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-rI7wFuvROLCRq-nt:-KI1zk0e5jBpNOvaCZO2:623946836) ([raw file](https://github.com/valloric/ycmd/blob/18fc8d8fbcb81cd29df53c2d753b9d3cef968d8f/cpp/ycm/LetterNode.cpp#L39)):*
> actually the memory is owned by `letters_` which when it will go out of scope it will free all the memory, look at the destructor of `LetterNodeListMap`
</details>
Oh right. That's fine then. Sorry I missed it. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/498)*

Valloric:
Nice! This should lead to a non-trivial reduction in memory usage. Could we get a before/after benchmark? Nothing too complicated: just load a bunch of (deterministically) generated candidates into ycmd and see before/after memory usage of the process.

Would be useful to have the framework for such a benchmark checked in.

Optional though.

:lgtm: 

> 
> [Coverage Status](https://coveralls.io/builds/6221746)
> 
> Coverage remained the same at 84.957% when pulling **045ac9eea026268bba0434c24486959a4041dcd4 on vheon:move-to-local** into **18fc8d8fbcb81cd29df53c2d753b9d3cef968d8f on Valloric:master**.
> 
</details>


---

---

*[cpp/ycm/LetterNode.cpp, line 35 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-qLmGqBWkwdkzJQWC:-KI4KWNraBtRWjjjyX-o:862996334) ([raw file](https://github.com/valloric/ycmd/blob/045ac9eea026268bba0434c24486959a4041dcd4/cpp/ycm/LetterNode.cpp#L35)):*
> It ok. I missed that we also put these in `letters_` so it cannot be changed to a vector of objects rather than pointers. 
</details>
I agree that we should specify the size of the vector on construction.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/498#-:-KI4L2r-B5Co95t9eUeh:1212092652)*

vheon:
> Nice! This should lead to a non-trivial reduction in memory usage.

That is what I thought initially but the expectation wasn't matched. The absolutely non scientific test:

- `cd clang_includes/include`
- `vim *.h`
- `:argdo e!`

So basically I've visited all the files in that directory and YCM build the Candidates and monitored the memory usage from System Monitor.

> Could we get a before/after benchmark? Nothing too complicated: just load a bunch of (deterministically) generated candidates into ycmd and see before/after memory usage of the process.

I will look into it, maybe trying the valgrind approach that @puremourning suggested.



---

---

*[cpp/ycm/LetterNode.cpp, line 35 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-qLmGqBWkwdkzJQWC:-KI4lNEgnfydUjgf1-EL:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/045ac9eea026268bba0434c24486959a4041dcd4/cpp/ycm/LetterNode.cpp#L35)):*
> I agree that we should specify the size of the vector on construction.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/498#-:-KI4lPbIRgjFO2dTehnV:322357090)*

asenac:

---

*[cpp/ycm/LetterNode.cpp, line 39 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-rI7wFuvROLCRq-nt:-KI4tFUSvQ9XcXCDltHY:-1338145455) ([raw file](https://github.com/valloric/ycmd/blob/18fc8d8fbcb81cd29df53c2d753b9d3cef968d8f/cpp/ycm/LetterNode.cpp#L39)):*
> Oh right. That's fine then. Sorry I missed it. 
</details>
This change introduces a memory leak. `LetterNodeListMap`'s destructor destroys   the lists of pointers to `LetterNode` instances, but not the nodes, so memory allocated by `new LetterNode` is never freed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/498)*

vheon:

---

*[cpp/ycm/LetterNode.cpp, line 39 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-rI7wFuvROLCRq-nt:-KI51NJ0_CVoXVE2zi62:-1259471942) ([raw file](https://github.com/valloric/ycmd/blob/18fc8d8fbcb81cd29df53c2d753b9d3cef968d8f/cpp/ycm/LetterNode.cpp#L39)):*
<details><summary><i>Previously, asenac (Andrés Senac) wrote…</i></summary>
> This change introduces a memory leak. `LetterNodeListMap`'s destructor destroys   the lists of pointers to `LetterNode` instances, but not the nodes, so memory allocated by `new LetterNode` is never freed.
</details>
you're right :( tomorrow i'll see if I can fix it in a clean way. Thanks! 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/498)*

vheon:

---

*[cpp/ycm/LetterNode.cpp, line 39 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-rI7wFuvROLCRq-nt:-KI53YxUuB1UK1EHiRzk:-1526067551) ([raw file](https://github.com/valloric/ycmd/blob/18fc8d8fbcb81cd29df53c2d753b9d3cef968d8f/cpp/ycm/LetterNode.cpp#L39)):*
> you're right :( tomorrow i'll see if I can fix it in a clean way. Thanks! 
</details>
I guess that a `shared_ptr<LetterNode *>` in the  `LetterNodeListMap` would suffice, would't it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/498)*

puremourning:

---

*[cpp/ycm/LetterNode.cpp, line 39 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-rI7wFuvROLCRq-nt:-KI5JdypemTPKvNTDfRj:94798810) ([raw file](https://github.com/valloric/ycmd/blob/18fc8d8fbcb81cd29df53c2d753b9d3cef968d8f/cpp/ycm/LetterNode.cpp#L39)):*
> I guess that a `shared_ptr<LetterNode *>` in the  `LetterNodeListMap` would suffice, would't it?
</details>
So yeah, thanks for pointing that out @asenac. Turns out I was half-right. Spider sense or something ;) The point is that the ownership of the data has been moved


In c++98, `boost::shared_ptr` is probably the best off the shelf choice for this `vector` (though i would personally probably just write the loop in the destructor and not pay the reference count overhead, but that's just me :))

Side note: we really gotta get the codebase off c++98. `unique_ptr` would make this whole thing simpler.

Side note 2: @puremourning adds an action item to make memory leak checking ycmd as simple as possible and writing another README :). I'm used to working in an environment where memory error/leak checking is mandatory prior to submit for review, but the standard rule applies: if it is too hard or too slow to do, it doesn't get done.

Side note 3: Incidentally, unless i'm going bonkers, `LetterNodeListMap` constructor has some dubious code:

```c++
LetterNodeListMap::LetterNodeListMap() {
  std::fill( letters_.begin(),
             letters_.end(),
             static_cast< std::list< LetterNode * >* >( NULL ) );
}
```

Someone tell me I'm wrong, but how could this code do anything? `letters_` is a `std::vector` which is default-constructed. That would mean its `.size()` was `0` and thus [`std::fill`](http://en.cppreference.com/w/cpp/algorithm/fill) cannot do anything (as `begin() == end()`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/498)*

Valloric:

---

*[cpp/ycm/LetterNode.cpp, line 39 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-rI7wFuvROLCRq-nt:-KI5a_pblvzldb9FBwDn:553274901) ([raw file](https://github.com/valloric/ycmd/blob/18fc8d8fbcb81cd29df53c2d753b9d3cef968d8f/cpp/ycm/LetterNode.cpp#L39)):*
> ```cpp
>     LetterNode *node = new LetterNode( letter, i );
>     letters_[ letter ].push_back( node );
>     letternode_per_text_index_[ i ] = boost::shared_ptr< LetterNode >( node );
> ```

> Side note: we really gotta get the codebase off c++98. unique_ptr would make this whole thing simpler.

MSVC dragging its feet was the major reason for the hold-up. Also old distros like redhat.

Might be feasible to move at this point, though it will certainly make some people unhappy.

The  `LetterNodeListMap` ctor stuff does seem weird.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/498)*

vheon:

---

*[cpp/ycm/LetterNode.cpp, line 39 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-rI7wFuvROLCRq-nt:-KI72mwqN1OlpC6Sb5ou:187114101) ([raw file](https://github.com/valloric/ycmd/blob/18fc8d8fbcb81cd29df53c2d753b9d3cef968d8f/cpp/ycm/LetterNode.cpp#L39)):*
> ```cpp
>     LetterNode *node = new LetterNode( letter, i );
>     letters_[ letter ].push_back( node );
>     letternode_per_text_index_[ i ] = boost::shared_ptr< LetterNode >( node );
> ```

> In c++98, boost::shared_ptr is probably the best off the shelf choice for this vector (though i would personally probably just write the loop in the destructor and not pay the reference count overhead, but that's just me :))

It was my initial thought too, and I think I'll do that.

> The LetterNodeListMap ctor stuff does seem weird.

I'll try to remove the weirdness and see how it goes :+1: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/498)*

vheon:

---

*[cpp/ycm/LetterNode.cpp, line 39 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-rI7wFuvROLCRq-nt:-KI9qH5eUNsZMksNfHnl:-1734886035) ([raw file](https://github.com/valloric/ycmd/blob/18fc8d8fbcb81cd29df53c2d753b9d3cef968d8f/cpp/ycm/LetterNode.cpp#L39)):*
> ```cpp
>     LetterNode *node = new LetterNode( letter, i );
>     letters_[ letter ].push_back( node );
>     letternode_per_text_index_[ i ] = boost::shared_ptr< LetterNode >( node );
> ```

@puremourning 
> Someone tell me I'm wrong, but how could this code do anything? letters_ is a std::vector 

Actually `letters_` is not a `std::vector` but a `boost::array` so it does have a size `!= 0` and the `std::fill` actually do some work.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/498)*

puremourning:

---

*[cpp/ycm/LetterNode.cpp, line 39 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/498#-KI-rI7wFuvROLCRq-nt:-KIIMs3qxJG9B-PP_HyH:-1586404513) ([raw file](https://github.com/valloric/ycmd/blob/18fc8d8fbcb81cd29df53c2d753b9d3cef968d8f/cpp/ycm/LetterNode.cpp#L39)):*
> @puremourning 
> > Someone tell me I'm wrong, but how could this code do anything? letters_ is a std::vector 
> 
> Actually `letters_` is not a `std::vector` but a `boost::array` so it does have a size `!= 0` and the `std::fill` actually do some work.
</details>
gotya.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/498)*

dgel:
Very late to the party (and not sure if I should be commenting here or on Reviewable), but I believe this introduces some problems in memory management of the letternodes.

The letters_ map inside each letternode keeps track of the occurrences of particular characters after the current letter, storing a pointer to each. For the root node (the one that the constructor from string is called on), storing the letternode pointers in the letters_ map and having it delete them after will work, but the subsequent letternodes for the same string contain pointers to these same nodes, so deleting in the LetterNodeListMap destructor will result in double deletes.

I've noticed that letternode_per_text_index was superfluous for most LetterNodes as well in #475 and have fixed it in another way in that PR:

letternode_per_text_index is moved behind a pointer to save space where it is not used and only instantiated for nodes constructed from a string. This array contains the actual memory, the letters_ map only points into that same array.

in the PR, I've used some tricks to avoid having the nodes in the array behind yet another pointer, which is probably some of the ugly c++98 code puremourning mentioned...

Speaking of #475 , I've not had the time to benchmark it recently, and don't think I'll be able to check in a well-organised benchmarking framework, but I'll do some testing locally and share them soon. There's a very nice c++11 header-only library that gives useful stats, I'll share the code I used for testing somewhere for anyone wanting to replicate on another machine... Sorry it's taking so long!

vheon:
@dgel Yeah once I've implemented it I noticed that there were problem managing the memory. I wanted to try to have an actual `RootLetterNode` which will own the memory and hence not having these memory management problems but I've still have to write the code 😝 

Regarding #475 I'm not fully convinced because the data structure is not that easy to understand (at least for me) compared to the current one, and that is because I tried this other way.

In any case until I see the data for the memory/performance I will not say that any approach is better than another.

> Sorry it's taking so long!

Don't worry about that 😃

dgel:
@vheon Yeah I tried splitting up the Letternode into a rootnode and letternode as well. It works ok, but it means a function in candidate.cpp needs to be written in an uglier way, and it's not really worth it for the small improvement in memory usage it gave over my existing changes, so I undid it.. If you want to go that way, it would definitely work though.

Regarding the structure in #475 it's really not all that different from what's already there. The std::list has been replaced with a dedicated struct so there's no need to search through it, the 'root' node contains an array of all the letternodes and the maps contain aforementioned struct with indices instead of pointers.
The one tricky bit is that you need the rootnode to find a LetterNode by index...

Perhaps I should abstract away all internal letternodes etc, and just present one object for searching... that would make client code nicer. I'll see if I can rework it a little, but first I'd like to get some hard numbers =)

vheon:
> @vheon Yeah I tried splitting up the Letternode into a rootnode and letternode as well. It works ok, but it means a function in candidate.cpp needs to be written in an uglier way, and it's not really worth it for the small improvement in memory usage it gave over my existing changes, so I undid it.. If you want to go that way, it would definitely work though.

Actually the next step that I wanted to make was to trim the lists:

- to one node if the first one is a capital letter
- to two node, one pointing to the first occurrence of the letter and another to the first capital one

That way I would've kept the data structure almost as it is right now but gaining in memory usage due to the lesser `LetterNode` to kept stored, and very little in performance since the capital letter is at most the second of a two element list.

>Regarding the structure in #475 it's really not all that different from what's already there. The std::list has been replaced with a dedicated struct so there's no need to search through it, the 'root' node contains an array of all the letternodes and the maps contain aforementioned struct with indices instead of pointers.
The one tricky bit is that you need the rootnode to find a LetterNode by index...

Yeah, once that I've read it carefully I've understand it, but the way it is now is really easy to read being object oriented like "you're in this letter and for the letter you're looking for follow this pointer" instead of "that is the index where in the original string you have to letter you're looking for". Again, I'm not saying is a bad data structure, is probably personal taste.

> I'll see if I can rework it a little, but first I'd like to get some hard numbers =)

Absolutely, we can make it pretty once we know is the right approach.

vheon:
I guess we can safely close this one since #475 has landed 👍

PR #507, opened 2016-05-23
Pipe Omnisharp server output to ycmd logs

mispencer:
Pipe the stdout and stderr of the Omnisharp server to the ycmd logs,
instead of to separate files. This makes tailing the logs easier, and
more importantly, allows seeing the Omnisharp server output in CI
logging output.

This is a change I implemented as part of https://github.com/Valloric/ycmd/pull/213 and was crucial to that implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/507)

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -30,7 +30,8 @@ import os
 import time
 import re
 from ycmd.completers.completer import Completer
-from ycmd.utils import ForceSemanticCompletion, CodepointOffsetToByteOffset
+from ycmd.utils import ( ForceSemanticCompletion, CodepointOffsetToByteOffset,
+                         ToUnicode )
 from ycmd import responses
 from ycmd import utils
 from ycmd.completers.completer_utils import GetFileContents
@@ -39,6 +40,8 @@ import urllib.parse
 import logging
 from . import solutiondetection
 import threading
+import traceback
+from subprocess import ( PIPE )
 
 SERVER_NOT_FOUND_MSG = ( 'OmniSharp server binary not found at {0}. ' +
                          'Did you compile it? You can do so by running ' +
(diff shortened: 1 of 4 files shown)


puremourning:
Hmm I originally did this for JavaScript completer but the consensus was that the team preferred separate log files. I feel like single log file is probably easier for debugging and issue reporting, so will leave ores to comment :)

<details><summary><i>Previously, mispencer wrote…</i></summary>
> #### Pipe Omnisharp server output to ycmd logs
> 
> Pipe the stdout and stderr of the Omnisharp server to the ycmd logs,
> 
> instead of to separate files. This makes tailing the logs easier, and
> 
> more importantly, allows seeing the Omnisharp server output in CI
> 
> logging output.
> 
> 
> 
> This is a change I implemented as part of https://github.com/Valloric/ycmd/pull/213 and was crucial to that implementation.
> 
> 
> 
</details>


---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/507#-:-KIU8lraxqquJhumGbzO:-299843738)*

Valloric:
I appreciate the PR, but the last time this was discussed the consensus was to have separate logs per sub-server. It makes it easier to figure out what's going on where.

> allows seeing the Omnisharp server output in CI
logging output.

Now that _is_ useful. If we could have this in some form for Travis only...

mispencer:
Could we revisit that discussion? This is the only way to see the Omnisharp logs in Travis/Appveyor, and sometime that is very important to debugging issues.

It's also useful locally when running the tests, since it puts the Omnisharp log lines inline next to the test that failed. Currently, to see them, you have find the specific section of the specific log file in the log folder, and there likely will be multiple log files - one per solution and isolated test, to search though and try to match to a specific ycmd event.

micbou:
> This is the only way to see the Omnisharp logs in Travis/Appveyor, and sometime that is very important to debugging issues.

Another way is to write the logfiles to stdout at the end of each test.

vheon:
> Another way is to write the logfiles to stdout at the end of each test.

It would be easier than incapsulate the Logging behaviour into a component to pass to the C# completer.

Valloric:
I certainly do want to address @mispencer's issue of debuggability on Travis. It's important we make that less painful.

I think we can configure our logger framework differently for tests. If that wouldn't work, I'd even be OK with putting all the log output in one pair of files with a startup option that would be used only during testing (so not exposed in YCM).

PR #510, opened 2016-05-29
[READY] Better elisp support for identifier completer

Valloric:
Elisp is a Lisp so it supports all kinds of crazy identifiers. AFAIR
using the same regexes as Clojure should work more than well enough
because Clojure is too a Lisp descendent.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/510)


micbou:
According to [this](http://hyperpolyglot.org/lisp#id), there are some differences between clojure and emacs lisp. Should we care?

> 
> [Coverage Status](https://coveralls.io/builds/6376697)
> 
> Coverage increased (+0.008%) to 84.965% when pulling **f2696ef3b19bcb580c32ece453ce6ce0d8796e52 on elisp** into **22f229c91f7e3e6b60da521104bc0a31fc72951d on master**.
> 
</details>


---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/510#-:-KIvo0rykTrkVQgNhs1R:292386363)*

vheon:
@valloric looks like someone is trying out Emacs; let me know how it goes, I was curious but never tried :stuck_out_tongue: 

> According to [this](http://hyperpolyglot.org/lisp#id), there are some differences between clojure and emacs lisp. Should we care?
</details>


---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/510#-:-KIwO6mOEcvv5FxZtDKW:1249720819)*

Valloric:
> According to this, there are some differences between clojure and emacs lisp.

Thanks for finding that, I was interested in what the differences are. Seems like the main thing is that emacs lisp will allow unicode chars in idents (which probably nobody will use) and Clojure has some special meaning for some chars (but they're allowed in both).

I don't think we care enough to tweak it further. Using the same regex should be good enough, at least for now.

> looks like someone is trying out Emacs; 

Correct! I've never touched it before and I like trying out new things. So far it's pretty good (with Evil + Spacemacs). 



---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/510#-:-KIx936EQ6dD0ErFgewb:552594696)*

vheon:
> I don't think we care enough to tweak it further. Using the same regex should be good enough, at least for now.

I'd say that if is good enough let's go for it as it is and tweak it when the need arise. :lgtm: 

> Correct! I've never touched it before and I like trying out new things. So far it's pretty good (with Evil + Spacemacs).

I was thinking to give Spacemacs (or at least Evil) a spin, so if you're saying that is good enough I could try the dark side 😝 



---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/510#-:-KIxMKpSZFGzckT4bo2Q:1020005347)*

micbou:
:lgtm:

@homu r+

> > I don't think we care enough to tweak it further. Using the same regex should be good enough, at least for now.
> 
> I'd say that if is good enough let's go for it as it is and tweak it when the need arise. :lgtm: 
> 
> > Correct! I've never touched it before and I like trying out new things. So far it's pretty good (with Evil + Spacemacs).
> 
> I was thinking to give Spacemacs (or at least Evil) a spin, so if you're saying that is good enough I could try the dark side 😝 
</details>


---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/510#-:-KIxOMCDiyFKvOTAfP_q:187772251)*

PR #528, opened 2016-06-18
[READY] Also checking that rust_src_path exists

Valloric:
We were only checking that the path was set to something, not that there
was something there. This blows up in the user's face if they have
a common editor config across machines that gets synced but they forget
to actually place the required files in that location.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/528)

### DIFF
--- a/ycmd/completers/rust/rust_completer.py
+++ b/ycmd/completers/rust/rust_completer.py
@@ -60,8 +60,8 @@ BINARY_NOT_FOUND_MESSAGE = ( 'racerd binary not found. Did you build it? ' +
                              '"./build.py --racer-completer".' )
 ERROR_FROM_RACERD_MESSAGE = (
   'Received error from racerd while retrieving completions. You did not '
-  'set the rust_src_path option, which is probably causing this issue. '
-  'See YCM docs for details.'
+  'set the rust_src_path option (or you did, but the path doesn\'t exist), '
+  'which is probably causing this issue. See YCM docs for details.'
 )
 
 
@@ -237,7 +237,8 @@ class RustCompleter( Completer ):
     try:
       completions = self._FetchCompletions( request_data )
     except requests.HTTPError:
-      if not self._rust_source_path:
+      if not self._rust_source_path or not os.path.exists(
+          self._rust_source_path ):
         raise RuntimeError( ERROR_FROM_RACERD_MESSAGE )
       raise
 
--- a/ycmd/tests/rust/get_completions_test.py
+++ b/ycmd/tests/rust/get_completions_test.py
@@ -23,11 +23,14 @@ from future import standard_library
 standard_library.install_aliases()
 from builtins import *  # noqa
 
-from hamcrest import assert_that, empty, has_entry, has_items, contains_string
+from hamcrest import assert_that, empty, has_entry, has_items
 from nose.tools import eq_
+from mock import patch
 
+from ycmd.completers.rust.rust_completer import ERROR_FROM_RACERD_MESSAGE
 from ycmd.tests.rust import IsolatedYcmd, PathToTestFile, SharedYcmd
-from ycmd.tests.test_utils import BuildRequest, CompletionEntryMatcher
+from ycmd.tests.test_utils import ( BuildRequest, CompletionEntryMatcher,
+                                    ErrorMatcher )
 from ycmd.utils import ReadFile
 import http.client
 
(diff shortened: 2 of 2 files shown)


vheon:
Reviewed 2 of 2 files at r1.

---

*[ycmd/tests/rust/get_completions_test.py, line 76 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/528#-KK_su4EXRKW9mDQHo7x:-KK_su4EXRKW9mDQHo7y:-1710297100) ([raw file](https://github.com/valloric/ycmd/blob/f3f4a9c3a337638ff93248c19a64352fb4754fdb/ycmd/tests/rust/get_completions_test.py#L76)):*
> ```Python
>   assert_that( response,
>                has_entry( 'message',
>                           contains_string( 'rust_src_path' ) ) )
> ```

I guess we should use the `ErrorMatcher( RuntimeError, ERROR_FROM_RACERD )`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/528)*

Valloric:

---

*[ycmd/tests/rust/get_completions_test.py, line 76 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/528#-KK_su4EXRKW9mDQHo7x:-KKagkhe6FT9pcsRinCr:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/f3f4a9c3a337638ff93248c19a64352fb4754fdb/ycmd/tests/rust/get_completions_test.py#L76)):*
> I guess we should use the `ErrorMatcher( RuntimeError, ERROR_FROM_RACERD )`
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/528)*

micbou:

---

*[ycmd/tests/rust/get_completions_test.py, line 76 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/528#-KK_su4EXRKW9mDQHo7x:-KKcUNRBjzDuO-tmruY3:-648409691) ([raw file](https://github.com/valloric/ycmd/blob/f3f4a9c3a337638ff93248c19a64352fb4754fdb/ycmd/tests/rust/get_completions_test.py#L76)):*
> Done.
</details>
I think it is better to write the actual message instead of using the constant defined in the completer:
 - if the message is changed, the test should fail;
 - if the constant name is changed but not the message, the test should still pass.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/528)*

Valloric:

---

*[ycmd/tests/rust/get_completions_test.py, line 76 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/528#-KK_su4EXRKW9mDQHo7x:-KKdLMEZu5G6WgbtC6a3:331687796) ([raw file](https://github.com/valloric/ycmd/blob/f3f4a9c3a337638ff93248c19a64352fb4754fdb/ycmd/tests/rust/get_completions_test.py#L76)):*
> I think it is better to write the actual message instead of using the constant defined in the completer:
>  - if the message is changed, the test should fail;
>  - if the constant name is changed but not the message, the test should still pass.
</details>
Fair points, but I disagree. By copy-pasting the error message we now have to maintain it in two places. What we actually want the test to enforce is that "the message for racerd error is used", not that the message has these exact characters. If someone rewords the error message, the test _should not_ fail. The idea the message was trying to convey will be the same.

I agree that the constant name changing would force a change in the test, but I'd rather have flexibility in the message text than in the constant name. The text is far more likely to change.

I can change this if you feel strongly about it, but I believe this to be the better approach.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/528)*

micbou:
:lgtm:

---

---

*[ycmd/tests/rust/get_completions_test.py, line 76 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/528#-KK_su4EXRKW9mDQHo7x:-KKdP48D0g56-x4ZufA2:-767092056) ([raw file](https://github.com/valloric/ycmd/blob/f3f4a9c3a337638ff93248c19a64352fb4754fdb/ycmd/tests/rust/get_completions_test.py#L76)):*
> Fair points, but I disagree. By copy-pasting the error message we now have to maintain it in two places. What we actually want the test to enforce is that "the message for racerd error is used", not that the message has these exact characters. If someone rewords the error message, the test _should not_ fail. The idea the message was trying to convey will be the same.
> 
> I agree that the constant name changing would force a change in the test, but I'd rather have flexibility in the message text than in the constant name. The text is far more likely to change.
> 
> I can change this if you feel strongly about it, but I believe this to be the better approach.
</details>
No, it's fine. Both approaches have their merits.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/528#-:-KKd_79iD--2VEB9ZCCX:1417088721)*

puremourning:
:lgtm:

@homu r=micbou

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/528#-:-KKds2kEpAuEgPzZ9Mr_:-522648114)*

PR #530, opened 2016-06-25
[READY] Check if Rust sources path exists at completer initialization

micbou:
This PR is a proof that the changes introduced in PR #528 are not working as expected. By waiting for the racerd server to be ready at the start of the tests instead of waiting in the `IsolatedYcmd` decorator, the `GetCompletions_RustSrcPathSetButDoesntExist` test does not pass the CI runs. The reason is that when calling the `WaitUntilRacerdServerReady` function, the Rust completer is initialized and thus patching the `RUST_SRC_PATH` environment variable has no effect (the `_GetRustSrcPath` method is already called). In other words, the `GetCompletions_RustSrcPathSetButDoesntExist` test in its current state is identical to the `GetCompletions_WhenStandardLibraryCompletionFails_MentionRustSrcPath` one.

When patching `RUST_SRC_PATH` before waiting for racerd, the server will not return an HTTP error but an empty list of completions. I can confirm this by setting the `RUST_SRC_PATH` variable before running Vim, editing the [`std_completions.rs` file from the test](https://github.com/Valloric/ycmd/blob/master/ycmd/tests/rust/testdata/std_completions.rs) and trying to complete the line 5. No error message appears in the status line.

From this, I am suggesting to either revert commit e75ff91fbbeb58cf488bb189815b86def6d995b5 or check for the existence of the Rust sources path when initializing the Rust completer (we could even check if some file exists). I can make these changes in this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/530)

### DIFF
--- a/ycmd/completers/rust/rust_completer.py
+++ b/ycmd/completers/rust/rust_completer.py
@@ -55,14 +55,16 @@ RACERD_BINARY_DEBUG = p.join( DIR_OF_THIRD_PARTY, 'racerd', 'target',
 RACERD_HMAC_HEADER = 'x-racerd-hmac'
 HMAC_SECRET_LENGTH = 16
 
-BINARY_NOT_FOUND_MESSAGE = ( 'racerd binary not found. Did you build it? ' +
-                             'You can do so by running ' +
-                             '"./build.py --racer-completer".' )
+BINARY_NOT_FOUND_MESSAGE = (
+  'racerd binary not found. Did you build it? '
+  'You can do so by running "./build.py --racer-completer".' )
+NON_EXISTING_RUST_SOURCES_PATH_MESSAGE = (
+  'Rust sources path does not exist. Check the value of the rust_src_path '
+  'option or the RUST_SRC_PATH environment variable.' )
 ERROR_FROM_RACERD_MESSAGE = (
   'Received error from racerd while retrieving completions. You did not '
-  'set the rust_src_path option (or you did, but the path doesn\'t exist), '
-  'which is probably causing this issue. See YCM docs for details.'
-)
+  'set the rust_src_path option, which is probably causing this issue. '
+  'See YCM docs for details.' )
 
 
 def FindRacerdBinary( user_options ):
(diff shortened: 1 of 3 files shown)


Valloric:
I'd rather we come up with a different solution than revert the commit, though reverting it would be better than leaving it as it currently does nothing (sadly).

micbou:
I went for the second solution: check that Rust sources path exists at completer initialization. Two tests were added: one when setting the `rust_src_path` option and another when using the `RUST_SRC_PATH` environment variable.

---

Reviewed 1 of 2 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/530#-:-KLB3K5610fuogEEAqYQ:2051726275)*

puremourning:
:lgtm:

---

Reviewed 1 of 2 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/530#-:-KLCSqyuXLPXinPH3N8_:1417088721)*

vheon:
Wouldn't be enough for the tests to swap the `@IsolatedYcmd` and `@patch.dict( ... )` lines instead of manually wait for the server to be up?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/530#-:-KLCZ3jtGHvkVEG6CX0_:1318997729)*

micbou:
> Wouldn't be enough for the tests to swap the @IsolatedYcmd and @patch.dict( ... ) lines instead of manually wait for the server to be up?

This would be enough for the `GetCompletions_NonExistingRustSrcPathFromEnvironmentVariable` test but not for the `GetCompletions_NonExistingRustSrcPathFromUserOption` one because of the `UserOption` context manager.

In addition, not waiting for the Rust completer in the `IsolatedYcmd` decorator is needed for the `DebugInfo` tests I am planning to add.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/530#-:-KLCc3oAnfDMFobGHLHf:-562002964)*

vheon:
I don't like the fact that we should explicitely wait in some test but not in others but I don't see an easy solution to that, except maybe creating another decorator that tell you explicitely that? like `@IsolatedYcmdNowait`. Do you think is worth it?

---

Reviewed 1 of 2 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/530#-:-KLD-g-fOSDcyRgL2YgA:633190595)*

micbou:
No, it will make things more complicated.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/530#-:-KLDKAkXioog9TB0-fgq:-1762830252)*

Valloric:
This is good stuff!

@homu r+

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/530#-:-KLDSRvVlKuPuVoGY9Eo:-1699326207)*

PR #532, opened 2016-06-26
[READY] Provide FixIt suggestions on notes attached to diagnostics

puremourning:
# Background

Certain diagnostic reports include "child" diagnostics. These are typically notes which either indicate the source of the error, point to an important definition, or in some cases include suggestions for fixing the issue.

A good example is the following snippet:

```c++
  bool x = true;
  if ( x = false ) {
    // do something
  }
```

clang reports a warning about the use of the assignment in the `if` expression, and includes 2 child diagnostics suggesting either to wrap the assignment in `( ... )` or to change to a comparison (`==`).

Incidentally, this is the approach that will be taken for incorporation of the `#include` fixer to clang. This tool adds a suggestion to typo correction to include the file which defines the potentially typo'd identifier.

# Implementation notes

`Diagnostic` objects now hold a 1-d `vector` of `FixIt` objects. This is actually more intuitive than the previous implementation where `FixIt` objects were only created for the purpose of populating replies.

To prevent scanning child diagnostics multiple times, the functionality of generating the "extended" diagnostic text and creating child `FixIt` objects have been combined.

You'll note that while the libclang API is recursive (child diagnostics can have child diagnostics), we generate a flat `vector` of fixits (in the same way that we generate a flat string for the text). This is primarily because our `FixIt` API is one dimensional and most user interfaces would display the options linearly anyway.

# API changes

The API has always supported reporting multiple FixIts for a given context.

In order to present the user with some meaningful options, we include the diagnostic text in the FixIt response.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/532)


Valloric:
:lgtm:

Thanks for this! Looks great with one minor comment.

---

Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/ycm/ClangCompleter/ClangHelpers.cpp, line 79 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/532#-KLDT17JhCLvcMDWg0Ju:-KLDT17Kuv2dPLRMn2aJ:-229624104) ([raw file](https://github.com/valloric/ycmd/blob/115b5fd7259c5753288c83d3b0b6654fd1e498d5/cpp/ycm/ClangCompleter/ClangHelpers.cpp#L79)):*
> ```cpp
> 
> // Warning: This method is re-entrant (recursive).
> std::string HandleChildDiags( std::vector< FixIt >& fixits,
> ```

The function name is somewhat vague. What does "handle" mean? I see that it returns a string, but what's in the string? It's not obvious from the function name what this returns.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/532#-:-KLDTvoDzA2pPrk7xRIl:869888917)*

puremourning:
Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/ycm/ClangCompleter/ClangHelpers.cpp, line 79 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/532#-KLDT17JhCLvcMDWg0Ju:-KLDnpF0TRKxP_TI2tIh:2008073286) ([raw file](https://github.com/valloric/ycmd/blob/115b5fd7259c5753288c83d3b0b6654fd1e498d5/cpp/ycm/ClangCompleter/ClangHelpers.cpp#L79)):*
> The function name is somewhat vague. What does "handle" mean? I see that it returns a string, but what's in the string? It's not obvious from the function name what this returns.
</details>
Done.

I gave it a more verbose name, changed it to take the string as an argument, and updated the method header comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/532)*

Valloric:
Reviewed 7 of 8 files at r1, 1 of 1 files at r2.

---

*[cpp/ycm/ClangCompleter/ClangHelpers.cpp, line 79 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/532#-KLDT17JhCLvcMDWg0Ju:-KLEhNUXNzkzH4aU2cZX:418723997) ([raw file](https://github.com/valloric/ycmd/blob/115b5fd7259c5753288c83d3b0b6654fd1e498d5/cpp/ycm/ClangCompleter/ClangHelpers.cpp#L79)):*
> Done.
> 
> I gave it a more verbose name, changed it to take the string as an argument, and updated the method header comment.
</details>
Much nicer, thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/532)*

vheon:
:lgtm:

---

Reviewed 7 of 8 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/532#-:-KMBLYEGMse-zHinZOe0:1417088721)*

micbou:
I tried with PR https://github.com/Valloric/YouCompleteMe/pull/2223 and it works great (at least with the example you provided).

Should we bump the core version? Users will need to recompile to use this feature.

---

Reviewed 8 of 8 files at r1, 1 of 1 files at r2.

---

*[ycmd/tests/clang/subcommands_test.py, line 763 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/532#-KLCzVPH7PGBVw-jSvsQ:-KLCzVPIzZu9M1nmVMon:1961745478) ([raw file](https://github.com/valloric/ycmd/blob/115b5fd7259c5753288c83d3b0b6654fd1e498d5/ycmd/tests/clang/subcommands_test.py#L763)):*
> ```Python
>     [ 21, 16, 'cpp11', ufile, FixIt_Check_unicode_Ins ],
> 
>     # FixIt attached to a "child" diagnistic (i.e. a Note)
> ```

`diagnostic` instead of `diagnistic`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/532#-:-KLJG_BWquPgDT7Sj51U:1309662771)*

puremourning:
Good point. Done.

---

Review status: 6 of 9 files reviewed at latest revision, 2 unresolved discussions.

---

*[ycmd/tests/clang/subcommands_test.py, line 763 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/532#-KLCzVPH7PGBVw-jSvsQ:-KMKvzvDESM4Xc6X59Q_:-499146009) ([raw file](https://github.com/valloric/ycmd/blob/115b5fd7259c5753288c83d3b0b6654fd1e498d5/ycmd/tests/clang/subcommands_test.py#L763)):*
> `diagnostic` instead of `diagnistic`.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/532#-:-KMKvdu7CqEvVY5dgBVg:-1622819128)*

vheon:
Still :lgtm: 

---

Reviewed 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/532#-:-KMLeYkc_JFMlDJKU8Br:-805494633)*

micbou:
:lgtm:

@homu r+

---

Reviewed 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/532#-:-KMLmm2u20IYUxhuUWQf:187772251)*

PR #533, opened 2016-06-27
[READY] Uniformize debug informations

micbou:
This PR makes the string returned by the `DebugInfo` method more consistent throughout completers. It also adds more informations:
 - status of the `.ycm_extra_conf.py` file for the C-family completer;
 - subserver PID when relevant (needed for the tests in PR #282);
 - executable(s) used by the completer;
 - current solution for the C♯ completer;
 - Rust sources path for the Rust completer.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/533)

### DIFF
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -36,6 +36,7 @@ from ycmd.completers.completer import Completer
 from ycmd.completers.completer_utils import GetIncludeStatementValue
 from ycmd.completers.cpp.flags import Flags, PrepareFlagsForClang
 from ycmd.completers.cpp.ephemeral_values_set import EphemeralValuesSet
+from ycmd.responses import NoExtraConfDetected, UnknownExtraConf
 
 import xml.etree.ElementTree
 
@@ -366,13 +367,24 @@ class ClangCompleter( Completer ):
 
   def DebugInfo( self, request_data ):
     filename = request_data[ 'filepath' ]
-    if not filename:
-      return ''
-    flags = self._FlagsForRequest( request_data ) or []
-    source = extra_conf_store.ModuleFileForSourceFile( filename )
-    return 'Flags for {0} loaded from {1}:\n{2}'.format( filename,
-                                                         source,
-                                                         list( flags ) )
+    try:
+      extra_conf = extra_conf_store.ModuleFileForSourceFile( filename )
+      flags = self._FlagsForRequest( request_data ) or []
+    except NoExtraConfDetected:
+      return ( 'C-family completer debug information:\n'
+               '  No configuration file found' )
+    except UnknownExtraConf as error:
+      return ( 'C-family completer debug information:\n'
+               '  Configuration file found but not loaded\n'
+               '  Configuration path: {0}'.format(
+                 error.extra_conf_file ) )
+    if not extra_conf:
+      return ( 'C-family completer debug information:\n'
+               '  No configuration file found' )
+    return ( 'C-family completer debug information:\n'
+             '  Configuration file found and loaded\n'
+             '  Configuration path: {0}\n'
+             '  Flags: {1}'.format( extra_conf, list( flags ) ) )
 
 
   def _FlagsForRequest( self, request_data ):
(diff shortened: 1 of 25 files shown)


vheon:
I know you're probably against this and I don't know if is worth it since later I guess you will change those to be json object instead of strings but right now we have something like:

```python
'TypeScript completer debug information:\n'
'  TSServer running\n'
'  TSServer process ID: {0}\n'
'  TSServer executable: {1}\n'
'  TSServer logfile: {2}'
```

in the implementation and

```python
'TypeScript completer debug information:\n'
'  TSServer running\n'
'  TSServer process ID: \d+\n'
'  TSServer executable: .+\n'
'  TSServer logfile: .+'
```

in the tests. Wouldn't be better to extract the one in the implementation into a variable and use it in the implementation as it is right now, and in the tests formatting the `{0}, {1}, ...` with the regex needed by that field? In the end, do we want to test that that kind of template is showed to the user or that that specific string is displayed? Personally I think the first is more valuable. What do you think?

---

Reviewed 3 of 26 files at r1.
Review status: 3 of 26 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/533#-:-KLHvf-p4qOzRPmblrr6:-14947601)*

Valloric:
This looks good to me, but I do share @vheon's concern that we seem to be returning structured data in a string. Our tests also enforce that it has this particular structure.

So I'm :lgtm: on this, but worth considering other options too. It's not easy to tell what would be the best approach here.

---

Reviewed 26 of 26 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/533#-:-KLIyPAvZu_mveS0lNyM:662511397)*

micbou:
@vheon This is the same matter as in PR #528: should we check that such constant is used or such string is returned? To be honest, I don't have a definite answer. In this case, I went for the easiest approach.

> This looks good to me, but I do share @vheon's concern that we seem to be returning structured data in a string. Our tests also enforce that it has this particular structure.

The plan is to return a dictionary instead of a string but this needs some thoughts on how to structure such dictionary and this will require changes in the clients (at least the Vim client), thus delaying even further PR #282. So I prefer to uniformize the string for now and return a dictionary later. It should not be too difficult to adapt the tests.

---

Reviewed 26 of 26 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/533#-:-KLJ6cURUb1NGoYfpsmb:-1109814844)*

Valloric:
OK, if this is temporary and we want to move to returning a dictionary long-term, then I'm satisfied. Let's unblock the work that is currently block and come back to this later.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/533#-:-KLJK9CPcF-t_5AxWxAy:-1346631485)*

vheon:
@micbou 
>This is the same matter as in PR #528:

That is why I said

> I know you're probably against this

:stuck_out_tongue_winking_eye: 

Anyway it was just to avoid the duplication, that's all.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/533#-:-KLMOYIT2ElfhIJ7FsK2:668358085)*

vheon:
:lgtm: 

---

Reviewed 23 of 26 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/533#-:-KLQeBkDabmq-kL96JrY:1267966159)*

vheon:
I guess we can call `homu` for this one right?

puremourning:
:lgtm:

---

Reviewed 26 of 26 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/533#-:-KLlpYmI5NziGIx5BnOt:1417088721)*

Valloric:
I see many LGTMs here. @homu r+

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/533#-:-KLlpgdu5ihs-cNy4urs:-1308769019)*

PR #535, opened 2016-06-29
Search for completers binaries in system path

z33ky:
The Rust and Typescript completers are already searching the system path if a binary is not installed alongside YCM.
This commit adds the same behavior to the C#, Go and Javascript completers.

This means one can build ycmd without bundling the specific completer engines, but being able to use them when installing them on the system.
I want to do this for an ArchLinux AUR package, so that a user only gets a YCM "core" installation and can install completer engines on a per-need basis. I imagine this would be favorable for other distribution's packaging as well.

I tested this for Go and Javascript (the existing completers for Rust and Typescript work like this already), but I did not test C#.
I am also unsure if any changes have to be made to work well on non-GNU/Linux systems.

Python completion is not adapted, because JediHTTP does currently not provide a facility to be installed on its own: vheon/JediHTTP#17.

For the C family this also seems to be inapplicable, since support for clang has to be built into the ycmd library.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/535)

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -51,6 +51,18 @@ PATH_TO_OMNISHARP_BINARY = os.path.join(
   'OmniSharp', 'bin', 'Release', 'OmniSharp.exe' )
 
 
+def FindOmniSharpBinary():
+  """Find the path to the OmniSharp binary.
+
+  If a version is installed with YCM, use that.
+  Otherwise try to find one installed in the system.
+  If that fails, return None."""
+
+  if os.path.exists( PATH_TO_OMNISHARP_BINARY ):
+    return PATH_TO_OMNISHARP_BINARY
+
+  return utils.PathToFirstExistingExecutable( [ 'OmniSharp', 'OmniSharp.exe' ] )
+
 class CsharpCompleter( Completer ):
   """
   A Completer that uses the Omnisharp server as completion engine.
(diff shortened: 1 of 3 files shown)


vheon:
As I mentioned in https://github.com/vheon/JediHTTP/issues/17 I don't like this because this would mean that uses would be capable to run versions of a completer that are outdated or that were not tested by our test suite.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/535#-:-KLR_0gcbOF7Er5QnbB-:-2089997085)*

puremourning:
I agree with Andrea.

Valloric:
Agreed with @puremourning and @vheon.

We appreciate the PR, but this is not the direction we want to go. We'd prefer to have every completer use our binaries, not the system ones. We go with the PATH search approach only when we must.

zsrkmyn:
Using your binaries means that users should have two or more copies of these libraries or completers, espcially libclang.so and many other third-party python libs, in my case. The purpose of shared libary is to decrese the duplicated codes and binaries in the one's system, as its name 'shared' suggests. If every software use their own binaris, how many duplicated things would be there in a one's computer? Think about the C library in Windows platform. I think, at least you should provide a way for user to choose which binaris (yours or system's) they use. If they have trouble in using system's binaries, then they can choose yours.

PR #537, opened 2016-07-05
Link to the correct Boost.Python variant if the system Boost is used

yan12125:
In typical Boost builds, the Python 3 variant of Boost.Python is built and installed as ```libboost_python3.so```. [1] For example Arch Linux's boost package. [2] Before, if ycmd is compiled with ```-DUSE_SYSTEM_BOOST=ON -DUSE_PYTHON2=OFF```, an dynamic linking error due to incompatible ABI between ```libboost_python.so``` and ```libpython3.x.so```. This patch fixes the problem.

Testing: tested on my Arch Linux, ycmd runs fine with both ```-DUSE_SYSTEM_BOOST=ON -DUSE_PYTHON2=OFF``` and ```-DUSE_SYSTEM_BOOST=ON -DUSE_PYTHON2=ON```.

[1] Jamfile of Boost.Python: http://svn.boost.org/svn/boost/trunk/libs/python/build/Jamfile.v2
[2] https://www.archlinux.org/packages/extra/x86_64/boost/files/

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/537)


puremourning:
How "typical" is this? I mean are there any existing systems which this change would break?
How can we confirm?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/537#-:-KMKx8p4sS_ZWeXfJmNC:-1831652024)*

yan12125:
According to the Jamfile of boost, the built library is always named as libboost_python3.so, no matter the default Python version is 2 or 3.

For distributions:
Fedora also have libboost_python3.so: http://koji.fedoraproject.org/koji/rpminfo?rpmID=7810915
Debian, unfortunately, has its own patches, so the file is named as libboost_python-py34.so: https://packages.debian.org/jessie/amd64/libboost-python1.55-dev/filelist. Using the system boost on Debian requires more tricks. Here's an similar report to this one: https://github.com/belltailjp/selective_search_py/issues/6.
I can't find an info page for OpenSUSE's boost packages.

After all, I can't find a distribution that names the Python 3 version of boost-python as libboost_python.so, the this change wouldn't break existing usages.

All conclusions above are extracted from information pages of individual distributions. I didn't test these distributions one by one.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/537#-:-KML1ODdRfOoZFAO2HJc:390876902)*

vheon:
> For distributions:
> Fedora also have libboost_python3.so: http://koji.fedoraproject.org/koji/rpminfo?rpmID=7810915
> Debian, unfortunately, has its own patches, so the file is named as libboost_python-py34.so: https://packages.debian.org/jessie/amd64/libboost-python1.55-dev/filelist. Using the system boost on Debian requires more tricks. Here's an similar report to this one: belltailjp/selective_search_py#6.
> I can't find an info page for OpenSUSE's boost packages.

So this means that currently users cannot build ycmd with system boost, right?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/537#-:-KMLfUUfk6TnEmT2txJr:123018167)*

micbou:
> So this means that currently users cannot build ycmd with system boost, right?

Not with Python 3.

For Debian-like distributions, you just need to create a symlink. For instance on Ubuntu 16.04:
```sh
cd /usr/lib/x86_64-linux-gnu
sudo ln -s libboost_python-py35.so libboost_python3.so
```

Anyway, since we don't support Boost system, it is not an issue if it doesn't work out of the box on some distributions (which use a non standard naming convention) so I am fine with the change.

:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/537#-:-KMLn-qfvk6dsZqfjX11:2018639243)*

yan12125:
Thanks @micbou. Yes the current configuration is broken for the system Boost + Python 3.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/537#-:-KMNII4hrSQs5YRa29HQ:326287185)*

vheon:
:lgtm: @homu r=micbou

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/537#-:-KMR2euTyDV0MPQyC6IB:-1885764372)*

yan12125:
Thanks!

PR #542, opened 2016-07-12
[READY] Remove unloaded_buffer parameter

### LINKED ISSUE #541: Why is clang_completer requesting `unloaded_buffer` key in OnBufferUnload
Would it be possible to just reuse `filepath` in `OnBufferUnload` in `clang_completer` instead of having to specify `unloaded_buffer`. The typescript completer is already using `filepath`. It would be nice to align both.

micbou:
Closes #541.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/542)

### DIFF
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -338,7 +338,7 @@ class ClangCompleter( Completer ):
 
   def OnBufferUnload( self, request_data ):
     self._completer.DeleteCachesForFile(
-        ToCppStringCompatible( request_data[ 'unloaded_buffer' ] ) )
+        ToCppStringCompatible( request_data[ 'filepath' ] ) )
 
 
   def GetDetailedDiagnostic( self, request_data ):
--- a/ycmd/tests/typescript/event_notification_test.py
+++ b/ycmd/tests/typescript/event_notification_test.py
@@ -1,0 +1,116 @@
+# Copyright (C) 2016 ycmd contributors
+#
+# This file is part of ycmd.
+#
+# ycmd is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ycmd is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from future import standard_library
+standard_library.install_aliases()
+from builtins import *  # noqa
+
+from hamcrest import assert_that, contains, has_entries
+
+from ycmd.tests.typescript import IsolatedYcmd, PathToTestFile
+from ycmd.tests.test_utils import ( BuildRequest, ClearCompletionsCache,
+                                    CompletionEntryMatcher )
+from ycmd.utils import ReadFile
+
+
+@IsolatedYcmd
+def EventNotification_OnBufferUnload_CloseFile_test( app ):
+  # Open main.ts file in a buffer.
+  main_filepath = PathToTestFile( 'buffer_unload', 'main.ts' )
+  main_contents = ReadFile( main_filepath )
+
+  event_data = BuildRequest( filepath = main_filepath,
+                             filetype = 'typescript',
+                             contents = main_contents,
+                             event_name = 'BufferVisit' )
+  app.post_json( '/event_notification', event_data )
+
+  # Complete in main.ts buffer an object defined in imported.ts.
+  completion_data = BuildRequest( filepath = main_filepath,
+                                  filetype = 'typescript',
+                                  contents = main_contents,
+                                  line_num = 3,
+                                  column_num = 10 )
+  response = app.post_json( '/completions', completion_data )
+  assert_that( response.json, has_entries( {
+    'completions': contains( CompletionEntryMatcher( 'method' ) ) } ) )
+  # FIXME: we should not have to clear the cache.
+  ClearCompletionsCache()
+
+  # Open imported.ts file in another buffer.
+  imported_filepath = PathToTestFile( 'buffer_unload', 'imported.ts' )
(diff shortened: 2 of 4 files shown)


puremourning:
That was my thought when i saw the ycmd issue :)

:lgtm:

Don't suppose there's any way we can make a simple test for this? I guess the only real way would be to mock out `_SendCommand` and check that it gets the right file name. Not a big deal for this PR if not.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/542#-:-KMVlpRarQRT_QUzA9fq:-1188401439)*

Valloric:
:lgtm: although like @puremourning, I'd like to see some sort of test. Not necessarily in this PR.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/542#-:-KMWTFVi8m8Mqw6Lelld:-436278223)*

micbou:
I'd also like to add a test for this. I think the best way would be to find an example that fails without this change where we open two files in TSServer, close one of them, and ask for completions in the other. I tried to find such example but was not successful. At this point, I even wonder if closing a file in TSServer is doing anything.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/542#-:-KMdnrggz9mxQHA4_WJP:486405762)*

micbou:
I gave some thoughts to the `unloaded_buffer` parameter and I think we should drop it. From my understanding, it exists because the current buffer is not necessarily the one being unloaded. One could argue that in this case the client should send a request about the buffer being unloaded and not the current buffer. But that's not the main reason to drop the parameter. The main reason is that if the filetype of the current buffer is different from the unloaded buffer one then the wrong completer will be called or no completer at all if the filetype is not supported. For instance, if we are editing a TypeScript file while unloading a C++ one, we will tell TSServer to close the C++ file, which is completely wrong.

So, what do you think? Should we drop the parameter in this PR or merge it in current state and open a new PR for that? 

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 3 of 3 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/542#-:-KQp_p0mQo1n1mSXGA2b:b-6ryl1t)*

puremourning:
That's certainly a good point. I wonder what the other client maintainers think...

Does the Vim client even send this parameter as anything other than the current file?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/542#-:-KQq6KgFGlQxHBQWgZXZ:b-lg5xyu)*

ptrv:
In Emacs we send the current file in the `BufferUnload` request

micbou:
> Does the Vim client even send this parameter as anything other than the current file?

YCM send [this parameter as the unloaded buffer path](https://github.com/Valloric/YouCompleteMe/blob/master/python/ycm/youcompleteme.py#L315).

> In Emacs we send the current file in the BufferUnload request

Would it be an issue to send the unloaded/closed buffer instead in the Emacs client?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/542#-:-KQqot8PX0FOggR3Hptb:bson3m8)*

ptrv:
> Would it be an issue to send the unloaded/closed buffer instead in the Emacs client?
That is the case. The current file is the one which is closed. So basically in Emacs the file for `unloaded_buffer` and `filepath` in the request data is the same.

micbou:
Thanks. I've looked at [the Atom client](https://github.com/Qusic/atom-youcompleteme) and it seems to do the same as the Emacs client. [The Sublime client](https://github.com/LuckyGeck/YcmdCompletion) is not implementing the `UnloadBuffer` event notification. @Qusic and @LuckyGeck may confirm this.

In conclusion, only the Vim client is affected by the issue I mentioned above (shame on us). So, we should first fix the bug in YCM by sending the request as the unloaded buffer instead of the current one for the `BufferUnload` event notification and then drop the `unloaded_buffer` parameter in ycmd.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/542#-:-KQudg2YiDKiidD_q6Zg:b-sdi3m3)*

Qusic:
> I've looked at the Atom client and it seems to do the same as the Emacs client.

Yes. So Atom client won't be affected.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/542#-:-KQxJi7uL8zzOlB04uJZ:b3oi1vy)*

micbou:
I updated the PR by completely removing the `unloaded_buffer` parameter. I also added a test covering the `BufferUnload` event of the TypeScript completer. This test illustrates the limits of our completions caching system and a deficiency in how the TypeScript completer deal with dirty buffers (see the `FIXME` annotations).

PR is now ready.

---

Reviewed 1 of 5 files at r4, 3 of 3 files at r5, 3 of 3 files at r7, 1 of 1 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/542#-:-KUsa6Zcue4480GVAIei:b-z32pv5)*

Valloric:
:lgtm: as before.

---

Review status: 3 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/542#-:-KQurWZhU6a9qF5nvxd3:b-hsusog)*

vheon:
I am a little puzzled by the title of the PR and the code of the PR so, we were already closing the right buffer in the Typescript completer?

@micbou  the code :lgtm: so if is only a matter of an "old" title feel free to fire `homu` yourself.

---

Review status: 3 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/542#-:-KUxefXoTqsaSYq48UrL:b-eum4tq)*

vheon:
Reviewed 1 of 5 files at r4, 2 of 3 files at r7, 1 of 1 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/542)*

micbou:
I forgot to update the title of the PR. Sorry for the confusion.

@homu r=vheon

---

Reviewed 1 of 1 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/542#-:-KUxh5FE4Xxr7EdboaBv:bjb7epv)*

vheon:
@homu retry

vheon:
@homu r+

micbou:
There is an issue with the TypeScript tests on Linux Travis. I'm on it.

micbou:
@homu retry

PR #545, opened 2016-07-13
RFC: Suppress value error when cpp flags not ready

stevenschimmel:
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/545)

### DIFF
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -317,6 +317,9 @@ class ClangCompleter( Completer ):
       raise ValueError( INVALID_FILE_MESSAGE )
 
     flags = self._FlagsForRequest( request_data )
+    # don't raise an error when flags are not ready yet
+    if flags == False:
+      return []
     if not flags:
       raise ValueError( NO_COMPILE_FLAGS_MESSAGE )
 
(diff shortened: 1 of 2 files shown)


puremourning:
Thanks for sending a PR!

Could you briefly explain the change and the use case or problem it is solving? We generally want users to know when there are no flags so that they know _why_ we aren't proving suggestions...

Valloric:
We appreciate the PR, but we explicitly want to notify users when flags aren't ready.

stevenschimmel:
There are a few reasons for the change. 1) The events in the editor that eventually trigger the ValueError in OnFileReadyToParse are not clearly a direct result of the user interacting with YCM. So when the error triggers, it is not immediately obvious to the user why it is showing up. 2) In some (large) build systems, it may take a while before flags are available for a file, leaving a long period in which the error can trigger. 3) In VIM, the message that is generated in response to the ValueError can be pretty disruptive, especially when it collides with other VIM messages generated around the same event, causing a "Press enter or type command to continue" prompt.

This change tries to strike a balance by suppressing the ValueError in OnFileReadyToParse (when the user didn't directly interact with YCM), but not in places where it is the result of the user interacting with YCM, for example in the completer subcommands.

A more elegant way of informing users if flags are available might be to implement a YcmStatus command, which returns a short string that describes the state of YCM for a given file. Users could add that status string to their editor's status bar.

micbou:
All of this is specific to the Vim client so it should be done there.

Valloric:
@stevenschimmel It might be worth thinking about a better way of informing the user of this error, but like @micbou said, the problem you're complaining is entirely a client issue (in your case, YCM) and should be handled there. The server (ycmd) should continue notifying about the error.

stevenschimmel:
Ok, that sounds fair enough. I'll think about a client-side solution. Thank you all for considering the PR.

PR #548, opened 2016-07-16
[READY] Always enable spell checking for better FixIts

### LINKED ISSUE #546: ycmd does not pass on many fixits
I'm finding that ycmd occasionally suggests fixits for things like missing semi-colons, but for many other things it does not suggest a fixit even though clang does. For instance, if I write `std::strng x;`, when compiling in clang it offers the correction to `string`, but ycmd does not offer this fixit. Is this a bug with ycmd, something wrong with my setup, or is the fixit feature constrained somehow?

puremourning:
Fixes #546 

We enable spell-checking to get better FixIt suggestions. The only time we don't is if the user's flags explicitly request it is disabled.

The change is in `PrepareFlagsForClang` because that is run both when getting the flags from the `FlagsForFile` function and when supplying them in the request. However, rather than having to update every test which does this to also expect `-fspell-checking` in the result, we use the same approach as `FlagsForFile` to allow the test system to disable adding it.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/548)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -99,7 +99,10 @@ class Flags( object ):
 
       if add_extra_clang_flags:
         flags += self.extra_clang_flags
-      sanitized_flags = PrepareFlagsForClang( flags, filename )
+
+      sanitized_flags = PrepareFlagsForClang( flags,
+                                              filename,
+                                              add_extra_clang_flags )
 
       if results[ 'do_cache' ]:
         self.flags_for_file[ filename ] = sanitized_flags
@@ -172,10 +175,12 @@ def _CallExtraConfFlagsForFile( module, filename, client_data ):
     return module.FlagsForFile( filename )
 
 
-def PrepareFlagsForClang( flags, filename ):
+def PrepareFlagsForClang( flags, filename, add_extra_clang_flags = True ):
   flags = _CompilerToLanguageFlag( flags )
   flags = _RemoveXclangFlags( flags )
   flags = _RemoveUnusedFlags( flags, filename )
+  if add_extra_clang_flags:
+    flags = _EnableTypoCorrection( flags )
   flags = _SanitizeFlags( flags )
   return flags
 
@@ -398,6 +403,24 @@ def _ExtraClangFlags():
   # for a similar issue.
   if OnWindows():
     flags.append( '-fno-delayed-template-parsing' )
+  return flags
+
+
+def _EnableTypoCorrection( flags ):
+  """Adds the -fspell-checking flag if the -fno-spell-checking flag is not
+  present"""
+
+  # "Typo correction" (aka spell checking) in clang allows it to produce
+  # hints (in the form of fix-its) in the case of certain diagnostics. A common
+  # example is "no type named 'strng' in namespace 'std'; Did you mean
+  # 'string'? (FixIt)". This is enabled by default in the clang driver (i.e. the
+  # 'clang' binary), but is not when using libclang (as we do). It's a useful
+  # enough feature that we just always turn it on unless the user explicitly
+  # turned it off in their flags (with -fno-spell-checking).
+  if '-fno-spell-checking' in flags:
+    return flags
+
+  flags.append( '-fspell-checking' )
   return flags
 
 
--- a/ycmd/tests/clang/flags_test.py
+++ b/ycmd/tests/clang/flags_test.py
@@ -185,6 +185,25 @@ def RemoveUnusedFlags_Depfiles_test():
                contains( *expected ) )
 
 
+def EnableTypoCorrection_Empty_test():
+  eq_( flags._EnableTypoCorrection( [] ), [ '-fspell-checking' ] )
+
+
+def EnableTypoCorrection_Trivial_test():
+  eq_( flags._EnableTypoCorrection( [ '-x', 'c++' ] ),
+                                    [ '-x', 'c++', '-fspell-checking' ] )
+
+
+def EnableTypoCorrection_Reciprocal_test():
+  eq_( flags._EnableTypoCorrection( [ '-fno-spell-checking' ] ),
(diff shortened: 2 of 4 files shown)


micbou:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---

*[ycmd/completers/cpp/flags.py, line 409 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/548#-KMptKhYDncJtyZ74fnh:-KMptKhZkzzujRiI_dqF:307176134) ([raw file](https://github.com/valloric/ycmd/blob/f7a31e9475f9c9056c14345164f09c431ef51e88/ycmd/completers/cpp/flags.py#L409)):*
> ```Python
>   # a feature useful enough that we just always turn it on. That is, unless
>   # the user explicitly turned it off (with -fno-spell-checking).
>   if '-fno-spell-checking' in flags:
> ```

We could check for `-fspell-checking` too so that we don't add it a second time.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/548#-:-KMpw1ub4LOUXm4xA_V8:1417088721)*

puremourning:

---

*[ycmd/completers/cpp/flags.py, line 409 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/548#-KMptKhYDncJtyZ74fnh:-KMpwmDemAndazgETo46:-345752202) ([raw file](https://github.com/valloric/ycmd/blob/f7a31e9475f9c9056c14345164f09c431ef51e88/ycmd/completers/cpp/flags.py#L409)):*
> We could check for `-fspell-checking` too so that we don't add it a second time.
</details>
true, i suppose, but i think adding it 2x is harmless.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/548)*

Valloric:
:lgtm: 

Thanks!

---

---

*[ycmd/completers/cpp/flags.py, line 406 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/548#-KMqUGU__PAfjt0mzMCV:-KMqUGU__PAfjt0mzMCW:1535222447) ([raw file](https://github.com/valloric/ycmd/blob/f7a31e9475f9c9056c14345164f09c431ef51e88/ycmd/completers/cpp/flags.py#L406)):*
> ```Python
>   # hints (in the form of fix-its) in the case of certain diagnostics. A common
>   # example is "no type named 'strng' in namespace 'std'; Did you mean
>   # 'string'? (FixIt)". This is enabled by default in the clang driver and is
> ```

It might be a bit confusing that you say that it's enabled by default and yet we add it ourselves. The difference is between the clang driver and calling libclang, but that might not be obvious to the reader.

I'd suggest clarifying this a bit to avoid confusion.

---

*[ycmd/completers/cpp/flags.py, line 409 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/548#-KMptKhYDncJtyZ74fnh:-KMqUVAqOiRLY-ml073H:328195815) ([raw file](https://github.com/valloric/ycmd/blob/f7a31e9475f9c9056c14345164f09c431ef51e88/ycmd/completers/cpp/flags.py#L409)):*
> true, i suppose, but i think adding it 2x is harmless.
</details>
Personally, I don't care either way since there's zero chance of the duplicate causing problems.

---

*[ycmd/tests/clang/testdata/FixIt_Clang_cpp11.cpp, line 68 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/548#-KMqUkw036gkEVJbIDdM:-KMqUkw036gkEVJbIDdN:5173007) ([raw file](https://github.com/valloric/ycmd/blob/f7a31e9475f9c9056c14345164f09c431ef51e88/ycmd/tests/clang/testdata/FixIt_Clang_cpp11.cpp#L68)):*
> ```cpp
> 
> namespace Typo {
>     struct SpellingIsNotMyStrongPoint;
> ```

:D

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/548#-:-KMqUnyLJyBndVUAvN2J:-244862327)*

vheon:
:lgtm: 

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/548#-:-KMsUGaHgcB8eyUNxKpZ:980077423)*

puremourning:
Rebased. I think this is ready to go.

---

Review status: 2 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 406 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/548#-KMqUGU__PAfjt0mzMCV:-KMt0NghWIKyp0y8SS9V:-219674365) ([raw file](https://github.com/valloric/ycmd/blob/f7a31e9475f9c9056c14345164f09c431ef51e88/ycmd/completers/cpp/flags.py#L406)):*
> It might be a bit confusing that you say that it's enabled by default and yet we add it ourselves. The difference is between the clang driver and calling libclang, but that might not be obvious to the reader.
> 
> I'd suggest clarifying this a bit to avoid confusion.
</details>
I thought 'clang driver' made that clear, but i have added a further clarification:

```
  # "Typo correction" (aka spell checking) in clang allows it to produce
  # hints (in the form of fix-its) in the case of certain diagnostics. A common
  # example is "no type named 'strng' in namespace 'std'; Did you mean
  # 'string'? (FixIt)". This is enabled by default in the clang driver (i.e. the
  # 'clang' binary), but is not when using libclang (as we do). It's a useful
  # enough feature that we just always turn it on unless the user explicitly
  # turned it off in their flags (with -fno-spell-checking).
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/548#-:-KMuOhfsjgiBwQEzIB5C:1527044772)*

Valloric:
Awesomesauce!

@homu r+

PR #550, opened 2016-07-17
[READY] Fix infinite loop leading to high cpu and disk space usage in TypeScript completer

micbou:
For each iteration in the `_ReaderLoop` method (run in a separate thread), we try to read lines from the TSServer stdout. This is working fine while the TSServer is running because `readline` wait for any output from the process. However, when the process is down, `readline` returns an empty string, which leads to an infinite loop with no interruption in `_ReaderLoop` where we catch and log the `RuntimeError` exception from `_ReadMessage`. This results in high cpu usage (full usage of one core) and a crazy disk space usage from the stderr logfile.

We prevent this by using an [event](https://docs.python.org/2/library/threading.html#event-objects) that is true when the server is running, false otherwise. We wait for this event to be true at each iteration in the `_ReaderLoop` method. If we catch a `RuntimeError` exception from `_ReadMessage`, we know that the server is not running and we set the event to false.

In addition, this PR improves the error message when sending a command or a request while TSServer is not running. We add a test for this.

I am marking this as RFC because I am not a multithreading expert and there is maybe a better approach to fix this issue.

Fixes https://github.com/Valloric/YouCompleteMe/issues/2161.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/550)

### DIFF
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -39,8 +39,9 @@ from ycmd import utils
 from ycmd.completers.completer import Completer
 from ycmd.completers.completer_utils import GetFileContents
 
-BINARY_NOT_FOUND_MESSAGE = ( 'tsserver not found. '
-                             'TypeScript 1.5 or higher is required' )
+BINARY_NOT_FOUND_MESSAGE = ( 'TSServer not found. '
+                             'TypeScript 1.5 or higher is required.' )
+SERVER_NOT_RUNNING_MESSAGE = 'TSServer is not running.'
 
 MAX_DETAILED_COMPLETIONS = 100
 RESPONSE_TIMEOUT_SECONDS = 10
@@ -124,8 +125,14 @@ class TypeScriptCompleter( Completer ):
 
     self._server_lock = threading.RLock()
 
-    # We first start the server and then the response-reading queue, so it will
-    # certainly find a responding TSServer.
+    # Used to read response only if TSServer is running.
+    self._tsserver_is_running = threading.Event()
+
+    # Start a thread to read response from TSServer.
+    self._thread = threading.Thread( target = self._ReaderLoop, args = () )
+    self._thread.daemon = True
+    self._thread.start()
+
     self._StartServer()
 
     # Used to map sequence id's to their corresponding DeferredResponse
@@ -135,11 +142,6 @@ class TypeScriptCompleter( Completer ):
     # Used to prevent threads from concurrently reading and writing to
     # the pending response dictionary
     self._pending_lock = threading.Lock()
-
-    # Start a thread to read response from TSServer.
-    self._thread = threading.Thread( target = self._ReaderLoop, args = () )
-    self._thread.daemon = True
-    self._thread.start()
 
     _logger.info( 'Enabling typescript completion' )
 
@@ -168,6 +170,8 @@ class TypeScriptCompleter( Completer ):
                                                stderr = subprocess.STDOUT,
                                                env = environ )
 
+      self._tsserver_is_running.set()
+
 
   def _ReaderLoop( self ):
     """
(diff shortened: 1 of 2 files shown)


vheon:
I started working on this but then didn't find the time to finish it up :sob: 

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/typescript/typescript_completer.py, line 187 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/550#-KMtQAvm-0KSydPN5Sat:-KMtQAvm-0KSydPN5Sau:1638812629) ([raw file](https://github.com/valloric/ycmd/blob/ca344a5df5a90812c7db34ab37d82bfd4a2d0822/ycmd/completers/typescript/typescript_completer.py#L187)):*
> ```Python
>       try:
>         message = self._ReadMessage()
>       except RuntimeError:
> ```

So we are assuming that the only way to get an exception is if the TSServer stops working?

---

*[ycmd/completers/typescript/typescript_completer.py, line 255 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/550#-KMtRPkE0gYPua4rQgaT:-KMtRPkE0gYPua4rQgaU:-1468919937) ([raw file](https://github.com/valloric/ycmd/blob/ca344a5df5a90812c7db34ab37d82bfd4a2d0822/ycmd/completers/typescript/typescript_completer.py#L255)):*
> ```Python
>   def _WriteRequest( self, request ):
>     """Write a request to TSServer stdin."""
> ```

We could also check here

```python
if not self._ServerIsRunning():
  _logger.exception( SERVER_NOT_RUNNING_MESSAGE )
  raise RuntimeError( SERVER_NOT_RUNNING_MESSAGE )
```

the catch down below is still needed because the server could crash rigth after we check so I don't know...

---

*[ycmd/completers/typescript/typescript_completer.py, line 262 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/550#-KMtQpfHt5k_SdV02dhi:-KMtQpfJlTkYQSE6STHk:856154665) ([raw file](https://github.com/valloric/ycmd/blob/ca344a5df5a90812c7db34ab37d82bfd4a2d0822/ycmd/completers/typescript/typescript_completer.py#L262)):*
> ```Python
>         self._tsserver_handle.stdin.flush()
>       # IOError is an alias of OSError in Python 3.
>       except IOError:
> ```

shouldn't we also set `_tsserver_is_running` here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/550#-:-KMtPxC7PSdBVBrdlvD4:690442098)*

micbou:
Reviewed 1 of 1 files at r2.
Review status: 1 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/typescript/typescript_completer.py, line 187 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/550#-KMtQAvm-0KSydPN5Sat:-KMtbEb_ZfgHpSkYGmMD:-1753205077) ([raw file](https://github.com/valloric/ycmd/blob/ca344a5df5a90812c7db34ab37d82bfd4a2d0822/ycmd/completers/typescript/typescript_completer.py#L187)):*
> So we are assuming that the only way to get an exception is if the TSServer stops working?
</details>
No, we only assume that a RuntimeError is raised if the server is down. If there are other exceptions, I think it is better to let them stop the loop.

---

*[ycmd/completers/typescript/typescript_completer.py, line 255 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/550#-KMtRPkE0gYPua4rQgaT:-KMtz-jSNi5365aZzJOs:-407885806) ([raw file](https://github.com/valloric/ycmd/blob/ca344a5df5a90812c7db34ab37d82bfd4a2d0822/ycmd/completers/typescript/typescript_completer.py#L255)):*
> We could also check here
> 
> ```python
> if not self._ServerIsRunning():
>   _logger.exception( SERVER_NOT_RUNNING_MESSAGE )
>   raise RuntimeError( SERVER_NOT_RUNNING_MESSAGE )
> ```
> 
> the catch down below is still needed because the server could crash rigth after we check so I don't know...
</details>
This is why I went for the [EAFP approach](https://docs.python.org/2/glossary.html#term-eafp) instead of the [LBYL approach](https://docs.python.org/2/glossary.html#term-lbyl). I don't think we need both approaches here.

---

*[ycmd/completers/typescript/typescript_completer.py, line 262 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/550#-KMtQpfHt5k_SdV02dhi:-KMtboo9R1NOtpZmwnmg:1345217290) ([raw file](https://github.com/valloric/ycmd/blob/ca344a5df5a90812c7db34ab37d82bfd4a2d0822/ycmd/completers/typescript/typescript_completer.py#L262)):*
> shouldn't we also set `_tsserver_is_running` here?
</details>
There is a very high chance that `_tsserver_is_running` is already cleared at this point. I am even wondering if it is worth to clear it in `_StopServer`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/550)*

Valloric:
:lgtm:

Nice!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/550#-:-KMzpstxs1VWkWFcGmkl:1300102931)*

micbou:
I removed clearing the `_tsserver_is_running` event when stopping TSServer. It was unnecessary.

Marking the PR as ready.

---

Reviewed 1 of 2 files at r1, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/550#-:-KN-1A3qDzj1XwM_By0F:1025407215)*

micbou:
@puremourning @vheon Thoughts on the PR?

---

Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/550#-:-KNEIY3BiN-FbArarFe4:-1195762657)*

puremourning:
:lgtm: can't see any data races or deadlocks, looks simple enough. Good stuff. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/550#-:-KNGf_hbpE-HXUlpbrhs:1084515563)*

puremourning:
Reviewed 1 of 2 files at r1, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/550)*

vheon:
:lgtm: @homu r=Valloric

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/550#-:-KNIdJVX574fas76Btq4:583508563)*

PR #551, opened 2016-07-17
[READY] Return True as JSON in /load_extra_conf_file and /ignore_extra_conf_file handlers

### LINKED ISSUE #536: ycmd should return valid JSON instead of empty HTML for 2 requests
/load_extra_conf_file and /ignore_extra_conf_file requests currently return an empty body
in case of success, which is not valid JSON. Instead, ycmd should return valid JSON body, for example just "true".

micbou:
See #536 for the rationale.

We choose to return `True` instead of an empty dictionary because it allows clients to use the response like this:
```python
response = ...
if response:
  # do something
```
while it would be awkward with an empty dictionary.

Adds one test for each handler and updates the existing ones.

Closes #536.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/551)

### DIFF
--- a/ycmd/handlers.py
+++ b/ycmd/handlers.py
@@ -188,12 +188,16 @@ def LoadExtraConfFile():
   request_data = RequestWrap( request.json, validate = False )
   extra_conf_store.Load( request_data[ 'filepath' ], force = True )
 
+  return _JsonResponse( True )
+
 
 @app.post( '/ignore_extra_conf_file' )
 def IgnoreExtraConfFile():
   _logger.info( 'Received extra conf ignore request' )
   request_data = RequestWrap( request.json, validate = False )
   extra_conf_store.Disable( request_data[ 'filepath' ] )
+
+  return _JsonResponse( True )
 
 
 @app.post( '/debug_info' )
--- a/ycmd/tests/misc_handlers_test.py
+++ b/ycmd/tests/misc_handlers_test.py
@@ -24,10 +24,10 @@ from future import standard_library
 standard_library.install_aliases()
 from builtins import *  # noqa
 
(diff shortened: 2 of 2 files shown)


puremourning:
Looks good. The only argument I can think of for a dictionary is that it allows us to enhance/change it later. Changing True to a dictionary is harder (not compatible). 

It's probably not a big deal though. What do you think?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/551#-:-KMvBFbcLv9xr7BbSlcv:1803470668)*

micbou:
I think we'll need to make incompatible changes to the API one day or another. We can't expect the API to never change. For instance, we are planning to change the `/debug_info` request to return a dictionary instead of a string. So, if one day, we really need to return more than a boolean, we'll make the change (preferably along other API changes). In addition, we are already returning `True` for the `/ready` and `/healthy` requests so this is not something new.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/551#-:-KMy8AbbAuVVkbEKx5-C:852780662)*

Valloric:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/551#-:-KMzqU-tpUVmIIUIudF_:1417088721)*

vheon:
:lgtm: @homu r=Valloric

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/551#-:-KN-4DyOdgX0oXnNUn3V:583508563)*

PR #555, opened 2016-07-23
[READY] Print a message equivalent to that printed by waitress.serve

puremourning:
#282 breaks emacs-ycmd as it parses the standard output to get the port that ycmd is listening on.

While it probably makes sense for emacs-ycmd to specify the port, this is a regression so we re-introduce the output equivalent to [`waitress.serve`](https://github.com/Pylons/waitress/blob/3850e4c09aa9a55dc5c61985981b4ef719d5ebb8/waitress/__init__.py#L13-L14)

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/555)

### DIFF
--- a/ycmd/wsgi_server.py
+++ b/ycmd/wsgi_server.py
@@ -38,6 +38,13 @@ class StoppableWSGIServer( TcpWSGIServer ):
   def Run( self ):
     """Wrapper of TcpWSGIServer run method. It prevents a traceback from
     asyncore."""
+
+    # Message for compatibility with clients who expect the output from
+    # waitress.serve here
+    print( 'serving on http://{0}:{1}'.format(
+      self.effective_host,
+      self.effective_port ) )
+
     try:
       self.run()
     except select.error:


puremourning:
cc @ptrv

vheon:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/555#-:-KNNM27mU_he6AUo6Imk:1417088721)*

micbou:
I was wondering why this message was not appearing anymore but didn't look further. I should have.

---

Reviewed 1 of 1 files at r1.

---

*[ycmd/__main__.py, line 188 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/555#-KNNJsgZuF3I5BV7itG3:-KNNJsg_XnCQSLDVJEgA:1839064648) ([raw file](https://github.com/valloric/ycmd/blob/cd7577da02739b7130a219cb59a288b2b14294ff/ycmd/__main__.py#L188)):*
> ```Python
>   print( 'serving on http://{0}:{1}'.format(
>     handlers.wsgi_server.effective_host,
>     handlers.wsgi_server.effective_port ) )
> ```

I would move this to the `StoppableWSGIServer` class because it depends on `waitress` internals.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/555#-:-KNNJT_7z3N3WlhChlNU:-847978147)*

ptrv:
Thanks for bringing back the old behavior. We probably need to change the way `emacs-ycmd` starts the server and how it checks for the port (probably specifying the port when starting the server) but that would need more effort to change compared to just adapt the new shutdown process.

Thanks again for your work!

puremourning:

---

*[ycmd/__main__.py, line 188 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/555#-KNNJsgZuF3I5BV7itG3:-KNNu9Dypys2U1NWL5gH:-212864262) ([raw file](https://github.com/valloric/ycmd/blob/cd7577da02739b7130a219cb59a288b2b14294ff/ycmd/__main__.py#L188)):*
> I would move this to the `StoppableWSGIServer` class because it depends on `waitress` internals.
</details>
Done.

You're quite right.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/555)*

micbou:
:lgtm:

@homu r+

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/555#-:-KNNw371ucG46nYQ6HCh:187772251)*

PR #560, opened 2016-07-26
[READY] Increase MEMFILE_MAX to 10MB

bijancn:
See discussion [here](https://github.com/Valloric/YouCompleteMe/issues/624).

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/560)

### DIFF
--- a/ycmd/handlers.py
+++ b/ycmd/handlers.py
@@ -41,7 +41,7 @@ from ycmd.completers.completer_utils import FilterAndSortCandidatesWrap
 
 # num bytes for the request body buffer; request.json only works if the request
 # size is less than this
-bottle.Request.MEMFILE_MAX = 1000 * 1024
+bottle.Request.MEMFILE_MAX = 10000 * 1024
 
 _server_state = None
 _hmac_secret = bytes()
--- a/ycmd/handlers.py
+++ b/ycmd/handlers.py
@@ -41,7 +41,7 @@ from ycmd.completers.completer_utils import FilterAndSortCandidatesWrap
 
 # num bytes for the request body buffer; request.json only works if the request
 # size is less than this
-bottle.Request.MEMFILE_MAX = 10000 * 1024
+bottle.Request.MEMFILE_MAX = 10 * 1024 * 1024
 
 _server_state = None
 _hmac_secret = bytes()
--- a/ycmd/tests/misc_handlers_test.py
+++ b/ycmd/tests/misc_handlers_test.py
@@ -51,8 +51,8 @@ def MiscHandlers_EventNotification_AlwaysJsonResponse_test( app ):
 
 @SharedYcmd
 def MiscHandlers_EventNotification_ReturnJsonOnBigFileError_test( app ):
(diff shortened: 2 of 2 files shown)


puremourning:
Hang on. Something is up with this. It says "1 of 2 LGTMs" so i think it is all confused and associated all the triggers, etc. with the other review. I suspect there is some issue with github workflow when merging master-master.

bijancn:
Mh nope. I see 0 of 2 LGTM here in #560

bijancn:
And now it shows 1 of 2.. Something is really strange here.

bijancn:
I keeps swapping the reviewable link between 559 and 560. But CI always point to 560.

puremourning:
I've closed the 559 review and done a little admin on this PR title. I hope you don't mind.

puremourning:
:lgtm: assuming the CI builds succeed!

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/560#-:-KNacCmvJ4_e5I9CqlEA:955259063)*

Valloric:
:lgtm: with minor comment

---

---

*[ycmd/handlers.py, line 44 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/560#-KNc5uQ8BGbzPGQEJBmQ:-KNc5uQ8BGbzPGQEJBmR:1390676870) ([raw file](https://github.com/valloric/ycmd/blob/e3a7dde6eff43237970ca8dfea376edbda705344/ycmd/handlers.py#L44)):*
> ```Python
> # num bytes for the request body buffer; request.json only works if the request
> # size is less than this
> bottle.Request.MEMFILE_MAX = 10000 * 1024
> ```

This should be `1024 * 1024 * 10` so that it's obvious it's 10 MB.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/560#-:-KNc60dOVzp6Y4L4Eegh:-1144525623)*

bijancn:
Can someone explain to me why one of these builds failed and what the differences of the builds are (beyond python version and USE_CLANG_COMPLETER)? 
https://travis-ci.org/Valloric/ycmd/builds/147418829
It seems that both .4 and .6 are using python 3.3 and USE_CLANG_COMPLETER=TRUE so I don't see why one fails.

bijancn:
How can I reduce coverage by 1% when I only change constants? ^^

puremourning:
> How can I reduce coverage by 1% when I only change constants? 
I think the coverage bot is a little flaky. Possibly the coverage goes down if the build or test fails.

puremourning:
> Can someone explain to me why one of these builds failed 
Flake? we can manually re-run if needed tho it looks like the build is re-running anyway (i guess you pushed your branch )

bijancn:
Travis now passed. So seems to have been a hickup. But in appveyor there are already fails again:
```AttributeError: 'FlakesChecker' object has no attribute 'with_doctest'```
Is this due to my changes?

Valloric:
@bijancn The appveyor fail is not because of your change.

Thanks for the PR!

@homu r+

PR #572, opened 2016-08-11
[READY] Properly remove logfiles at program termination

micbou:
With this PR, we now remove `ycmd` logfiles at program termination and not only when receiving the `SIGTERM` or `SIGINT` signals. This fixes the long-standing bug where logfiles were never removed on Windows and the regression where they were not removed anymore on Unix platforms when receiving the `shutdown` request or through the watchdog plugin.

In addition, when shutting down a sub-server in a completer, we always wait for its process to be terminated before removing the logfiles. We need this on Windows because we cannot remove a file that it still used by a process on this platform. Most of the completers were already waiting but not the Python and C♯ completers.

Shutdown tests are updated to check if the logfiles are removed.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/572)

### DIFF
--- a/ycmd/__main__.py
+++ b/ycmd/__main__.py
@@ -32,6 +32,7 @@ from future import standard_library
 standard_library.install_aliases()
 from builtins import *  # noqa
 
+import atexit
 import sys
 import logging
 import json
@@ -53,29 +54,31 @@ def YcmCoreSanityCheck():
 
 # We manually call sys.exit() on SIGTERM and SIGINT so that atexit handlers are
 # properly executed.
-def SetUpSignalHandler( stdout, stderr, keep_logfiles ):
+def SetUpSignalHandler():
   def SignalHandler( signum, frame ):
-    # We reset stderr & stdout, just in case something tries to use them
-    if stderr:
-      tmp = sys.stderr
-      sys.stderr = sys.__stderr__
-      tmp.close()
-    if stdout:
-      tmp = sys.stdout
-      sys.stdout = sys.__stdout__
-      tmp.close()
-
-    if not keep_logfiles:
-      if stderr:
-        utils.RemoveIfExists( stderr )
-      if stdout:
-        utils.RemoveIfExists( stdout )
-
     sys.exit()
 
   for sig in [ signal.SIGTERM,
                signal.SIGINT ]:
     signal.signal( sig, SignalHandler )
+
+
+def CleanUpLogfiles( stdout, stderr, keep_logfiles ):
+  # We reset stderr & stdout, just in case something tries to use them
+  if stderr:
+    tmp = sys.stderr
+    sys.stderr = sys.__stderr__
+    tmp.close()
+  if stdout:
+    tmp = sys.stdout
+    sys.stdout = sys.__stdout__
+    tmp.close()
+
+  if not keep_logfiles:
+    if stderr:
+      utils.RemoveIfExists( stderr )
+    if stdout:
+      utils.RemoveIfExists( stdout )
 
 
 def PossiblyDetachFromTerminal():
@@ -172,7 +175,13 @@ def Main():
   handlers.UpdateUserOptions( options )
   handlers.SetHmacSecret( hmac_secret )
   handlers.KeepSubserversAlive( args.check_interval_seconds )
-  SetUpSignalHandler( args.stdout, args.stderr, args.keep_logfiles )
+  SetUpSignalHandler()
+  # Functions registered by the atexit module are called at program termination
+  # in last in, first out order.
+  atexit.register( CleanUpLogfiles, args.stdout,
+                                    args.stderr,
+                                    args.keep_logfiles )
+  atexit.register( handlers.ServerCleanup )
   handlers.app.install( WatchdogPlugin( args.idle_suicide_seconds,
                                         args.check_interval_seconds ) )
   handlers.app.install( HmacPlugin( hmac_secret ) )
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -27,7 +27,6 @@ from future.utils import itervalues
 
 from collections import defaultdict
 import os
-import time
 import re
 from ycmd.completers.completer import Completer
 from ycmd.utils import ForceSemanticCompletion, CodepointOffsetToByteOffset
@@ -72,10 +71,9 @@ class CsharpCompleter( Completer ):
 
 
   def Shutdown( self ):
-    if ( self.user_options[ 'auto_stop_csharp_server' ] ):
+    if self.user_options[ 'auto_stop_csharp_server' ]:
       for solutioncompleter in itervalues( self._completer_per_solution ):
-        if solutioncompleter._ServerIsRunning():
-          solutioncompleter._StopServer()
+        solutioncompleter._StopServer()
 
 
   def SupportedFiletypes( self ):
@@ -411,40 +409,20 @@ class CsharpSolutionCompleter( object ):
   def _StopServer( self ):
     """ Stop the OmniSharp server using a lock. """
     with self._server_state_lock:
-      if not self._ServerIsRunning():
-        return
-
-      self._logger.info( 'Stopping OmniSharp server' )
-
-      self._TryToStopServer()
-
-      # Terminate it if it's still up
       if self._ServerIsRunning():
-        self._logger.info( 'Terminating OmniSharp server' )
-        self._omnisharp_phandle.terminate()
+        self._logger.info( 'Stopping OmniSharp server with PID {0}'.format(
+                               self._omnisharp_phandle.pid ) )
+        self._GetResponse( '/stopserver' )
         self._omnisharp_phandle.wait()
+        self._logger.info( 'Stopped OmniSharp server' )
 
-      self._CleanupAfterServerStop()
-
-      self._logger.info( 'Stopped OmniSharp server' )
-
-
-  def _TryToStopServer( self ):
-    for _ in range( 5 ):
-      try:
-        self._GetResponse( '/stopserver', timeout = .1 )
-      except:
-        pass
-      for _ in range( 10 ):
-        if not self._ServerIsRunning():
-          return
-        time.sleep( .1 )
+      self._CleanUp()
 
 
-  def _CleanupAfterServerStop( self ):
+  def _CleanUp( self ):
     self._omnisharp_port = None
     self._omnisharp_phandle = None
-    if ( not self._keep_logfiles ):
+    if not self._keep_logfiles:
       if self._filename_stdout:
         utils.RemoveIfExists( self._filename_stdout )
         self._filename_stdout = None
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -418,10 +418,11 @@ class CsharpSolutionCompleter( object ):
 
       self._TryToStopServer()
 
-      # Kill it if it's still up
+      # Terminate it if it's still up
       if self._ServerIsRunning():
-        self._logger.info( 'Killing OmniSharp server' )
-        self._omnisharp_phandle.kill()
+        self._logger.info( 'Terminating OmniSharp server' )
+        self._omnisharp_phandle.terminate()
+        self._omnisharp_phandle.wait()
 
       self._CleanupAfterServerStop()
 
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -413,8 +413,12 @@ class CsharpSolutionCompleter( object ):
         self._logger.info( 'Stopping OmniSharp server with PID {0}'.format(
                                self._omnisharp_phandle.pid ) )
         self._GetResponse( '/stopserver' )
-        self._omnisharp_phandle.wait()
-        self._logger.info( 'Stopped OmniSharp server' )
+        try:
+          utils.WaitUntilProcessIsTerminated( self._omnisharp_phandle,
+                                              timeout = 5 )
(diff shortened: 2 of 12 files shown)


Valloric:
:lgtm:

Nice!

---

Reviewed 6 of 6 files at r1.

---

*[ycmd/__main__.py, line 180 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/572#-KOw5ZSBTVvg3Nd8AR0T:-KOw5ZSBTVvg3Nd8AR0U:b-6777f) ([raw file](https://github.com/valloric/ycmd/blob/10b1c8964aec380fe897a695d8c246fab2ef25fd/ycmd/__main__.py#L180)):*
> ```Python
>   # Functions registered by the atexit module are called at program termination
>   # in last in, first out order.
>   atexit.register( LogfilesCleanup, args.stdout,
> ```

We've had some issues with atexit handlers in the past. We used to use them in ycmd (or maybe YCM; can't remember) but we move away from them.

Can't recall the details. Just keep an eye open.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/572#-:-KOw6POowgIilzAsV1BK:bli1pub)*

Valloric:

---

*[ycmd/__main__.py, line 180 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/572#-KOw5ZSBTVvg3Nd8AR0T:-KOw6XpInx7KnkStDpT0:bqdjln6) ([raw file](https://github.com/valloric/ycmd/blob/10b1c8964aec380fe897a695d8c246fab2ef25fd/ycmd/__main__.py#L180)):*
> We've had some issues with atexit handlers in the past. We used to use them in ycmd (or maybe YCM; can't remember) but we move away from them.
> 
> Can't recall the details. Just keep an eye open.
</details>
Sorry, I meant we used them for other stuff, not just ServerCleanup.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/572)*

puremourning:
:lgtm:

Nice work! I am curious about how effective `atexit` handlers really are. I also have some recollection that they were problematic for some reason in the past. I googled it and read the python docs, but the caveats don't seem to be such a deal for us:

- we don't use `os._exit()` and i hope/don't think any of our libs do
- we don't use `sys.exitfunc` and i hope/don't think any of our libs do

I found this useful: http://grodola.blogspot.com/2016/02/how-to-always-execute-exit-functions-in-py.html

I wonder if we want to handle signals _and_ `atexit`? Is there any reason not to? We know the signal handlers don't work on windows, and there doesn't appear to be a reliable way to implement one, but it would cover a bunch of other cases for unix systems (such as os shutdown, manual kill, etc.).

---

Reviewed 3 of 6 files at r1, 3 of 3 files at r2.

---

*[ycmd/__main__.py, line 66 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/572#-KPbfic0LSEicKBELb68:-KPbfic0LSEicKBELb69:b-jf0ngc) ([raw file](https://github.com/valloric/ycmd/blob/affe9438ff84b970e4c0feed6178af70c55bd3a9/ycmd/__main__.py#L66)):*
> ```Python
> 
> 
> def LogfilesCleanup( stdout, stderr, keep_logfiles ):
> ```

I think a better name would be `CleanUpLogFiles` and that is verb-noun rather than noun-verb (more natural in English)

---

*[ycmd/completers/cs/cs_completer.py, line 424 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/572#-KPbg1wyjIcwzAWUzdBI:-KPbg1wyjIcwzAWUzdBJ:b-h1g86) ([raw file](https://github.com/valloric/ycmd/blob/affe9438ff84b970e4c0feed6178af70c55bd3a9/ycmd/completers/cs/cs_completer.py#L424)):*
> ```Python
>       if self._ServerIsRunning():
>         self._logger.info( 'Terminating OmniSharp server' )
>         self._omnisharp_phandle.terminate()
> ```

We have a `utils.TerminateProcess` which works on a pid. I wonder if we should change it to check if it is passed an instance of `phandle` (or whatever this is) and do `terminate()/wait()`. Or maybe add a `TerminatePHandle` so that we do this consistently for subserves?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/572#-:-KPbhbFJJVBEe0pAm09C:b-sag5a4)*

vheon:
Reviewed 2 of 6 files at r1, 3 of 3 files at r2, 4 of 7 files at r3.
Review status: 6 of 10 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/572)*

vheon:
:lgtm: just a minor comment :+1: 

---

Reviewed 1 of 6 files at r1, 3 of 7 files at r3, 3 of 3 files at r4.

---

*[ycmd/completers/typescript/typescript_completer.py, line 550 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/572#-KPddNJh6b0IVDJ3SwpG:-KPddNJi0jYgckIqSSpr:b66awhl) ([raw file](https://github.com/valloric/ycmd/blob/9d2bc3c27609f6e6325812eb8fedaaa367f617b0/ycmd/completers/typescript/typescript_completer.py#L550)):*
> ```Python
> 
>   def _CleanUp( self ):
>     self._tsserver_handle = None
> ```

Shouldn't we keep this to restore the internal state? we seems to do it for every completer :confused: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/572#-:-KPddc9PaYx5WbrGiBoX:b-v4eizu)*

vheon:
Review status: 5 of 12 files reviewed at latest revision, 5 unresolved discussions.

---

*[ycmd/utils.py, line 284 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/572#-KPdr2S3TLbukEsslihz:-KPdr2S3TLbukEsslii-:b-up4txe) ([raw file](https://github.com/valloric/ycmd/blob/73a93af51ec8ca551689b37b0e0bc276e853b4a4/ycmd/utils.py#L284)):*
> ```Python
> 
> 
> def WaitUntilProcessIsTerminated( handle, timeout = 5 ):
> ```

Why is this needed now? couldn't we just use the `timeout` parameter of `wait`?

---

*[ycmd/utils.py, line 285 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/572#-KPdqyxc0O5hAXzT8Wrz:-KPdqyxc0O5hAXzT8Ws-:bxb6h57) ([raw file](https://github.com/valloric/ycmd/blob/0d8a4c354a6e7ef269b803d719a65735d1108ff6/ycmd/utils.py#L285)):*
> ```Python
> 
> # From here: http://stackoverflow.com/a/8536476/1672783
> def TerminateProcess( pid ):
> ```

Is this not needed anymore?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/572)*

micbou:
I updated the `_StopServer` methods to be more consistent through all completers.

> I found this useful: http://grodola.blogspot.com/2016/02/how-to-always-execute-exit-functions-in-py.html

That's interesting. According to [this comment](http://grodola.blogspot.com/2016/02/how-to-always-execute-exit-functions-in-py.html?google_comment_id=z13lgx1welfndn424222urkqauy2yl1ey), we are doing the right thing in `SetUpSignalHandler`.

> I wonder if we want to handle signals and atexit? Is there any reason not to?

We already take care of `SIGTERM` and `SIGINT` in `SetUpSignalHandler`. Are there more signals to handle?

> We know the signal handlers don't work on windows, and there doesn't appear to be a reliable way to implement one, but it would cover a bunch of other cases for unix systems (such as os shutdown, manual kill, etc.).

A manual kill would be difficult to handle: this is [a `SIGKILL` signal](https://en.wikipedia.org/wiki/Unix_signal#SIGKILL).

---

Reviewed 2 of 6 files at r1, 2 of 3 files at r2, 7 of 7 files at r3, 1 of 3 files at r4, 6 of 7 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 1 of 1 files at r8.

---

*[ycmd/__main__.py, line 66 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/572#-KPbfic0LSEicKBELb68:-KPcTdI5S784x51DtuuS:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/affe9438ff84b970e4c0feed6178af70c55bd3a9/ycmd/__main__.py#L66)):*
> I think a better name would be `CleanUpLogFiles` and that is verb-noun rather than noun-verb (more natural in English)
</details>
Done.

---

*[ycmd/utils.py, line 284 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/572#-KPdr2S3TLbukEsslihz:-KPdrbH7ds99hYGlmjp4:bh8yfgm) ([raw file](https://github.com/valloric/ycmd/blob/73a93af51ec8ca551689b37b0e0bc276e853b4a4/ycmd/utils.py#L284)):*
> Why is this needed now?

We don't want to be indefinitely stuck waiting for the process to be down.

>  couldn't we just use the timeout parameter of wait?

Unfortunately, the `timeout` parameter is only available since Python 3.3 so we need our own implementation.

---

*[ycmd/utils.py, line 285 \[r5\]](https://reviewable.io:443/reviews/valloric/ycmd/572#-KPdqyxc0O5hAXzT8Wrz:-KPdrZJsT-ybOwUI5xnW:b-exasd3) ([raw file](https://github.com/valloric/ycmd/blob/0d8a4c354a6e7ef269b803d719a65735d1108ff6/ycmd/utils.py#L285)):*
> Is this not needed anymore?
</details>
Yes, I've searched through the code base and it is not used anywhere.

---

*[ycmd/completers/cs/cs_completer.py, line 424 \[r2\]](https://reviewable.io:443/reviews/valloric/ycmd/572#-KPbg1wyjIcwzAWUzdBI:-KPcGIsxi91fUgMJ5a-T:b-91uxw2) ([raw file](https://github.com/valloric/ycmd/blob/affe9438ff84b970e4c0feed6178af70c55bd3a9/ycmd/completers/cs/cs_completer.py#L424)):*
> We have a `utils.TerminateProcess` which works on a pid. I wonder if we should change it to check if it is passed an instance of `phandle` (or whatever this is) and do `terminate()/wait()`. Or maybe add a `TerminatePHandle` so that we do this consistently for subserves?
</details>
The thing is that we don't always `terminate()/wait()` when stopping a subserver. When a subserver exposes a shutdown method (like OmniSharp and gocode), we prefer to use it over `terminate` because it is likely to work better on Windows. So, it seems difficult to wrap this in one function for all subservers.

---

*[ycmd/completers/typescript/typescript_completer.py, line 550 \[r3\]](https://reviewable.io:443/reviews/valloric/ycmd/572#-KPddNJh6b0IVDJ3SwpG:-KPdoBcXzhQrKqrqfBty:b-mt2r47) ([raw file](https://github.com/valloric/ycmd/blob/9d2bc3c27609f6e6325812eb8fedaaa367f617b0/ycmd/completers/typescript/typescript_completer.py#L550)):*
> Shouldn't we keep this to restore the internal state? we seems to do it for every completer :confused: 
</details>
I initially removed it because it was raising an `AttributeError` in the `_WriteRequest` method. But catching this exception is not an issue (it's even a good idea) so it's back.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/572#-:-KPcDhT1oe3hxHt6cMu8:bc28gq2)*

vheon:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/572#-:-KPiogTZOelfU2qGfird:bnfp4nl)*

micbou:
Codecov is complaining that the diff part of this PR is not covered enough. Unfortunately, the non-covered code is really difficult to test: it mostly involves sub-servers timing out. I could exclude this code from coverage by adding `# pragma: no cover` comments but I am not a huge fan of this approach. The total coverage is still increased so we could simply ignore the check and merge the PR. Thoughts?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/572#-:-KQvc_1JL6qnm8udn_6d:b-2kk5u0)*

puremourning:
:lgtm:

I'm not personally worried about the diff %, given that this overall increases coverage and fixes a pretty big issue on windows :)

---

Reviewed 1 of 7 files at r3, 1 of 3 files at r4, 5 of 7 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 1 of 1 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/572#-:-KQwOWHh6HuFr8MZljtN:b-3uojrd)*

vheon:
> I'm not personally worried about the diff %, given that this overall increases coverage and fixes a pretty big issue on windows :)

I agree. Still :lgtm: 

---

Reviewed 5 of 7 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 1 of 1 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/572#-:-KQzW79DUsXQgp-ZFJk3:b-hvjgbc)*

micbou:
I am merging then.

@homu r+

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/572#-:-KR0GS6ZVQlcjzdUZtGL:b-db2z3c)*

micbou:
@homu retry

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/572#-:-KR0W4WnJPlb_KDvD5jM:b8khjrm)*

PR #448, opened 2016-04-01
[READY] Add python-future to sys.path just before packages path

micbou:
### Problem

Modules from our `python-future` package may be overridden by ones from already installed packages: a package using the same name for its modules (ex: `builtins`) or a different version of `python-future`.

### How to reproduce

With a package installing a `builtins` module:
```sh
pip install pies2overrides
```
results in:
```
Traceback (most recent call last):
  File "<string>", line 24, in <module>
  File "C:\\Users\\micbou\\.vim\\bundle\\YouCompleteMe\\autoload\..\python\ycm\setup.py", line 49, in SetUpYCM
    base.LoadJsonDefaultsIntoVim()
  File "C:\\Users\\micbou\\.vim\\bundle\\YouCompleteMe\\autoload\..\python\ycm\base.py", line 55, in LoadJsonDefaultsIntoVim
    defaults = user_options_store.DefaultOptions()
  File "C:\Users\micbou\.vim\bundle\YouCompleteMe\python\ycm\..\..\third_party\ycmd\ycmd\user_options_store.py", line 55, in DefaultOptions
    options = json.loads( ReadFile( settings_path ) )
  File "C:\Users\micbou\.vim\bundle\YouCompleteMe\python\ycm\..\..\third_party\ycmd\ycmd\utils.py", line 55, in ReadFile
    with open( filepath, encoding = 'utf8' ) as f:
TypeError: 'encoding' is an invalid keyword argument for this function
```

With a different version of `python-future`:
```sh
pip install future==0.10.0
```
gives this error:
```
Traceback (most recent call last):
  File "<string>", line 22, in <module>
  File "C:\\Users\\micbou\\.vim\\bundle\\YouCompleteMe\\autoload\..\python\ycm\base.py", line 23, in <module>
    standard_library.install_aliases()
AttributeError: 'module' object has no attribute 'install_aliases'
```

### Solution

Instead of appending `python-future` to `sys.path`, insert it just before the first `site-packages` (`dist-packages` on Debian-like distributions) path. If no `site/dist-packages` are found, fall back to appending it.

Tested on Windows 7, Ubuntu 14.04, and OS X 10.11.

Fixes Valloric/YouCompleteMe#2024
Closes #446

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/448)

PR #577, opened 2016-08-17
[READY] Update gocode submodule

micbou:
Our version of Gocode is not working anymore with [the new Go release](https://blog.golang.org/go1.7). This makes the [Travis runs](https://travis-ci.org/Valloric/ycmd/builds/152665592) fail on OS X where Go was updated to 1.7 but not on Linux which is still using 1.4.1. The issue is fixed by commit https://github.com/nsf/gocode/commit/f535dc686130fcc7b942c504ce5903222a205ca3 in the Gocode repository. We update the Gocode submodule to its latest version.

With this new version, the Gocode daemon now automatically shut down after a period of inactivity (30 minutes by default). Since we don't want subservers to shut down for inactivity while ycmd is up, we need to add some logic to keep them alive. This is implemented in PR #576.

PR #576, opened 2016-08-17
[READY] Keep subservers alive while ycmd is up

micbou:
We don't want subservers from filetype completers to automatically shut down for inactivity while ycmd is still up so we periodically check if they are healthy (by calling the `ServerIsHealthy` method for each loaded filetype completer), which should effectively keep them alive.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/576)

PR #578, opened 2016-08-17
[READY] Insert python-future module just after the standard library path

micbou:
In PR #448, I made the assumption that the `site-packages` paths are always placed after the standard library path so that if we insert the `python-future` module just before the first `site-packages` path, it would necessary be after the standard library. Turns out that it is not true when a `site-packages` path is added to the `PYTHONPATH` environment variable, e.g, when using the software [ROS](http://www.ros.org/). See issue https://github.com/Valloric/YouCompleteMe/issues/2186. When this happens, the `python-future` module will raise the following exception on Python 3:
```python
ImportError: This package should not be accessible on Python 3.
Either you are trying to run from the python-future src folder or
your installation of python-future is corrupted.
```

We prevent this by looking for the standard library path in `sys.path` and by inserting the `python-future` module just after it. If we can't find it, we raise an exception since YCM and ycmd cannot work without it.

Fixes https://github.com/Valloric/YouCompleteMe/issues/2186.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/578)

### DIFF
--- a/ycmd/server_utils.py
+++ b/ycmd/server_utils.py
@@ -22,6 +22,7 @@ from __future__ import absolute_import
 # No other imports from `future` because this module is loaded before we have
 # put our submodules in sys.path
 
+from distutils import sysconfig
 import io
 import logging
 import os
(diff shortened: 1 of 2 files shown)


Valloric:
:lgtm:

Hopefully Python gives us a goddamn break at some point...

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/578#-:-KPONu--g8l3AlPPaRcV:b-83crqy)*

vheon:
:lgtm: @homu r=Valloric

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/578#-:-KPOdTJkUX_2HyJztN4t:b9nem37)*

vheon:
I just forgot that we first need #577 and #576 😭

vheon:
@homu r-

micbou:
Only #577 is needed to pass the tests.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/578#-:-KPOfQ7IABhmEiTzFj9L:b-sepjc6)*

vheon:
This should be good now. @homu retry

puremourning:
@homu r=vheon

PR #579, opened 2016-08-23
[READY] Search os.py file to find standard library path in sys.path

micbou:
It turns out that using `get_python_lib` from the `sysconfig` module to find the standard library path in `sys.path` is not reliable. See issue https://github.com/Valloric/YouCompleteMe/issues/2293. Instead, we now look for the `os.py` file through all paths of `sys.path`. You may ask why `os.py` and not another file from the standard library. Python itself is using this file to [find the `sys.prefix` path](https://github.com/python/cpython/blob/master/Modules/getpath.c#L109) so it is a safe bet.

Fixes https://github.com/Valloric/YouCompleteMe/issues/2293.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/579)

### DIFF
--- a/ycmd/server_utils.py
+++ b/ycmd/server_utils.py
@@ -22,7 +22,6 @@ from __future__ import absolute_import
 # No other imports from `future` because this module is loaded before we have
 # put our submodules in sys.path
 
-from distutils import sysconfig
 import io
 import logging
 import os
@@ -146,10 +145,8 @@ def PathToNearestThirdPartyFolder( path ):
 
 
 def GetStandardLibraryIndexInSysPath():
-  standard_library_path = os.path.normcase(
-    sysconfig.get_python_lib( standard_lib = True ) )
   for path in sys.path:
-    if os.path.normcase( path ) == standard_library_path:
+    if os.path.isfile( os.path.join( path, 'os.py' ) ):
       return sys.path.index( path )
   raise RuntimeError( 'Could not find standard library path in Python path.' )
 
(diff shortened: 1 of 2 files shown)


puremourning:
According to the comments in the python source file you linked, python looks for something matching `lib/python<PYTHON_VERSION>/os.py`.

Perhaps we should do the same? I mean it is *unlikely* that anyone would have their own `os.py` in a random project, but it isn't beyond the realm of possibility (considering the nonsense i've seen with "fast" json routines overriding the standard library and making it an order of magnitude slower!)

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/579#-:-KPqn1H1B77Im4VW99a8:bwizrv6)*

micbou:
Unfortunately, the `lib/python<PYTHON_VERSION>/os.py` path would only work for UNIX platforms. On Windows, we would need to match `Lib/os.py` instead. As the current solution is really simple, I think it is better to keep it that way for now and consider the approach you suggested if users are still reporting similar issues with the `python-future` module.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/579#-:-KPrVDTIol9UbcOpljro:b-88q5)*

puremourning:
Solid argument

:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/579#-:-KPrhnwbmSkVcxdvWx2E:b-rqn6m5)*

Valloric:
:lgtm:

Oh Python, I hate you so, so much.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/579#-:-KPs2qmTQTa3WP9QhYfX:brmac3r)*

vheon:
:lgtm: @homu r=puremourning 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/579#-:-KPsJex5R6ArlO6nacMC:bpw2qui)*

PR #580, opened 2016-08-23
[READY] Update racerd to master

puremourning:
Fixes https://github.com/Valloric/YouCompleteMe/issues/2222#issuecomment-241806666 by updating racerd to master.

WIP because @jwilm thinks appveyor might barf.

Edit: Technically only fixes above issue when ycmd submodule updated in YCM

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/580)

PR #581, opened 2016-08-23
Update racerd.

rillian:
Use v0.3.0 of racerd for rust lookup. This bumps the version
of the racer engine to 0.7.10 and updates other dependencies
to work with near-future (1.12) rust releases.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/581)


micbou:
Thanks for the PR but we already have #580.

PR #589, opened 2016-09-01
Make it possible to run on cygwin with windows python

Hernrup:
If we're running vim in cygwin, while JediHTTP is currently running a
windows python, we need to convert the path's we pass to the windows
python before when we start the server and also convert the returned
paths so that GoTo commands will work.

(Partial credit to @joakimkarlsson)

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/589)

### DIFF
--- a/ycmd/completers/python/jedi_completer.py
+++ b/ycmd/completers/python/jedi_completer.py
@@ -72,8 +72,9 @@ class JediCompleter( Completer ):
     self._python_binary_path = sys.executable
 
     self._UpdatePythonBinary( user_options.get( 'python_binary_path' ) )
-    self._StartServer()
+    self._is_windows_python = utils.IsWindowsPython(self._python_binary_path)
 
+    self._StartServer()
 
   def _UpdatePythonBinary( self, binary ):
     if binary:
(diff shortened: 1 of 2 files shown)


mispencer:
FYI - I have been planning a [general Cygwin pull request](https://github.com/mispencer/ycmd/commit/fcfa1a3e9e96934eebff233cd3a6d58fc80a741a) but I haven't had time to complete it.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/589#-:-KQa0nGZZ6Hd2n4S-KRC:bcqg5ti)*

Hernrup:
That rely nice @mispencer. The cygwin side of things can use some love ;)
I might be missing something but I think my solution is the minimal effort needed to make all features functional. Perhaps we should merge the proposals and get this done once and for all?

Apparently there is also some test that does not agree with me, dont know what thats all about. Will have to look into it.

Valloric:
Thanks for the PR! Cygwin is not currently supported and we don't currently have plans for adding such support. If we did, it would be a bigger endeavor where it would have to go into our test matrix etc.

So we appreciate the interest and the work you put into it, but we won't be merging this.

PR #590, opened 2016-09-02
[READY] Enable KeepGoing option when creating translation units in Clang completer

### LINKED ISSUE #421: Investigate whether or not the new 'keep-going' flag would be useful
Reference: http://reviews.llvm.org/rL262318

This new flag `CXTranslationUnit_KeepGoing` allows the `libclang` parser to attempt to continue, even after fatal errors:

> Optionally demote fatal errors to non-fatal errors.

> This behavior is enabled when the new CXTranslationUnit_KeepGoing
option is passed to clang_parseTranslationUnit{,2}. It is geared
towards use by IDEs and similar consumers of the clang-c API where
fatal errors may arise when parsing incomplete code mid-edit, or
when include paths are not properly configured yet. In such situations
one still wants to get as much information as possible about a TU.
Previously, the semantic analysis would not instantiate templates
or report additional fatal errors after the first fatal error was
encountered.

It's probably worth taking a look to see if this improves the ycmd feature set.

micbou:
I am creating this PR to try out the new `KeepGoing` option mentioned by @puremourning in issue #421. The option is now available with Clang 3.9.0. Here's its documentation:
> When fatal errors are encountered while parsing a translation unit, semantic analysis is typically stopped early when compiling code. A common source for fatal errors are unresolvable include files. For the  purposes of an IDE, this is undesirable behavior and as much information as possible should be reported. Use this flag to enable this behavior.

I did some tests and my conclusion is that enabling this option is a no-brainer. There are not a lot of different fatal errors and most of them never happen in practice but the one raised when an include file is missing is really common.

Closes #421.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/590)


puremourning:
> my conclusion is that enabling this option is a no-brainer

I agree. We need to rebase this, and maybe add a quick test maybe (i.e. prove that we get a diagnostic with a `#include <this_file_does_not_exist>` in it ?

---

Review status: 0 of 59 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/590#-:-KQwkltq4XQo0JyiL_oZ:bv1c4k1)*

micbou:
PR rebased. I added a test with two missing includes. An error is reported for the second missing include when the `KeepGoing` option is set.

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/590#-:-KQxhBL5KFzRtaVDyX5n:b-iuefff)*

vheon:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/590#-:-KQzWmlXNA3diASPypst:bnfp4nl)*

puremourning:
:lgtm:

@homu r+

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/590#-:-KRJH-xvHT1vuHab1OnY:b33slzv)*

puremourning:
Strangely, i just tested this #include thing without this patch and it already behaves like described.

So, while this flags still seems like a no-brainer, I'm curious what it is changing internally. I might go poke around and see if I can find a test case.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/590#-:-KRJLeLsIM0713vZP99u:b-u0jwmo)*

vheon:
@puremourning 

> edit: I was wrong, the multiple includes thing totally does change, i'm just a fool.

What??

puremourning:
Oh yeah, oops. I replied saying that the change didn't actually change what the test was testing, but then 2 minutes later i realised i was being really dumb so edited it! *rolleyes*

PR #592, opened 2016-09-04
[READY] Make the do_cache item returned by FlagsForFile optional

micbou:
The `do_cache` item returned by the `FlagsForFile` function in `.ycm_extra_conf.py` is generally always set to `True` so let's make it optional with a default value of `True`. This simplifies a little the `.ycm_extra_conf.py` file and should be less confusing for users trying to understand it (cache + filename + flags? Sorry, can't compute).

This change is backward compatible.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/592)

### DIFF
--- a/cpp/ycm/.ycm_extra_conf.py
+++ b/cpp/ycm/.ycm_extra_conf.py
@@ -182,7 +182,4 @@ def FlagsForFile( filename, **kwargs ):
     relative_to = DirectoryOfThisScript()
     final_flags = MakeRelativePathsInFlagsAbsolute( flags, relative_to )
 
-  return {
-    'flags': final_flags,
-    'do_cache': True
-  }
+  return { 'flags': final_flags }
--- a/examples/.ycm_extra_conf.py
+++ b/examples/.ycm_extra_conf.py
@@ -139,8 +139,4 @@ def FlagsForFile( filename, **kwargs ):
     relative_to = DirectoryOfThisScript()
     final_flags = MakeRelativePathsInFlagsAbsolute( flags, relative_to )
 
-  return {
-    'flags': final_flags,
-    'do_cache': True
-  }
-
+  return { 'flags': final_flags }
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -104,7 +104,7 @@ class Flags( object ):
                                               filename,
                                               add_extra_clang_flags )
 
-      if results[ 'do_cache' ]:
+      if results.get( 'do_cache', True ):
         self.flags_for_file[ filename ] = sanitized_flags
       return sanitized_flags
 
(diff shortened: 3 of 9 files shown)


puremourning:
:lgtm:

fantastic, thanks!

---

Reviewed 9 of 9 files at r1.

---

*[ycmd/completers/cpp/flags.py, line 107 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/592#-KQq30shlQepGgtpVRTO:-KQq30shlQepGgtpVRTP:b-a79x43) ([raw file](https://github.com/valloric/ycmd/blob/5a998f8441775fa26f178c69f03a650ba6acef81/ycmd/completers/cpp/flags.py#L107)):*
> ```Python
>                                               add_extra_clang_flags )
> 
>       if 'do_cache' not in results or results[ 'do_cache' ]:
> ```

Elsewhere we used `results.get( 'do_cache', True )`. I don't feel the need to change this, just an observation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/592#-:-KQq3JpDcTb49ecG-ej9:b-rk64qu)*

micbou:
Reviewed 9 of 9 files at r1, 1 of 1 files at r2.

---

*[ycmd/completers/cpp/flags.py, line 107 \[r1\]](https://reviewable.io:443/reviews/valloric/ycmd/592#-KQq30shlQepGgtpVRTO:-KQtB5P1UbHgGsdTAQ5r:b-pw7xam) ([raw file](https://github.com/valloric/ycmd/blob/5a998f8441775fa26f178c69f03a650ba6acef81/ycmd/completers/cpp/flags.py#L107)):*
> Elsewhere we used `results.get( 'do_cache', True )`. I don't feel the need to change this, just an observation.
</details>
This is shorter and more Pythonic so let's write it this way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/592)*

Valloric:
Nice! 

@homu r+

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/592#-:-KQuocUlq68c1T7zXM40:b-l4266t)*

DanielArndt:
@micbou @Valloric Just as a heads up that if a user follows the current completion instructions for cpp, and follows the link for [See YCM's own `.ycm_extra_conf.py`](https://github.com/Valloric/ycmd/blob/master/cpp/ycm/.ycm_extra_conf.py) the user is taken to this version of the `.ycm_extra_conf.py` which is incompatible with the submodule version referenced in YouCompleteMe (currently `ycmd @ b3643ce`).

This might cause some "false" bug reports of where users are getting `KeyError: 'do_cache'` -- when they really should use the `.ycm_extra_conf.py` version @ `b3643ce` (until the submodule is updated).

If this is worthy of an issue (YouCompleteMe or ycmd) let me know. It seemed more like a (subtle) user error to me, but had me doing some digging.

micbou:
Thanks. I've sent [a PR updating the ycmd submodule](https://github.com/Valloric/YouCompleteMe/pull/2318).

PR #594, opened 2016-09-08
[READY] Fix shutdown request in example client

micbou:
When shutting down the server in the example client, we get the following traceback:
```python
  File "./examples/example_client.py", line 494, in <module>
    Main()
  File "./examples/example_client.py", line 490, in Main
    server.Shutdown()
  File "./examples/example_client.py", line 116, in Shutdown
    self.PostToHandlerAndLog( 'shutdown' )
  File "./examples/example_client.py", line 120, in PostToHandlerAndLog
    self._CallHttpie( 'post', handler, data )
  File "./examples/example_client.py", line 249, in _CallHttpie
    data )
  File "./examples/example_client.py", line 220, in _HmacForRequest
    self._hmac_secret ) ), 'utf8' )
  File "./examples/example_client.py", line 276, in CreateRequestHmac
    body = ToBytes( body )
  File "./examples/example_client.py", line 266, in ToBytes
    return bytes( value, encoding = 'utf-8' )
TypeError: encoding or errors without a string argument
```
It happens because we are trying to convert a `None` body to bytes to create the HMAC key while preparing the shutdown request. Handle this case by returning an empty bytes (like we do in [ycmd ToBytes](https://github.com/Valloric/ycmd/blob/master/ycmd/utils.py#L103)).

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/594)

### DIFF
--- a/examples/example_client.py
+++ b/examples/example_client.py
@@ -257,6 +257,8 @@ class YcmdHandle( object ):
 
 
 def ToBytes( value ):
+  if not value:
+    return bytes()
   if isinstance( value, bytes ):
     return value
   if isinstance( value, int ):


puremourning:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/594#-:-KRAdX8lDjdWpE5TxR4Y:bnfp4nl)*

vheon:
:lgtm: @homu r=puremourning 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/594#-:-KRFdDkuOA3BSxPyomid:bpw2qui)*

PR #598, opened 2016-09-12
Add VSCode youcompleteme

richard1122:
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/598)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Known ycmd clients:
 - [you-complete-me][atom-you-complete-me]: Atom client.
 - [YcmdCompletion][sublime-ycmd]: Sublime client
 - [kak-ycmd][]: Kakoune client.
+- [you-complete-me][vscode-you-complete-me]: VSCode client.
 
 Feel free to send a pull request adding a link to your client here if you've
 built one.
@@ -321,3 +322,4 @@ This software is licensed under the [GPL v3 license][gpl].
 [dev-setup]: https://github.com/Valloric/ycmd/blob/master/DEV_SETUP.md
 [test-setup]: https://github.com/Valloric/ycmd/blob/master/TESTS.md
 [extra-conf-vim-data-doc]: https://github.com/Valloric/YouCompleteMe#the-gycm_extra_conf_vim_data-option
+[vscode-you-complete-me]: https://marketplace.visualstudio.com/items?itemName=RichardHe.you-complete-me


richard1122:
And What's the appropriate API to call if I want to get method(signature help)? Is `GetDoc`?

What's the correct parameters of GetDoc? Is it the same as `GetType`, `GoToDefinition`?

Thanks.

vheon:
> What's the correct parameters of GetDoc? Is it the same as GetType, GoToDefinition?

Since is a subcommand as anyone else it gets the same parameters. Did you tried and it didn't work?

vheon:
I don't know if you knew this: http://puremourning.github.io/ycmd-1/#path--run_completer_command

Is a WIP documentation for the API (thanks to @puremourning) 👍

richard1122:
Thanks @puremourning @vheon , it helps a lot.

richard1122:
> Did you tried and it didn't work?
@vheon When I'm trying to get doc from the following code:
image

cursor position is after the `begin(`, It reports 
```
raise ValueError( NO_DOCUMENTATION_MESSAGE )\\nValueError: No documentation available for current context\\n\", \"message\": \"No documentation available for current context\"}"
```

It's also same for other functions like `vector.push_back` and more...

I'll read that doc to see If I missed anything.

vheon:
What happen if you instead are with the cursor in the middle of `begin`?

richard1122:
@vheon It's the same...

richard1122:
I'm not sure is it relative to platform? I'm using win10 with mingw64 header files.

vheon:
Now that I think about it IIRC the `GetDoc` subcommand returns the documentation that libclang extract from the source file, like doxygen. So unless the header file are documented in that way I don't think that will work. I just tried on my mac and I get the same results `ValueError: No documentation available for current context`

richard1122:
Well, I guess you're right. I'll have a try on a linux virtual machine. 
Thank you very much!

puremourning:
The c-familhy language completer returns that error when libclang returns no 'documentation' associated. @vheon is spot on that if the method signature doesn't have *doxygen* comments associated with it, then libclang returns nothing.
I've been thinking that we could *at least* try and return the signature (or type) info in that case, but it is surprisingly tricky to do.

puremourning:
Wait correction, that's not the whole truth. We actually throw that error only if we can't parse the XML version. It's possible to return something meaningful anyway, so I will take a look at doing that (this has been bugging me too)

Valloric:
I see discussion going on here, but it's not related to the PR, right? Can we merge this?

richard1122:
Yes. Thanks.
And I will test get doc on linux.

micbou:
A new client supporting ycmd. I am all for it.

:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/598#-:-KRSvflfkA20OhKDTdxv:bnnqn0f)*

Valloric:
Always happy to see a new client! :)

@homu r=micbou

PR #601, opened 2016-09-16
[READY][API] Return debugging information as a dictionary instead of a string

micbou:
Currently, the data returned by the `debug_info` handler is a block of text. This makes it difficult for clients to retrieve a specific information on the server (e.g. the path of the logfiles, the process identifier of a sub-server, etc.) or to customize the way it is displayed to the user.

We improve this by returning a dictionary instead of a string. Its structure is fully explained in [the new `debug_info` API documentation](http://micbou.github.io/ycmd/#path--debug_info).

There are two important changes on its content:
- add information on the Python interpreter used to start the ycmd server: its path and its version.
- move information on the `.ycm_extra_conf.py` file (its path and if it is loaded or not) from the Clang completer to ycmd itself since it is not supposed to be specific to this completer.

This is obviously an API-breaking change but not a big one since it only affects the `debug_info` handler.

@abingham @Qusic @LuckyGeck @mawww @richard1122

---

This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/601)

### DIFF
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -1,4 +1,5 @@
-# Copyright (C) 2011, 2012 Google Inc.
+# Copyright (C) 2011-2012 Google Inc.
+#               2017      ycmd contributors
 #
 # This file is part of ycmd.
 #
@@ -30,7 +31,6 @@ import re
 import os.path
 import textwrap
 from ycmd import responses
-from ycmd import extra_conf_store
 from ycmd.utils import ToCppStringCompatible, ToUnicode
 from ycmd.completers.completer import Completer
 from ycmd.completers.completer_utils import GetIncludeStatementValue
@@ -371,52 +371,31 @@ class ClangCompleter( Completer ):
 
 
   def DebugInfo( self, request_data ):
-    filename = request_data[ 'filepath' ]
-    try:
-      extra_conf = extra_conf_store.ModuleFileForSourceFile( filename )
-    except UnknownExtraConf as error:
-      return ( 'C-family completer debug information:\n'
-               '  Configuration file found but not loaded\n'
-               '  Configuration path: {0}'.format(
-                 error.extra_conf_file ) )
-
     try:
       # Note that it only raises NoExtraConfDetected:
       #  - when extra_conf is None and,
       #  - there is no compilation database
-      flags = self._FlagsForRequest( request_data )
-    except NoExtraConfDetected:
-      # No flags
-      return ( 'C-family completer debug information:\n'
-               '  No configuration file found\n'
-               '  No compilation database found' )
-
-    # If _FlagsForRequest returns None or raises, we use an empty list in
-    # practice.
-    flags = flags or []
-
-    if extra_conf:
-      # We got the flags from the extra conf file
-      return ( 'C-family completer debug information:\n'
-               '  Configuration file found and loaded\n'
-               '  Configuration path: {0}\n'
-               '  Flags: {1}'.format( extra_conf, list( flags ) ) )
+      flags = self._FlagsForRequest( request_data ) or []
+    except ( NoExtraConfDetected, UnknownExtraConf ):
+      # If _FlagsForRequest returns None or raises, we use an empty list in
+      # practice.
+      flags = []
 
     try:
-      database = self._flags.FindCompilationDatabase(
-          os.path.dirname( filename ) )
+      database_directory = self._flags.FindCompilationDatabase(
+          os.path.dirname( request_data[ 'filepath' ] ) ).database_directory
     except NoCompilationDatabase:
-      # No flags
-      return ( 'C-family completer debug information:\n'
-               '  No configuration file found\n'
-               '  No compilation database found' )
+      database_directory = None
 
-    # We got the flags from the compilation database
-    return ( 'C-family completer debug information:\n'
-             '  No configuration file found\n'
-             '  Using compilation database from: {0}\n'
-             '  Flags: {1}'.format( database.database_directory,
-                                    list( flags ) ) )
+    database_item = responses.DebugInfoItem(
+      key = 'compilation database path',
+      value = '{0}'.format( database_directory ) )
+    flags_item = responses.DebugInfoItem(
+      key = 'flags', value = '{0}'.format( list( flags ) ) )
+
+    return responses.BuildDebugInfoResponse( name = 'C-family',
+                                             items = [ database_item,
+                                                       flags_item ] )
 
 
   def _FlagsForRequest( self, request_data ):
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -1,5 +1,6 @@
-# Copyright (C) 2011, 2012 Chiel ten Brinke <ctenbrinke@gmail.com>
-#                          Google Inc.
+# Copyright (C) 2011-2012 Chiel ten Brinke <ctenbrinke@gmail.com>
+#                         Google Inc.
+#               2017      ycmd contributors
 #
 # This file is part of ycmd.
 #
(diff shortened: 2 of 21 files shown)


vheon:
Review status: 0 of 21 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/tests/clang/debug_info_test.py, line 67 at r1](https://reviewable.io:443/reviews/valloric/ycmd/601#-KRnwkUocj95LwMAhab5:-KRnwkUocj95LwMAhab6:b-xhgulj) ([raw file](https://github.com/valloric/ycmd/blob/909175ebe2ce2610337eb0af285d4b69bd423024/ycmd/tests/clang/debug_info_test.py#L67)):*
> ```Python
>     } ) )
>   )
>   # Second time, an empty list of flags is returned.
> ```

isn't this the same response that we have above? I think that since you moved the extra_conf stuff to be always returned we should at least rename the test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601)*

vheon:
Review status: 0 of 21 files reviewed at latest revision, 2 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/601#-KRnzYmJsSlxf7w-DFF8:-KRnzYmKXjNyZmKEq6PX:b8r0kep) (no related file):*
I understand that now we return all the information all the time since debug info is now a structure but before, at least in the typescript, rust, python, javascript tests, we were actually testing that also the values had a meaning: for example here we were testing that the string reporting the status of the server were actually correct (running/not running) while now we check only the structure of the `debug_info` handle. Is this something worth keeping?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601)*

micbou:
Reviewed 20 of 21 files at r1, 1 of 1 files at r2.

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/601#-KRnzYmJsSlxf7w-DFF8:-KRoRxIJWEm9pXm_cGol:bdggsdb) (no related file):*
> I understand that now we return all the information all the time since debug info is now a structure but before, at least in the typescript, rust, python, javascript tests, we were actually testing that also the values had a meaning: for example here we were testing that the string reporting the status of the server were actually correct (running/not running) while now we check only the structure of the `debug_info` handle. Is this something worth keeping?
</details>
I think it is still useful to check if the `debug_info[ 'completer' ]` structure is correct. Also, without those tests, the `DebugInfo` method in the completers would not be covered.

---

*[ycmd/tests/clang/debug_info_test.py, line 67 at r1](https://reviewable.io:443/reviews/valloric/ycmd/601#-KRnwkUocj95LwMAhab5:-KRoTh8PmYMUGLR2ry_W:b-4j0r14) ([raw file](https://github.com/valloric/ycmd/blob/909175ebe2ce2610337eb0af285d4b69bd423024/ycmd/tests/clang/debug_info_test.py#L67)):*
> isn't this the same response that we have above? I think that since you moved the extra_conf stuff to be always returned we should at least rename the test.
</details>
Yes, this is the same response but the executed code is different. When sending the first request, the `FlagsForFile` function raises a `NoExtraConfDetected` exception that we catch in `DebugInfo`. Second request, the `FlagsForFile` function returns `None`. We handle both cases by setting the flags to an empty list.

I updated and renamed the tests to make it clear that we are checking the flags.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601)*

vheon:

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/601#-KRnzYmJsSlxf7w-DFF8:-KRod2IOPqESK3zWPOjK:bvdkhta) (no related file):*
> I think it is still useful to check if the `debug_info[ 'completer' ]` structure is correct. Also, without those tests, the `DebugInfo` method in the completers would not be covered.
</details>
I'm not saying that we should not check the structure, I'm saying that I'm feeling that we're loosing some coverage because we're not actually checking behaviour anymore because all we check is that the structure is the right one and that's it. But I think that is just a feeling because there is no actual behaviour to test anymore since we have removed all the `if` statement in the `DebugInfo` method :confused: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601)*

micbou:

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/601#-KRnzYmJsSlxf7w-DFF8:-KRof7tjRFd4AVvP22li:b-weq3sp) (no related file):*
> I'm not saying that we should not check the structure, I'm saying that I'm feeling that we're loosing some coverage because we're not actually checking behaviour anymore because all we check is that the structure is the right one and that's it. But I think that is just a feeling because there is no actual behaviour to test anymore since we have removed all the `if` statement in the `DebugInfo` method :confused: 
</details>
Yes, this is why I removed the tests. Now that there is only one branch in `DebugInfo` (except for the Clang and C# completers, which have more tests), we only need one test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601)*

Valloric:

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/601#-KRnzYmJsSlxf7w-DFF8:-KRooI6JDGw5hqkI-uAE:b-n180dh) (no related file):*
> Yes, this is why I removed the tests. Now that there is only one branch in `DebugInfo` (except for the Clang and C# completers, which have more tests), we only need one test.
</details>
:lgtm: Looks nice!

I keep seeing these draft API docs; they look really sweet! When can we see some PRs for that? :) 

---

*[ycmd/handlers.py, line 234 at r2](https://reviewable.io:443/reviews/valloric/ycmd/601#-KRoo30jKZuRAm8l-MTL:-KRoo30jKZuRAm8l-MTM:bmb54fe) ([raw file](https://github.com/valloric/ycmd/blob/fd756a380330166e7601b5a84fc46b1f5ab9a14c/ycmd/handlers.py#L234)):*
> ```Python
>     'extra_conf': {
>       'path': extra_conf_path,
>       'loaded': loaded
> ```

I'd rather make this something like `is_loaded`. More obvious it's a bool.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601)*

micbou:

---

*[ycmd/handlers.py, line 234 at r2](https://reviewable.io:443/reviews/valloric/ycmd/601#-KRoo30jKZuRAm8l-MTL:-KRopYuCi3ehQdsL5Bov:b-njqt2) ([raw file](https://github.com/valloric/ycmd/blob/fd756a380330166e7601b5a84fc46b1f5ab9a14c/ycmd/handlers.py#L234)):*
> I'd rather make this something like `is_loaded`. More obvious it's a bool.
</details>
I'll change it. Should we replace `support` by `has_support` in Clang info?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601)*

Valloric:
Review status: 19 of 21 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/handlers.py, line 234 at r2](https://reviewable.io:443/reviews/valloric/ycmd/601#-KRoo30jKZuRAm8l-MTL:-KRtcOSC7HrpuAxWP1-d:b-llc7ev) ([raw file](https://github.com/valloric/ycmd/blob/fd756a380330166e7601b5a84fc46b1f5ab9a14c/ycmd/handlers.py#L234)):*
> I'll change it. Should we replace `support` by `has_support` in Clang info?
</details>
Yep, that sounds like a good change too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601)*

micbou:
Reviewed 2 of 2 files at r3.

---

*[ycmd/handlers.py, line 234 at r2](https://reviewable.io:443/reviews/valloric/ycmd/601#-KRoo30jKZuRAm8l-MTL:-KRtnlKETjzK8TytOK7-:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/fd756a380330166e7601b5a84fc46b1f5ab9a14c/ycmd/handlers.py#L234)):*
> Yep, that sounds like a good change too.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601)*

vheon:
@micbou What is blocking this PR?

micbou:
@vheon Nothing. I am going to fix the conflicts then I'll mark it as ready.

micbou:
Updated with minor changes:
 - use [`platform.python_version()`](https://docs.python.org/2/library/platform.html#platform.python_version) instead of [`sys.version`](https://docs.python.org/2/library/sys.html#sys.version) to get the Python version; `sys.version` returns additional information that's not really useful (date of compilation and compiler used);
 - rename `description` to `key` in `DebugInfoItem` object.

[API documentation](http://micbou.github.io/ycmd/#path--debug_info) updated as well.

PR is now ready.

---

Reviewed 8 of 8 files at r4, 12 of 12 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601#-:-KVGZay3RcRpfQdXYzO-:b-idnw4l)*

Valloric:
> platform.python_version()

I think we used to use that at some point but moved away from it. Can't really remember why. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601#-:-KVGofvOg5TZc4jVYKvg:b-or7u3i)*

micbou:
I looked at [the Python sources](https://github.com/python/cpython/blob/master/Lib/platform.py#L1234) and this function simply parses the string returned by `sys.version` so it should be completely safe.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601#-:-KVGrSAqq6NAVV-IZAGz:b-vyb7es)*

Valloric:
@micbou I think it might be related to the Python interpreter embedded inside Vim. Especially for the Windows Vim version.

Anyway, if you're confident this is solid, I'm fine with it. Can't really provide an argument based on vague memory. :)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601#-:-KVIHZj9viJuSBKQPQaT:b-txz30q)*

micbou:
If this is related to Vim then we shouldn't worry since we are doing that in the ycmd server.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601#-:-KVK_ic_7sJs1nkkVO-n:b-vm11nw)*

micbou:
I've fixed the conflicts. I think we can merge this PR if nobody is against it. This new API is really convenient for adding more debugging information in completers which I am planning to do for the Python completer (it will be most likely information on the project root and the preloaded modules). In addition, this makes information about the `.ycm_extra_conf.py` file not tied anymore to the Clang completer but to ycmd itself so I will not have to add this information in the Python completer.

---

Reviewed 3 of 3 files at r6, 2 of 2 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601#-:-KX1Lc3vkzcnD1sbj1MT:b-chr3du)*

Valloric:
Still a fan of this PR. :)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601#-:-KX8bw7cR9vw_QCIqocV:b3jka6d)*

micbou:
Fixed the conflicts and updated the copyrights (new year).

---

Reviewed 2 of 2 files at r8, 21 of 21 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601#-:-KaSd9zd9u4NWUbjOK-7:b-l2cfke)*

vheon:
:lgtm:

---

Reviewed 21 of 21 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601#-:-KaSqhE5IIE0TXwXXvEu:bnfp4nl)*

Valloric:

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/601#-KRnzYmJsSlxf7w-DFF8:-KaiJC-lzRRWIdANMCim:b-7ennmq) (no related file):*

:lgtm: Looks nice!

I keep seeing these draft API docs; they look really sweet! When can we see some PRs for that? :) 
</blockquote></details>

Let's go! :)

@homu r+

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/601)*

PR #602, opened 2016-09-17
only skip flag that is actually a file

comicfans:
this improves compatibility with msvc-cl style compile flags

---

This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/602)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -306,7 +306,8 @@ def _RemoveUnusedFlags( flags, filename ):
     # remove that.
     if ( not current_flag_starts_with_dash and
           ( not previous_flag_starts_with_dash or
-            ( not previous_flag_is_include and '/' in flag ) ) ):
+            ( not previous_flag_is_include and '/' in flag ) ) 
+          and (os.path.realpath(flag) == flag ) ):
       continue
 
     new_flags.append( flag )


comicfans:
Hi, I'm trying to make YouCompleteMe works with MSVC style arguments better,  these arguments starts with '/' but currently YouCompleteMe thinks these're file flag and throws them, this patch add a addition check to make sure arguments to throw is actually a file argument ,not MSVC style argument

puremourning:
Thanks for sending a PR!

We will need tests updated for this. I don't think we have ever tested MSVC style flags and tbh I didn't realise clang accepted them. 

So I think this is a larger job :/

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/602#-:-KRsYqHp5Ppz9sAiJC2O:b-ic73lr)*

vheon:
I'm not sure I understood the problem, but this for sure needs a test to be written.

micbou:
Libclang only accepts MSVC style flags when the driver mode is set to `cl` (`--driver-mode=cl` flag). So, I think a better approach would be to look at the `--driver-mode` flag and accept MSVC flags (flags starting with a `/`) only if its value is `cl`.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/602#-:-KRsZonrlyr3hmOS2Bec:bo5g64e)*

Valloric:
Agreed with @micbou. Accepting CL-style flags is not a common case for clang. The drawback of looking for `--driver-mode` is that it might change in the future, although that's probably unlikely.

Either way, extensive tests are necessary.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/602#-:-KRtd4LWUlRIGH_2HQYW:bccilha)*

comicfans:
I've created a modified [msbuild](https://github.com/comicfans/msbuild) to generate compile_commands.json for existing MSVC project and adds --driver-mode=cl to every compile commands,  makes this work .
I think even looking for ```--driver-mode```,  we still need to determine if flag is file or not (cl mode with unix-like file path),so this condition check may be still needed.

micbou:

---

*[ycmd/completers/cpp/flags.py, line 312 at r1](https://reviewable.io:443/reviews/valloric/ycmd/602#-KRwNR-FrYN3n2pDEJxH:-KRwNR-FrYN3n2pDEJxI:b84rkx7) ([raw file](https://github.com/valloric/ycmd/blob/c269a21d1fa61311f62c94da7001d76741d021b9/ycmd/completers/cpp/flags.py#L312)):*
> ```Python
>           ( not previous_flag_starts_with_dash or
>             ( not previous_flag_is_include and '/' in flag ) ) 
>           and is_file):
> ```

This is wrong. We are supposed to ignore any file, not just the one currently edited.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/602)*

comicfans:
is_file does not mean the one currently edited, but every flag in flags iterates.

micbou:
In `_RemoveUnusedFlags` function, `filename` is the file being passed to Clang so `is_file` is `True` if the flag points to the same path as the current file, `False` otherwise. What you want (at least if I understand what you are trying to achieve) is to be `True` for any file. That's not the same thing.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/602#-:-KRwXLRT030VR067ppQP:b-9pd5mh)*

comicfans:
Yes you're right, I misuse this. what I want to achiveve is make sure it to be true for any file. so should I correct this , or waiting YCM implements ```--driver-mode=cl``` logic ?

micbou:
Closed in favor of https://github.com/Valloric/ycmd/pull/789.

PR #609, opened 2016-09-22
[READY] Support GetTypeQuick completer subcommand in clang completer

puremourning:
This sacrifices accuracy for speed, so that it can be used from, say, a vim auto command.

I've had this for a _long_ time in my fork, and a number of people have requested PR for it, so here it is.

I know there was some discussion that is should be GetTypeImprecise, but personally I think that's a bit of a mouthful (fistful?) to type... Open to thoughts, though all my bindings are to `GetTypeQuick` so I am totally biassed.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/609)

### DIFF
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -131,6 +131,10 @@ class ClangCompleter( Completer ):
          self._ClearCompilationFlagCache() ),
       'GetType'                  : ( lambda self, request_data, args:
          self._GetSemanticInfo( request_data, func = 'GetTypeAtLocation' ) ),
+      'GetTypeImprecise'         : ( lambda self, request_data, args:
+         self._GetSemanticInfo( request_data,
+                                func = 'GetTypeAtLocation',
+                                reparse = False ) ),
       'GetParent'                : ( lambda self, request_data, args:
          self._GetSemanticInfo( request_data,
                                 func = 'GetEnclosingFunctionAtLocation' ) ),
--- a/ycmd/tests/clang/subcommands_test.py
+++ b/ycmd/tests/clang/subcommands_test.py
@@ -385,6 +385,14 @@ def Subcommands_GetType_test():
             test,
             [ 'GetType' ] )
 
+  # For every practical scenario, GetTypeImprecise is the same as GetType (it
+  # just skips the reparse)
+  for test in tests:
+    yield ( RunGetSemanticTest,
+            'GetType_Clang_test.cc',
+            test,
+            [ 'GetTypeImprecise' ] )
+
 
 def Subcommands_GetParent_test():
   tests = [


vheon:
I don't have a strong opinion around the name, especially because a command like this will be used through a mapping by whatever client uses ycmd. :lgtm: 

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/609#-:-KSIoaUwoOf1xXQs8ZeU:b-9zwxe)*

Valloric:
It has to be `GetTypeImprecise` instead of `GetTypeQuick` because we already have `GoToImprecise` and we should be consistent. The *Imprecise "modifier" stands for "without reparsing" and I don't want to have another way of saying the same thing. We could in theory change `GoToImprecise` to `GoToQuick` but that would be a breaking change (unless we made the old name an alias) and probably not worth the hassle.

So let's just call it `GetTypeImprecise`.

With that, :lgtm: 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/609#-:-KSJQ1UZYevhjm4Dtf4d:bo0q400)*

puremourning:
OK, I will do this, but I would feel bad if i didn't give my two further closing arguments (your honour!)  ...  :)

- "Imprecise" makes sense for a `GoTo` command, because the result might not be quite in the right location (might be off by a line or two, but otherwise probably useful). `GetTypeImprecise` on the other hand implies that the type returned might not be completely correct (e.g. say missing a `const` qualifier or something). That's not the actual behaviour: it's more likely to get a completely incompatible type, or no type at all. I confess that `GetTypeQuick` doesn't suggest that either though...
- We already have `GetDocQuick` which is more similar to `GetType` in terms of behaviour

Valloric:
> We already have GetDocQuick which is more similar to GetType in terms of behaviour

We do? Grrr. 

Well it's inconsistent already, so proceed with GetTypeQuick if you wish.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/609#-:-KSc_LmgfxuJCt_qWXxB:bhxezuv)*

puremourning:
Heh, i already pushed a change to make it `GetTypeImprecise`  :) TBH I think you're probably right and we should be consistent...

agauniyal:
Is there anything stopping this PR to get merged, I would love to get this feature since it would allow me to use vscode plugin with ycm 😄

micbou:
We'll need to rename the `GetDocQuick` command to `GetDocImprecise` too for consistency but we can do this in another PR. Also, this new command needs to be documented.

@homu r+

---

Reviewed 2 of 2 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/609#-:-KT4Of2ak8q1a-8B4mUr:bbi5p0s)*

PR #623, opened 2016-10-10
[READY] Do not compile ycm_extra_conf.py files

### LINKED ISSUE #618: Processing .ycm_extra_conf.py creates __pycache__ directory
I'm not sure if this is the intended behaviour. When YCM reads configuration creates a compiled version in `__pycache__`. I know that this behaviour can be disabled passing to `python` the `-B` argument or setting `PYTHONDONTWRITEBYTECODE=1` environmental variable. I don't want to disable global bytecode generation but I want to disable for `.ycm_extra_conf.py` because I feel it pollutes my project directory. 

Is there a easy/reliable way to disable it in the YCM config?

micbou:
See issue #618 and the comment.

I've verified that it does not affect the ycmd sources as we care a lot about ycmd loading speed. They are still compiled with this change.

Resolves #618.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/623)

### DIFF
--- a/ycmd/extra_conf_store.py
+++ b/ycmd/extra_conf_store.py
@@ -146,7 +146,21 @@ def Load( module_file, force = False ):
   # anymore, but there are a lot of old ycm_extra_conf.py files that we don't
   # want to break.
   sys.path.insert( 0, _PathToCppCompleterFolder() )
-  module = LoadPythonSource( _RandomName(), module_file )
+
+  # By default, the Python interpreter compiles source files into bytecode to
+  # load them faster next time they are run. These *.pyc files are generated
+  # along the source files prior to Python 3.2 or in a __pycache__ folder for
+  # newer versions. We disable the generation of these files when loading
+  # ycm_extra_conf.py files as users do not want them inside their projects.
+  # The drawback is negligible since ycm_extra_conf.py files are generally small
+  # files thus really fast to compile and only loaded once by editing session.
+  old_dont_write_bytecode = sys.dont_write_bytecode
+  sys.dont_write_bytecode = True
+  try:
+    module = LoadPythonSource( _RandomName(), module_file )
+  finally:
+    sys.dont_write_bytecode = old_dont_write_bytecode
+
   del sys.path[ 0 ]
 
   with _module_for_module_file_lock:


vheon:

---

*[ycmd/extra_conf_store.py, line 159 at r1](https://reviewable.io:443/reviews/valloric/ycmd/623#-KTjeICVtXjnOxtWU31W:-KTjeICWj8LjuvO4VSip:b-k9xndt) ([raw file](https://github.com/valloric/ycmd/blob/c6b2daad33b9591382170b1f84cc91a6bd75c053/ycmd/extra_conf_store.py#L159)):*
> ```Python
>   sys.dont_write_bytecode = True
>   module = LoadPythonSource( _RandomName(), module_file )
>   sys.dont_write_bytecode = False
> ```

Isn't possible that `sys.dont_write_bytecode` was `True` to begin with? If so we should get the original value and restore it here, otherwise nevermind :stuck_out_tongue_closed_eyes: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/623)*

micbou:

---

*[ycmd/extra_conf_store.py, line 159 at r1](https://reviewable.io:443/reviews/valloric/ycmd/623#-KTjeICVtXjnOxtWU31W:-KTjogTTl2tj-Bkhm0bE:bvo2qzw) ([raw file](https://github.com/valloric/ycmd/blob/c6b2daad33b9591382170b1f84cc91a6bd75c053/ycmd/extra_conf_store.py#L159)):*
> Isn't possible that `sys.dont_write_bytecode` was `True` to begin with? If so we should get the original value and restore it here, otherwise nevermind :stuck_out_tongue_closed_eyes: 
</details>
You are absolutely right. It can be `True` if Python is started with the `-B` parameter or the `PYTHONDONTWRITEBYTECODE` environment variable is set. See [the `sys.dont_write_bytecode` documentation](https://docs.python.org/3/library/sys.html#sys.dont_write_bytecode). I've changed it to restore the old value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/623)*

vheon:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/623#-:-KTkI2mylvAWjNf6Ea9z:bnfp4nl)*

Valloric:

---

*[ycmd/extra_conf_store.py, line 159 at r1](https://reviewable.io:443/reviews/valloric/ycmd/623#-KTjeICVtXjnOxtWU31W:-KTkc4qk5R0Tj58YBQ71:bvwnhgl) ([raw file](https://github.com/valloric/ycmd/blob/c6b2daad33b9591382170b1f84cc91a6bd75c053/ycmd/extra_conf_store.py#L159)):*
> You are absolutely right. It can be `True` if Python is started with the `-B` parameter or the `PYTHONDONTWRITEBYTECODE` environment variable is set. See [the `sys.dont_write_bytecode` documentation](https://docs.python.org/3/library/sys.html#sys.dont_write_bytecode). I've changed it to restore the old value.
</details>
This should use try-finally so that we still reset dont_write_bytecode in the face of exceptions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/623)*

micbou:
Reviewed 1 of 1 files at r3.

---

*[ycmd/extra_conf_store.py, line 159 at r1](https://reviewable.io:443/reviews/valloric/ycmd/623#-KTjeICVtXjnOxtWU31W:-KTkg78JmG0WAm1H2e3g:b-7uzlr4) ([raw file](https://github.com/valloric/ycmd/blob/c6b2daad33b9591382170b1f84cc91a6bd75c053/ycmd/extra_conf_store.py#L159)):*
> This should use try-finally so that we still reset dont_write_bytecode in the face of exceptions.
</details>
I didn't think of that. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/623)*

Valloric:
Thanks!

@homu r=vheon

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/623#-:-KTq9W6UAc8Pfps4JlkQ:b-fef2al)*

PR #626, opened 2016-10-15
Allow using tsserver with JavaScript

dannysu:
TypeScript's ["Salsa" JavaScript language service](https://github.com/Microsoft/TypeScript/wiki/Salsa) allows tsserver to be used with JavaScript files as well. tsserver starts to work much faster than tern.js.

This change uses tsserver (if it exists) as the fallback when tern.js isn't enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/626)

### DIFF
--- a/ycmd/completers/javascript/hook.py
+++ b/ycmd/completers/javascript/hook.py
@@ -26,9 +26,15 @@ from builtins import *  # noqa
 from ycmd.completers.javascript.tern_completer import (
   ShouldEnableTernCompleter, TernCompleter )
 
+from ycmd.completers.typescript.typescript_completer import (
+  ShouldEnableTypescriptCompleter, TypeScriptCompleter )
+
 
 def GetCompleter( user_options ):
-  if not ShouldEnableTernCompleter():
-    return None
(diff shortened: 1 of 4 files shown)


vheon:
I don't personally like having more than one semantic engine for a single language for a couple of reasons:

- How do you let decide the user which engine to use? This usually means that we have to add more configuration options to ycmd and I don't like that.

> This change uses tsserver (if it exists) as the fallback when tern.js isn't enabled.

If you do this you are assuming that the completion and services that the TSServer provides for JavaScript is better than what Tern.js is providing. Do you have proof of that or is just an assumption?

Moreover is it feature parity with Tern.js?

dannysu:
I'll close this PR. TSServer seemed to work well, but upon further experiment with my projects once I got the plugin working showed that it is less useful than tern.js at this time.

vheon:
@dannysu Thanks for PR anyway! Personally I don't use JavaScript so I don't follow the tools around it that much, but if you do and find that one tools show up to be better than the one we use as a default you can always let us know 👍

PR #627, opened 2016-10-17
add .ycm_extra_conf_advanced.py

ObiWahn:
Hi,

I have created a slightly modified version of your extra conf. I hope you like the some of the changes and tell me what I need to to do to get them integrated.

Regards

Jan

---

This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/627)

### DIFF
--- a/cpp/ycm/.ycm_extra_conf_advanced.py
+++ b/cpp/ycm/.ycm_extra_conf_advanced.py
@@ -115,6 +115,8 @@ def flags_for_closest_include(filename):
 
 def find_database(filename):
     compilation_db_path = find_closest_path(filename, 'compile_commands.json')
+    if not compilation_db_path:
+        return None
     compilation_db_dir = os.path.dirname(compilation_db_path)
     logging.info("Set compilation database directory to " + compilation_db_dir)
     compilation_db =  ycm_core.CompilationDatabase(compilation_db_dir)
--- a/cpp/ycm/.ycm_extra_conf_advanced.py
+++ b/cpp/ycm/.ycm_extra_conf_advanced.py
@@ -1,0 +1,276 @@
+#!/usr/bin/python
+# This file is NOT licensed under the GPLv3, which is the license for the rest
+# of YouCompleteMe.
+#
+# Here's the license text for this file:
+#
+# This is free and unencumbered software released into the public domain.
+#
+# Anyone is free to copy, modify, publish, use, compile, sell, or
+# distribute this software, either in source code form or as a compiled
+# binary, for any purpose, commercial or non-commercial, and by any
+# means.
+#
+# In jurisdictions that recognize copyright laws, the author or authors
+# of this software dedicate any and all copyright interest in the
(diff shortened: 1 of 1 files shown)


Valloric:
Thanks for sending a PR!

What are the benefits of your new extra conf file? Note that the `.ycm_extra_conf.py` file in ycmd is ycmd's own extra conf file; it just so happens to be an example file as well. We don't really want to add more extra conf files here.

puremourning:
I'm also curious as to what the purpose of the feature is. In particular, I currently maintain a branch which adds a bunch of "default" behaviour as a "default ycm extra conf". This has come up on the tracker a few times and I've personally found it useful. There is understandably however only lukewarm appetite to maintain it as part of the official repo.

https://github.com/Valloric/ycmd/compare/master...puremourning:default-extra-conf?expand=1

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/627#-:-KUIzZYYCBkLkjn8JPrR:beawon1)*

bstaletic:
I took a look at this extra conf file and as far as I can see it tries to do a few things differently. I also tested it quickly in my current project.
Here's how this works:

- Tries to find a compilation database in the current directory of the extra conf file. If it fails, it looks for `compile_commands.json` in its parent directory. This has provided correct flags for my project.
- When finding the `compile_commands.json` eventually fails (I removed the database to test), it looks for "additional files". There's a list named `additional_flag_files=['.clang_complete']`. The extra conf loops through this list trying to find the listed files in the same way it tried to find `compile_commands`.
  - The additional file ares supposed to have a separate flag on each line.
- Having not found any additional files it sets default flags in a few steps.
  - First, it reads the `fallback_flags` list.
  - Second, it tries to be smart about include path and looks for a dir named `include` in the same way it looked for other files before.
  - Finally, it appends base_flags list to whatever flags it found beforehand. The default is `base_flags=['-xc++']`

One of additional files could look like this:
```
-DSTM32F4
-Ilibopencm3/include
-Llibopencm3/lib
-xc
-std=c99
```
In case I wasn't clear, if one of the previous ways of setting the flags managed to produce any kind of flags it won't try to use the other ways, so if it finds a database it won't look for "additional files" etc.

The default flags it set didn't work for me, but my project is an embedded one with a cross toolchain and a lot of weird defines, so it was expected.

A notable feature of this is the logger it uses to make records of what it is doing. It does issue a few useful warning visible in ycmd's stderr log.

If anyone has any questions I'll be glad to answer.

micbou:
@bstaletic Thanks for the detailed explanation.

Now that PR #680 is merged, the compilation database part is directly implemented in ycmd. The rest doesn't seem particularly useful (and if it was, should be done in ycmd code) so I think we can close the PR.

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/627#-:-KaDtssn_jlmMp2uozpN:b-83y1gd)*

PR #634, opened 2016-10-28
[READY] Avoid to wrap around when searching for previous identifier

vheon:
I was looking at the `identifier_completer` and its tests and I something do not adds up: from implementation of `_PreviousIdentifier` it seems that we are looking for the whole identifier before the cursor but there is a test that expects the function to return the identifier under the cursor. I think that the test is wrong also because for the identifier under the cursor we have a separate function `AddIdentifierUnderCursor`. Moreover the `PreviousIdentifier_ColumnInMiddleStillWholeIdent_test` test is passing because in the `_PreviousIdentifier` function we wrap the buffer; so in the test we first check for the previous identifier before the cursor, we don't find it because it ends after the `column` and then we check the previous line but since the buffer has only one line we check it again but this time with `column` as the line length therefore the test pass.

In this first commit I've pushed only the new failing test, after I'll push the fix with a new test explicitly stating that we don't want to wrap the buffer for the previous identifier.

I hope I got all of this right 😝

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/634)

### DIFF
--- a/ycmd/completers/all/identifier_completer.py
+++ b/ycmd/completers/all/identifier_completer.py
@@ -199,7 +199,12 @@ def _PreviousIdentifier( min_num_candidate_size_chars, request_data ):
       return ''
     return ident
 
-  prev_line = contents_per_line[ line_num - 1 ]
+  line_num = line_num - 1
+
+  if line_num < 0:
+    return ''
+
+  prev_line = contents_per_line[ line_num ]
   ident = PreviousIdentifierOnLine( prev_line, len( prev_line ), filetype )
   if len( ident ) < min_num_candidate_size_chars:
     return ''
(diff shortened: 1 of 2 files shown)


vheon:
The test failed (as expected) so I've pushed the new test and the fix for it. I think I'm going to refactor the function a little since it got a little big 😕

micbou:
Subtle bug.

:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/634#-:-KVBaIwU74U7oRorCRxf:b-l856ss)*

vheon:
> I think I'm going to refactor the function a little since it got a little big 😕

I was going to refactor the function but I was touching too much code, so if you all are ok with the way it is right now I'm going to refactor this in another PR.

Valloric:
Wow this is subtle as hell. The `-1` to indexing works because Python uses that to mean "last item in list". 

Nice find!

@homu r=micbou

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/634#-:-KVC2W3ah5B9-U36JVfi:b-o1r13d)*

PR #635, opened 2016-10-29
[READY] Fix warnings when resetting process handle

micbou:
This PR suppresses the `ResourceWarning` exceptions we get when running our tests on Python 3. For example:

``` sh
/home/travis/build/Valloric/ycmd/ycmd/completers/javascript/tern_completer.py:314: ResourceWarning: unclosed file <_io.BufferedWriter name=7>
  self._server_handle = None
```

They are caused by setting to `None` a process handle while its standard piped streams are still open.

---

This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/635)

### DIFF
--- a/ycmd/completers/javascript/tern_completer.py
+++ b/ycmd/completers/javascript/tern_completer.py
@@ -28,7 +28,6 @@ import logging
 import os
 import requests
 import threading
-import traceback
 
 from subprocess import PIPE
 from ycmd import utils, responses
@@ -134,11 +133,12 @@ class TernCompleter( Completer ):
     # server.
     self._server_paths_relative_to = None
 
-    with self._server_state_mutex:
-      self._server_stdout = None
-      self._server_stderr = None
-      self._Reset()
-      self._StartServer()
+    self._server_handle = None
+    self._server_port = None
+    self._server_stdout = None
+    self._server_stderr = None
+
+    self._StartServer()
 
 
   def _WarnIfMissingTernProject( self ):
@@ -301,20 +301,6 @@ class TernCompleter( Completer ):
       return False
 
 
-  def _Reset( self ):
-    with self._server_state_mutex:
-      if not self._server_keep_logfiles:
-        if self._server_stdout:
-          utils.RemoveIfExists( self._server_stdout )
-          self._server_stdout = None
-        if self._server_stderr:
-          utils.RemoveIfExists( self._server_stderr )
-          self._server_stderr = None
-
-      self._server_handle = None
-      self._server_port   = 0
-
-
   def _PostRequest( self, request, request_data ):
     """Send a raw request with the supplied request block, and
     return the server's response. If the server is not running, it is started.
@@ -440,11 +426,10 @@ class TernCompleter( Completer ):
                                                   stdout = stdout,
                                                   stderr = stderr )
       except Exception:
-        _logger.warning( 'Unable to start Tern server: '
-                        + traceback.format_exc() )
-        self._Reset()
+        _logger.exception( 'Unable to start Tern server' )
+        self._CleanUp()
 
-      if self._server_port > 0 and self._ServerIsRunning():
+      if self._server_port and self._ServerIsRunning():
         _logger.info( 'Tern Server started with pid: ' +
                       str( self._server_handle.pid ) +
                       ' listening on port ' +
@@ -478,7 +463,18 @@ class TernCompleter( Completer ):
         except RuntimeError:
           _logger.exception( 'Error while stopping Tern server' )
 
-      self._Reset()
+      self._CleanUp()
+
+
+  def _CleanUp( self ):
+    utils.CloseStandardStreams( self._server_handle )
+    self._server_handle = None
+    self._server_port = None
+    if not self._server_keep_logfiles:
+      utils.RemoveIfExists( self._server_stdout )
+      self._server_stdout = None
+      utils.RemoveIfExists( self._server_stderr )
+      self._server_stderr = None
 
 
   def _ServerIsRunning( self ):
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -553,6 +553,7 @@ class TypeScriptCompleter( Completer ):
 
 
   def _CleanUp( self ):
+    utils.CloseStandardStreams( self._tsserver_handle )
     self._tsserver_handle = None
     if not self.user_options[ 'server_keep_logfiles' ]:
       utils.RemoveIfExists( self._logfile )
--- a/ycmd/tests/client_test.py
+++ b/ycmd/tests/client_test.py
@@ -42,9 +42,9 @@ from ycmd.hmac_utils import CreateHmac, CreateRequestHmac, SecureBytesEqual
 from ycmd.tests import PathToTestFile
 from ycmd.tests.test_utils import BuildRequest
 from ycmd.user_options_store import DefaultOptions
-from ycmd.utils import ( CreateLogfile, GetUnusedLocalhostPort, ReadFile,
-                         RemoveIfExists, SafePopen, SetEnviron, ToBytes,
-                         ToUnicode )
+from ycmd.utils import ( CloseStandardStreams, CreateLogfile,
+                         GetUnusedLocalhostPort, ReadFile, RemoveIfExists,
+                         SafePopen, SetEnviron, ToBytes, ToUnicode )
 
 HEADERS = { 'content-type': 'application/json' }
 HMAC_HEADER = 'x-ycm-hmac'
@@ -63,6 +63,7 @@ class Client_test( object ):
     self._servers = []
     self._logfiles = []
     self._options_dict = DefaultOptions()
+    self._popen_handle = None
 
 
   def setUp( self ):
@@ -75,6 +76,7 @@ class Client_test( object ):
     for server in self._servers:
       if server.is_running():
         server.terminate()
+    CloseStandardStreams( self._popen_handle )
     for logfile in self._logfiles:
       RemoveIfExists( logfile )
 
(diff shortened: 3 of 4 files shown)


micbou:
Reviewed 4 of 4 files at r1.

---

*[ycmd/completers/javascript/tern_completer.py, line 137 at r1](https://reviewable.io:443/reviews/valloric/ycmd/635#-KVHsHuSuYCof4a8Ppx2:-KVHsHuTHTonX9Pv9-GR:b-h6e1l0) ([raw file](https://github.com/valloric/ycmd/blob/c3daaacd28decaf8250ee7feaa956d0519eaa52d/ycmd/completers/javascript/tern_completer.py#L137)):*
> ```Python
>     self._server_paths_relative_to = None
> 
>     with self._server_state_mutex:
> ```

This lock is not needed because when the completer is initialized, [it is already under a lock](https://github.com/Valloric/ycmd/blob/master/ycmd/server_state.py#L61-L78).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/635)*

Valloric:
:lgtm:

Thanks!

---

---

*[ycmd/completers/javascript/tern_completer.py, line 137 at r1](https://reviewable.io:443/reviews/valloric/ycmd/635#-KVHsHuSuYCof4a8Ppx2:-KVIHF_43ELINe0iI4ug:bdwj7w9) ([raw file](https://github.com/valloric/ycmd/blob/c3daaacd28decaf8250ee7feaa956d0519eaa52d/ycmd/completers/javascript/tern_completer.py#L137)):*
> This lock is not needed because when the completer is initialized, [it is already under a lock](https://github.com/Valloric/ycmd/blob/master/ycmd/server_state.py#L61-L78).
</details>
Solid explanation, but needs to be a code comment. :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/635#-:-KVIHMg8PG5HGv9v_EYn:b9benhj)*

micbou:
Reviewed 1 of 1 files at r2.

---

*[ycmd/completers/javascript/tern_completer.py, line 137 at r1](https://reviewable.io:443/reviews/valloric/ycmd/635#-KVHsHuSuYCof4a8Ppx2:-KVK_D4InpsK49eDBAn2:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/c3daaacd28decaf8250ee7feaa956d0519eaa52d/ycmd/completers/javascript/tern_completer.py#L137)):*
> Solid explanation, but needs to be a code comment. :)
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/635)*

vheon:

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/635#-KVQLycoWv9AOUBZ40Vm:-KVQLycoWv9AOUBZ40Vn:bnicdfm) (no related file):*
I'll put the comment here since is related to the whole change. It seems like every time we call `CloseStandardStreams` we set the handle to `None` right after so it looks like the two things are strictly related; wouldn't make more sense to include that into the function as well? So something like `CloseHandle` which first closes the streams and than set the handle to `None`? If you think that the `CloseStandardStreams` makes sense on its own in the near future we can keep it but right now I don't think it does. What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/635)*

micbou:

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/635#-KVQLycoWv9AOUBZ40Vm:-KVQROnsQhTG81bpKxoa:b-bb3xw7) (no related file):*
> I'll put the comment here since is related to the whole change. It seems like every time we call `CloseStandardStreams` we set the handle to `None` right after so it looks like the two things are strictly related; wouldn't make more sense to include that into the function as well? So something like `CloseHandle` which first closes the streams and than set the handle to `None`? If you think that the `CloseStandardStreams` makes sense on its own in the near future we can keep it but right now I don't think it does. What do you think?
</details>
I thought about that but it's not possible because function parameters are local variables. Setting the handle to `None` will not change its value outside the function.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/635)*

puremourning:

---

*[ycmd/completers/javascript/tern_completer.py, line 137 at r1](https://reviewable.io:443/reviews/valloric/ycmd/635#-KVHsHuSuYCof4a8Ppx2:-KVR90qLrKrCqi-3vRKw:b-hy3or3) ([raw file](https://github.com/valloric/ycmd/blob/c3daaacd28decaf8250ee7feaa956d0519eaa52d/ycmd/completers/javascript/tern_completer.py#L137)):*
> when the completer is initialized, it is already under a lock
I'm not super keen on having a specific mutex for the internal state of this object, and eliding it here due to some nature and specific implementation of the _user_ of this object. Seems brittle.

That said, I think the point is that you don't need to hold the lock, as no user of the object can interact with it until `__init__` has been returned (because they necessarily don't have the `self` pointer yet). So we can elide locking because the `__init__` method inherently has exclusive access to member variables.

---

*[ycmd/completers/javascript/tern_completer.py, line 432 at r1](https://reviewable.io:443/reviews/valloric/ycmd/635#-KVRAg_xClDGGR9jAWUl:-KVRAg_xClDGGR9jAWUm:bsicaef) ([raw file](https://github.com/valloric/ycmd/blob/3e8f230fe9eee789eb6464cf9f00a6d3b837930a/ycmd/completers/javascript/tern_completer.py#L432)):*
> ```Python
>         _logger.warning( 'Unable to start Tern server: '
>                         + traceback.format_exc() )
>         self._Reset()
> ```

Am I missing something, or did `self._Reset()` get renamed in this diff, and so this would cause exception-ception (exception within an exception)? It might be that `_Reset` is still defined but I can't see it in reviewable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/635)*

micbou:
Reviewed 1 of 1 files at r3.

---

*[ycmd/completers/javascript/tern_completer.py, line 137 at r1](https://reviewable.io:443/reviews/valloric/ycmd/635#-KVHsHuSuYCof4a8Ppx2:-KVRKC5KwJB0THCliGYc:b-7gb9x1) ([raw file](https://github.com/valloric/ycmd/blob/c3daaacd28decaf8250ee7feaa956d0519eaa52d/ycmd/completers/javascript/tern_completer.py#L137)):*
> > when the completer is initialized, it is already under a lock
> I'm not super keen on having a specific mutex for the internal state of this object, and eliding it here due to some nature and specific implementation of the _user_ of this object. Seems brittle.
> 
> That said, I think the point is that you don't need to hold the lock, as no user of the object can interact with it until `__init__` has been returned (because they necessarily don't have the `self` pointer yet). So we can elide locking because the `__init__` method inherently has exclusive access to member variables.
</details>
I think you are right and the comment doesn't make sense then. I removed it.

---

*[ycmd/completers/javascript/tern_completer.py, line 432 at r1](https://reviewable.io:443/reviews/valloric/ycmd/635#-KVRAg_xClDGGR9jAWUl:-KVRKlCKrmARx3T-ouw7:b-mae0g) ([raw file](https://github.com/valloric/ycmd/blob/3e8f230fe9eee789eb6464cf9f00a6d3b837930a/ycmd/completers/javascript/tern_completer.py#L432)):*
> Am I missing something, or did `self._Reset()` get renamed in this diff, and so this would cause exception-ception (exception within an exception)? It might be that `_Reset` is still defined but I can't see it in reviewable.
</details>
Good catch. I renamed `_Reset` to be consistent with other completers.

---

*[ycmd/completers/javascript/tern_completer.py, line 429 at r3](https://reviewable.io:443/reviews/valloric/ycmd/635#-KVRNYo-56EEYbLNrNOQ:-KVRNYo0nbBcmS3j9v7q:b-32mwxh) ([raw file](https://github.com/valloric/ycmd/blob/c0450ecb43e70ec3884a55ada9cade1f8524ce1d/ycmd/completers/javascript/tern_completer.py#L429)):*
> ```Python
>                                                   stderr = stderr )
>       except Exception:
>         _logger.exception( 'Unable to start Tern server' )
> ```

This is exactly what [`Logger.exception`](https://docs.python.org/2/library/logging.html#logging.Logger.exception) is for.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/635)*

puremourning:
:lgtm:

Nice, thanks.

---

Reviewed 1 of 1 files at r3.

---

*[ycmd/completers/javascript/tern_completer.py, line 429 at r3](https://reviewable.io:443/reviews/valloric/ycmd/635#-KVRNYo-56EEYbLNrNOQ:-KVRbKHFn-GJnO8hyl5g:bb7slsi) ([raw file](https://github.com/valloric/ycmd/blob/c0450ecb43e70ec3884a55ada9cade1f8524ce1d/ycmd/completers/javascript/tern_completer.py#L429)):*
> This is exactly what [`Logger.exception`](https://docs.python.org/2/library/logging.html#logging.Logger.exception) is for.
</details>
yup!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/635#-:-KVRcVwkZTC4k0gMaers:b-kdl37h)*

vheon:
:lgtm: @homu r=Valloric

---

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/635#-KVQLycoWv9AOUBZ40Vm:-KVRqe8eQEM1NTswjIUp:b-8fx2u7) (no related file):*
> I thought about that but it's not possible because function parameters are local variables. Setting the handle to `None` will not change its value outside the function.
</details>
:+1: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/635#-:-KVRqpYRtntULlzQIkmx:b9nem37)*

PR #642, opened 2016-11-12
[RFC] Add keep_logfiles option

micbou:
Currently, we have two switches to control the removal of logfiles at shutdown:
 - the `--keep_logfiles` parameter for the ycmd logfiles;
 - the `server_keep_logfiles` option for the completers logfiles.

It would be simpler to have one global switch instead.

This is done by deprecating the `--keep-logfiles` parameter and the `server_keep_logfiles` option in favor of the new `keep_logfiles` option.

The deprecation is implemented as follows:
 - if `keep_logfiles` is given, ignore `server_keep_logfiles` and `--keep_logfiles`;
 - otherwise, `server_keep_logfiles` takes precedence over `--keep_logfiles`:
 - raise a warning and log it if `server_keep_logfiles` and/or `--keep_logfiles` are given.

Here's an example of how the logs look like if both deprecated options are used:
```
2016-11-11 18:17:07,913 - WARNING - __main__.py:147: DeprecationWarning: The '--keep_logfiles' argument is deprecated in favor of the 'keep_logfiles' option
  DeprecationWarning )

2016-11-11 18:17:07,913 - WARNING - __main__.py:153: DeprecationWarning: The 'server_keep_logfiles' option is deprecated in favor of the 'keep_logfiles' option
  DeprecationWarning )

```

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/642)

### DIFF
--- a/ycmd/__main__.py
+++ b/ycmd/__main__.py
@@ -21,8 +21,8 @@ from __future__ import print_function
 from __future__ import division
 # Other imports from `future` must be placed after SetUpPythonPath.
 
-import sys
 import os
+import sys
 
 sys.path.insert( 0, os.path.dirname( os.path.abspath( __file__ ) ) )
 from server_utils import SetUpPythonPath, CompatibleWithCurrentCore
@@ -32,14 +32,14 @@ from future import standard_library
 standard_library.install_aliases()
 from builtins import *  # noqa
 
-import atexit
-import sys
-import logging
-import json
 import argparse
-import signal
-import os
+import atexit
 import base64
+import json
+import logging
+import os
+import signal
+import warnings
 
 from ycmd import extra_conf_store, user_options_store, utils
 from ycmd.hmac_plugin import HmacPlugin
@@ -114,12 +114,18 @@ def ParseArguments():
                        help = 'optional file to use for stdout' )
   parser.add_argument( '--stderr', type = str, default = None,
                        help = 'optional file to use for stderr' )
+  # This argument is deprecated.
   parser.add_argument( '--keep_logfiles', action = 'store_true', default = None,
-                       help = 'retain logfiles after the server exits' )
+                       help = argparse.SUPPRESS )
   return parser.parse_args()
 
 
 def SetupLogging( log_level ):
+  def LogWarning( message, category, filename, lineno ):
+    warning = warnings.formatwarning( message, category, filename, lineno )
+    logger = logging.getLogger( 'py.warnings' )
+    logger.warning( warning )
+
   numeric_level = getattr( logging, log_level.upper(), None )
   if not isinstance( numeric_level, int ):
     raise ValueError( 'Invalid log level: %s' % log_level )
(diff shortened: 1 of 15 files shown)


micbou:
Reviewed 15 of 15 files at r1.

---

*[ycmd/__main__.py, line 36 at r1](https://reviewable.io:443/reviews/valloric/ycmd/642#-KWKunn2G3nXC7eUIK-a:-KWKunn2G3nXC7eUIK-b:b-6g4yvr) ([raw file](https://github.com/valloric/ycmd/blob/1496bdec0e1bf59cfd18204117289b298d339f44/ycmd/__main__.py#L36)):*
> ```Python
> 
> import atexit
> import sys
> ```

We were importing the `sys` module twice.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/642)*

Valloric:
Not sure how I feel about this change. At first, I didn't like it. Thinking about it more, I can see the benefits, but I'm still not fully sold on this.  We'd have a non-trivial amount of code (and complexity) to ensure backwards compatibility for a problem I'm not convinced is worth solving.

So I'm on the fence and could go either way. @vheon @puremourning Thoughts?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/642#-:-KWL09_342HVnyywHoni:b-jidbff)*

vheon:
I think that using a command line flag to keep the logfiles for `ycmd` and an option in the config file for the completers doesn't make much sense, plus I don't know how useful is to have the two separate and independent from each other.

> We'd have a non-trivial amount of code (and complexity) to ensure backwards compatibility

That's the counterpart for this decision, now the "non-trivial amount of code" is actually only the `EnsureBackwardCompatibilityForOptions` function, so if we add a 100% coverage for that I'm more than fine with this change.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/642#-:-KWP-o7RGrUYBPIextDm:b-iix0sc)*

Valloric:
Give that there's two of you for it, let's do it then. :)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/642#-:-KWPlJKErCwMJwXWxDJO:b-na6yjg)*

puremourning:
For what it is worth, my personal opinion is that this does not net _decrease_ complexity. It perhaps feels like having fewer options is less complex, but this is naive because we now actually just have 1 extra option (as we still maintain interface that uses al of the previous 2 plus the new one). I don't honestly think the additional code complexity and necessary additional test cases warrant the perceived benefit here. :/

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/642#-:-KWPmwnvwoq0A_MOUnIn:b82nhu9)*

Valloric:
OK, so we're 2 for 2 here.

Since opinion seems split, let's go with the status quo and let it be.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/642#-:-KWRHRxuVhbEF_Xk9MKO:brub3pe)*

micbou:
I am closing this then and I'll update PR https://github.com/Valloric/YouCompleteMe/pull/2430 to convert the `keep_logfiles` option into `server_keep_logfiles`.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/642#-:-KWTSxZD_d3H9p5qJehE:b-7zuzn9)*

PR #645, opened 2016-11-16
[READY] Catch exceptions from global extra conf

### LINKED ISSUE #569: exit code not correct if importing ycm_core in global config fails
Hi,

I am not sure this is a real bug, but I encountered this while implementing handling of exit code in `emacs-ycmd`.

I had a `import ycm_core` in my global config. If importing fails there the line with `code = CompatibleWithCurrentCore()` in `__main__.py` will never be reached to return the correct exit code and then I just get an exit code 1.

micbou:
ycmd crashes when an error occurs while calling the `YcmCorePreload`, `VimLeave`, and `Shutdown` methods from the global extra conf file because these calls don't happen through a request but before starting the HTTP server (`YcmCorePreload`) or after stopping it (`VimLeave` and `Shutdown`). We need to catch these errors.

I added tests that cover all the `extra_conf_store` module except the `CallGlobalExtraConfYcmCorePreloadIfExists` and `Shutdown` functions which are covered by the shutdown tests.

Fixes #569.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/645)

### DIFF
--- a/ycmd/extra_conf_store.py
+++ b/ycmd/extra_conf_store.py
@@ -1,4 +1,5 @@
-# Copyright (C) 2011, 2012 Google Inc.
+# Copyright (C) 2011-2012 Google Inc.
+#               2016      ycmd contributors
 #
 # This file is part of ycmd.
 #
@@ -36,6 +37,8 @@ from ycmd.responses import UnknownExtraConf, YCM_EXTRA_CONF_FILENAME
 from ycmd.utils import LoadPythonSource, PathsToAllParentFolders
 from fnmatch import fnmatch
 
+
+_logger = logging.getLogger( __name__ )
 
 # Singleton variables
 _module_for_module_file = {}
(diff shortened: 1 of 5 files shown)


Valloric:
:lgtm:

As usual, fantastic stuff. Thanks so much!

---

Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/645#-:-KWkjXi7XjR7X5Z9uHf7:b-k1leht)*

vheon:
:lgtm: @homu r=Valloric

---

Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/645#-:-KWml2L0YMu8IGz-w10H:b9nem37)*

PR #650, opened 2016-11-17
Remove unused CustomAssert

vheon:
Looking that the coverage this file looks like is never used and I prefer to remove stuff we don't use.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/650)


puremourning:
I noticed this too! I thought I decided it was used by the tests but clearly not. However I do wonder if @valloric remembers why it was added?

vheon:
I'm not sure if @Valloric will remember since it was 2012 😝  https://github.com/Valloric/YouCompleteMe/commit/3e56aea6e3e0878d873024d736f2d57c86c9269a and https://github.com/Valloric/YouCompleteMe/commit/f88c9feb4f29354d5f646a14b795005248e2f7e3 was the first commit that I found where is used 😕

Valloric:
Valloric actually does remember! :D

I use that file in various C++ projects because it implements a really good custom assert. Thought I used it here; apparently not. If we ever need it again, it's in the git history.

LGTM on killing it.

puremourning:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/650#-:-KWsr6v3rDpiojgSqusT:bnfp4nl)*

micbou:
:lgtm:

@homu r+

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/650#-:-KWo_TDHNPd3Z3_p_mEX:b33slzv)*

PR #651, opened 2016-11-18
[READY] Do not reload extra conf file

micbou:
When the extra conf file is already loaded as a module and we attempt to force load it, instead of reusing the module, we reload it and lose its current state. This makes impossible to share variables between the `YcmCorePreload`, `VimLeave`, and `Shutdown` methods of the same global extra conf file.

For now, this PR only adds tests to show the issue. Once the builds have failed, I'll update it with one possible fix.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/651)

### DIFF
--- a/ycmd/extra_conf_store.py
+++ b/ycmd/extra_conf_store.py
@@ -147,14 +147,13 @@ def Load( module_file, force = False ):
   if not module_file:
     return None
 
-  if not force:
-    with _module_for_module_file_lock:
-      if module_file in _module_for_module_file:
-        return _module_for_module_file[ module_file ]
+  with _module_for_module_file_lock:
+    if module_file in _module_for_module_file:
+      return _module_for_module_file[ module_file ]
 
-    if not _ShouldLoad( module_file ):
-      Disable( module_file )
-      return None
+  if not force and not _ShouldLoad( module_file ):
+    Disable( module_file )
+    return None
 
   # This has to be here because a long time ago, the ycm_extra_conf.py files
   # used to import clang_helpers.py from the cpp folder. This is not needed
--- a/ycmd/tests/clang/get_completions_test.py
+++ b/ycmd/tests/clang/get_completions_test.py
@@ -283,7 +283,7 @@ def GetCompletions_FilteredNoResults_Fallback_test( app ):
   } )
 
 
-@SharedYcmd
+@IsolatedYcmd
 def GetCompletions_WorksWithExplicitFlags_test( app ):
   app.post_json(
     '/ignore_extra_conf_file',
@@ -317,7 +317,7 @@ int main()
   eq_( 7, response_data[ 'completion_start_column' ] )
 
 
-@SharedYcmd
+@IsolatedYcmd
 def GetCompletions_NoCompletionsWhenAutoTriggerOff_test( app ):
   with UserOption( 'auto_trigger', False ):
     app.post_json(
--- a/ycmd/tests/extra_conf_store_test.py
+++ b/ycmd/tests/extra_conf_store_test.py
@@ -26,7 +26,8 @@ from builtins import *  # noqa
 import inspect
 from mock import patch
 
-from hamcrest import assert_that, calling, equal_to, has_length, none, raises
+from hamcrest import ( assert_that, calling, equal_to, has_length, none, raises,
+                       same_instance )
 from ycmd import extra_conf_store
 from ycmd.responses import UnknownExtraConf
 from ycmd.tests import PathToTestFile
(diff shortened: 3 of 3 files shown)


vheon:
How is it possible that the tests are passing on Linux/Mac but failing on Windows? I would expect that it would fail on every platform :confused:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/651#-:-KWrXCr2GpeQrQ0q2-a8:bg9vv6o)*

micbou:
This is because the `extra_conf_store_test.py` file mode is executable and nose ignore executable files. PR #652 will fix that.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/651#-:-KWrXpQfSysiFg3OGGyQ:bqy1d4f)*

micbou:
Here's the fix. Note that with this change an ignored extra conf cannot be loaded anymore even when forced.

---

Reviewed 1 of 1 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/651#-:-KWtePtLBF8562rVbUoX:bggt5ey)*

Valloric:
:lgtm:

> Note that with this change an ignored extra conf cannot be loaded anymore even when forced.

Unfortunate, but I'd still rather have the fix than that.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/651#-:-KWuZPK6cTop6d340D1E:b80vb2s)*

vheon:
Reviewed 1 of 1 files at r1, 2 of 2 files at r2.

---

*[ycmd/tests/extra_conf_store_test.py, line 145 at r2](https://reviewable.io:443/reviews/valloric/ycmd/651#-KWwEyIKMnj06cFGXZXn:-KWwEyIKMnj06cFGXZXo:b-7xbjv7) ([raw file](https://github.com/valloric/ycmd/blob/0b73069d46510dc1f42a41ed9abc290c19e3a415/ycmd/tests/extra_conf_store_test.py#L145)):*
> ```Python
> 
> 
>   def Load_DoNotReloadExtraConf_NoForce_test( self ):
> ```

These two tests looks identical aside the `force = ...`. Wouldn't be better to make a `RunLoad( self, force = False )` and call it 2 times, one with `True` and one without?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/651)*

micbou:

---

*[ycmd/tests/extra_conf_store_test.py, line 145 at r2](https://reviewable.io:443/reviews/valloric/ycmd/651#-KWwEyIKMnj06cFGXZXn:-KWwQT-v_rj0xJSwPZOx:bqh410j) ([raw file](https://github.com/valloric/ycmd/blob/0b73069d46510dc1f42a41ed9abc290c19e3a415/ycmd/tests/extra_conf_store_test.py#L145)):*
> These two tests looks identical aside the `force = ...`. Wouldn't be better to make a `RunLoad( self, force = False )` and call it 2 times, one with `True` and one without?
</details>
We could but one downside of this approach is that we would not test anymore that the `force` parameter is optional when calling `Load`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/651)*

micbou:

---

*[ycmd/tests/extra_conf_store_test.py, line 145 at r2](https://reviewable.io:443/reviews/valloric/ycmd/651#-KWwEyIKMnj06cFGXZXn:-KWwT8-ixb2Jo8JNSGW2:b8bd44s) ([raw file](https://github.com/valloric/ycmd/blob/0b73069d46510dc1f42a41ed9abc290c19e3a415/ycmd/tests/extra_conf_store_test.py#L145)):*
> We could but one downside of this approach is that we would not test anymore that the `force` parameter is optional when calling `Load`.
</details>
It's more complicated than that. We wouldn't test that `force` is optional the second time we call `Load`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/651)*

vheon:
:lgtm: @homu r=Valloric

---

---

*[ycmd/tests/extra_conf_store_test.py, line 145 at r2](https://reviewable.io:443/reviews/valloric/ycmd/651#-KWwEyIKMnj06cFGXZXn:-KWwocDM7ZuOh-aiGGwC:b-8vxsur) ([raw file](https://github.com/valloric/ycmd/blob/0b73069d46510dc1f42a41ed9abc290c19e3a415/ycmd/tests/extra_conf_store_test.py#L145)):*
> It's more complicated than that. We wouldn't test that `force` is optional the second time we call `Load`.
</details>
fair enough :+1: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/651#-:-KWwogrPCYiu943tEPVU:b9nem37)*

PR #658, opened 2016-11-30
[WIP] Fix compile error with gcc 4.4

alexchenl:
The code doesn't compile with gcc version 4.4.5 (Debian 4.4.5-8), see this issue: https://github.com/Valloric/YouCompleteMe/issues/2227

It seems that `boost::movelib::unique_ptr< NearestLetterNodeArray > letters_;` has a private copy construct, thus meets the condition of [Deleted implicitly-declared copy assignment operator](http://en.cppreference.com/w/cpp/language/copy_assignment) scenario:

> T has a non-static data member or a direct or virtual base class that cannot be copy-assigned (overload resolution for the copy assignment fails, or selects a deleted or inaccessible function);

So, I tried to add the definition of the missing copy assignment operator.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/658)

### DIFF
--- a/cpp/ycm/LetterNodeListMap.cpp
+++ b/cpp/ycm/LetterNodeListMap.cpp
@@ -52,6 +52,7 @@ LetterNodeListMap::LetterNodeListMap( const LetterNodeListMap &other ) {
 LetterNodeListMap &LetterNodeListMap::operator=( const LetterNodeListMap &other ) {
   if ( this != &other && other.letters_ )
     letters_.reset( new NearestLetterNodeArray( *other.letters_ ) );
+  return *this;
 }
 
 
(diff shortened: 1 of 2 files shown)


Valloric:
This looks good and passes the test suite. Thanks!

Could we get a few simple tests covering this new function? With the tests, this would be good to merge.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/658#-:-KXs_FeA6dVAvHuL0fjL:b-4scktq)*

micbou:
Thanks for the PR but we are dropping support for GCC versions prior to 4.6. See PR #697.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/658#-:-Kbp38I4MEab_bbZ2mc2:b-njonkn)*

PR #661, opened 2016-12-02
Add insertion_snippet field for clang completer

Qusic:
Some context
* https://github.com/Valloric/ycmd/issues/121
* https://github.com/Valloric/ycmd/pull/137

---

It turned out that a breaking refactor is not suitable for this project and this time I try to implement parameter placeholders in completions in a much less intrusive way.

Basically a field named `insertion_snippet` is added to completion candidate object as an alternative to `insertion_text`.
Take a function completion for `fun(int x)` as an example, we can have:
```
insertion_text: 'fun',
insertion_snippet: 'fun(${1:int x})'
```

## TextMate Snippet Format
This field is only adopted in clang completer and all snippet placeholders will be parameters of functions or templates, numbered from 1. Yet this [snippet format](https://manual.macromates.com/en/snippets) is much more capable and also widespread.
Though it may not be fully supported by all editors, the simple usage above works indeed in UltiSnips, Atom and VSCode without parsing and reconstructing.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/661)

### DIFF
--- a/cpp/ycm/ClangCompleter/CompletionData.cpp
+++ b/cpp/ycm/ClangCompleter/CompletionData.cpp
@@ -148,6 +148,54 @@ std::string OptionalChunkToString( CXCompletionString completion_string,
 }
 
 
+std::string ChunkToSnippet( CXCompletionString completion_string,
+                            uint chunk_num,
+                            uint &idx_placeholder ) {
+  std::string snippet;
+
+  if ( !completion_string )
+    return snippet;
+
+  CXCompletionChunkKind kind = clang_getCompletionChunkKind(
+                                 completion_string, chunk_num );
+
+  if ( kind == CXCompletionChunk_Optional ) {
+    CXCompletionString optional_completion_string =
+      clang_getCompletionChunkCompletionString (
+        completion_string, chunk_num );
+
+    if ( optional_completion_string ) {
+      uint optional_num_chunks = clang_getNumCompletionChunks(
+                                   optional_completion_string );
+
+      for ( uint j = 0; j < optional_num_chunks; ++j ) {
+        snippet.append( ChunkToSnippet(
+          optional_completion_string, j, idx_placeholder ) );
+      }
+    }
+  }
+
+  else if ( kind == CXCompletionChunk_Informative ) {
+  }
+
+  else if ( kind == CXCompletionChunk_Placeholder ) {
+    idx_placeholder++;
+    snippet.append( "${" )
+    .append( std::to_string( idx_placeholder ) )
+    .append( ":" )
+    .append( ChunkToString( completion_string, chunk_num ) )
+    .append( "}" );
+  }
+
+  else {
+    snippet.append(
+      ChunkToString( completion_string, chunk_num ) );
+  }
+
+  return snippet;
+}
+
+
 // foo( -> foo
 // foo() -> foo
 std::string RemoveTrailingParens( std::string text ) {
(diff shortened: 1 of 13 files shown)


vheon:
> It turned out that a breaking refactor is not suitable for this project

What are you referring to?

I'm not sure about coupling the API to a Snippet format 😕  Why did we move away from returning the chunks as a json object?

bstaletic:
I remember at least @oblitum objecting the parameter hinting to be done using snippets. The reason behind his objection was that if you move away from the snippet you've lost parameter hinting. While it works for initial writing of a function call it would be useless in case a user needs to change what parameters are passed to the function. I may be wrong but I also think oblitum wasn't the only one saying that snippets were a bad way to do this.

Qusic:
In previous PR, I removed `insertion_text` in favor of `completion_chunks`. However all completers were affected and the changes were too large.
I also believe chunks array is the better way, but do you have any suggestions on how can I make such a PR easy to review and not to be abandoned?

puremourning:
Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/661#-KY3CGuml3rMkA2m4E7E:-KY3CGunw6y09oMPbi4H:b8zh8c8) (no related file):*
The major problem with this (which for me makes it unmergable as is) is that it doesn't work in a sensible way for overloads. If a method is overloaded, you get a snippet for simply one (arbitrary) overload. 

When I prototyped something similar, i changed ycmd to return all overloads as completion suggestions, but that has its own drawbacks.

Out of curiosity, which client is being updated to use this? FWIW, I found that it isn't possible to do this in Vim because you don't have the extra context to actually fire the snippet expansion after selection of a completion. I prototyped a change to Vim to add this additional context to completions, and use CompletionDone autocmd, but the result was unsatisfactory in a number of ways.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/661)*

bstaletic:
@puremourning 
I've just tied what does oblitum's fork do for overloaded constructors and when user selects the appropriate class from a semantic C++ completer it opens a scratch buffer with all available constructors listed and even the hinted parameter in completion menu are actually a list containing all available constructors. The most interesting case is what it does for cases like when there is both `testConstructor(int a)` and `testConstructor(int a, char b)`. In this case until a user enter ',' it will show only `int a`, but the moment user type a comma character it will display `char b`.  I know this may not apeal to everyone, but the idea was to introduce the least intrusive way of handling parameter hinting.

puremourning:
I know exactly what @oblitum's fork does. I was referring to this PR.

vheon:
> In previous PR, I removed insertion_text in favor of completion_chunks. However all completers were affected and the changes were too large.

Can't we do it in steps? So, adding the new interface side by side with the old one and then remove the old one. This way we would not break anything and I think PRs would be more manageble. I don't know if it is a good idea or what the others think though 😕  @Valloric @micbou @puremourning

Valloric:
>  So, adding the new interface side by side with the old one and then remove the old one. 

We could, though I don't see a reason why since all clients pin to a specific version of ycmd. If we just added the new API and removed the old one, clients could still migrate at their leisure (though we would need to inform them).

My concern is not so much the breaking changes as it's making the changes in reviewable chunks and not mixing in unrelated stuff (like refactoring, which is great, but should be a separate PR). And not one PR that does everything, but logical, separate changes as separate PRs.

Qusic:
> When I prototyped something similar, i changed ycmd to return all overloads as completion suggestions, but that has its own drawbacks.

Maybe we can add another field similar to `TypedText` in clang for clients to group completions as needed.

PR #662, opened 2016-12-03
Fix decoding error

Winiex:
I encounter a decoding error when I'm testing YCMD with code from [https://github.com/7sDream/zhihu-py3/tree/master/zhihu](https://github.com/7sDream/zhihu-py3/tree/master/zhihu) which includes python 3 code with Chinese inside.

Adding decode('utf-8') to contents will fix it.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/662)

### DIFF
--- a/ycmd/utils.py
+++ b/ycmd/utils.py
@@ -414,6 +414,9 @@ def SplitLines( contents ):
   # ''.splitlines() returns [], but we want [ '' ]
   if contents == '':
     return [ '' ]
+  
+  # Fix decoding error
+  contents = contents.decode('utf-8')
 
   lines = contents.splitlines()
 


Winiex:
P.S. The exception is raised in this line: [https://github.com/Valloric/ycmd/blob/master/ycmd/utils.py#L432](https://github.com/Valloric/ycmd/blob/master/ycmd/utils.py#L432)

CI has failed on flake8, I'm trying to figure out why.

Winiex:
If content is a unicode string, code in [https://github.com/Valloric/ycmd/blob/master/ycmd/utils.py#L415](https://github.com/Valloric/ycmd/blob/master/ycmd/utils.py#L415) will raise a warning:

UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal.

My solution doesn't solve this problem.

puremourning:

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/662#-KY4GkquBwDBbew-I6JZ:-KY4GkqvcI7dg572PtUM:b-vdnwzh) (no related file):*
Thanks for sending a PR!

Firstly, the change needs tests  :) I would recommend getting the actual code which caused the error and using it as a regression test.

Secondly, I suspect this isn't the correct solution. From memory, I think any string passed to that method should already appropriately be unicode-type string, and we'd need to look at the callers of the method to see why that isn't the case (if indeed it isn't). In any case, we need to support PY2 and PY3 stuff, so I think the correct call would be `ToUnicode( contents )`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/662)*

micbou:
> I encounter a decoding error when I'm testing YCMD with code from https://github.com/7sDream/zhihu-py3/tree/master/zhihu which includes python 3 code with Chinese inside.

I tried to complete the code from your repository with YCM and wasn't able to raise any decoding error. Could you provide more detail on the issue? Which client are you using? Which Python is ycmd running on? Can you give us a list of steps that reproduces the issue?

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/662#-:-KYEjnLkvT_N8gdTdoeC:b-ugpq4e)*

Winiex:
Thanks for your support! :)

This error was raised in my attempt to port the bottle based APIs to tornado based ones, which will read file contents in YCMD process with a acceptable performance and a clear API.

After hours of researching of the source code in handlers.py and related code, I tried to make a minimal situation to reproduce the exception, resulting in everything is fine. So the error is raised just in my ported code, for sake of not using ToUnicode method.

This pull request may be closed by now.

But is it a problem that method in [https://github.com/Valloric/ycmd/blob/master/ycmd/utils.py#L415](https://github.com/Valloric/ycmd/blob/master/ycmd/utils.py#L415) should assert or cast contents into a unicode string anyway? That might make the code more robust, I think.

puremourning:
Ok thanks

PR #667, opened 2016-12-06
Support objc import module

### LINKED ISSUE #659: module support for objective-c
I noticed the latest `YouCompleteMe` now supports `clang 3.9.0`.
So I did a little experiment (https://github.com/haifengkao/libclang-experiments-py).
To my surprise, the `libclang` built by `ycmd` can handle the objective-c modules successfully.

But `:YcmDiag` still fails with the annoying error `could not build module 'Foundation'`.

Since `libclang` now supports objective modules, what should I do to make it work with `YouCompleteMe`?

*update*
The YCM flags can be found in https://github.com/haifengkao/libclang-experiments-py/blob/master/.ycm_extra_conf.py

haifengkao:
Fixes #659 

It turns out that only the libclang in Xcode 8 can handle the `@import Foundation;` syntax correctly.
It means the ycmd should be built  by `install.py --clang-completer --system-libclang`. 
Besides, I got to remove some tweaks in `CMakeLists.txt` and `flags.py` for the system's libclang to work.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/667)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -407,7 +407,7 @@ endif()
 # present in the same directory as our ycm_core.so is used.
 # And the extra `rpath` is helpful to find libclang.dylib's dependencies
 # if it is built with `--enable-shared` flag.
-if ( EXTERNAL_LIBCLANG_PATH AND APPLE )
+if ( EXTERNAL_LIBCLANG_PATH AND APPLE AND NOT USE_SYSTEM_LIBCLANG)
   get_filename_component(EXTRA_LIB_PATH ${EXTERNAL_LIBCLANG_PATH} DIRECTORY)
 
   set_target_properties(${PROJECT_NAME}
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -181,8 +181,11 @@ def PrepareFlagsForClang( flags, filename, add_extra_clang_flags = True ):
   flags = _RemoveUnusedFlags( flags, filename )
   if add_extra_clang_flags:
     flags = _EnableTypoCorrection( flags )
-  flags = _SanitizeFlags( flags )
-  return flags
+
+  vector = ycm_core.StringVector()
+  for flag in flags:
+    vector.append( ToCppStringCompatible( flag ) )
+  return vector
 
 
 def _RemoveXclangFlags( flags ):
@@ -203,16 +206,6 @@ def _RemoveXclangFlags( flags ):
     sanitized_flags.append( flag )
 
   return sanitized_flags
-
-
-def _SanitizeFlags( flags ):
-  """Drops unsafe flags. Currently these are only -arch flags; they tend to
-  crash libclang."""
-
-  vector = ycm_core.StringVector()
-  for flag in flags:
-    vector.append( ToCppStringCompatible( flag ) )
-  return vector
 
 
 def _RemoveFlagsPrecedingCompiler( flags ):
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -209,22 +209,8 @@ def _SanitizeFlags( flags ):
   """Drops unsafe flags. Currently these are only -arch flags; they tend to
   crash libclang."""
 
-  sanitized_flags = []
-  saw_arch = False
-  for i, flag in enumerate( flags ):
-    if flag == '-arch':
-      saw_arch = True
-      continue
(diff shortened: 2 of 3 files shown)


Valloric:
Thanks for the PR!

Tests are failing. You'll need to fix that.

haifengkao:
The tests are failing because the `-arch` flag is not being removed in `_SanitizeFlags` anymore.
I don't really know why this feature exists in the first place, but if you like, I can update the test case to ensure that the flag will not be removed in the future.

haifengkao:
Travis CI shows 
```
FAIL: ycmd.tests.typescript.debug_info_test.DebugInfo_ServerIsNotRunning_LogfilesDoNotExist_test

  File "/Users/travis/.pyenv/versions/2.7.8/Python.framework/Versions/2.7/lib/python2.7/site-packages/nose/case.py", line 197, in runTest

    self.test(*self.arg)

  File "/Users/travis/build/Valloric/ycmd/ycmd/tests/typescript/__init__.py", line 86, in Wrapper

    test( app, *args, **kwargs )

  File "/Users/travis/build/Valloric/ycmd/ycmd/tests/typescript/debug_info_test.py", line 64, in DebugInfo_ServerIsNotRunning_LogfilesDoNotExist_test

    matches_regexp( 'TypeScript completer debug information:\n'

AssertionError: 

Expected: a string matching 'TypeScript completer debug information:

  TSServer is not running

  TSServer executable: .+'

     but: was 'Server has Clang support compiled in: True\nClang version: clang version 3.9.0 (tags/RELEASE_390/final)\nTypeScript completer debug information:\n  TSServer no longer running\n  TSServer executable: /usr/local/bin/tsserver\n  TSServer logfile: /var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/tsserver_5hv4EO.log'
```

How to solve this one?

Valloric:
> The tests are failing because the -arch flag is not being removed in _SanitizeFlags anymore.
I don't really know why this feature exists in the first place, but if you like, I can update the test case to ensure that the flag will not be removed in the future.

We've had issues in the past with the `-arch` flag causing crashes in libclang. That might be fixed by this point. Is that flag needed for objc imports? If so, it's fine to stop removing it, but you'll need to update the tests as well (probably remove the ones that test the flag is being removed).

> How to solve this one?

If you didn't touch the tsserver stuff (I don't think you did), it's probably a transient failure. You'll get another travis run every time you update your branch.

haifengkao:
> We've had issues in the past with the -arch flag causing crashes in libclang. That might be fixed by this point. Is that flag needed for objc imports?

Yes.

>  If so, it's fine to stop removing it, but you'll need to update the tests as well (probably remove the ones that test the flag is being removed).

I have updated the test to ensure the flag will not be removed. Is it acceptable?

vheon:
Reviewed 1 of 2 files at r1.
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[ycmd/completers/cpp/flags.py, line 208 at r3](https://reviewable.io:443/reviews/valloric/ycmd/667#-KYonOtsehhY6kjBOxup:-KYonOtsehhY6kjBOxuq:bgb4hte) ([raw file](https://github.com/valloric/ycmd/blob/644aae40cb5c2279326bee53efa6f18387fda5ad/ycmd/completers/cpp/flags.py#L208)):*
> ```Python
> 
> 
> def _SanitizeFlags( flags ):
> ```

Looks like this function is not needed anymore. At least not to "sanitize" flags. All it does is to prepare the flags for the C++ layer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/667)*

haifengkao:
@vheon any suggestions?

Valloric:
This looks good to merge! :lgtm: 

@haifengkao For future reference, github doesn't send notifications when a PR author adds new commits to a PR branch. You have to post a comment for us to know to take another look. :) 

---

Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/667#-:-KZbkJwGP9iY8za_rd43:b-b83igz)*

vheon:
@haifengkao sorry for the delayed response. Anyway this :lgtm: too. @homu r=Valloric

---

Reviewed 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/667#-:-KZcBxBHAYV7nquV6l3H:b-lahxuf)*

PR #668, opened 2016-12-14
Make IdentifierAtIndex more accurate

Winiex:
In the old implementation, IdentifierAtIndex function will return the next identifier right after the space character when cursor with specified index is on it. I add a restriction on the index so IdentifierAtIndex will handle that situation well.

There might be other problems I don't take into consideration, what's your opinions?

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/668)

### DIFF
--- a/ycmd/identifier_utils.py
+++ b/ycmd/identifier_utils.py
@@ -157,6 +157,9 @@ def IdentifierAtIndex( text, index, filetype = None ):
     return ''
 
   for match in IdentifierRegexForFiletype( filetype ).finditer( text ):
-    if match.end() > index:
+    if match.start() > index:
+      return ''
+    
+    if match.start() < index < match.end():
       return match.group()
   return ''


puremourning:
What problem is this attempting to solve? Can you write steps to reproduce? We also need to add tests for this :/

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/668#-:-KYwcYjd_vOyeR1FCzgp:b-b218l2)*

micbou:
> IdentifierAtIndex function will return the next identifier right after the space character when cursor with specified index is on it.

This behavior is correct.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/668#-:-KYwfNcMc2PGXe5p1KR6:bbe455c)*

Issue #670, opened 2016-12-20
Shutdown tests are failing for some reason


onur:
Hi. I am maintainer of vim-youcompleteme/ycmd packages for Debian and Ubuntu. I am making an updated version of packages before sid freezes but `ycmd.tests.shutdown_test.Shutdown_test` started giving errors for an unknown reason and there isn't much output other than: `<h1>Critical error while processing request: /ready</h1>`.

I hope you can tell the actual reason I can't figure out myself and all tests must pass before this package get into Debian.

All other tests are working fine (ignore `GetCompletions_UnicodeInLineFilter_test` for now).

Output of `./run_tests.py`:

```
$ HOME=«PKGBUILDDIR» LC_ALL=C.UTF-8 ./run_tests.py --skip-build --no-byte-compile --exclude='^cs$' --exclude='UnknownExtraConfException' --exclude='javascript' --exclude='rust' --exclude='go' --exclude='AddNearestThirdPartyFoldersToSysPath_'
...
#63 ycmd.tests.clang.get_completions_test.GetCompletions_UnicodeInLineFilter_test ... FAIL
...
#349 ycmd.tests.shutdown_test.Shutdown_test.FromHandlerWithSubservers_test ... ERROR
#350 ycmd.tests.shutdown_test.Shutdown_test.FromHandlerWithoutSubserver_test ... ERROR
#351 ycmd.tests.shutdown_test.Shutdown_test.FromWatchdogWithSubservers_test ... ERROR
#352 ycmd.tests.shutdown_test.Shutdown_test.FromWatchdogWithoutSubserver_test ... ERROR
...

======================================================================
ERROR: ycmd.tests.shutdown_test.Shutdown_test.FromHandlerWithSubservers_test
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "«PKGBUILDDIR»/ycmd/tests/client_test.py", line 268, in Wrapper
    test( self, *args )
  File "«PKGBUILDDIR»/ycmd/tests/shutdown_test.py", line 47, in FromHandlerWithSubservers_test
    self.Start()
  File "«PKGBUILDDIR»/ycmd/tests/client_test.py", line 123, in Start
    self._WaitUntilReady()
  File "«PKGBUILDDIR»/ycmd/tests/client_test.py", line 145, in _WaitUntilReady
    if self._IsReady( filetype ):
[...]

PR #673, opened 2016-12-21
Update bottle to 0.12.13

onur:
Fixes: #670

Currently ycmd is only working with bottle that comes with ycmd.
It is really an old version of bottle and ycmd is not compatible with
recent version.

A Debian package must use system-dependencies instead of third_party modules.

This patch is updating bottle and making ycmd to compatible with latest
bottle.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/673)

### DIFF
--- a/ycmd/bottle_utils.py
+++ b/ycmd/bottle_utils.py
@@ -23,7 +23,7 @@ from __future__ import absolute_import
 from builtins import *  # noqa
 
 from future.utils import PY2
-from ycmd.utils import ToBytes, ToUnicode
+from ycmd.utils import ToCppStringCompatible, ToUnicode
 import bottle
 
 
@@ -38,5 +38,6 @@ import bottle
 # making life easier for codebases that work across versions, thus preventing
 # tracebacks in the depths of WSGI server frameworks.
 def SetResponseHeader( name, value ):
-  name = ToBytes( name ) if PY2 else ToUnicode( name )
-  bottle.response.set_header( name, ToUnicode( value ) )
+  name = ToCppStringCompatible( name ) if PY2 else ToUnicode( name )
+  value = ToCppStringCompatible( value ) if PY2 else ToUnicode( value )
+  bottle.response.set_header( name, value )


Valloric:
Thanks for the PR!

I added one minor inline comment, but conceptually this PR is good to go.

---

---

*[ycmd/bottle_utils.py, line 43 at r1](https://reviewable.io:443/reviews/valloric/ycmd/673#-KZ_VSgdWezrnfLa_-ET:-KZ_VSgdWezrnfLa_-EU:by1eb2z) ([raw file](https://github.com/valloric/ycmd/blob/d323691f59c90904788b14b1d9c0b76e5f4ac542/ycmd/bottle_utils.py#L43)):*
> ```Python
> def SetResponseHeader( name, value ):
>   name = name.encode( 'utf8' ) if PY2 else ToUnicode( name )
>   value = value.encode( 'utf8' ) if PY2 else ToUnicode( value )
> ```

Why are you removing the use of the `ToBytes` function? We want to use our wrappers because we know exactly how they behave; they encapsulate complexity.

This appears like it will work with `ToBytes` instead of `encode('utf8')`, and if not, it needs to be changed. We don't want to have raw calls to `encode` anywhere in our codebase outside of the wrapper functions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/673#-:-KZ_VtB-fCkTAF-bdIz2:b-qzsksp)*

onur:
*[ycmd/bottle_utils.py, line 43 at r1](https://reviewable.io:443/reviews/valloric/ycmd/673#-KZ_VSgdWezrnfLa_-ET:-KZaGw39q4BdHrNCxtA1:b-d9k2wi) ([raw file](https://github.com/valloric/ycmd/blob/d323691f59c90904788b14b1d9c0b76e5f4ac542/ycmd/bottle_utils.py#L43)):*

Why are you removing the use of the `ToBytes` function? We want to use our wrappers because we know exactly how they behave; they encapsulate complexity.

This appears like it will work with `ToBytes` instead of `encode('utf8')`, and if not, it needs to be changed. We don't want to have raw calls to `encode` anywhere in our codebase outside of the wrapper functions.
</blockquote></details>

`bottle.response.set_header` is no longer working with `ToBytes` (even older versions of bottle is not working with `ToBytes`). Looks like `ToBytes` is only working on bottle that comes with ycmd and it was the primary cause of #670. Also set_header worked only with `encode`, but idk if there is a better way to use it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/673)*

puremourning:

---

*[ycmd/bottle_utils.py, line 43 at r1](https://reviewable.io:443/reviews/valloric/ycmd/673#-KZ_VSgdWezrnfLa_-ET:-KZaxDU3DyANynPLjcSS:b-tbmx43) ([raw file](https://github.com/valloric/ycmd/blob/d323691f59c90904788b14b1d9c0b76e5f4ac542/ycmd/bottle_utils.py#L43)):*
<details><summary><i>Previously, onur (Onur Aslan) wrote…</i></summary><blockquote>

`bottle.response.set_header` is no longer working with `ToBytes` (even older versions of bottle is not working with `ToBytes`). Looks like `ToBytes` is only working on bottle that comes with ycmd and it was the primary cause of #670. Also set_header worked only with `encode`, but idk if there is a better way to use it.
</blockquote></details>

Its very important for us to understand the reason why this has changed, i.e. What in bottle actually changed which meant that this i no longer working as we expect? If bottle expects instances of `str` on PY2 and `str` on PY3, then the original code should have done that. If, however, something internally to bottle has changed what it is expecting, then it's important to understand and document why. This PY2/PY3 compatibility stuff is a monumental PITA and the development team here have some "rules of thumb" that we apply. If any rule is not always working, then we really want to understand it to improve our heuristics.

Looking at the actual error, it is the HMAC header that bottle it choking on. I'd like to investigate further to be sure that we're not missing some root-cause here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/673)*

onur:
*[ycmd/bottle_utils.py, line 43 at r1](https://reviewable.io:443/reviews/valloric/ycmd/673#-KZ_VSgdWezrnfLa_-ET:-KZb8BrUkjInjCkStZeu:bqmnq0f) ([raw file](https://github.com/valloric/ycmd/blob/d323691f59c90904788b14b1d9c0b76e5f4ac542/ycmd/bottle_utils.py#L43)):*

Its very important for us to understand the reason why this has changed, i.e. What in bottle actually changed which meant that this i no longer working as we expect? If bottle expects instances of `str` on PY2 and `str` on PY3, then the original code should have done that. If, however, something internally to bottle has changed what it is expecting, then it's important to understand and document why. This PY2/PY3 compatibility stuff is a monumental PITA and the development team here have some "rules of thumb" that we apply. If any rule is not always working, then we really want to understand it to improve our heuristics.

Looking at the actual error, it is the HMAC header that bottle it choking on. I'd like to investigate further to be sure that we're not missing some root-cause here.
</blockquote></details>

In bottle `set_header` changed like this:

```patch
     def set_header(self, name, value):
         ''' Create a new response header, replacing any previously defined
             headers with the same name. '''
-        self._headers[_hkey(name)] = [str(value)]
+        self._headers[_hkey(name)] = [_hval(value)]
```

And `_hval` function added:

```python
def _hval(value):
    value = value if isinstance(value, unicode) else str(value)
    if '\n' in value or '\r' in value or '\0' in value:
        raise ValueError("Header value must not contain control characters: %r" % value)
    return value
```

But for some reason ycmd's `ToUnicode` is not working on python2.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/673)*

micbou:

---

*[ycmd/bottle_utils.py, line 43 at r1](https://reviewable.io:443/reviews/valloric/ycmd/673#-KZ_VSgdWezrnfLa_-ET:-KZb6WJpbkp2X5J6ZtJR:b-qbj93n) ([raw file](https://github.com/valloric/ycmd/blob/d323691f59c90904788b14b1d9c0b76e5f4ac542/ycmd/bottle_utils.py#L43)):*

Its very important for us to understand the reason why this has changed, i.e. What in bottle actually changed which meant that this i no longer working as we expect? If bottle expects instances of `str` on PY2 and `str` on PY3, then the original code should have done that. If, however, something internally to bottle has changed what it is expecting, then it's important to understand and document why. This PY2/PY3 compatibility stuff is a monumental PITA and the development team here have some "rules of thumb" that we apply. If any rule is not always working, then we really want to understand it to improve our heuristics.

Looking at the actual error, it is the HMAC header that bottle it choking on. I'd like to investigate further to be sure that we're not missing some root-cause here.
</blockquote></details>

I did a bisect and the incompatibility is introduced by commit https://github.com/bottlepy/bottle/commit/3f838db73f7488a108dd8eea308fcc1188303371. This is one of the last change made in Bottle.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/673)*

puremourning:

---

*[ycmd/bottle_utils.py, line 43 at r1](https://reviewable.io:443/reviews/valloric/ycmd/673#-KZ_VSgdWezrnfLa_-ET:-KZbMx0M6MBVJISiA-gc:buab4sz) ([raw file](https://github.com/valloric/ycmd/blob/d323691f59c90904788b14b1d9c0b76e5f4ac542/ycmd/bottle_utils.py#L43)):*

I did a bisect and the incompatibility is introduced by commit https://github.com/bottlepy/bottle/commit/3f838db73f7488a108dd8eea308fcc1188303371. This is one of the last change made in Bottle.
</blockquote></details>


Thanks both.

What's strange is that would imply that it is expecting a unicode type and therefore the result of `ToUnicode`, which returns some form of `str` type on Python 3 and some form of `unicode` type on python 2. Or perhaps we actually want to always use the "native" string type and thus `ToCppStringCompatible`.

`ToBytes` otoh returns some form of `bytes` type on python 3 and some form of `str` type on python 2.

What;s strange is that in python 2 `'\n'` is of type `str` not `unicode`, so i'm not sure how the check exactly works (`'\n' in a_unicode_instance`, though maybe it does).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/673)*

Valloric:

onur:
Is there anything else you want me to add?

Valloric:
@onur There's one inline discussion on the code that hasn't yet been resolved. Thanks for bearing with us! :)

---

---

*[ycmd/bottle_utils.py, line 43 at r1](https://reviewable.io:443/reviews/valloric/ycmd/673#-KZ_VSgdWezrnfLa_-ET:-KkYbTfjpb0Il0VVIGmD:b6qz0ng) ([raw file](https://github.com/valloric/ycmd/blob/d323691f59c90904788b14b1d9c0b76e5f4ac542/ycmd/bottle_utils.py#L43)):*


Thanks both.

What's strange is that would imply that it is expecting a unicode type and therefore the result of `ToUnicode`, which returns some form of `str` type on Python 3 and some form of `unicode` type on python 2. Or perhaps we actually want to always use the "native" string type and thus `ToCppStringCompatible`.

`ToBytes` otoh returns some form of `bytes` type on python 3 and some form of `str` type on python 2.

What;s strange is that in python 2 `'\n'` is of type `str` not `unicode`, so i'm not sure how the check exactly works (`'\n' in a_unicode_instance`, though maybe it does).
</blockquote></details>

Like @puremourning said, we need to know why this is necessary. We have a fairly good set of integration tests that cover python 2 & 3 on Linux, Mac and Windows. If using `ToCppStringCompatible` works here, we should use that instead.

Either way, there needs to be a comment in the code describing why this is being done that way because it's not obvious. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/673#-:-KkYbKd9sLox588OVoZR:b-f5hvj5)*

onur:
*[ycmd/bottle_utils.py, line 43 at r1](https://reviewable.io:443/reviews/valloric/ycmd/673#-KZ_VSgdWezrnfLa_-ET:-Kk_E0zoA2cg6jpwwUg8:bjpr2l9) ([raw file](https://github.com/valloric/ycmd/blob/d323691f59c90904788b14b1d9c0b76e5f4ac542/ycmd/bottle_utils.py#L43)):*

Like @puremourning said, we need to know why this is necessary. We have a fairly good set of integration tests that cover python 2 & 3 on Linux, Mac and Windows. If using `ToCppStringCompatible` works here, we should use that instead.

Either way, there needs to be a comment in the code describing why this is being done that way because it's not obvious. 
</blockquote></details>

Sure I can add a comment, but I am not sure why this change is required in first place. I only managed to make it work with trial and error :-) Maybe @puremourning or @micbou can explain why.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/673)*

bstaletic:
Guys, take a look at this:
https://travis-ci.org/bstaletic/ycmd/builds/249172726
Newset bottle works with `ToCppStringCompatibe()` instead of raw `.encode( 'utf8' )` on python2.

bstaletic:
With the last message I made a mistake of using bottle 0.12.9.
[Here's](https://travis-ci.org/bstaletic/ycmd/builds/249179362) how it looks with bottle 0.12.13.

And here's the function in question:
```py
def SetResponseHeader( name, value ):
  name = ToCppStringCompatible( name ) if PY2 else ToUnicode( name )
  value = ToCppStringCompatible( value ) if PY2 else ToUnicode( value )
  bottle.response.set_header( name, value )
```

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/673#-:-KnzkCZelTP1SyrE3Q7I:b-d42re3)*

Valloric:
@bstaletic That sounds great then; might need to open a separate PR because this one seems stalled.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/673#-:-Ko3ugueglAa3w5i45bM:bpvijsy)*

onur:
This seems much better, I'll update this PR.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/673#-:-Ko4l3e9z0YDSt9-I3x8:brz2taq)*

bstaletic:
:lgtm:

---

Reviewed 2 of 2 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/673#-:-Ko55DANx6uC8laOAESu:bnfp4nl)*

bstaletic:
@zzbot retry

bstaletic:
This PR is missing another LGTM. That's why I canceled the second @zzbot test.

Valloric:
@bstaletic When a second reviewer triggers zzbot after a previous reviewer has given an LGTM, that means the second reviewer is also giving an LGTM. It saves us a round-trip through Reviewable. :)

:lgtm:

@zzbot retry

bstaletic:
@Valloric Alright, I'll keep tht in mind.

Issue #489, opened 2016-05-13
Why do I need a .ycm_extra_conf.py?


mprobst:
I just set up YCM with both VIM and Sublime Text. It works, hurray!

That being said, I'm a bit mystified by the setup process. I copied some `.ycm_extra_conf.py` file from somewhere, removed some code I don't understand, added some code I don't understand, but overall the entire file looks completely plain vanilla to me. There is nothing in it that would be specific to my setup, machine, or projects.

Couldn't this just work out of the box, assuming I have a `compile_commands.json` file?

Valloric:
Please refer to [the relevant section in the README](https://github.com/Valloric/YouCompleteMe#c-family-semantic-completion) that explains why you need it if writing C/C++/ObjC code.

Short story: _every_ project has custom compilation flags and you need to provide them to YCM so that it can compile your code.

mprobst:
I'm using Clang's compile database format. Would it make sense to look for that by default? Getting that python file (a language I'm not comfortable with) into the right place was a substantial hurdle for adoption for me.

Having something that works (even just a bit more) out of the box would make using YCM a lot easier.

TypeScript seems to be a good example, where the service just picks up a `tsconfig.json` it finds and does the right thing automatically.

puremourning:
Right, if you wrote a tsconfig.json and put it in the right place. Just as arbitrary IMO

mprobst:
@puremourning YMMV. `tsconfig.json` is a declarative format that's understood by all TS tools, and very succinct. As it stands now, for YCM I need a Python file I don't understand, plus the compile lines from Clang, which seems worse in any case.

Valloric:
@mprobst 

> tsconfig.json is a declarative format that's understood by all TS tools

That's the problem; C and C++ don't have that. For every user using the clang compilation database, there's 1000 more that aren't and they also use YCM. This is the reason why we have _one_ file for all of these users. You can use it to integrate YCM with whatever your build system may be.

mprobst:
@Valloric the flexibility makes sense to me, but I think you could help a lot of users substantially by having some sane defaults, e.g. by probing for a `compile_commands.json` and just doing the right thing (tm) if it exists.

puremourning:
Exists... where?

mprobst:
@puremourning in an ancestor directory of whatever file you're touching.

puremourning:
[...]

Issue #26, opened 2014-08-27
How to make ycmd use the clang compilation database


abingham:
I'm having trouble getting ycmd to use my clang compilation database. I figure I must need to configure it somehow, but I don't know what needs to be done.

When I send a completion request I get this:

```
Traceback (most recent call last):
  File "/Users/sixtynorth/projects/ycmd/third_party/bottle/bottle.py", line 861, in _handle
    return route.call(**args)
  File "/Users/sixtynorth/projects/ycmd/third_party/bottle/bottle.py", line 1734, in wrapper
    rv = callback(*a, **ka)
  File "/Users/sixtynorth/projects/ycmd/ycmd/../ycmd/watchdog_plugin.py", line 100, in wrapper
    return callback( *args, **kwargs )
  File "/Users/sixtynorth/projects/ycmd/ycmd/../ycmd/hmac_plugin.py", line 54, in wrapper
    body = callback( *args, **kwargs )
  File "/Users/sixtynorth/projects/ycmd/ycmd/../ycmd/handlers.py", line 100, in GetCompletions
    completer.ComputeCandidates( request_data ),
  File "/Users/sixtynorth/projects/ycmd/ycmd/../ycmd/completers/completer.py", line 170, in ComputeCandidates
    candidates = self._GetCandidatesFromSubclass( request_data )
  File "/Users/sixtynorth/projects/ycmd/ycmd/../ycmd/completers/completer.py", line 185, in _GetCandidatesFromSubclass
    raw_completions = self.ComputeCandidatesInner( request_data )
  File "/Users/sixtynorth/projects/ycmd/ycmd/../ycmd/completers/cpp/clang_completer.py", line 85, in ComputeCandidatesInner
    raise RuntimeError( NO_COMPILE_FLAGS_MESSAGE )
RuntimeError: Still no compile flags, no completions yet.
```

I've verified that the compilation database exists and works. Tools like 'clang-check' work on the same file, so it should all be in place. 
So, do I need to do anything to explicitly ask ycmd to use the compilation database? Or am I perhaps doing something else incorrectly?

PR #680, opened 2017-01-01
[READY] Automatically load a compilation database if found

puremourning:
Alternative implementation to : https://github.com/Valloric/ycmd/pull/679

See #679 for explanation and rationale.

Fixes: https://github.com/Valloric/ycmd/issues/489
Fixes: https://github.com/Valloric/YouCompleteMe/issues/2310 (sort of)
Fixes: https://github.com/Valloric/ycmd/issues/26
Fixes: https://github.com/Valloric/YouCompleteMe/issues/1981
Fixes: https://github.com/Valloric/YouCompleteMe/issues/1623 (sort of)
Fixes: https://github.com/Valloric/YouCompleteMe/issues/1622
Partly implements: https://github.com/Valloric/YouCompleteMe/issues/927
Fixes: https://github.com/Valloric/YouCompleteMe/issues/664
Fixes: https://github.com/Valloric/YouCompleteMe/issues/487
Fixes (discussion on): https://github.com/Valloric/YouCompleteMe/issues/174

This implementation is almost identical to the default-extra-conf version, except:

- the code lives in `flags.py`
- it re-uses `INCLUDE_FLAGS` from `flags.py`
- it provides the directory of the compilation database in the `clang_completer` debug info instead of the name of the default extra conf module.

On reflection, I think I might prefer this implementation.

For now, the tests are identical to the other PR (you can now see why I added that set of tests!)

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/680)

### DIFF
--- a/cpp/ycm/ClangCompleter/CompilationDatabase.cpp
+++ b/cpp/ycm/ClangCompleter/CompilationDatabase.cpp
@@ -41,11 +41,11 @@ remove_pointer< CXCompileCommands >::type > CompileCommandsWrap;
 
 CompilationDatabase::CompilationDatabase(
   const boost::python::object &path_to_directory )
-  : is_loaded_( false ) {
+  : is_loaded_( false )
+  , path_to_directory_( GetUtf8String( path_to_directory ) ) {
   CXCompilationDatabase_Error status;
-  std::string path_to_directory_string = GetUtf8String( path_to_directory );
   compilation_database_ = clang_CompilationDatabase_fromDirectory(
-                            path_to_directory_string.c_str(),
+                            path_to_directory_.c_str(),
                             &status );
   is_loaded_ = status == CXCompilationDatabase_NoError;
 }
--- a/cpp/ycm/ClangCompleter/CompilationDatabase.h
+++ b/cpp/ycm/ClangCompleter/CompilationDatabase.h
@@ -54,9 +54,14 @@ public:
   CompilationInfoForFile GetCompilationInfoForFile(
     const boost::python::object &path_to_file );
 
+  std::string GetDatabaseDirectory() {
+    return path_to_directory_;
+  }
+
 private:
 
   bool is_loaded_;
+  std::string path_to_directory_;
   CXCompilationDatabase compilation_database_;
   boost::mutex compilation_database_mutex_;
 };
--- a/cpp/ycm/ycm_core.cpp
+++ b/cpp/ycm/ycm_core.cpp
@@ -197,7 +197,9 @@ BOOST_PYTHON_MODULE(ycm_core)
     .def( "AlreadyGettingFlags",
           &CompilationDatabase::AlreadyGettingFlags )
     .def( "GetCompilationInfoForFile",
-          &CompilationDatabase::GetCompilationInfoForFile );
+          &CompilationDatabase::GetCompilationInfoForFile )
+    .def_readonly( "database_directory",
+                   &CompilationDatabase::GetDatabaseDirectory );
 
   class_< CompilationInfoForFile,
       boost::shared_ptr< CompilationInfoForFile > >(
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -34,7 +34,8 @@ from ycmd import extra_conf_store
 from ycmd.utils import ToCppStringCompatible, ToUnicode
 from ycmd.completers.completer import Completer
 from ycmd.completers.completer_utils import GetIncludeStatementValue
-from ycmd.completers.cpp.flags import Flags, PrepareFlagsForClang
+from ycmd.completers.cpp.flags import ( Flags, PrepareFlagsForClang,
+                                        NoCompilationDatabase )
 from ycmd.completers.cpp.ephemeral_values_set import EphemeralValuesSet
 from ycmd.responses import NoExtraConfDetected, UnknownExtraConf
 
@@ -373,22 +374,49 @@ class ClangCompleter( Completer ):
     filename = request_data[ 'filepath' ]
     try:
       extra_conf = extra_conf_store.ModuleFileForSourceFile( filename )
-      flags = self._FlagsForRequest( request_data ) or []
-    except NoExtraConfDetected:
-      return ( 'C-family completer debug information:\n'
-               '  No configuration file found' )
     except UnknownExtraConf as error:
       return ( 'C-family completer debug information:\n'
                '  Configuration file found but not loaded\n'
                '  Configuration path: {0}'.format(
                  error.extra_conf_file ) )
-    if not extra_conf:
+
+    try:
+      # Note that it only raises NoExtraConfDetected:
+      #  - when extra_conf is None and,
+      #  - there is no compilation database
+      flags = self._FlagsForRequest( request_data )
+    except NoExtraConfDetected:
+      # No flags
       return ( 'C-family completer debug information:\n'
-               '  No configuration file found' )
+               '  No configuration file found\n'
+               '  No compilation database found' )
+
+    # If _FlagsForRequest returns None or raises, we use an empty list in
+    # practice.
+    flags = flags or []
+
+    if extra_conf:
+      # We got the flags from the extra conf file
+      return ( 'C-family completer debug information:\n'
+               '  Configuration file found and loaded\n'
+               '  Configuration path: {0}\n'
+               '  Flags: {1}'.format( extra_conf, list( flags ) ) )
+
+    try:
+      database = self._flags.FindCompilationDatabase(
+          os.path.dirname( filename ) )
+    except NoCompilationDatabase:
+      # No flags
+      return ( 'C-family completer debug information:\n'
+               '  No configuration file found\n'
+               '  No compilation database found' )
+
+    # We got the flags from the compilation database
     return ( 'C-family completer debug information:\n'
-             '  Configuration file found and loaded\n'
-             '  Configuration path: {0}\n'
-             '  Flags: {1}'.format( extra_conf, list( flags ) ) )
+             '  No configuration file found\n'
+             '  Using compilation database from: {0}\n'
+             '  Flags: {1}'.format( database.database_directory,
+                                    list( flags ) ) )
 
 
   def _FlagsForRequest( self, request_data ):
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -30,12 +30,17 @@ import re
 from future.utils import PY2, native
 from ycmd import extra_conf_store
 from ycmd.utils import ( ToCppStringCompatible, OnMac, OnWindows, ToUnicode,
-                         ToBytes )
+                         ToBytes, PathsToAllParentFolders )
 from ycmd.responses import NoExtraConfDetected
+
 
 INCLUDE_FLAGS = [ '-isystem', '-I', '-iquote', '-isysroot', '--sysroot',
                   '-gcc-toolchain', '-include', '-include-pch', '-iframework',
                   '-F', '-imacros' ]
+
+# --sysroot= must be first (or at least, before --sysroot) because the latter is
+# a prefix of the former (and the algorithm checks prefixes)
+PATH_FLAGS =  [ '--sysroot=' ] + INCLUDE_FLAGS
 
 # We need to remove --fcolor-diagnostics because it will cause shell escape
 # sequences to show up in editors, which is bad. See Valloric/YouCompleteMe#1421
@@ -59,6 +64,23 @@ FILE_FLAGS_TO_SKIP = set( [ '-MF',
 # See Valloric/ycmd#266
 CPP_COMPILER_REGEX = re.compile( r'\+\+(-\d+(\.\d+){0,2})?$' )
 
+# List of file extensions to be considered "header" files and thus not present
+# in the compilation database. The logic will try and find an associated
+# "source" file (see SOURCE_EXTENSIONS below) and use the flags for that.
+HEADER_EXTENSIONS = [ '.h', '.hxx', '.hpp', '.hh' ]
+
+# List of file extensions which are considered "source" files for the purposes
+# of heuristically locating the flags for a header file.
+SOURCE_EXTENSIONS = [ '.cpp', '.cxx', '.cc', '.c', '.m', '.mm' ]
+
+EMPTY_FLAGS = {
+  'flags': [],
+}
+
+
+class NoCompilationDatabase( Exception ):
+  pass
+
 
 class Flags( object ):
   """Keeps track of the flags necessary to compile a file.
@@ -71,42 +93,70 @@ class Flags( object ):
     self.extra_clang_flags = _ExtraClangFlags()
     self.no_extra_conf_file_warning_posted = False
 
+    # We cache the compilation database for any given source directory
+    # Keys are directory names and values are ycm_core.CompilationDatabase
+    # instances or None. Value is None when it is known there is no compilation
+    # database to be found for the directory.
+    self.compilation_database_dir_map = dict()
+
+    # Sometimes we don't actually know what the flags to use are. Rather than
+    # returning no flags, if we've previously found flags for a file in a
+    # particular directory, return them. These will probably work in a high
+    # percentage of cases and allow new files (which are not yet in the
+    # compilation database) to receive at least some flags.
+    # Keys are directory names and values are ycm_core.CompilationInfo
+    # instances. Values may not be None.
+    self.file_directory_heuristic_map = dict()
+
 
   def FlagsForFile( self,
                     filename,
                     add_extra_clang_flags = True,
                     client_data = None ):
+
+    # The try-catch here is to avoid a synchronisation primitive. This method
+    # may be called from multiple threads, and python gives us
+    # 1-python-statement synchronisation for "free" (via the GIL)
     try:
       return self.flags_for_file[ filename ]
     except KeyError:
-      module = extra_conf_store.ModuleForSourceFile( filename )
-      if not module:
-        if not self.no_extra_conf_file_warning_posted:
-          self.no_extra_conf_file_warning_posted = True
-          raise NoExtraConfDetected
-        return None
+      pass
 
-      results = _CallExtraConfFlagsForFile( module,
+    module = extra_conf_store.ModuleForSourceFile( filename )
+    try:
+      results = self._GetFlagsFromExtraConfOrDatabase( module,
+                                                       filename,
+                                                       client_data )
+    except NoCompilationDatabase:
+      if not self.no_extra_conf_file_warning_posted:
+        self.no_extra_conf_file_warning_posted = True
+        raise NoExtraConfDetected
+      return None
+
+    if not results or not results.get( 'flags_ready', True ):
+      return None
+
+    flags = _ExtractFlagsList( results )
+    if not flags:
+      return None
+
+    if add_extra_clang_flags:
+      flags += self.extra_clang_flags
+
+    sanitized_flags = PrepareFlagsForClang( flags,
                                             filename,
-                                            client_data )
+                                            add_extra_clang_flags )
 
-      if not results or not results.get( 'flags_ready', True ):
-        return None
+    if results.get( 'do_cache', True ):
+      self.flags_for_file[ filename ] = sanitized_flags
+    return sanitized_flags
 
-      flags = _ExtractFlagsList( results )
-      if not flags:
-        return None
 
-      if add_extra_clang_flags:
-        flags += self.extra_clang_flags
+  def _GetFlagsFromExtraConfOrDatabase( self, module, filename, client_data ):
+    if not module:
+      return self._GetFlagsFromCompilationDatabase( filename )
 
-      sanitized_flags = PrepareFlagsForClang( flags,
-                                              filename,
-                                              add_extra_clang_flags )
-
-      if results.get( 'do_cache', True ):
-        self.flags_for_file[ filename ] = sanitized_flags
-      return sanitized_flags
+    return _CallExtraConfFlagsForFile( module, filename, client_data )
 
 
   def UserIncludePaths( self, filename, client_data ):
@@ -148,6 +198,77 @@ class Flags( object ):
 
   def Clear( self ):
     self.flags_for_file.clear()
+    self.compilation_database_dir_map.clear()
+    self.file_directory_heuristic_map.clear()
+
+
+  def _GetFlagsFromCompilationDatabase( self, file_name ):
+    file_dir = os.path.dirname( file_name )
+    file_root, file_extension = os.path.splitext( file_name )
+
+    database = self.FindCompilationDatabase( file_dir )
+    compilation_info = _GetCompilationInfoForFile( database,
+                                                   file_name,
+                                                   file_extension )
+
+    if not compilation_info:
+      # Note: Try-catch here synchronises access to the cache (as this can be
+      # called from multiple threads).
+      try:
+        # We previously saw a file in this directory. As a guess, just
+        # return the flags for that file. Hopefully this will at least give some
+        # meaningful compilation.
+        compilation_info = self.file_directory_heuristic_map[ file_dir ]
+      except KeyError:
+        # No cache for this directory and there are no flags for this file in
+        # the database.
+        return EMPTY_FLAGS
+
+    # If this is the first file we've seen in path file_dir, cache the
+    # compilation_info for it in case we see a file in the same dir with no
+    # flags available.
+    # The following updates file_directory_heuristic_map if and only if file_dir
+    # isn't already there. This works around a race condition where 2 threads
+    # could be executing this method in parallel.
+    self.file_directory_heuristic_map.setdefault( file_dir, compilation_info )
+
+    return {
+      'flags': _MakeRelativePathsInFlagsAbsolute(
+        compilation_info.compiler_flags_,
+        compilation_info.compiler_working_dir_ ),
+    }
+
+
+  # Return a compilation database object for the supplied path. Raises
+  # NoCompilationDatabase if no compilation database can be found.
+  def FindCompilationDatabase( self, file_dir ):
+    # We search up the directory hierarchy, to first see if we have a
+    # compilation database already for that path, or if a compile_commands.json
+    # file exists in that directory.
+    for folder in PathsToAllParentFolders( file_dir ):
+      # Try/catch to syncronise access to cache
+      try:
+        database = self.compilation_database_dir_map[ folder ]
+        if database:
+          return database
+
+        raise NoCompilationDatabase
+      except KeyError:
+        pass
+
+      compile_commands = os.path.join( folder, 'compile_commands.json' )
+      if os.path.exists( compile_commands ):
+        database = ycm_core.CompilationDatabase( folder )
+
+        if database.DatabaseSuccessfullyLoaded():
+          self.compilation_database_dir_map[ folder ] = database
+          return database
+
+    # Nothing was found. No compilation flags are available.
+    # Note: we cache the fact that none was found for this folder to speed up
+    # subsequent searches.
+    self.compilation_database_dir_map[ file_dir ] = None
+    raise NoCompilationDatabase
 
 
 def _ExtractFlagsList( flags_for_file_output ):
@@ -417,3 +538,67 @@ def _SpecialClangIncludes():
   libclang_dir = os.path.dirname( ycm_core.__file__ )
   path_to_includes = os.path.join( libclang_dir, 'clang_includes' )
   return [ '-resource-dir=' + path_to_includes ]
+
+
+def _MakeRelativePathsInFlagsAbsolute( flags, working_directory ):
+  if not working_directory:
+    return list( flags )
+  new_flags = []
+  make_next_absolute = False
+  for flag in flags:
+    new_flag = flag
+
+    if make_next_absolute:
+      make_next_absolute = False
+      if not os.path.isabs( new_flag ):
+        new_flag = os.path.join( working_directory, flag )
+      new_flag = os.path.normpath( new_flag )
+    else:
+      for path_flag in PATH_FLAGS:
+        # Single dash argument alone, e.g. -isysroot <path>
+        if flag == path_flag:
+          make_next_absolute = True
+          break
+
+        # Single dash argument with inbuilt path, e.g. -isysroot<path>
+        # or double-dash argument, e.g. --isysroot=<path>
+        if flag.startswith( path_flag ):
+          path = flag[ len( path_flag ): ]
+          if not os.path.isabs( path ):
+            path = os.path.join( working_directory, path )
+          path = os.path.normpath( path )
+
+          new_flag = '{0}{1}'.format( path_flag, path )
+          break
+
+    if new_flag:
+      new_flags.append( new_flag )
+  return new_flags
+
+
+# Find the compilation info structure from the supplied database for the
+# supplied file. If the source file is a header, try and find an appropriate
+# source file and return the compilation_info for that.
+def _GetCompilationInfoForFile( database, file_name, file_extension ):
+  # The compilation_commands.json file generated by CMake does not have entries
+  # for header files. So we do our best by asking the db for flags for a
+  # corresponding source file, if any. If one exists, the flags for that file
+  # should be good enough.
+  if file_extension in HEADER_EXTENSIONS:
+    for extension in SOURCE_EXTENSIONS:
+      replacement_file = os.path.splitext( file_name )[ 0 ] + extension
+      compilation_info = database.GetCompilationInfoForFile(
+        replacement_file )
+      if compilation_info and compilation_info.compiler_flags_:
+        return compilation_info
+
+    # No corresponding source file was found, so we can't generate any flags for
+    # this header file.
+    return None
+
+  # It's a source file. Just ask the database for the flags.
+  compilation_info = database.GetCompilationInfoForFile( file_name )
+  if compilation_info.compiler_flags_:
+    return compilation_info
+
+  return None
--- a/ycmd/tests/clang/__init__.py
+++ b/ycmd/tests/clang/__init__.py
@@ -25,8 +25,13 @@ from builtins import *  # noqa
 
 import functools
 import os
+import tempfile
+import contextlib
+import json
+import shutil
 
 from ycmd import handlers
+from ycmd.utils import ToUnicode
 from ycmd.tests.test_utils import ClearCompletionsCache, SetUpApp
 
 shared_app = None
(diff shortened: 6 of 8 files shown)


Valloric:
@r4nt I remember you wanting to see this implemented in ycmd. Curious what your thoughts are; check out both this PR and the other implementation (linked in PR description).

Valloric:
Gave a light review pass. I'm leaning towards this approach as well now that I've seen both. The other approach is a default extra conf file, but the we load it by default and the user would be unwise to change that file directly... at that point we might as well just have this default logic hard-coded into ycmd, because it effectively is.

As usual, thanks for the PR! :)

---

Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[cpp/ycm/ClangCompleter/CompilationDatabase.h, line 57 at r1](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_Mg2GjYFe8w4yqm-zD:-K_Mg2GjYFe8w4yqm-zE:b-6qgpxx) ([raw file](https://github.com/valloric/ycmd/blob/affc3a0438b4d8bf38a97ca4f6e0d6de5dc8933b/cpp/ycm/ClangCompleter/CompilationDatabase.h#L57)):*
> ```objc
>     const boost::python::object &path_to_file );
> 
>   const char * GetDatabaseDirectory()
> ```

I'm pretty sure you can just make this return an `std::string` and Boost.Python will know what to do with it. 

Let's avoid returning raw pointers if possible.

---

*[ycmd/completers/cpp/flags.py, line 96 at r1](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_MgLAFaFt6hPSKB4PA:-K_MgLAFaFt6hPSKB4PB:b-vulwm8) ([raw file](https://github.com/valloric/ycmd/blob/affc3a0438b4d8bf38a97ca4f6e0d6de5dc8933b/ycmd/completers/cpp/flags.py#L96)):*
> ```Python
>     # We cache the database for any given source directory
>     self.compilation_database_dir_map = dict()
>     # Sometimes we don't actually know what the flags to use are. Rather than
> ```

Empty line before comment block?

---

*[ycmd/completers/cpp/flags.py, line 232 at r1](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_MgqQ3eWAeA30O2Yct:-K_MgqQ3eWAeA30O2Ycu:byxwf0r) ([raw file](https://github.com/valloric/ycmd/blob/affc3a0438b4d8bf38a97ca4f6e0d6de5dc8933b/ycmd/completers/cpp/flags.py#L232)):*
> ```Python
>   # database already for that path, or if a compile_commands.json file exists in
>   # that directory.
>   def FindCompilationDatabase( self, wd ):
> ```

Let's expand `wd` to a longer name.

---

*[ycmd/completers/cpp/flags.py, line 235 at r1](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_MgxOWfccb1CQWHUu0:-K_MgxOWfccb1CQWHUu1:boilh7i) ([raw file](https://github.com/valloric/ycmd/blob/affc3a0438b4d8bf38a97ca4f6e0d6de5dc8933b/ycmd/completers/cpp/flags.py#L235)):*
> ```Python
>     # Find all the ancestor directories of the supplied directory
>     for folder in PathsToAllParentFolders( wd ):
>       # Did we already cache a database for this path?
> ```

Maybe cut down on some of these obvious comments now that this isn't in a user-facing file. :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680#-:-K_MhMld0lnzsImRmMEy:br2pljj)*

puremourning:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[cpp/ycm/ClangCompleter/CompilationDatabase.h, line 57 at r1](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_Mg2GjYFe8w4yqm-zD:-K_MjEDH4WMWhxQXx2rp:b6ky3uj) ([raw file](https://github.com/valloric/ycmd/blob/affc3a0438b4d8bf38a97ca4f6e0d6de5dc8933b/cpp/ycm/ClangCompleter/CompilationDatabase.h#L57)):*

I'm pretty sure you can just make this return an `std::string` and Boost.Python will know what to do with it. 

Let's avoid returning raw pointers if possible.
</blockquote></details>

Old habit. Trying to avoid a string copy (`std::string&` and `const std::string&` don't work). But truth is I didn't check whether the underlying implementation doesn't copy the string anyway, so meh, changed to `std::string` return.

---

*[ycmd/completers/cpp/flags.py, line 96 at r1](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_MgLAFaFt6hPSKB4PA:-K_MjyU_89VSEZDA1T6y:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/affc3a0438b4d8bf38a97ca4f6e0d6de5dc8933b/ycmd/completers/cpp/flags.py#L96)):*

Empty line before comment block?
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/flags.py, line 232 at r1](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_MgqQ3eWAeA30O2Yct:-K_Mk-D11hN21hEhYJWN:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/affc3a0438b4d8bf38a97ca4f6e0d6de5dc8933b/ycmd/completers/cpp/flags.py#L232)):*

Let's expand `wd` to a longer name.
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/flags.py, line 235 at r1](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_MgxOWfccb1CQWHUu0:-K_MkpzkLNl__T2GJaJb:b613ko0) ([raw file](https://github.com/valloric/ycmd/blob/affc3a0438b4d8bf38a97ca4f6e0d6de5dc8933b/ycmd/completers/cpp/flags.py#L235)):*

Maybe cut down on some of these obvious comments now that this isn't in a user-facing file. :)
</blockquote></details>

Yep, for this revision i literally copy/pasted the whole code block :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680)*

vheon:
Review status: 0 of 6 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cpp/clang_completer.py, line 37 at r3](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_QA8uAtZJoXQ4ABaBy:-K_QA8uAtZJoXQ4ABaBz:b21a449) ([raw file](https://github.com/valloric/ycmd/blob/91558c0bd585f3c140d164bc64748ae6e5c441b1/ycmd/completers/cpp/clang_completer.py#L37)):*
> ```Python
> from ycmd.completers.completer import Completer
> from ycmd.completers.completer_utils import GetIncludeStatementValue
> from ycmd.completers.cpp.flags import ( Flags, PrepareFlagsForClang )
> ```

why the parenthesis here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680)*

puremourning:
Review status: 0 of 6 files reviewed at latest revision, 5 unresolved discussions.

---

*[ycmd/completers/cpp/clang_completer.py, line 37 at r3](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_QA8uAtZJoXQ4ABaBy:-K_Qns7U3WbkzbR0t7iB:b-60uwgc) ([raw file](https://github.com/valloric/ycmd/blob/91558c0bd585f3c140d164bc64748ae6e5c441b1/ycmd/completers/cpp/clang_completer.py#L37)):*

why the parenthesis here?
</blockquote></details>

I think because i changed this then changed it back. Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680)*

r4nt:
Yep, agree that built-in is probably better. Very happy to see projects that use the compilation db work just out of the box. Thanks so much for this.
If I had one wish, it would be longest common suffix matching for .h/.cpp paths, but that might be too brittle :)

puremourning:
> If I had one wish, it would be longest common suffix matching for .h/.cpp paths, but that might be too brittle

Interesting. Thanks! We were chatting about this last night. Our plan is go out something simple live then improve the heuristics based on feedback and/or inspiration. We considered more caches and some path manipulations for header files (the latter being the key goal). 

Do you have experience with longest common suffix? Do you know that it works well? If so, that could be our first feedback/inspiration point. :)

Valloric:
:lgtm: Nice stuff! Some minor inline comments, otherwise great.

---

Review status: 0 of 8 files reviewed at latest revision, 4 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 97 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_VyBmMDp1Y-4SpFr0V:-K_VyBmMDp1Y-4SpFr0W:b-840ppg) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/completers/cpp/flags.py#L97)):*
> ```Python
> 
>     # We cache the compilation database for any given source directory
>     self.compilation_database_dir_map = dict()
> ```

Since this is python, I'd like to see docs stating what the type of key and value are.

---

*[ycmd/completers/cpp/flags.py, line 104 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_VyHiTJG_B2YW0ZQyD:-K_VyHiTJG_B2YW0ZQyE:bpy1o8p) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/completers/cpp/flags.py#L104)):*
> ```Python
>     # percentage of cases and allow new files (which are not yet in the
>     # compilation database) to receive at least some flags
>     self.file_directory_heuristic_map = dict()
> ```

Same as above; I'd like to see docs stating what the type of key and value are.

God, I hate dynamic languages...

---

*[ycmd/tests/clang/__init__.py, line 104 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_W5lPaGF_xFZTt8R6K:-K_W5lPaGF_xFZTt8R6L:b-3w2w76) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/tests/clang/__init__.py#L104)):*
> ```Python
> 
> @contextlib.contextmanager
> def TemporaryProject( tmp_dir, compile_commands ):
> ```

"Project" seems to be used here to imply a `compile_command.json` connection. That's not very obvious to me. Maybe "ClangProject"? Something better than "project" which is too ambiguous. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680#-:-K_W6RTfu1JfD7khaDXj:b1sjhd0)*

puremourning:
Thanks for the review!

---

Review status: 0 of 8 files reviewed at latest revision, 5 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 97 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_VyBmMDp1Y-4SpFr0V:-K_WEuFTtc1jHWvbHl0D:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/completers/cpp/flags.py#L97)):*

Since this is python, I'd like to see docs stating what the type of key and value are.
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/flags.py, line 104 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_VyHiTJG_B2YW0ZQyD:-K_WEurwMQ2uhAUuNsbC:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/completers/cpp/flags.py#L104)):*

Same as above; I'd like to see docs stating what the type of key and value are.

God, I hate dynamic languages...
</blockquote></details>

Done.

---

*[ycmd/tests/clang/__init__.py, line 104 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_W5lPaGF_xFZTt8R6K:-K_WFxVu9zxb01pcG8q8:b-9svjn8) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/tests/clang/__init__.py#L104)):*

"Project" seems to be used here to imply a `compile_command.json` connection. That's not very obvious to me. Maybe "ClangProject"? Something better than "project" which is too ambiguous. 
</blockquote></details>

Done. ClangProject seems ok.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680#-:-K_WFzZzzhNUdzD-YufO:b-9qdoz3)*

vheon:
Reviewed 2 of 6 files at r1, 1 of 3 files at r2, 1 of 5 files at r4, 1 of 5 files at r5.
Review status: 4 of 8 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cpp/clang_completer.py, line 388 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_Vz2Swa-qG8NS8phfR:-K_Vz2SxdEZVKGjP1vzV:bpj2469) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/completers/cpp/clang_completer.py#L388)):*
> ```Python
>       database = self._flags.FindCompilationDatabase(
>           os.path.dirname( filename ) )
>       if database is None or not database.DatabaseSuccessfullyLoaded():
> ```

can't we write this as `if not database`?

---

*[ycmd/completers/cpp/flags.py, line 104 at r5](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_WRbFthVTkYQfxtyMr:-K_WRbFuMCi9a7P9fN0N:b8gl2pn) ([raw file](https://github.com/valloric/ycmd/blob/b868dd265ff652fe131f8914a5139e7564046ba2/ycmd/completers/cpp/flags.py#L104)):*
> ```Python
>     # Sometimes we don't actually know what the flags to use are. Rather than
>     # returning no flags, if we've previously found flags for a file in a
>     # particular directory, return them. The will probably work in a high
> ```

Just a nitpick but is "The" right here?

---

*[ycmd/completers/cpp/flags.py, line 198 at r5](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_WSNQTbLFzZHNWPoii:-K_WSNQTbLFzZHNWPoij:b-exwxxj) ([raw file](https://github.com/valloric/ycmd/blob/b868dd265ff652fe131f8914a5139e7564046ba2/ycmd/completers/cpp/flags.py#L198)):*
> ```Python
> 
>     database = self.FindCompilationDatabase( file_dir )
>     if database is None or not database.DatabaseSuccessfullyLoaded():
> ```

We call this function in two places and in both places we do this check. Couldn't we `raise` directly in `FindCompilationDatabase` and adapt the other call to `try: catch` this? I'm not sure that is better but since this check is done twice I leaning to think that it deserves at least a name (even though is a simple condition)?

---

*[ycmd/completers/cpp/flags.py, line 205 at r5](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_WT8l-Lz4myt47JlB6:-K_WT8l0ph-6SQ8c1GcY:b-ck3bgz) ([raw file](https://github.com/valloric/ycmd/blob/b868dd265ff652fe131f8914a5139e7564046ba2/ycmd/completers/cpp/flags.py#L205)):*
> ```Python
>                                                    file_extension )
> 
>     if compilation_info is None:
> ```

I think is more Pythonic to write it as `if not compilation_info`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680)*

micbou:
Reviewed 3 of 6 files at r1, 3 of 3 files at r2, 2 of 2 files at r3, 5 of 5 files at r4, 4 of 5 files at r5.
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[cpp/ycm/ClangCompleter/CompilationDatabase.h, line 58 at r2](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_OxCzZwQAMW3nejQ5X:-K_OxCz_txXQL1Zj2_xf:b-f13dnn) ([raw file](https://github.com/valloric/ycmd/blob/0bc399561300a7d73f17bb821fb83dc8af9671c6/cpp/ycm/ClangCompleter/CompilationDatabase.h#L58)):*
> ```objc
> 
>   std::string GetDatabaseDirectory()
>   {
> ```

Isn't the open bracket supposed to be on the first line according to our coding style?

---

*[ycmd/completers/cpp/clang_completer.py, line 402 at r2](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_Oxofu_qL5FPHz3E4G:-K_Oxofu_qL5FPHz3E4H:b-a61z1q) ([raw file](https://github.com/valloric/ycmd/blob/0bc399561300a7d73f17bb821fb83dc8af9671c6/ycmd/completers/cpp/clang_completer.py#L402)):*
> ```Python
>                  '  No configuration file found\n'
>                  '  No compilation database file found\n' )
>       else:
> ```

This `else:` is not needed since we are returning in the first branch.

---

*[ycmd/completers/cpp/clang_completer.py, line 391 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_TYWDubVEMPaZcTKtD:-K_TYWDubVEMPaZcTKtE:b7eknor) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/completers/cpp/clang_completer.py#L391)):*
> ```Python
>         return ( 'C-family completer debug information:\n'
>                  '  No configuration file found\n'
>                  '  No compilation database found\n' )
> ```

Last `\n` should be removed.

---

*[ycmd/completers/cpp/clang_completer.py, line 397 at r5](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_WUlPaT4mnqhevzVPJ:-K_WUlPbhWEWqyXei8XH:bw24i6u) ([raw file](https://github.com/valloric/ycmd/blob/b868dd265ff652fe131f8914a5139e7564046ba2/ycmd/completers/cpp/clang_completer.py#L397)):*
> ```Python
>                  '  Using compilation database from: {0}\n'
>                  '  Flags: {1}'.format( database.database_directory,
>                                         list( flags ) ) )
> ```

It's not obvious why `flags`is always defined in this case and the one below. I think we should move the `flags` assignment in its own `try/catch` block like this:
```python
try:
  extra_conf = extra_conf_store.ModuleFileForSourceFile( filename )
except UnknownExtraConf as error:
  return ...

try:
  flags = self._FlagsForRequest( request_data ) or []
except NoExtraConfDetected:
  return ...

database = self._flags.compilation_database_dir_map.get( 
    os.path.dirname( filename ) )

if database:
  return ... flags ...
return ... flags ...
```
This way, it is clear that `flags` is always defined.

---

*[ycmd/completers/cpp/flags.py, line 186 at r2](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_P1reTJzmNf8edlTVn:-K_P1reUpv81Ec7qdXCY:bkbsnxx) ([raw file](https://github.com/valloric/ycmd/blob/0bc399561300a7d73f17bb821fb83dc8af9671c6/ycmd/completers/cpp/flags.py#L186)):*
> ```Python
> 
>   def Clear( self ):
>     self.flags_for_file.clear()
> ```

I think this function should also clear the `compilation_database_dir_map` and `file_directory_heuristic_map` dictionaries.

---

*[ycmd/completers/cpp/flags.py, line 191 at r2](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_OytiA58qzByk4be0K:-K_OytiA58qzByk4be0L:by9ifnv) ([raw file](https://github.com/valloric/ycmd/blob/0bc399561300a7d73f17bb821fb83dc8af9671c6/ycmd/completers/cpp/flags.py#L191)):*
> ```Python
>   def _GetFlagsFromCompilationDatabase( self, file_name ):
>     file_dir = os.path.dirname( file_name )
>     ( file_root, file_extension ) = os.path.splitext( file_name )
> ```

Parentheses are not required.

---

*[ycmd/completers/cpp/flags.py, line 248 at r5](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_WcIzfrCGIK_vqu2wV:-K_WcIzfrCGIK_vqu2wW:b-vcl197) ([raw file](https://github.com/valloric/ycmd/blob/b868dd265ff652fe131f8914a5139e7564046ba2/ycmd/completers/cpp/flags.py#L248)):*
> ```Python
>     # Nothing was found. No compilation flags are available.
>     # Note: we cache the fact that none was found for this folder to speed up
>     # subsequent searches..
> ```

One dot can be removed.

---

*[ycmd/tests/clang/__init__.py, line 133 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_TZAIJe7dntXcoiUeR:-K_TZAIJe7dntXcoiUeS:b-ai34kw) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/tests/clang/__init__.py#L133)):*
> ```Python
>     yield
>   finally:
>     os.unlink( path )
> ```

We use `os.remove` in other places so we should also use it here to be consistent ([`os.remove` and `os.unlink` are identicals](https://docs.python.org/2/library/os.html#os.unlink)).

---

*[ycmd/tests/clang/flags_test.py, line 513 at r2](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_P-lZOR9zZuCvcdWj4:-K_P-lZPkaUMYNwLJvgs:bug6i0m) ([raw file](https://github.com/valloric/ycmd/blob/0bc399561300a7d73f17bb821fb83dc8af9671c6/ycmd/tests/clang/flags_test.py#L513)):*
> ```Python
> 
> 
> def _MakeRelativePathsInFlagsAvsoluteTest( test ):
> ```

Typo: `Avsolute` instead of `Absolute`.

---

*[ycmd/tests/clang/flags_test.py, line 513 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_TZg9AJr-jWTNrMn4l:-K_TZg9AJr-jWTNrMn4m:b32behq) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/tests/clang/flags_test.py#L513)):*
> ```Python
> 
> 
> def CompilationDatabase_MakeRelativePathsInFlagsAbsolute_test():
> ```

This test is not directly related to compilation databases so I would not prefix its name with `CompilationDatabase`. Same comment for the tests below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680)*

puremourning:
Review status: 3 of 8 files reviewed at latest revision, 17 unresolved discussions.

---

*[cpp/ycm/ClangCompleter/CompilationDatabase.h, line 58 at r2](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_OxCzZwQAMW3nejQ5X:-K_Wh7MwwL4QkqXeoY0N:b-ghzh4t) ([raw file](https://github.com/valloric/ycmd/blob/0bc399561300a7d73f17bb821fb83dc8af9671c6/cpp/ycm/ClangCompleter/CompilationDatabase.h#L58)):*

Isn't the open bracket supposed to be on the first line according to our coding style?
</blockquote></details>

it is indeed.

---

*[ycmd/completers/cpp/clang_completer.py, line 402 at r2](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_Oxofu_qL5FPHz3E4G:-K_WhFt54D8B3nHq2kOM:b-b1sx4z) ([raw file](https://github.com/valloric/ycmd/blob/0bc399561300a7d73f17bb821fb83dc8af9671c6/ycmd/completers/cpp/clang_completer.py#L402)):*

This `else:` is not needed since we are returning in the first branch.
</blockquote></details>

Done.

Removed by refactor of exceptions.

---

*[ycmd/completers/cpp/clang_completer.py, line 388 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_Vz2Swa-qG8NS8phfR:-K_Wh8weDhGQ0ABMDQ6S:b-alwgpu) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/completers/cpp/clang_completer.py#L388)):*

can't we write this as `if not database`?
</blockquote></details>

yeah, i just hate dynamic languages. None behaves strangely.

---

*[ycmd/completers/cpp/clang_completer.py, line 391 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_TYWDubVEMPaZcTKtD:-K_WhEjHc2CRSR83OGKA:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/completers/cpp/clang_completer.py#L391)):*

Last `\n` should be removed.
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/clang_completer.py, line 397 at r5](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_WUlPaT4mnqhevzVPJ:-K_WhIWuJnhWN9RGBwGX:b-pz7p6k) ([raw file](https://github.com/valloric/ycmd/blob/b868dd265ff652fe131f8914a5139e7564046ba2/ycmd/completers/cpp/clang_completer.py#L397)):*

It's not obvious why `flags`is always defined in this case and the one below. I think we should move the `flags` assignment in its own `try/catch` block like this:
```python
try:
  extra_conf = extra_conf_store.ModuleFileForSourceFile( filename )
except UnknownExtraConf as error:
  return ...

try:
  flags = self._FlagsForRequest( request_data ) or []
except NoExtraConfDetected:
  return ...

database = self._flags.compilation_database_dir_map.get( 
    os.path.dirname( filename ) )

if database:
  return ... flags ...
return ... flags ...
```
This way, it is clear that `flags` is always defined.
</blockquote></details>

I've changed the control flow here to make it more obvious and added a few comments.

---

*[ycmd/completers/cpp/flags.py, line 186 at r2](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_P1reTJzmNf8edlTVn:-K_WmTbyauZk16wHgTkK:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/0bc399561300a7d73f17bb821fb83dc8af9671c6/ycmd/completers/cpp/flags.py#L186)):*

I think this function should also clear the `compilation_database_dir_map` and `file_directory_heuristic_map` dictionaries.
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/flags.py, line 191 at r2](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_OytiA58qzByk4be0K:-K_WmUH5NfW7S1vsvxpM:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/0bc399561300a7d73f17bb821fb83dc8af9671c6/ycmd/completers/cpp/flags.py#L191)):*

Parentheses are not required.
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/flags.py, line 104 at r5](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_WRbFthVTkYQfxtyMr:-K_WmSFvA9tNgo6xyW0J:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/b868dd265ff652fe131f8914a5139e7564046ba2/ycmd/completers/cpp/flags.py#L104)):*

Just a nitpick but is "The" right here?
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/flags.py, line 198 at r5](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_WSNQTbLFzZHNWPoii:-K_WX1lRcQ6oFWNg-4jt:btsvvhl) ([raw file](https://github.com/valloric/ycmd/blob/b868dd265ff652fe131f8914a5139e7564046ba2/ycmd/completers/cpp/flags.py#L198)):*

We call this function in two places and in both places we do this check. Couldn't we `raise` directly in `FindCompilationDatabase` and adapt the other call to `try: catch` this? I'm not sure that is better but since this check is done twice I leaning to think that it deserves at least a name (even though is a simple condition)?
</blockquote></details>

Yeah, I have it this way in another change. It is better to have `FindCompilatoinDatabase` raise `NoCompilationDatabase` This required a bit of refactoring.

---

*[ycmd/completers/cpp/flags.py, line 205 at r5](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_WT8l-Lz4myt47JlB6:-K_WmXfahkgbuCMYkDOT:b-p7jzdz) ([raw file](https://github.com/valloric/ycmd/blob/b868dd265ff652fe131f8914a5139e7564046ba2/ycmd/completers/cpp/flags.py#L205)):*

I think is more Pythonic to write it as `if not compilation_info`?
</blockquote></details>

grr. done :)

---

*[ycmd/completers/cpp/flags.py, line 248 at r5](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_WcIzfrCGIK_vqu2wV:-K_Wm_MSEr7WJQrIg0se:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/b868dd265ff652fe131f8914a5139e7564046ba2/ycmd/completers/cpp/flags.py#L248)):*

One dot can be removed.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680)*

Valloric:
Review status: 3 of 8 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cpp/flags.py, line 97 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_VyBmMDp1Y-4SpFr0V:-K_XLjbtCHQmARSf45_f:bjxbdkl) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/completers/cpp/flags.py#L97)):*

Done.
</blockquote></details>

Maybe use "key" instead of "LHS" and "value" instead of "RHS"? Seems more obvious. 

---

*[ycmd/completers/cpp/flags.py, line 104 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_VyHiTJG_B2YW0ZQyD:-K_XLzZ-2LBuYxDPsWOu:b-y2gvxr) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/completers/cpp/flags.py#L104)):*

Done.
</blockquote></details>

Same as above. RHS is a needless abbreviation, just use "value".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680)*

Valloric:
Reviewed 2 of 6 files at r1, 1 of 5 files at r5, 5 of 5 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680)*

puremourning:
Review status: 7 of 8 files reviewed at latest revision, 9 unresolved discussions.

---

*[ycmd/tests/clang/__init__.py, line 133 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_TZAIJe7dntXcoiUeR:-K_ZcCGc80N7mVc7xhHL:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/tests/clang/__init__.py#L133)):*

We use `os.remove` in other places so we should also use it here to be consistent ([`os.remove` and `os.unlink` are identicals](https://docs.python.org/2/library/os.html#os.unlink)).
</blockquote></details>

Done.

---

*[ycmd/tests/clang/flags_test.py, line 513 at r2](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_P-lZOR9zZuCvcdWj4:-K_ZcCr-_-lsfevaJwPY:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/0bc399561300a7d73f17bb821fb83dc8af9671c6/ycmd/tests/clang/flags_test.py#L513)):*

Typo: `Avsolute` instead of `Absolute`.
</blockquote></details>

Done.

---

*[ycmd/tests/clang/flags_test.py, line 513 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_TZg9AJr-jWTNrMn4l:-K_ZcDUg8U-NnRZ64FZa:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/tests/clang/flags_test.py#L513)):*

This test is not directly related to compilation databases so I would not prefix its name with `CompilationDatabase`. Same comment for the tests below.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680)*

puremourning:
Review status: 6 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 97 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_VyBmMDp1Y-4SpFr0V:-K_Zctr79_kG2itl5f8U:bb6un1j) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/completers/cpp/flags.py#L97)):*

Maybe use "key" instead of "LHS" and "value" instead of "RHS"? Seems more obvious. 
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680)*

puremourning:
Review status: 6 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 104 at r4](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_VyHiTJG_B2YW0ZQyD:-K_ZcyDke6whgOJ6SRj0:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/bf093d32403f61707677d10fd4972d024806fc10/ycmd/completers/cpp/flags.py#L104)):*

Same as above. RHS is a needless abbreviation, just use "value".
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680)*

Valloric:
Reviewed 2 of 2 files at r7.

---

*[ycmd/tests/clang/debug_info_test.py, line 98 at r5](https://reviewable.io:443/reviews/valloric/ycmd/680#-K_bgU-qv9CL-luIeheP:-K_bgU-qv9CL-luIeheQ:bkwaq9u) ([raw file](https://github.com/valloric/ycmd/blob/b868dd265ff652fe131f8914a5139e7564046ba2/ycmd/tests/clang/debug_info_test.py#L98)):*
> ```Python
>                         '  No configuration file found\n'
>                         '  Using compilation database from: {0}\n'
>                         '  Flags: .+-Wall.+'.format( tmp_dir ) ) )
> ```

Oh nice!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680)*

micbou:
I tested it on our code base and, to my surprise, it worked great. I only needed to copy the `compile_commands.json` file into the root project. We could even drop our `.ycm_extra_conf.py` file with this PR (or maybe not).

:lgtm:

---

Reviewed 5 of 5 files at r6, 2 of 2 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680#-:-K_g7Ym32MMxJmg5Uv7R:bhirhwi)*

puremourning:
I've rebased and just tweaked a few comments which were wrong or contained typos.

Is this ready to merge now? I was thinking we could merge this, merge my documentation updates to YCM and update ycmd submodule. This would release python 3.6 support, a few bug fixes and this change. Major release! :)

---

Review status: 7 of 8 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/680#-:-K_tNJ5L4Bv6xv8vJwwS:b-x6p61z)*

Valloric:
Merge it! :)

puremourning:
@homu r=micbou

Issue #671, opened 2016-12-21
Javascript strings completion does not include the initial quote


arthurflachs:
Autocompleting javascript string (including required modules) with YCM does not happen as expected in Vim. The same behavior should be met with other clients.

**Steps to reproduce**
* Enter the following : require('re<C-Space>
* Select 'react'
*Expected result:* require('react'
*Actual result:* require(''react'

*Solution to the problem*
The ternjs server includes the quote in the completions result. Therefore the first call to the omnifunc function should return the index of the quote, not the index of the fist alphabetic character.
The part to change to make this working is to include the quote in the javascript FILETYPE_TO_IDENTIFIER_REGEX here : [identifiers_utils.py](https://github.com/Valloric/ycmd/blob/4c6c66d7a6c494fef56976fa61e019bf00c211a6/ycmd/identifier_utils.py#L71)  

This solves the problem in this specific case in Vim, but I don't know if this might have unexpected side effects.

puremourning:
Could you post your `.tern-project` file ?

I can't get tern to complete the paths in `require` in my testing.

arthurflachs:
The node plugin needs to be enabled

```
{
  "plugins": {
    "node": {}
  }
}
```

puremourning:
OK thanks, I can reproduce.

I suspect we should actually strip the `'` because i don't think `'` is actually valid in a javascript identifier (i.e. you can't have a variable named `'`)

puremourning:
Steps to reproduce:

- `mkdir ~/tmp/js`
- create `~/tmp/js/.tern-project` with the following contents:

```json
{
  "plugins": {
    "node": {}
  }
}
```

- save/quit
- `vim test.js`
- insert `require( 're` and press `<ctrl+space>`
- select an item "`repl`"
- expect: `require( 'repl'`
- actual `require( ''repl'`

bstaletic:
I'm not a javascript developer, so I am not quite sure my patch is sufficient, but it does solve the ssue described in the previous @puremourning's post.

Here's what I came up with:
```diff
diff --git a/ycmd/identifier_utils.py b/ycmd/identifier_utils.py
index 6f51fe6b..e64cf4a5 100644
--- a/ycmd/identifier_utils.py
+++ b/ycmd/identifier_utils.py
@@ -68,7 +68,7 @@ FILETYPE_TO_IDENTIFIER_REGEX = {
     # Spec:
[...]

PR #681, opened 2017-01-03
[READY] Allow completers to set the start column 

puremourning:
We always calculate the start column based on our own identifier semantics.

However, some completion engines might have a different interpretation, and indeed the javascript completer does exactly that for `require( '`

Fixes: #671 

This PR is just to demonstrate the approach for comments. It isn't really tested yet (the tests will certainly fail).

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/681)


Valloric:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[ycmd/request_wrap.py, line 112 at r1](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_bf9VSw2q-KP63Q_bh:-K_bf9VSw2q-KP63Q_bi:b-ger6h8) ([raw file](https://github.com/valloric/ycmd/blob/06cbaf68b371c41885b9eeb96e41bf320bce3b4c/ycmd/request_wrap.py#L112)):*
> ```Python
> 
> 
>   def SetCompletionStartColumn( self, column_num ):
> ```

Seems like you want to implement `__setitem__` so that `request_data[ 'start_codepoint' ] = 123` ends up calling `SetCompletionStartColumn`, which would be private.

With the current approach we implement a nice API for reading computed fields but not for setting them, which is weird.

This would need a bit of re-architecting since `_computed_key` only allows storing a getter function, but not a setter.

---

*[ycmd/completers/javascript/tern_completer.py, line 206 at r1](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_beYOOLGpKpkmYXc1_:-K_beYOOLGpKpkmYXc1a:b-72r8g9) ([raw file](https://github.com/valloric/ycmd/blob/06cbaf68b371c41885b9eeb96e41bf320bce3b4c/ycmd/completers/javascript/tern_completer.py#L206)):*
> ```Python
> 
>     completions = response.get( 'completions', [] )
>     start_codepoint = response[ 'start' ][ 'ch' ] + 1
> ```

The +1 needs a comment.

---

*[ycmd/completers/javascript/tern_completer.py, line 208 at r1](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_bfzYjcgV32GlHr8di:-K_bfzYjcgV32GlHr8dj:b-rjegr3) ([raw file](https://github.com/valloric/ycmd/blob/06cbaf68b371c41885b9eeb96e41bf320bce3b4c/ycmd/completers/javascript/tern_completer.py#L208)):*
> ```Python
>     start_codepoint = response[ 'start' ][ 'ch' ] + 1
> 
>     request_data.SetCompletionStartCodepoint( start_codepoint )
> ```

This would need an extensive comment as I don't really understand why you're doing this. :D

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

puremourning:
Thanks for the review!

I've updated it according to Val's suggestion and added some tests.

I should also note, this is only fixed in the Vim client with https://github.com/Valloric/YouCompleteMe/pull/2489 which adds some more valuable context.

---

Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/request_wrap.py, line 112 at r1](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_bf9VSw2q-KP63Q_bh:-K_eXA-uXqbojuZoNdao:b-evaozd) ([raw file](https://github.com/valloric/ycmd/blob/06cbaf68b371c41885b9eeb96e41bf320bce3b4c/ycmd/request_wrap.py#L112)):*

Seems like you want to implement `__setitem__` so that `request_data[ 'start_codepoint' ] = 123` ends up calling `SetCompletionStartColumn`, which would be private.

With the current approach we implement a nice API for reading computed fields but not for setting them, which is weird.

This would need a bit of re-architecting since `_computed_key` only allows storing a getter function, but not a setter.
</blockquote></details>

OK sure :) Done.

---

*[ycmd/completers/javascript/tern_completer.py, line 206 at r1](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_beYOOLGpKpkmYXc1_:-K_eXElWXisD58SVMI6d:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/06cbaf68b371c41885b9eeb96e41bf320bce3b4c/ycmd/completers/javascript/tern_completer.py#L206)):*

The +1 needs a comment.
</blockquote></details>

Done.

---

*[ycmd/completers/javascript/tern_completer.py, line 208 at r1](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_bfzYjcgV32GlHr8di:-K_eXFb0_fx1Yy8QS-6d:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/06cbaf68b371c41885b9eeb96e41bf320bce3b4c/ycmd/completers/javascript/tern_completer.py#L208)):*

This would need an extensive comment as I don't really understand why you're doing this. :D
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681#-:-K_eXYif1fxZtNrVqNUz:b-g185i0)*

Valloric:
:lgtm: with minor comments

Thanks for the PR!

---

Review status: 0 of 8 files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[ycmd/request_wrap.py, line 45 at r2](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_gW_wRZ8rg7eMjBFHI:-K_gW_wRZ8rg7eMjBFHJ:bmdyn79) ([raw file](https://github.com/valloric/ycmd/blob/ac9c57a5adf90a1a19d8bfbc03e54f98c1a4b095/ycmd/request_wrap.py#L45)):*
> ```Python
>     self._request = request
> 
>     # Keys are keys for this object (returned by __getitem__) and values are a
> ```

Keys are keys?! :D :P

---

*[ycmd/request_wrap.py, line 91 at r2](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_gWp7Wpj084Idh7oKe:-K_gWp7Wpj084Idh7oKf:boxnnfg) ([raw file](https://github.com/valloric/ycmd/blob/ac9c57a5adf90a1a19d8bfbc03e54f98c1a4b095/ycmd/request_wrap.py#L91)):*
> ```Python
>       return self._cached_computed[ key ]
>     if key in self._computed_key:
>       getter, setter = self._computed_key[ key ]
> ```

Since you don't actually want the setter, you should use `_` for the name. That's the Pythonic way of saying "I just one of these two". 

---

*[ycmd/request_wrap.py, line 100 at r2](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_gWyeHyz6uOTMRqfos:-K_gWyeHyz6uOTMRqfot:b8hiugl) ([raw file](https://github.com/valloric/ycmd/blob/ac9c57a5adf90a1a19d8bfbc03e54f98c1a4b095/ycmd/request_wrap.py#L100)):*
> ```Python
>   def __setitem__( self, key, value ):
>     if key in self._computed_key:
>       getter, setter = self._computed_key[ key ]
> ```

Same as above, `_` for getter.

---

*[ycmd/completers/javascript/tern_completer.py, line 208 at r1](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_bfzYjcgV32GlHr8di:-K_gXC0rYRcFEEPrwFUQ:b-c6fjtc) ([raw file](https://github.com/valloric/ycmd/blob/06cbaf68b371c41885b9eeb96e41bf320bce3b4c/ycmd/completers/javascript/tern_completer.py#L208)):*

Done.
</blockquote></details>

Wow, you really took "extensive" to heart! Thanks for explaining this so thoroughly! :)

---

*[ycmd/completers/javascript/tern_completer.py, line 209 at r2](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_gXJzIDbMBiq0i-MxO:-K_gXJzIDbMBiq0i-MxP:b-o0m6rn) ([raw file](https://github.com/valloric/ycmd/blob/ac9c57a5adf90a1a19d8bfbc03e54f98c1a4b095/ycmd/completers/javascript/tern_completer.py#L209)):*
> ```Python
> 
>     # Tern returns the range of the word in the file which it is replacing. This
>     # may not be the same rainge that our "completion start column" calculation
> ```

typo: "rainge"

---

*[ycmd/completers/javascript/tern_completer.py, line 216 at r2](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_gXPkijTPizkTwqFr0:-K_gXPkjYXNKOgTJnQou:b9g7rm6) ([raw file](https://github.com/valloric/ycmd/blob/ac9c57a5adf90a1a19d8bfbc03e54f98c1a4b095/ycmd/completers/javascript/tern_completer.py#L216)):*
> ```Python
>     #
>     # with the cursor on |, tern returns something like 'test' (i.e. including
>     # the single-quotes. Single-quotes are not a JavaScript identifier, so
> ```

Missing closing paren.

---

*[ycmd/completers/javascript/tern_completer.py, line 218 at r2](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_gXS0lIFLrlDcWK1dX:-K_gXS0lIFLrlDcWK1dY:b-i6inw6) ([raw file](https://github.com/valloric/ycmd/blob/ac9c57a5adf90a1a19d8bfbc03e54f98c1a4b095/ycmd/completers/javascript/tern_completer.py#L218)):*
> ```Python
>     # the single-quotes. Single-quotes are not a JavaScript identifier, so
>     # should not normally be considered an identifier character, but by using
>     # our aown start_codepoint calculation, the inserted string would be:
> ```

aown

I recommend turning on spell checking in Vim :P

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681#-:-K_gXijvbdoReypfj7Sv:b-o9tsr8)*

vheon:
:lgtm:

---

Reviewed 1 of 3 files at r1, 7 of 7 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681#-:-K_kddWbrhYs_VxWz2-M:bnfp4nl)*

puremourning:
now possible with https://github.com/Valloric/YouCompleteMe/pull/2657!

puremourning:
@zzbot try

appveyor flaked?

bstaletic:
:lgtm: with a minor comment.

---

Reviewed 2 of 7 files at r2, 6 of 6 files at r3.

---

*[ycmd/request_wrap.py, line 61 at r3](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkbHXcttBXDk0DDcwjj:-KkbHXcttBXDk0DDcwjk:bqirh0p) ([raw file](https://github.com/valloric/ycmd/blob/93707d2b2c952f495f921fa40ec1ac491fe2741f/ycmd/request_wrap.py#L61)):*
> ```Python
>         ByteOffsetToCodepointOffset( self[ 'line_bytes' ],
>                                      self[ 'column_num' ] ) ),
>         None ),
> ```

This should be aligned with `(lambda:`. Also `(lambda:` should have a space after opening pren.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681#-:-KkbvLCrFkcIqnwvUcI-:bshw4f8)*

puremourning:

---

*[ycmd/request_wrap.py, line 45 at r2](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_gW_wRZ8rg7eMjBFHI:-Ke_hiTvs0SrIFBBwOZt:b3ej2v7) ([raw file](https://github.com/valloric/ycmd/blob/ac9c57a5adf90a1a19d8bfbc03e54f98c1a4b095/ycmd/request_wrap.py#L45)):*

Keys are keys?! :D :P
</blockquote></details>

I meant keys of `_computed_key` are the keys of _this object_ (accessed via `__getitem__` / `[]` operator). I have reworded to make more sense!

---

*[ycmd/request_wrap.py, line 91 at r2](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_gWp7Wpj084Idh7oKe:-Ke_hsTodTUIs-C9v8IS:b-7rl0jd) ([raw file](https://github.com/valloric/ycmd/blob/ac9c57a5adf90a1a19d8bfbc03e54f98c1a4b095/ycmd/request_wrap.py#L91)):*

Since you don't actually want the setter, you should use `_` for the name. That's the Pythonic way of saying "I just one of these two". 
</blockquote></details>

done.

---

*[ycmd/request_wrap.py, line 100 at r2](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_gWyeHyz6uOTMRqfos:-Ke_i17vwZfFWUXlBKzu:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/ac9c57a5adf90a1a19d8bfbc03e54f98c1a4b095/ycmd/request_wrap.py#L100)):*

Same as above, `_` for getter.
</blockquote></details>

Done.

---

*[ycmd/request_wrap.py, line 61 at r3](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkbHXcttBXDk0DDcwjj:-Kkc2V7YB-djFbzyb__a:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/93707d2b2c952f495f921fa40ec1ac491fe2741f/ycmd/request_wrap.py#L61)):*

This should be aligned with `(lambda:`. Also `(lambda:` should have a space after opening pren.
</blockquote></details>

Done.

---

*[ycmd/completers/javascript/tern_completer.py, line 209 at r2](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_gXJzIDbMBiq0i-MxO:-Ke_iEULqRge_1YOsRIs:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/ac9c57a5adf90a1a19d8bfbc03e54f98c1a4b095/ycmd/completers/javascript/tern_completer.py#L209)):*

typo: "rainge"
</blockquote></details>

Done.

---

*[ycmd/completers/javascript/tern_completer.py, line 216 at r2](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_gXPkijTPizkTwqFr0:-Ke_iE7WYTp5WrItNT9F:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/ac9c57a5adf90a1a19d8bfbc03e54f98c1a4b095/ycmd/completers/javascript/tern_completer.py#L216)):*

Missing closing paren.
</blockquote></details>

Done.

---

*[ycmd/completers/javascript/tern_completer.py, line 218 at r2](https://reviewable.io:443/reviews/valloric/ycmd/681#-K_gXS0lIFLrlDcWK1dX:-Ke_iH2opyF0m0L4-SIq:bohjkzf) ([raw file](https://github.com/valloric/ycmd/blob/ac9c57a5adf90a1a19d8bfbc03e54f98c1a4b095/ycmd/completers/javascript/tern_completer.py#L218)):*

aown

I recommend turning on spell checking in Vim :P
</blockquote></details>

You probably won't believe me, but I actually do use `:set spell`.... :/

If only there was an autocomplete plugin for Vim which supported offering dictionary suggestions when typing in comments.... OK kill me. I think that's fair.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

bstaletic:
@puremourning This has no `[READY]` tag. Is there anything else missing? I also don't see any unresolved discussions.

---

Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681#-:-KkcXB6kJmn8F2_1F-zR:by7em8h)*

puremourning:
I honk it is g2g from server POV. when we're happy with https://github.com/Valloric/YouCompleteMe/issues/2657 then we can merge this. It's safe to merge without it (I think) - will check

micbou:
Reviewed 7 of 7 files at r2, 5 of 6 files at r3, 1 of 1 files at r4.

---

*[ycmd/request_wrap.py, line 135 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkowaNm-K6dNIjxTLvt:-KkowaNm-K6dNIjxTLvu:bgcf4e2) ([raw file](https://github.com/valloric/ycmd/blob/f19b4c3765867c84755141a13921ec51e6cf61c0/ycmd/request_wrap.py#L135)):*
> ```Python
>     self._cached_computed[ 'start_codepoint' ] = ByteOffsetToCodepointOffset(
>       self[ 'line_value' ],
>       column_num )
> ```

Why don't we just remove the `start_codepoint` key from the cache instead so that we only re-compute it if needed? Also, since `query` depends on `start_codepoint` (and indirectly on `start_column`), we should either update its cached value or as suggested, remove its key from the cache.

---

*[ycmd/request_wrap.py, line 148 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-Kkowftk587XRiVWUMSC:-Kkowftk587XRiVWUMSD:benpsw9) ([raw file](https://github.com/valloric/ycmd/blob/f19b4c3765867c84755141a13921ec51e6cf61c0/ycmd/request_wrap.py#L148)):*
> ```Python
>     self._cached_computed[ 'start_column' ] = CodepointOffsetToByteOffset(
>       self[ 'line_value' ],
>       codepoint_offset )
> ```

Same comment as above.

---

*[ycmd/tests/request_wrap_test.py, line 284 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkoupxjHXtDLoDmA5X_:-KkoupxjHXtDLoDmA5Xa:bnx1b4) ([raw file](https://github.com/valloric/ycmd/blob/f19b4c3765867c84755141a13921ec51e6cf61c0/ycmd/tests/request_wrap_test.py#L284)):*
> ```Python
> def Calculated_SetMethod_test():
>   assert_that(
>     calling( RequestWrap( PrepareJson( ) ).__setitem__).with_args(
> ```

Missing space after `__setitem__`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

puremourning:

---

*[ycmd/request_wrap.py, line 135 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkowaNm-K6dNIjxTLvt:-KnfFO6XQr7h9FEwBUTK:bffjzy9) ([raw file](https://github.com/valloric/ycmd/blob/608f53ecb082c04eb9eb97590fced810c2013524/ycmd/request_wrap.py#L135)):*

Why don't we just remove the `start_codepoint` key from the cache instead so that we only re-compute it if needed? Also, since `query` depends on `start_codepoint` (and indirectly on `start_column`), we should either update its cached value or as suggested, remove its key from the cache.
</blockquote></details>

Fair point. Done.

---

*[ycmd/request_wrap.py, line 148 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-Kkowftk587XRiVWUMSC:-KnfFQ7kNHKAMchzxcGR:bffjzy9) ([raw file](https://github.com/valloric/ycmd/blob/608f53ecb082c04eb9eb97590fced810c2013524/ycmd/request_wrap.py#L148)):*

Same comment as above.
</blockquote></details>

Fair point. Done.

---

*[ycmd/tests/request_wrap_test.py, line 284 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkoupxjHXtDLoDmA5X_:-KnfFY3y_eBtUfYkYXXH:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/608f53ecb082c04eb9eb97590fced810c2013524/ycmd/tests/request_wrap_test.py#L284)):*

Missing space after `__setitem__`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

puremourning:
Hang on. I marked comments as done, but i haven't done them. Senile moment. Lemme see what happened.

puremourning:

---

*[ycmd/request_wrap.py, line 135 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkowaNm-K6dNIjxTLvt:-KnfYFtfneDewSrA1-pH:b-j1k1qh) ([raw file](https://github.com/valloric/ycmd/blob/608f53ecb082c04eb9eb97590fced810c2013524/ycmd/request_wrap.py#L135)):*

Fair point. Done.
</blockquote></details>

Correction. I tried this and it just doesn't work. It's not super intuitive, but I'll explain:

The problem is that the calculation of the fields `_GetCompletionStartColumn` and `_GetCompletionStartCodepoint` are by default  dependent only on the 'column number' in the original request. This makes sense because otherwise there would be an (arbitrary) dependency ordering between the start column and start codepoint. When we override the 'start column' (be it byte offset or codepoint), we are breaking the association between 'request column number' and 'start column'. The mechanism used to enforce this is to pre-compute the dependent value, because we don't store the 'new start column' (in whatever units the user called us) anywhere on the object.

Basically:
 - if nobody ever calls the `__setitem__`, then the values calculated are _independent_ of each other; they are calculated based on the `column_num`
 - if anybody calls the `__setitem__`, then they are now dependent on that value that was just _set_. As that value is not available at a later time, we must pre-compute the now-dependent value

I hope that makes sense... It took me a few mins to work it out :)

---

*[ycmd/request_wrap.py, line 148 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-Kkowftk587XRiVWUMSC:-Knf_NA_idRQ4L7hVTN2:b-qy15nk) ([raw file](https://github.com/valloric/ycmd/blob/608f53ecb082c04eb9eb97590fced810c2013524/ycmd/request_wrap.py#L148)):*

Fair point. Done.
</blockquote></details>

See above. I've added a brief comment to explain.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

puremourning:
OK panic over, I wasn't going mad. I reverted it because it was wrong :)

puremourning:
I don't know what codecov is moaning about, coverage seems good to me.

bstaletic:
> I don't know what codecov is moaning about,

Codecov is moaning (I like that word in this context) about [this](https://codecov.io/gh/Valloric/ycmd/compare/99f469e1339996c04a133f0c4c9285cdb531d445...4fe692cba1472b9733ca53583cc323e0fcafd272/tree?path=ycmd%2Fcompleters%2Fjavascript). At least that's the only file with  negative coverage diff according to codecov, but I can't understand why.

---

Reviewed 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681#-:-Knh75zgvQREnIH-of00:b-h34dpf)*

micbou:
Reviewed 1 of 2 files at r5.

---

*[ycmd/request_wrap.py, line 135 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkowaNm-K6dNIjxTLvt:-KnfcYHBlqDF_sCzzMBN:bl7e542) ([raw file](https://github.com/valloric/ycmd/blob/d6a5700ebbf797a266f8f93675eba02cbf4e52ca/ycmd/request_wrap.py#L135)):*

Correction. I tried this and it just doesn't work. It's not super intuitive, but I'll explain:

The problem is that the calculation of the fields `_GetCompletionStartColumn` and `_GetCompletionStartCodepoint` are by default  dependent only on the 'column number' in the original request. This makes sense because otherwise there would be an (arbitrary) dependency ordering between the start column and start codepoint. When we override the 'start column' (be it byte offset or codepoint), we are breaking the association between 'request column number' and 'start column'. The mechanism used to enforce this is to pre-compute the dependent value, because we don't store the 'new start column' (in whatever units the user called us) anywhere on the object.

Basically:
 - if nobody ever calls the `__setitem__`, then the values calculated are _independent_ of each other; they are calculated based on the `column_num`
 - if anybody calls the `__setitem__`, then they are now dependent on that value that was just _set_. As that value is not available at a later time, we must pre-compute the now-dependent value

I hope that makes sense... It took me a few mins to work it out :)
</blockquote></details>

Right. I didn't think this through. What about `query`? It needs to be updated so that candidates are correctly filtered. See your example in https://github.com/Valloric/YouCompleteMe/pull/2489: `test` and `testing` should be filtered against `this_is_a_test.`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

puremourning:

---

*[ycmd/request_wrap.py, line 135 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkowaNm-K6dNIjxTLvt:-Kno8E0ilXZdnSgSofTb:bsk85vc) ([raw file](https://github.com/valloric/ycmd/blob/d6a5700ebbf797a266f8f93675eba02cbf4e52ca/ycmd/request_wrap.py#L135)):*

Right. I didn't think this through. What about `query`? It needs to be updated so that candidates are correctly filtered. See your example in https://github.com/Valloric/YouCompleteMe/pull/2489: `test` and `testing` should be filtered against `this_is_a_test.`.
</blockquote></details>

Hmm do we really want to change the `query` in response to a completion request? I'm concerned that might lead to unpredictable results (it's a sort of feedback loop).

The original examples that i remember for the omnicompleter were:
 - javacomplete2 completion of `@annonations` - we would end up with 2 @ signs, but the filtering was fine IIRC
 - eclim completion of import declarations (eclipse insists on returning the `import` keyword for inexplicable reasons. eclim does a hearty job of squashing this but iirc there were cases where it wasn't perfect. anyway, again i think the filtering in that case would be screwy if we changed the `query` to be the `import statement`)

That said, I just tested it out with the javacomplete2 @annotation thing and it isn't working correctly (it's replacing in the right position, but the @ symbol is removed). Will need to drill into that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

micbou:
Reviewed 1 of 2 files at r5.

---

*[ycmd/request_wrap.py, line 135 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkowaNm-K6dNIjxTLvt:-Knoe0PucQHY0BdIscy4:b-9c7j4o) ([raw file](https://github.com/valloric/ycmd/blob/d6a5700ebbf797a266f8f93675eba02cbf4e52ca/ycmd/request_wrap.py#L135)):*
> Hmm do we really want to change the query in response to a completion request? I'm concerned that might lead to unpredictable results (it's a sort of feedback loop).

This is already the case for the JavaScript completer with the proposed changes. The query is still not set when the start column is updated. It's only computed when filtering the candidates. This means that when changing the start column, we are also changing the query. Try to access the query before computing the candidates. This leads to ycmd returning all candidates when inserting the second punctuation mark in a `require` statement:



This is because the query is empty in that case (the start column is at the second punctuation mark). However, if we access the query after computing the candidates (and thus after updating the start column), the query is `'os'` as expected.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

puremourning:

---

*[ycmd/request_wrap.py, line 135 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkowaNm-K6dNIjxTLvt:-KnolC84Z2jEvRG_vKAN:b-qb9az6) ([raw file](https://github.com/valloric/ycmd/blob/d6a5700ebbf797a266f8f93675eba02cbf4e52ca/ycmd/request_wrap.py#L135)):*
Actually i have debugged this, and it's the omnifunc that's wrong. For this test case:

```java
public class Test {
  @SuppressW
  public static void main(String[] args) {
  }
}
```

The omnifunc returns 2 for the `findstart` and `SuppressWarnings` for the text, so our behaviour is correct and javacomplete2 is telling us lies.

```
2017-06-29 17:53:40,454 - DEBUG - Omnifunc returned start: 2, values: [
  {
    "menu": "com.sun.xml.internal.ws.api.message", 
    "word": "SuppressAutomaticWSARequestHeadersFeature", 
    "type": "c"
  }, 
  {
    "menu": "java.lang", 
    "word": "SuppressWarnings", 
    "type": "c"
  }
]
```

For confirmation, the result is:

```java
public class Test {
  SuppressWarnings
  public static void main(String[] args) {
  }
}
```

and the Vim help says (of the findstart result):

> On the first invocation the arguments are:
   a:findstart  1
   a:base	empty

> The function must return the column where the completion starts.  It must be a
number between *zero and the cursor column "col('.')"*.  This involves looking
at the characters just before the cursor and including those characters that
could be part of the completed item.  The text between this column and the
cursor column will be replaced with the matches.

So a `findstart` of 2 represents the `@` and the replacement text is `SuppressWarnings`

Can you check my logic on that?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

puremourning:

---

*[ycmd/request_wrap.py, line 135 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkowaNm-K6dNIjxTLvt:-KnongFeHI74SkImpPCy:b-4x1lu4) ([raw file](https://github.com/valloric/ycmd/blob/d6a5700ebbf797a266f8f93675eba02cbf4e52ca/ycmd/request_wrap.py#L135)):*
> This is because the query is empty in that case (the start column is at the second punctuation mark). However, if we access the query after computing the candidates (and thus after updating the start column), the query is 'os' as expected.

You're right. I think we should clear the query when forcing the start codepoint/ start column. as it's calculated based on the actual return of the `__getitem__`, we don't need to precompute it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

puremourning:

---

*[ycmd/request_wrap.py, line 135 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkowaNm-K6dNIjxTLvt:-KnooGjL0HSkgUpIZlEX:bdor69j) ([raw file](https://github.com/valloric/ycmd/blob/d6a5700ebbf797a266f8f93675eba02cbf4e52ca/ycmd/request_wrap.py#L135)):*

> This is because the query is empty in that case (the start column is at the second punctuation mark). However, if we access the query after computing the candidates (and thus after updating the start column), the query is 'os' as expected.

You're right. I think we should clear the query when forcing the start codepoint/ start column. as it's calculated based on the actual return of the `__getitem__`, we don't need to precompute it.
</blockquote></details>

actually i can't repro your behaviour with or without the change to 'query'

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

puremourning:

---

*[ycmd/request_wrap.py, line 135 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkowaNm-K6dNIjxTLvt:-Knp-ajDf0rCJq0EdlMJ:b-pr9yhi) ([raw file](https://github.com/valloric/ycmd/blob/d6a5700ebbf797a266f8f93675eba02cbf4e52ca/ycmd/request_wrap.py#L135)):*

actually i can't repro your behaviour with or without the change to 'query'
</blockquote></details>

Going back to the omnifunc thing. Testing without YCM, (going straight to the omnifunc) I get more sensible results, so something is certainly fishy with the interaction with the omnifunc.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

puremourning:

---

*[ycmd/request_wrap.py, line 135 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkowaNm-K6dNIjxTLvt:-Knp0d_bhj8xqxALHVOC:bvv2hd) ([raw file](https://github.com/valloric/ycmd/blob/d6a5700ebbf797a266f8f93675eba02cbf4e52ca/ycmd/request_wrap.py#L135)):*

Going back to the omnifunc thing. Testing without YCM, (going straight to the omnifunc) I get more sensible results, so something is certainly fishy with the interaction with the omnifunc.
</blockquote></details>

Fixed it. The two issues are totally related. The solution was to fix the 'query' parameter as you mentioned, and to assign the start column before calling the omnifunc.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

puremourning:
Review status: 6 of 8 files reviewed at latest revision, 6 unresolved discussions.

---

*[ycmd/request_wrap.py, line 135 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkowaNm-K6dNIjxTLvt:-Knp35p2slMeh12BHE3L:b-hj0oj3) ([raw file](https://github.com/valloric/ycmd/blob/d6a5700ebbf797a266f8f93675eba02cbf4e52ca/ycmd/request_wrap.py#L135)):*

Fixed it. The two issues are totally related. The solution was to fix the 'query' parameter as you mentioned, and to assign the start column before calling the omnifunc.
</blockquote></details>

OK this is now done. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681)*

micbou:
:lgtm_strong:

---

Reviewed 2 of 2 files at r6.

---

*[ycmd/request_wrap.py, line 135 at r4](https://reviewable.io:443/reviews/valloric/ycmd/681#-KkowaNm-K6dNIjxTLvt:-KnpDP_Kz8LmupGd4G61:bccpe83) ([raw file](https://github.com/valloric/ycmd/blob/d6a5700ebbf797a266f8f93675eba02cbf4e52ca/ycmd/request_wrap.py#L135)):*

OK this is now done. 
</blockquote></details>

Awesome!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681#-:-KnpDfvrl1sq3OKgZOuc:b-jbsf8z)*

micbou:
Given the number of :lgtm:s, I think we can merge this.

@zzbot r+

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681#-:-KnqPR2_UPvvPHn-9RV9:blit5px)*

micbou:
This bot hates me.

@zzbot retry

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/681#-:-KnqQwJ1xWZGBTgXZdzE:b-yhsxbc)*

PR #682, opened 2017-01-03
Fix Type Error in Flags.FlagsForFile@completers/cpp/flags.py

sasdf:
In `completers/cpp/flags.py`, `UserIncludePaths` expect `self.FlagsForFile` to return a `list`. But it return `None` when there's no flags available, causing `TypeError` and flickering in vim.

This PR changes three `return None` to `return []` in `FlagsForFile`.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/682)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -131,14 +131,14 @@ class Flags( object ):
       if not self.no_extra_conf_file_warning_posted:
         self.no_extra_conf_file_warning_posted = True
         raise NoExtraConfDetected
-      return None
+      return []
 
     if not results or not results.get( 'flags_ready', True ):
-      return None
+      return []
 
     flags = _ExtractFlagsList( results )
     if not flags:
-      return None
+      return []
 
     if add_extra_clang_flags:
       flags += self.extra_clang_flags
--- a/ycmd/tests/clang/flags_test.py
+++ b/ycmd/tests/clang/flags_test.py
@@ -32,7 +32,21 @@ from ycmd.tests.test_utils import MacOnly
 from ycmd.responses import NoExtraConfDetected
 from ycmd.tests.clang import TemporaryClangProject, TemporaryClangTestDir
 
-from hamcrest import assert_that, calling, contains, raises, none
+from hamcrest import assert_that, calling, contains, raises
+
+
+@patch( 'ycmd.extra_conf_store.ModuleForSourceFile', return_value = Mock() )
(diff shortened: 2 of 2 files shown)


puremourning:
Thanks for sending a PR!

However, I'm not sure this is the correct approach. I thing that `UserIncludePaths` should support `flags` returning none, e.g.

```python
flags = self.FlagsForFile( ... )
if not flags:
  # Handle the case that flags is None or empty list

...
```

Also, this change needs either new tests or existing tests to be updated. The tests for this module are mostly in `ycmd/tests/clang/flags_test.py`. You can see info on setting up in `DEV_SETUP.md`.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/682#-:-K__q8jFIqU66PZ9i3EG:b-rpmcez)*

micbou:
:lgtm: since `None` and `[]` have the same [truth value](https://docs.python.org/2/library/stdtypes.html#truth-value-testing).

About the flickering, I can't reproduce in gVim. It's probably a redraw issue from Vim.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/682#-:-K__w8k8Y8cXurmnTwPz:bfqw6fc)*

Valloric:
If this is fixing a bug, then some tests either need to change or need to be written. Our current tests weren't catching this issue (obviously).

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/682#-:-K_aQH4lV53NSpSBKayQ:blpu3dn)*

sasdf:
It seems happened when extra conf set incorrectly (or not set). I think maybe it should raise a more readable error or ignore the error.

sasdf:
Tests added and conflict fixed, but I have no idea why AppVeyor with 64bit+python36 failed.
```
ERROR: Failure: ImportError (version conflict: 'c:\\python36-x64\\lib\\site-packages\\psutil\\_psutil_windows.cp36-win_amd64.pyd' C extension module was built for another version of psutil (5.0.2 instead of 5.0.1); you may try to 'pip uninstall psutil', manually remove c:\python36-x64\lib\site-packages\psutil\_psutil_windows.cp36-win_amd64.pyd or clean the virtual env somehow, then reinstall)
```

Valloric:
The CI errors seem like flakes. (_Sigh_)

This seems fine to me. Thanks for the PR!

@homu r=micbou

PR #643, opened 2016-11-13
[READY] Add regex for JavaScript and TypeScript identifiers

micbou:
Fixes https://github.com/Valloric/YouCompleteMe/issues/2438.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/643)

PR #684, opened 2017-01-07
[READY] Use non-capturing group in JavaScript identifier regex

micbou:
In PR https://github.com/Valloric/ycmd/issues/643, we added a regular expression with a capturing group for JavaScript (and TypeScript) identifiers. Unfortunately, this makes [`re.findall` returns matches of this group instead of the whole pattern](https://docs.python.org/2/library/re.html#re.findall). Since this function is used in `ExtractIdentifiersFromText`, it breaks the extraction of identifiers in JS and TS files. We fix the issue by replacing the `(...)` pattern with `(:?...)`.

Added a test that fails without the fix:
```
AssertionError: ['var', 'foo', 'require', 'bar'] != ['v', 'f', 'r', 'b']
```

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/684)

### DIFF
--- a/ycmd/identifier_utils.py
+++ b/ycmd/identifier_utils.py
@@ -68,7 +68,7 @@ FILETYPE_TO_IDENTIFIER_REGEX = {
     # Spec:
     # http://www.ecma-international.org/ecma-262/6.0/#sec-names-and-keywords
     # Default identifier plus the dollar sign.
-    'javascript': re.compile( r"([^\W\d]|\$)[\w$]*", re.UNICODE ),
+    'javascript': re.compile( r"(?:[^\W\d]|\$)[\w$]*", re.UNICODE ),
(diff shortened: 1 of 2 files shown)


Valloric:
:lgtm:

Nice, thanks!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/684#-:-K_r0tQdF7Uv4h_wLnZI:b-kdl37u)*

puremourning:
:lgtm: @homu r+

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/684#-:-K_t8C11cS2RJJUoTnar:b-2cbojr)*

PR #686, opened 2017-01-07
Update JediHTTP for Python 3.6 support

hexchain:
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/686)

PR #687, opened 2017-01-07
Update Python version in CI environments to 3.6

hexchain:
This is the follow up of https://github.com/Valloric/ycmd/pull/686#issuecomment-271075945.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/687)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,23 +7,24 @@ environment:
   USE_CLANG_COMPLETER: true
   COVERAGE: true
   matrix:
-  # We only test MSVC 11 and 12 with Python 3.5 on 64 bits.
+  # We only test MSVC 11 with Python 3.5 on 64 bits.
+  # MSVC 11 cannot build Python 3.6 extensions.
   - msvc: 11
     arch: 64
     python: 35
   - msvc: 12
     arch: 64
-    python: 35
+    python: 36
   - msvc: 14
     arch: 32
-    python: 35
+    python: 36
   # We only test Python 2.7 with MSVC 14 on 64 bits.
   - msvc: 14
     arch: 64
     python: 27
   - msvc: 14
     arch: 64
-    python: 35
+    python: 36
 install:
   - ci\appveyor\appveyor_install.bat
 build_script:


Valloric:

---

*[.travis.yml, line 30 at r1](https://reviewable.io:443/reviews/valloric/ycmd/687#-K_uKPJIJrZYpQpm5vi5:-K_uKPJIJrZYpQpm5vi6:b-wr8zro) ([raw file](https://github.com/valloric/ycmd/blob/1bb78bc67c5c297cde37f3e068ff0e088aa58e67/.travis.yml#L30)):*
> ```YAML
>     - USE_CLANG_COMPLETER=true YCMD_PYTHON_VERSION=2.6
>     - USE_CLANG_COMPLETER=true YCMD_PYTHON_VERSION=2.7
>     - USE_CLANG_COMPLETER=true YCMD_PYTHON_VERSION=3.6
> ```

Not sure this is the best approach; the minimum py3 version we support is 3.3 and we want to ensure we work with that.

I believe @micbou was talking about bumping the Windows Python version for appveyor, not travis.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/687)*

micbou:
> Python 3.6 extensions does not build with MSVC 11 so it is removed from build matrix.

Let's keep it in the build matrix with Python 3.5 for now. We'll remove it when we decide to drop its support. We should just leave a comment explaining why we don't test it with Python 3.6.

---

Reviewed 3 of 3 files at r1.

---

*[.travis.yml, line 30 at r1](https://reviewable.io:443/reviews/valloric/ycmd/687#-K_uKPJIJrZYpQpm5vi5:-K_uVgA7HeGNd8EHze_h:b9kfqkm) ([raw file](https://github.com/valloric/ycmd/blob/1bb78bc67c5c297cde37f3e068ff0e088aa58e67/.travis.yml#L30)):*

Not sure this is the best approach; the minimum py3 version we support is 3.3 and we want to ensure we work with that.

I believe @micbou was talking about bumping the Windows Python version for appveyor, not travis.
</blockquote></details>

Yes, the reason we prefer testing against the last version of Python on Windows is that, since there is no default version of Python installed on this platform, users are likely to download and install the last one so we want to be sure that it works on this version.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/687#-:-K_ujHvkl3KUYfG5-f2-:bymstnc)*

Valloric:

---

*[.travis.yml, line 30 at r1](https://reviewable.io:443/reviews/valloric/ycmd/687#-K_uKPJIJrZYpQpm5vi5:-K_voO7ks6uPhgA-Ff4i:b-k2cgk7) ([raw file](https://github.com/valloric/ycmd/blob/1bb78bc67c5c297cde37f3e068ff0e088aa58e67/.travis.yml#L30)):*

Yes, the reason we prefer testing against the last version of Python on Windows is that, since there is no default version of Python installed on this platform, users are likely to download and install the last one so we want to be sure that it works on this version.
</blockquote></details>

That makes sense. So we're on the same page that the travis file shouldn't be changed then.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/687)*

micbou:
:lgtm:

---

Reviewed 2 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/687#-:-K_yAPeRGUSbSfVIiumo:bnfp4nl)*

puremourning:
:lgtm: Thanks again!

@homu r+

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/687#-:-K_zFbwLNlz13X5hNRVM:b-5om8j1)*

puremourning:
Reviewed 2 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/687)*

PR #676, opened 2016-12-22
Update racerd to 0.3.2

jwilm:
Has fix for issue with builds on newest rust compiler.

cc jwilm/racerd#60

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/676)

PR #689, opened 2017-01-11
Update to racerd 0.3.3.

rillian:
Pick up racer 2.0 and fix build issues with rust 1.15.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/689)


rillian:
I see an update to 0.3.2, which fixes the same build failure I hit, is blocked on a ci failure in #676. We'll see what travis says about this one.

PR #696, opened 2017-01-27
[WIP] Minimise boost library usage by using C++11 feautres

bstaletic:
This pull request makes use of C++11 featuure, by mostly switching from `boost::` to `std::`. In some parts of the code some non-trivial refactoring needed to be done as well.

Changes planned:

- [x] Change `boost::regex` to its `std` counterpart
- [x] Change `boost::function` to its `std` counterpart
- [x] Remove `boost::assign::list_of`
- [x] Change `boost::shared_ptr`, `boost::make_shared` and `boost::remove_ptr` to its `std` counterparts
- [x] Change `boost::mutex`, `boost::unique_lock` and `boost::try_to_lock_t` to its `std` counterpart
- [x] Change `boost::unique_ptr` to its `std` counterpart
- [x] Change `boost::unordered_set` to its `std` counterpart
- [x] Change `boost::scoped_ptr` to `std::unique_ptr` - [Reference](https://stackoverflow.com/questions/21852612/alternatives-to-boostscoped-ptr-in-c-11/21852657#21852657)
- [x] Change `boost::alogrithm` to its `std` counterpart
- [x] Change `boost::unordered_map` to its `std` counterpart
- [x] Remove `boost::exception`
- [x] Use range based for instead of BOOST_FOREACH
- [ ] Change `boost::hash` to `std::hash`
- [ ] Don't depend on boost for noncopyable classes
- [ ] Change `CMakeLists.txt` to reflect the new changes
- [ ] Bump ycmd version

Boost headers left in the code:

- `boost/python.hpp`
- `boost/filesystem.hpp`

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/696)

### DIFF
--- a/cpp/ycm/Candidate.cpp
+++ b/cpp/ycm/Candidate.cpp
@@ -19,18 +19,21 @@
 #include "Candidate.h"
 #include "Result.h"
 
-#include <boost/algorithm/string.hpp>
+#include <algorithm>
 #include <cctype>
 #include <locale>
 
-using boost::algorithm::all;
-using boost::algorithm::is_lower;
-using boost::algorithm::is_print;
+using std::all_of;
+using std::islower;
+using std::isprint;
 
 namespace YouCompleteMe {
 
 bool IsPrintable( const std::string &text ) {
-  return all( text, is_print( std::locale::classic() ) );
+  for ( auto character : text )
+    if ( !isprint( character ) )
+      return false;
+  return true;
 }
 
 
@@ -75,9 +78,17 @@ Candidate::Candidate( const std::string &text )
   :
   text_( text ),
   word_boundary_chars_( GetWordBoundaryChars( text ) ),
-  text_is_lowercase_( all( text, is_lower() ) ),
   letters_present_( LetterBitsetFromString( text ) ),
   root_node_( new LetterNode( text ) ) {
+    text_is_lowercase_ = true;
+    for ( auto character : text )
+    {
+      if ( islower( character ) )
+      {
+        text_is_lowercase_ = false;
+	break;
+      }
+    }
 }
 
 
--- a/cpp/ycm/Candidate.cpp
+++ b/cpp/ycm/Candidate.cpp
@@ -63,7 +63,7 @@ std::string GetWordBoundaryChars( const std::string &text ) {
 Bitset LetterBitsetFromString( const std::string &text ) {
   Bitset letter_bitset;
 
-  foreach ( char letter, text ) {
+  for ( char letter : text ) {
     int letter_index = IndexForLetter( letter );
 
     if ( IsInAsciiRange( letter_index ) )
@@ -97,7 +97,7 @@ Result Candidate::QueryMatchResult( const std::string &query,
   LetterNode *node = root_node_.get();
   int index_sum = 0;
 
-  foreach ( char letter, query ) {
+  for ( char letter : query ) {
     const NearestLetterNodeIndices *nearest =
       node->NearestLetterNodesForLetter( letter );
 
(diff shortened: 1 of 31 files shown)


bstaletic:
Comment on the unfinished stuff:

####boost::hash

This is already removed everywhere but in `TranslationUnitStore.cpp`. In that file `boost::hash<std::vector<std::string>>` is used and the `std::hash` doesn't provide the necessary specialisation of the class for `hash<vector<T>`.

####noncopyable

I have tried making a `Noncopyable.h` in `cpp/ycm` directory and making every class inhherit that instead of the boost provided variant. This made boost::python unhappy and complained about a missing boost typedef.

#### Failing tests
Three tests are failing when run from `vagrant` invironment.
Tests:

- IdentifierCompleterTest.TagsEndToEndWorks
- IdentifierUtilsTest.ExtractIdentifiersFromTagsFileWorks
- LetterNodeTest.ThrowOnNonAsciiCharacters

The first failing test I have no idea why would it fail. I tried `vim a.cpp` and typing "fo". I got offered "foosy" and "fooaaa" as identifires. Tags file in use was the same as `base.tags`.

The second test, very strangely, fails only on C#.

For the third I have no idea how to make it throw an exception. The function in question returns a pointer to, what looks to me, a valid address.

bstaletic:
The commit which introduced LetterNodeTest failure is 0ff8121.

This commit literally only changes `std::` to `boost::` and makes appropriate changes to `#include` directives. Currently I have no idea why would it behave differently than the boost version.

vheon:
For starters thanks for the PR ;) Anyway since the way this is going I think that making smaller PRs would be better. I don't know what the others think but I would do it in this kind of steps:

- Check and force for a C++11 compiler
- Switch to C++11 counterpart where replacing `boost::` with `std::` with the proper headers is enough
- Switch to C++11 counterpart where a trivial refactoring is needed
- Switch to C++11 where refactoring is non trivial

I would do the first 2 step in one PR and the rest as single PR (especially the last one where actually switching to the `std` counterpart would require an assessment and discussion).

This way actually reviewing the code would be easier and having the first point will ensure that the tests in the build environment are actually passing.

@micbou @puremourning @Valloric thoughts?

Valloric:
Thanks for the PR! This must have been a ton of work and we very much appreciate it.

Can't really disagree with @vheon's proposed plan; it sounds very reasonable.

I would like to add that we really do must know why the tests that fail do so. Even if we decided to change the expectation of an assert, we must do so with informed knowledge of what changed and why.

---

Review status: 0 of 30 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/696#-:-KbYfuT-qFVWPNbpqJX5:b8nit6x)*

Valloric:
Also, if these imports/refactorings/changes impact code that doesn't have coverage (the C++ parts don't have great coverage), we should really have a test that passes before the change and after it. These swaps from boost to std undeniably come with behavioral changes and I wouldn't be very comfortable switching to the std variant without tests proving that nothing broke.

The current code may not have tests, but it does have several years of production mileage on it.

---

Review status: 0 of 30 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/696#-:-KbYgrfEB1CA_sDMHoLm:b7v8jfr)*

bstaletic:
Okay, I'll tart with a new PR containing only the simplest of changes. I won't be including any changes that make tests fail for the first PR.

As for coverage, I'll try to cover as much of the code as I can.

Closing this PR and will open another as soon as I have time.

PR #697, opened 2017-01-28
Switch to C++11 from boost where only the namespace change is enough

bstaletic:
As a continuation of #696 this makes only the most trivial part of the transition.

No new tests added, but I will get back to those. On the up side, the existing tests aren't failing.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/697)


bstaletic:
Okay, so build fails on appveyor because "Your C++ compiler does NOT support C++11". I have obviously missed something important in `appveyor.yml`.

And travis fails on OSX reporting a lot of errors in system files, complaining about syntax errors? That can't be right.

vheon:
> Okay, so build fails on appveyor because "Your C++ compiler does NOT support C++11"

Looks like we never se `CPP11_AVAILABLE` for MSVC so that is why.

vheon:
https://cmake.org/cmake/help/v3.7/variable/MSVC_VERSION.html this is the variable to get the version of Visual Studio

vheon:
But looking at CMake more I think we should use [CXX_STANDARD_REQUIRED](https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD_REQUIRED.html) and [CXX_STANDARD](https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html) instead.

vheon:
I gave a preliminary review on this :) One thing that I didn't see used in this first pass the use of `auto`. I'm don't like `auto` everywhere but I love it where 

- the the type is actually repeated on the same line:

```cpp
std::shared_ptr<Type> t = std::make_shared<Type>()
// instead I prefer
auto t = std::make_shared<Type>()
```
- the type is not actually clear in the first case:

```cpp
for ( const FilepathToCandidates::value_type & path_and_candidates :  *it->second )

// instead of
for ( const auto& path_and_candidates: *it->second )
```


One thing that should also be updated is the `update_boost.py` script to remove the boost library no longer needed.

in this case I prefer the shorter version :)

Of course this is only my opinion so also here I would like to hear from the others @micbou @puremourning @valloric 

---

Reviewed 23 of 23 files at r1.
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[cpp/CMakeLists.txt, line 185 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_YKG20ce-aGYrlj_8:-Kb_YKG20ce-aGYrlj_9:b5xxfzt) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/CMakeLists.txt#L185)):*
> ```CMake
>   endif()
> else()
>   message( FATAL_ERROR
> ```

As said in the main discussion, I think we should use CXX_STANDARD_REQUIRED and CXX_STANDARD.

---

*[cpp/ycm/CandidateRepository.cpp, line 20 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_ZtQGGvSNbZ1LsBbx:-Kb_ZtQHyKAjTlnRyiv9:b-x5ieno) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/CandidateRepository.cpp#L20)):*
> ```cpp
> #include "CandidateRepository.h"
> #include "Candidate.h"
> #include "standard.h"
> ```

I think this is not needed anymore in this file.

---

*[cpp/ycm/CandidateRepository.cpp, line 24 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_Yx1RrnwiTAAN3gg6:-Kb_Yx1RrnwiTAAN3gg7:bkrxi5o) ([raw file](https://github.com/valloric/ycmd/blob/717b36a76ebdb962bc1a886d262ff013519ceeff/cpp/ycm/CandidateRepository.cpp#L24)):*
> ```cpp
> 
> #include <boost/thread/locks.hpp>
> #include <boost/algorithm/string.hpp>
> ```

Wasn't this header used at all here or I am missing something?

---

*[cpp/ycm/IdentifierCompleter.cpp, line 19 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb__Q_izpBT9FNuRppp:-Kb__Q_izpBT9FNuRppq:b-x5ieno) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/IdentifierCompleter.cpp#L19)):*
> ```cpp
> 
> #include "IdentifierCompleter.h"
> #include "standard.h"
> ```

I think this is not needed anymore in this file.

---

*[cpp/ycm/IdentifierDatabase.cpp, line 19 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb__XdVezxa9G1xoabB:-Kb__XdVezxa9G1xoabC:b-x5ieno) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/IdentifierDatabase.cpp#L19)):*
> ```cpp
> 
> #include "IdentifierDatabase.h"
> #include "standard.h"
> ```

I think this is not needed anymore in this file.

---

*[cpp/ycm/IdentifierUtils.cpp, line 20 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_beIErYIDNA4I_PkV:-Kb_beIErYIDNA4I_PkW:b-n1kz9y) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/IdentifierUtils.cpp#L20)):*
> ```cpp
> #include "IdentifierUtils.h"
> #include "Utils.h"
> #include "standard.h"
> ```

I'm not sure why this is here in the first place :confused: 

---

*[cpp/ycm/PythonSupport.cpp, line 19 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_bxAGkJNUc_pyUJcb:-Kb_bxAGkJNUc_pyUJcc:bel7fa5) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/PythonSupport.cpp#L19)):*
> ```cpp
> 
> #include "PythonSupport.h"
> #include "standard.h"
> ```

I think this is not needed here anymore.

---

*[cpp/ycm/Result.cpp, line 19 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_cAetQSFq0v8Yl4rW:-Kb_cAetQSFq0v8Yl4rX:bfxx65q) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/Result.cpp#L19)):*
> ```cpp
> 
> #include "Result.h"
> #include "standard.h"
> ```

also here I'm not sure why this is here :confused: 

---

*[cpp/ycm/ClangCompleter/ClangCompleter.cpp, line 35 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_cjTQREPmLGEo5vQB:-Kb_cjTQREPmLGEo5vQC:bngx1jy) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/ClangCompleter/ClangCompleter.cpp#L35)):*
> ```cpp
> 
> using boost::shared_ptr;
> using std::unordered_map;
> ```

actually I'm not sure `unordered_map` is actually used in this TU

---

*[cpp/ycm/ClangCompleter/TranslationUnit.h, line 29 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_eTaxGZ7b33B7hWWn:-Kb_eTayVI9t-0I2hcSf:bcuwv5u) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/ClangCompleter/TranslationUnit.h#L29)):*
> ```objc
> #include <clang-c/Index.h>
> #include <boost/utility.hpp>
> #include <boost/shared_ptr.hpp>
> ```

why add this header here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697#-:-Kb__my8FMNCj_fsgB2C:bsmbi7p)*

bstaletic:
Review status: 15 of 23 files reviewed at latest revision, 10 unresolved discussions.

---

*[cpp/CMakeLists.txt, line 185 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_YKG20ce-aGYrlj_8:-Kb_wP-4scJx0WAtfA99:bb8fzzl) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/CMakeLists.txt#L185)):*

As said in the main discussion, I think we should use CXX_STANDARD_REQUIRED and CXX_STANDARD.
</blockquote></details>

Is it enough to just remove that block and the block that checks for `CPP11_AVAILABLE`?
As far as I can tell it would be enough, as `CXX_STANDARD` first tries to set `-std=gnu++11` and gracefully degrades if that is not available.

---

*[cpp/ycm/CandidateRepository.cpp, line 20 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_ZtQGGvSNbZ1LsBbx:-Kb_xfAXWCLhAo94IzgZ:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/CandidateRepository.cpp#L20)):*

I think this is not needed anymore in this file.
</blockquote></details>

Done.

---

*[cpp/ycm/CandidateRepository.cpp, line 24 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_Yx1RrnwiTAAN3gg6:-Kb_xfqxyD0cqRhQ4uyU:b-49kzad) ([raw file](https://github.com/valloric/ycmd/blob/717b36a76ebdb962bc1a886d262ff013519ceeff/cpp/ycm/CandidateRepository.cpp#L24)):*

Wasn't this header used at all here or I am missing something?
</blockquote></details>

It wasn't used. Removed.

---

*[cpp/ycm/IdentifierCompleter.cpp, line 19 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb__Q_izpBT9FNuRppp:-Kb_xly8GTGSWwihxjNv:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/IdentifierCompleter.cpp#L19)):*

I think this is not needed anymore in this file.
</blockquote></details>

Done.

---

*[cpp/ycm/IdentifierDatabase.cpp, line 19 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb__XdVezxa9G1xoabB:-Kb_xn-5OzegyQhIEl5i:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/IdentifierDatabase.cpp#L19)):*

I think this is not needed anymore in this file.
</blockquote></details>

Done.

---

*[cpp/ycm/IdentifierUtils.cpp, line 20 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_beIErYIDNA4I_PkV:-Kb_xqW6AzzMKDTWvVX8:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/IdentifierUtils.cpp#L20)):*

I'm not sure why this is here in the first place :confused: 
</blockquote></details>

Done.

---

*[cpp/ycm/PythonSupport.cpp, line 19 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_bxAGkJNUc_pyUJcb:-Kb_xsqBw3Hz-dGNZIQZ:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/PythonSupport.cpp#L19)):*

I think this is not needed here anymore.
</blockquote></details>

Done.

---

*[cpp/ycm/Result.cpp, line 19 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_cAetQSFq0v8Yl4rW:-Kb_xv4FLyiY0URQxQDP:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/Result.cpp#L19)):*

also here I'm not sure why this is here :confused: 
</blockquote></details>

Done.

---

*[cpp/ycm/ClangCompleter/ClangCompleter.cpp, line 35 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_cjTQREPmLGEo5vQB:-Kb_xvdTgPRQ8n0nLca0:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/ClangCompleter/ClangCompleter.cpp#L35)):*

actually I'm not sure `unordered_map` is actually used in this TU
</blockquote></details>

Done.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.h, line 29 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_eTaxGZ7b33B7hWWn:-Kb_xxAsx_lkSesELILq:bup6g3k) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/ycm/ClangCompleter/TranslationUnit.h#L29)):*

why add this header here?
</blockquote></details>

GCC failed to compile because of the (also unused) typedef a few line lower. Removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697)*

bstaletic:
Pushed a commit that cleans up the include statements.
Still need to fix the CMakeLists.txt and update_boost.py

---

Review status: 15 of 23 files reviewed at latest revision, 10 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697#-:-Kb_z1yiaWvv9S4-Irgp:b-dseikl)*

Valloric:
Looks great so far!

What I'm utterly puzzled by is the 1.42% fall in coverage. Why would that happen? Are we including boost code in our coverage runs?

Agreed with @vheon on `auto` usage, though I don't think we need to gate these PRs on adding `auto` (we could do that later). I see these PRs as getting us onto C++11 and off of boost where we can. 

---

Review status: 15 of 23 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697#-:-KbaAXabISUr0wV-VviT:by1ij84)*

vheon:
Reviewed 8 of 8 files at r2.

---

*[cpp/CMakeLists.txt, line 185 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_YKG20ce-aGYrlj_8:-KbaCYmSSUuW_9zZQOH1:bbz9ksg) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/CMakeLists.txt#L185)):*

Is it enough to just remove that block and the block that checks for `CPP11_AVAILABLE`?
As far as I can tell it would be enough, as `CXX_STANDARD` first tries to set `-std=gnu++11` and gracefully degrades if that is not available.
</blockquote></details>

well we can't use `CXX_STANDARD` without `CXX_STANDARD_REQUIRED` because we're using C++11 features. Anyway I think that if we use these two variables yes we can remove these two blocks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697)*

bstaletic:
@Valloric
The only file where codecov reports actual misses for the chaanged code are two lines in `CandidateRepository.cpp`.
More precisely it's the `std::vector< const Candidate * > GetCandidatesForString( std::vector< CompletionData > &datas )`

Should I try to write a test for that function? It isn't covered at all.

---

Review status: 22 of 23 files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/CMakeLists.txt, line 185 at r1](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kb_YKG20ce-aGYrlj_8:-KbaTuaoMrRv7zf3vxnm:b-2ag69r) ([raw file](https://github.com/valloric/ycmd/blob/864c1f1813d26a0c5136c894ce5edd80561cf7f2/cpp/CMakeLists.txt#L185)):*

well we can't use `CXX_STANDARD` without `CXX_STANDARD_REQUIRED` because we're using C++11 features. Anyway I think that if we use these two variables yes we can remove these two blocks.
</blockquote></details>

Made that change. Let's see if appveyor is satisfied with it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697#-:-KbaU7ZS94_hdqggmXdV:bo3gqt9)*

bstaletic:
Travis on linux is not using `-std=c++11`, while it works on OSX.
This error got introduced with the usage of `CXX_STANDARD` and `CXX_STANDARD_REQUIRED`. It also started failing on vagrant virtual machine.

Oddly enough, it is working on my x86_64 linuxmachine.

bstaletic:
`CMAKE_CXX_STANDARD` requires `cmake` 3.1 or newer. `ycmd`'s top level `CMakeLists.txt` says the newest `cmake` on RHEL is 2.8.11. So, are we sure we should use the `CMAKE_CXX_STANDARD`?

vheon:
> The only file where codecov reports actual misses for the chaanged code are two lines in CandidateRepository.cpp.
> More precisely it's the std::vector< const Candidate * > GetCandidatesForString( std::vector< CompletionData > &datas )
> Should I try to write a test for that function? It isn't covered at all.

Don't, I actually noticed that the function wasn't covered at all and looked at it and in fact is not used at all so I don't think there is a need to cover it (I was actually going to open a PR to delete it entirely).

> CMAKE_CXX_STANDARD requires cmake 3.1 or newer. ycmd's top level CMakeLists.txt says the newest cmake on RHEL is 2.8.11. So, are we sure we should use the CMAKE_CXX_STANDARD?

Mh, looks like I didn't check the version of CMake required by that :confused: I don't feel the actual need to require a bump in the CMake version; maybe we bring back the `CPP11_AVAILABLE` block of code and do the check on the various compilers ourselves?

---

Reviewed 1 of 2 files at r3, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697#-:-KbbVF7VC9-2fMRJjtZ6:b-62gj7y)*

bstaletic:
> I don't feel the actual need to require a bump in the CMake version; maybe we bring back the CPP11_AVAILABLE

My thoughts exactly. I just didn't want to create spam.

Before I push the change, does this look okay?
```cmake
if ( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" )
  execute_process(
    COMMAND ${CMAKE_CXX_COMPILER} -dumpversion OUTPUT_VARIABLE GCC_VERSION)
  if ( GCC_VERSION VERSION_GREATER 4.6 OR GCC_VERSION VERSION_EQUAL 4.6 )
    set( CPP11_AVAILABLE true )
  endif()
elseif( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" )
    set( CPP11_AVAILABLE true )
    set( CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LANGUAGE_STANDARD "c++11" )
    set( CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LIBRARY "libc++" )
elseif( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC" )
  if ( "${MSVC_VERSION}" EQUAL 1800 OR "${MSVC_VERSION}" GREATER 1800 )
    set( CPP11_AVAILABLE true )
  endif()
endif()
```

And then deciding on the flags like this:
```cmake
if ( CPP11_AVAILABLE )
  # Cygwin needs its hand held a bit; see issue #473
  if ( CYGWIN AND CMAKE_COMPILER_IS_GNUCXX )
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
  else()
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
  endif()
else()
  message( FATAL_ERROR
    "Your C++ compiler does NOT support C++11." )
endif()
```

Should we also check that clang is at least version 3.3? Before that clang didn't have full C++11 feature set available.

vheon:
>Should we also check that clang is at least version 3.3?

Well I think we should.

bstaletic:
Made a typo. Will fix soon. A quick question. What flag should be passed to MSVC to enable C++11? I can't seem to find anything. It can't be just not setting anything and settling on the correct version, could it?

bstaletic:
Okay, what's going on with travis on OSX?
I can't make sense of its errors.

When these cmake annoyances are resolved, I will squash all the tiny commits that only touch CMakeLists.txt.

I have just tried changing almost every smart pointer from boost to std and all tests are passing. So after the squash I have another useful commit prepared.

bstaletic:
I tried to compile ycmd with these changes on FreeBSD and it went without and problems. So I can't reproduce the error seen on travis on OSX.

Valloric:
> What flag should be passed to MSVC to enable C++11?

I don't think there's a flag in MSVC to turn on C++11; the version of the compiler that support C++11 just do so out-of-the-box.

WRT OS X errors... they do look weird. It might be that the wrong version of the C++ standard library is being used. AFAIR, clang can work with both the gcc C++ std library (libstdc++) or the clang C++ std library (libc++), but the clang one is the better supported one.

Valloric:
Reviewed 14 of 23 files at r1, 8 of 8 files at r2, 1 of 2 files at r3, 1 of 1 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697)*

puremourning:
Xcode provides the standard library on OSX. I'm sorry I can't try it out here because I'm on restricted typing due to RSI. :/

Valloric:
> I'm sorry I can't try it out here because I'm on restricted typing due to RSI. :/

Dude, I feel so bad for you! I had RSI issues and I wouldn't wish them on anyone. Rest up & take care of yourself!

bstaletic:
Looking around the internet I have found [this](https://github.com/Marqt/ViZDoom/issues/56), an issue on a random github repository with the exact issue this pull request has.

Should I try that solution? 

Also [this](https://bugs.python.org/issue10910) seems to suggest a conflict between boost::python (actually python - specifically `pyport.h`) on one side and clang and C++11 on another. `pyport.h` is included in in `Python.h` which is included in `boost/python.h`. 

> The problem is that pyport tries to replace the ctypes definion[s]

Strangely enough, by the end of that thread, they say the bug is resolved.

bstaletic:
Also, seems like it was fixed for python3.6 and python3.5. I don't feel like we should just drop support for earlier python versions.

bstaletic:
Looking up latest python versions for RHEL and Debian stable, I really don't think that dropping support for python2.7.12 and earlier is an option, becuse that means also dropping the support for those two distributions.

Also dropping support for python3.5.2 and earlier means dropping support for every BSD out there, as well as Ubuntu.

This data is taken from distrowatch, so I am not sure it is completely accurate.

Valloric:
Good find!

The issue report you link to has a link to a fix: 

> It seems that `<iostream>` is necessary and it has to be included before `<Python.h>`

Why can't we just do that? Seems like it might solve the problem.

bstaletic:
I'll do that right now. I was just waiting for someone to give approval, as that solution seems too hackish to my liking and I don't like making hackish solutions without consulting others.

puremourning:
Iostream include seems to fix the build at least :D

bstaletic:
I'll now squash the commits regarding the cmake standard and compilation issue. Then I'll push another useful commit.

What's up with coverage? It is reporting changes in ycmd directory which I have never touched.

bstaletic:
Hmm... The same workaround didn't work on ycm_core.cpp.
I will now try the workaround proposed in the python bug tracker.

bstaletic:
Okay, that didn't work as well.

So, `pyport.h` is redefining things assuming they are macros set in `ctype.h`, but C++ redefines these as functions in`cctype`.

The first time, we solved it with `iostream` which includes `cctype`.

@vheon @Valloric @micbou 
What do you suggest?

Valloric:
Ugh, this sucks... can importing Python.h first help here, like python bug tracker suggested? Agreed these are all hacks, but we don't seem to have a choice here. Either we add the hacks, we don't switch to C++11 or we drop support for the older versions of Python. The last one is a deal breaker.

@bstaletic Thanks for putting all this work in BTW!

bstaletic:
@Valloric 

> can importing Python.h first help

I thought I did that laready, but I realise I made a mistake. I'll push a new commit right now.

bstaletic:
Why would linking fail on appveyor?
It's complaining about not being able to find a certain file.
On python2.7:

> cannot open file 'python27_d.lib'

On python3.6:

> cannot open file 'python36_d.lib'

Should I expect that the following is a random error that shouldn't be repeatable?

```
ERROR: ycmd.tests.shutdown_test.Shutdown_test.FromHandlerWithSubservers_test

  File "/home/travis/.pyenv/versions/2.7.8/lib/python2.7/site-packages/nose/case.py", line 197, in runTest

    self.test(*self.arg)

  File "/home/travis/build/Valloric/ycmd/ycmd/tests/client_test.py", line 260, in Wrapper

    test( self, *args )

  File "/home/travis/build/Valloric/ycmd/ycmd/tests/shutdown_test.py", line 56, in FromHandlerWithSubservers_test

    self.StartSubserverForFiletype( filetype )

  File "/home/travis/build/Valloric/ycmd/ycmd/tests/client_test.py", line 165, in StartSubserverForFiletype

    self._WaitUntilReady( filetype )

  File "/home/travis/build/Valloric/ycmd/ycmd/tests/client_test.py", line 142, in _WaitUntilReady

    'aborting.'.format( server, timeout ) )

RuntimeError: Waited for the cs subserver to be ready for 5 seconds, aborting.
```

I'll try again with <iostream> as the very first included file.


By the way, can this even be called programming if we're just trying to find the specific order of include statements that would just happen to work?

bstaletic:
Well, the builds are finally passing! 😃 

And can anyone explain to me what's going on with codecov?
I most certainly didn't add 5000 lines.

puremourning:
Codecov is probably bugging. It's possible that there are leftover gcda files or something. But that shouldn't happen. We could ask their Gitter channel. They are pretty good folks

micbou:
Thanks a lot for doing this work.

We probably want to remove [the `-Wc++98-compat` flag from our `.ycm_extra_conf.py`](https://github.com/Valloric/ycmd/blob/master/cpp/ycm/.ycm_extra_conf.py#L41).

> One thing that should also be updated is the update_boost.py script to remove the boost library no longer needed.

It should be done in a separate PR to not pollute this one. I'll look into it.

---

Reviewed 8 of 23 files at r1, 6 of 8 files at r2, 1 of 2 files at r3, 1 of 1 files at r6, 12 of 13 files at r7, 1 of 1 files at r10.

---

*[appveyor.yml, line 10 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-KbovAYW9HZCvKpZRqkF:-KbovAYW9HZCvKpZRqkG:bnwqjhu) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/appveyor.yml#L10)):*
> ```YAML
>   COVERAGE: true
>   matrix:
>   # MSVC 11 not tested since we require C++11 features.
> ```

I tried these changes with MSVC 11 and it's compiling fine so, even if its C++11 support is really limited, we should keep it in the build matrix while it's still working.

---

*[cpp/CMakeLists.txt, line 117 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbonh1ZiNPTx2mXKpcy:-Kbonh1ZiNPTx2mXKpcz:brsj4va) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/CMakeLists.txt#L117)):*
> ```CMake
> # Determining the presence of C++11 support in the compiler
> set( CPP11_AVAILABLE false )
> if ( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" )
> ```

Why don't we still use `CMAKE_COMPILER_IS_GNUCXX` and `COMPILER_IS_CLANG`?

---

*[cpp/CMakeLists.txt, line 127 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbonu9fUVuhc4vmUuLR:-Kbonu9gDE6lObzxkG3i:bx0h9d2) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/CMakeLists.txt#L127)):*
> ```CMake
>     set( CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LIBRARY "libc++" )
>   endif()
> elseif( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC" )
> ```

This condition can be replaced with just `MSVC`.  This will fix the following warning with CMake 3.7:
```
CMake Warning (dev) at CMakeLists.txt:117 (if):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.
  Quoted variables like "MSVC" will no longer be dereferenced when the policy
  is set to NEW.  Since the policy is not set the OLD behavior will be used.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

---

*[cpp/CMakeLists.txt, line 131 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbowr8FNgYMiI5md_50:-Kbowr8FNgYMiI5md_51:b-6x10y9) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/CMakeLists.txt#L131)):*
> ```CMake
>     set( CPP11_AVAILABLE true )
>   endif()
> endif()
> ```

Missing blank line.

---

*[cpp/CMakeLists.txt, line 188 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbowsroiy_e7UyEQuoY:-Kbowsroiy_e7UyEQuoZ:b-cj4ygb) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/CMakeLists.txt#L188)):*
> ```CMake
> else()
>   message( FATAL_ERROR
>     "Your C++ compiler does NOT support C++11." )
> ```

There is now enough place to put this on one line.

---

*[cpp/ycm/standard.h, line 18 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbp-u5PG3AY0OFeYW-_:-Kbp-u5PG3AY0OFeYW-a:b-ywj6f5) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/ycm/standard.h#L18)):*
> ```objc
> // along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
> 
> typedef unsigned int uint;
> ```

Having a header file just for that looks dumb. Should we throw away this `typedef` and always use `unsigned int` or move it elsewhere?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697#-:-Kboxaou8NpyIRDm9XeA:b-tyib1b)*

bstaletic:
Review status: 27 of 30 files reviewed at latest revision, 6 unresolved discussions.

---

*[appveyor.yml, line 10 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-KbovAYW9HZCvKpZRqkF:-KbpBD51hlNZD50De2yX:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/appveyor.yml#L10)):*

I tried these changes with MSVC 11 and it's compiling fine so, even if its C++11 support is really limited, we should keep it in the build matrix while it's still working.
</blockquote></details>

Done.

---

*[cpp/CMakeLists.txt, line 117 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbonh1ZiNPTx2mXKpcy:-Kbp3OeEEQIkdJ6V6Ng7:blft7r2) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/CMakeLists.txt#L117)):*

Why don't we still use `CMAKE_COMPILER_IS_GNUCXX` and `COMPILER_IS_CLANG`?
</blockquote></details>

I made that change because there is no `COMPILER_IS_MSVC` and I thought this was a more consistent way of checking.

Considering a warining from CMake 3.7 (which I did not notice) I'll change it back.

---

*[cpp/CMakeLists.txt, line 127 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbonu9fUVuhc4vmUuLR:-KbpBJirAVwIGynMKFCr:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/CMakeLists.txt#L127)):*

This condition can be replaced with just `MSVC`.  This will fix the following warning with CMake 3.7:
```
CMake Warning (dev) at CMakeLists.txt:117 (if):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.
  Quoted variables like "MSVC" will no longer be dereferenced when the policy
  is set to NEW.  Since the policy is not set the OLD behavior will be used.
This warning is for project developers.  Use -Wno-dev to suppress it.
```
</blockquote></details>

Done.

---

*[cpp/CMakeLists.txt, line 131 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbowr8FNgYMiI5md_50:-KbpBLvj8km2Cnph2Iex:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/CMakeLists.txt#L131)):*

Missing blank line.
</blockquote></details>

Done.

---

*[cpp/CMakeLists.txt, line 188 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbowsroiy_e7UyEQuoY:-KbpBSVsbNrGXB4kzd0q:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/CMakeLists.txt#L188)):*

There is now enough place to put this on one line.
</blockquote></details>

Done.

---

*[cpp/ycm/standard.h, line 18 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbp-u5PG3AY0OFeYW-_:-KbpBV2qCKJA_MyPsT9G:b4t2b02) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/ycm/standard.h#L18)):*

Having a header file just for that looks dumb. Should we throw away this `typedef` and always use `unsigned int` or move it elsewhere?
</blockquote></details>

I agree that a header file for only one typedef is silly.

I may have a better proposal. Why shoulnd't `size_t` be used? It is supposed to represent an unsigned value for variables representing things like lengths and sizes, which naturally can't be negative.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697)*

bstaletic:
@puremourning

I posted a question on their gitter. Now I'm waiting for a reply.

---

Review status: 27 of 30 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697#-:-KbpCLL0YOq7b7bq8TVs:b-bkleyv)*

Valloric:
I'm :lgtm: on this. @micbou pointed out some more  issues, but they're all minor.  Travis is having an outage for mac builds so that's why stuff is failing, but when that gets sorted out, I'm fine with merging this (once the minor stuff is resolved). Don't worry too much about coveralls.

@bstaletic Thank you for this PR! The C++11 port has been a long time coming and it was dreary stuff few were excited about. So your work is _highly_ appreciated.

---

Reviewed 12 of 13 files at r7, 1 of 1 files at r10, 3 of 3 files at r11.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697#-:-KbvCmCrKv7UtkeeA9lc:brh62lg)*

bstaletic:
@Valloric 
I got a response from Steve about codecov. He asked for a commit in question. He asked about the commit that causes the issues and later fixed it.

But, I thought he was asking for the frst commit that was misbehaving, so I said it was 98c9507. Now, to me, it seems like I should have said 8e04288 as it is the lates one.

So how do we proceed?

bstaletic:

---

*[cpp/ycm/standard.h, line 18 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbp-u5PG3AY0OFeYW-_:-KbvHV568OvBjPAB5kaO:b-qkmru3) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/ycm/standard.h#L18)):*

I agree that a header file for only one typedef is silly.

I may have a better proposal. Why shoulnd't `size_t` be used? It is supposed to represent an unsigned value for variables representing things like lengths and sizes, which naturally can't be negative.
</blockquote></details>

So, `size_t` or `unsigned int`?
Those should be defined to the same thing, the only question is which one is more clear.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697)*

vheon:
Reviewed 1 of 1 files at r6, 12 of 13 files at r7, 1 of 1 files at r10, 3 of 3 files at r11.

---

*[appveyor.yml, line 10 at r11](https://reviewable.io:443/reviews/valloric/ycmd/697#-KbvNvkB_ibThN-67NJg:-KbvNvkB_ibThN-67NJh:b-v4uvw0) ([raw file](https://github.com/valloric/ycmd/blob/717b36a76ebdb962bc1a886d262ff013519ceeff/appveyor.yml#L10)):*
> ```YAML
>   COVERAGE: true
>   matrix:
>   # We only test MSVC 11 with Python 3.5 on 64 bits.
> ```

I believe at this point it doesn't make sense to change this file at all :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697)*

bstaletic:

---

*[appveyor.yml, line 10 at r11](https://reviewable.io:443/reviews/valloric/ycmd/697#-KbvNvkB_ibThN-67NJg:-KbvTgeOBa799assqYEg:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/717b36a76ebdb962bc1a886d262ff013519ceeff/appveyor.yml#L10)):*

I believe at this point it doesn't make sense to change this file at all :)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697)*

micbou:
At least the badge in the README.md was fixed (it was red). Let's wait for the tests to complete and see if the codecov report is still bugged.

---

Reviewed 3 of 3 files at r11, 1 of 1 files at r12.

---

*[cpp/ycm/standard.h, line 18 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbp-u5PG3AY0OFeYW-_:-KbvRUkabSW6Hyd0gMaJ:b-k2gzl9) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/ycm/standard.h#L18)):*

So, `size_t` or `unsigned int`?
Those should be defined to the same thing, the only question is which one is more clear.
</blockquote></details>

`size_t` is the right choice but I think we should do that in another PR. This one is already large enough. For now, we can just add a `TODO` comment about replacing it with `size_t` and removing the whole header.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697#-:-KbvU7fo76yiYQJkx-Kw:b-94a3et)*

micbou:

---

*[cpp/ycm/standard.h, line 18 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbp-u5PG3AY0OFeYW-_:-KbvaKRD5sC2Nclyi8sQ:bmochh7) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/ycm/standard.h#L18)):*

Then how about not adding a comment in this pull request either?
Since this transition is going to have at least two more PRs I could either make the comment in the next PR, or nake the actual change in there.
The second part should be much smaller, since it will contain only the changes with minimal refactoring needed.
</blockquote></details>

I am fine with that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697)*

vheon:
Build and coverage aside this :lgtm: 

---

---

*[cpp/ycm/standard.h, line 18 at r10](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbp-u5PG3AY0OFeYW-_:-KbvjBBwN7lv2jVtsFYv:b-4letwr) ([raw file](https://github.com/valloric/ycmd/blob/98c9507191d8355a2e5a27962880245454520a18/cpp/ycm/standard.h#L18)):*

I am fine with that.
</blockquote></details>

I agree that removing this header file is the way to go and I also think that it should be a separate PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697#-:-KbvjMIbQCG4qGZ7BPjl:b-jsz366)*

micbou:

---

*[cpp/CMakeLists.txt, line 118 at r12](https://reviewable.io:443/reviews/valloric/ycmd/697#-KbvjnZ69eCXTwrTSAJh:-KbvjnZ69eCXTwrTSAJi:bs77o6v) ([raw file](https://github.com/valloric/ycmd/blob/e1f72ea15c52dcf82f844e99cd040e4a22f313b3/cpp/CMakeLists.txt#L118)):*
> ```CMake
> set( CPP11_AVAILABLE false )
> if ( CMAKE_COMPILER_IS_GNUCXX )
>   if ( NOT ( CMAKE_CXX_COMPILER_VERSION VERSION_LESS  4.6 ) )
> ```

The parentheses for this `NOT` and the ones below can be removed. There is also an extra space after `VERSION_LESS`.

---

*[cpp/ycm/CandidateRepository.cpp, line 120 at r12](https://reviewable.io:443/reviews/valloric/ycmd/697#-KbvlEkmS9JeOgx2emcr:-Kbvl7qy4g0wp-tcuaIB:b-29p2w8) ([raw file](https://github.com/valloric/ycmd/blob/e1f72ea15c52dcf82f844e99cd040e4a22f313b3/cpp/ycm/CandidateRepository.cpp#L120)):*
> ```cpp
> CandidateRepository::~CandidateRepository() {
>   for ( const CandidateHolder::value_type & pair :
>             candidate_holder_ ) {
> ```

Can be put on one line.

---

*[cpp/ycm/IdentifierDatabase.cpp, line 47 at r12](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbvlstw8HUSwuebJybc:-Kbvlstw8HUSwuebJybd:bj8r8fo) ([raw file](https://github.com/valloric/ycmd/blob/e1f72ea15c52dcf82f844e99cd040e4a22f313b3/cpp/ycm/IdentifierDatabase.cpp#L47)):*
> ```cpp
>   for ( const FiletypeIdentifierMap::value_type & filetype_and_map :
>             filetype_identifier_map ) {
>     for( const FilepathToIdentifiers::value_type & filepath_and_identifiers :
> ```

Missing space after `for`.

---

*[cpp/ycm/PythonSupport.cpp, line 105 at r12](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbvm8F4hbbHxN0LP0hp:-Kbvm8F5lG9rJWIl_0KH:b-o0qzfh) ([raw file](https://github.com/valloric/ycmd/blob/e1f72ea15c52dcf82f844e99cd040e4a22f313b3/cpp/ycm/PythonSupport.cpp#L105)):*
> ```cpp
> 
>   for ( const ResultAnd< int > &result_and_object :
>             result_and_objects ) {
> ```

Can fit on one line.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 427 at r12](https://reviewable.io:443/reviews/valloric/ycmd/697#-KbvmJEg5AErhclfYlh3:-KbvmJEg5AErhclfYlh4:bj8r8fo) ([raw file](https://github.com/valloric/ycmd/blob/e1f72ea15c52dcf82f844e99cd040e4a22f313b3/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L427)):*
> ```cpp
>     unique_lock< mutex > lock( diagnostics_mutex_ );
> 
>     for( const Diagnostic& diagnostic : latest_diagnostics_ ) {
> ```

Missing space after `for`.

---

*[cpp/ycm/ClangCompleter/TranslationUnitStore.cpp, line 93 at r12](https://reviewable.io:443/reviews/valloric/ycmd/697#-KbvmRpMJv98FepLl5Sd:-KbvmRpMJv98FepLl5Se:b-dstfjx) ([raw file](https://github.com/valloric/ycmd/blob/e1f72ea15c52dcf82f844e99cd040e4a22f313b3/cpp/ycm/ClangCompleter/TranslationUnitStore.cpp#L93)):*
> ```cpp
>                                                   unsaved_files,
>                                                   flags,
>                                                   clang_index_ );
> ```

These three parameters should be aligned with the first one.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697)*

bstaletic:
Review status: 23 of 29 files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/CMakeLists.txt, line 118 at r12](https://reviewable.io:443/reviews/valloric/ycmd/697#-KbvjnZ69eCXTwrTSAJh:-KbvpUJ6G6JAmXXCa3cW:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/e1f72ea15c52dcf82f844e99cd040e4a22f313b3/cpp/CMakeLists.txt#L118)):*

The parentheses for this `NOT` and the ones below can be removed. There is also an extra space after `VERSION_LESS`.
</blockquote></details>

Done.

---

*[cpp/ycm/CandidateRepository.cpp, line 120 at r12](https://reviewable.io:443/reviews/valloric/ycmd/697#-KbvlEkmS9JeOgx2emcr:-KbvpSgcpqoauGH0FLWG:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/e1f72ea15c52dcf82f844e99cd040e4a22f313b3/cpp/ycm/CandidateRepository.cpp#L120)):*

Can be put on one line.
</blockquote></details>

Done.

---

*[cpp/ycm/IdentifierDatabase.cpp, line 47 at r12](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbvlstw8HUSwuebJybc:-KbvpQTgjjcWU_bWjH_m:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/e1f72ea15c52dcf82f844e99cd040e4a22f313b3/cpp/ycm/IdentifierDatabase.cpp#L47)):*

Missing space after `for`.
</blockquote></details>

Done.

---

*[cpp/ycm/PythonSupport.cpp, line 105 at r12](https://reviewable.io:443/reviews/valloric/ycmd/697#-Kbvm8F4hbbHxN0LP0hp:-KbvpP6qVjzIFUqaEoSv:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/e1f72ea15c52dcf82f844e99cd040e4a22f313b3/cpp/ycm/PythonSupport.cpp#L105)):*

Can fit on one line.
</blockquote></details>

Done.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 427 at r12](https://reviewable.io:443/reviews/valloric/ycmd/697#-KbvmJEg5AErhclfYlh3:-KbvpN4sDCP-07jHSF-X:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/e1f72ea15c52dcf82f844e99cd040e4a22f313b3/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L427)):*

Missing space after `for`.
</blockquote></details>

Done.

---

*[cpp/ycm/ClangCompleter/TranslationUnitStore.cpp, line 93 at r12](https://reviewable.io:443/reviews/valloric/ycmd/697#-KbvmRpMJv98FepLl5Sd:-KbvpLLfoaNLdFNR53gd:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/e1f72ea15c52dcf82f844e99cd040e4a22f313b3/cpp/ycm/ClangCompleter/TranslationUnitStore.cpp#L93)):*

These three parameters should be aligned with the first one.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697)*

micbou:
Ok, coverage report seems correct now. There is a small decrease but it's mostly partial branches so I don't think it's worth adding new tests for that.

:lgtm:

@homu r+

---

Reviewed 6 of 6 files at r13.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/697#-:-KbwFLPWdK_TfbMrklRJ:bco23mr)*

PR #699, opened 2017-02-02
[READY] Update racerd to 0.3.3

micbou:
Fixes `GetCompletions_NoCompletionsFound` test and updates Travis and AppVeyor configurations by using `rustup` to install `rustc` and `cargo`.

Closes #676.
Closes #689.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/699)

PR #701, opened 2017-02-02
[READY] Remove boost where only simple refactoring is needed

bstaletic:
This pull request continues with removing boost components with std counterparts. Only simple refactoring is supposed to be done here. Current changes:

- Removes boost/algorithm
- Removes boost/exceptions
- Removes boost/assign
- Removes standard.h in favour of using `size_t`, or `unsigned int`
- Minimises boost/noncopyable, leaving it only in one file

#### Things on a TODO list:

~~Write tests if coverage complains~~
- [x] Bump ycmd core version (Is this needed?)

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/701)


bstaletic:
Travis is passing, but appveyor is stalling on MSVC12.

Should we move away from `size_t` in order to silence appveyor warnings?

Also, `-std=c++11` on MSVC is an unknown option, which is rightfully ignored, but MSVC emits a warning. Should we fix that too?

bstaletic:
Also the errors on MSVC11 are caused by replacing boost::assign.

But the syntax I used is the same that is valid in C when one needs to assign multidimensional array of characters aand is valid in C99.

vheon:
I think we have to check which version of the `islower` `isupper` we should use, because in this change we're using the one from `ctypes` while there are counterparts in `locale`. For the use of `isprint` was clear that we were missing the locale but what about the other ones?

---

Reviewed 16 of 21 files at r1.
Review status: 16 of 21 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[cpp/ycm/Candidate.cpp, line 34 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-f6zb8t1IFbDBhsjY:-Kc-f6zb8t1IFbDBhsjZ:b-h8hp0w) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/Candidate.cpp#L34)):*
> ```cpp
>   return all_of( text.cbegin(),
>                  text.cend(),
>                  []( char c ) { return isprint( c ); } );
> ```

Didn't we loose the locale here? Shouldn't we use [this version](http://en.cppreference.com/w/cpp/locale/isprint) of isprint? So maybe instead of a lambda we could do:

```cpp
namespace detail
{
const locale& loc = std::locale::classic();
bool isprint( char c ) {
  return std::locale::isprint( c, loc );
}
}
...
return all_of( text.begin(), text.end(), detail::isprint );
```

Or if we want to keep the lambda just use `std::locale::isprint( c, std::locale::classic() )`.

---

*[cpp/ycm/Candidate.cpp, line 81 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-gkVK-nmPez0RX4RG:-Kc-gkVK-nmPez0RX4RH:bvpr2u5) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/Candidate.cpp#L81)):*
> ```cpp
>   text_is_lowercase_( all_of( text.cbegin(),
>                               text.cend(),
>                               []( char c ) { return islower( c ); } ) ),
> ```

If we go changin the lambda above we could make a proxy function for `islower` in the `detail` namespace as well so this could be a one-liner that is easy to read.

---

*[cpp/ycm/IdentifierDatabase.cpp, line 89 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-hYyvVH6ZoNpw3c80:-Kc-hYyvVH6ZoNpw3c81:bidu7xy) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/IdentifierDatabase.cpp#L89)):*
> ```cpp
>   bool query_has_uppercase_letters = any_of( query.cbegin(),
>                                              query.cend(),
>                                              []( char c ) { return isupper( c ); } );
> ```

I think this is longer than 80 column.

---

*[cpp/ycm/ClangCompleter/Location.h, line 35 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-lZy47GRq8aD2EFyv:-Kc-lZy51Knrihv3tGA8:b9fl9r4) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/ClangCompleter/Location.h#L35)):*
> ```objc
>       filename_( "" ) {}
> 
>   Location( const std::string &filename, unsigned int line, unsigned int column )
> ```

this is now longer than 80 columns

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 25 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-lmkso_44r-zstdET:-Kc-lmkso_44r-zstdEU:bo6zejd) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L25)):*
> ```cpp
> 
> #include <cstdlib>
> #include <algorithm>
> ```

I think I'm missing something: what are those header for?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701#-:-Kc-lvsIBxPw6lP22Ret:bjnagto)*

Valloric:
> Also, -std=c++11 on MSVC is an unknown option, which is rightfully ignored, but MSVC emits a warning. Should we fix that too?

Yes.

> But the syntax I used is the same that is valid in C when one needs to assign multidimensional array of characters aand is valid in C99.

MSVC does not support C99 yet; I think they're working on improving that in future versions.

> replicating that header in our own NonCopyable.h is as simple as making a class with only constructor, destructor and copy constructor.

Sounds like a plan! Needs to work though. :)

---

Review status: 16 of 21 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701#-:-Kc-vvx4vUKQ3WcKcP82:b-s0zuz8)*

Valloric:
Review status: 16 of 21 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[cpp/ycm/exceptions.h, line 34 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-wPVt3Ga-uTWDT2sN:-Kc-wPVt3Ga-uTWDT2sO:b2aqvhh) ([raw file](https://github.com/valloric/ycmd/blob/bacfdcc940423d5bdbb237d2a94b46e3494fdbf0/cpp/ycm/exceptions.h#L34)):*
> ```objc
>  * The common base for all exceptions.
>  */
> struct ExceptionBase: virtual std::exception, virtual boost::exception {};
> ```

Hm, not sure is this a safe change. Then again, was never a huge fan of boost::exception so even if we lose some of its functionality, I don't care too much.

---

*[cpp/ycm/Result.cpp, line 222 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-wdN8yfslwAyEmcud:-Kc-wdN8yfslwAyEmcue:b8kj3rz) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/Result.cpp#L222)):*
> ```cpp
> 
> 
> bool Result::QueryIsPrefix( std::string text, std::string query ) {
> ```

You probably want to take params by const ref, not by value. The latter introduces copies.

So: `const std::string& text` and same for the other param.

---

*[cpp/ycm/Result.h, line 52 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-wrGjmGcPRcGr-FOl:-Kc-wrGk2Tub13SPVQd8:b74ywi5) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/Result.h#L52)):*
> ```objc
>     const std::string &word_boundary_chars );
> 
>   bool QueryIsPrefix( std::string text, std::string query );
> ```

const ref (see previous comment)

---

*[cpp/ycm/ClangCompleter/CompletionData.cpp, line 149 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-wvwcLbePHPJQ6KHu:-Kc-wvwcLbePHPJQ6KHv:brw46c9) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/ClangCompleter/CompletionData.cpp#L149)):*
> ```cpp
> 
> 
> bool IdentifierEndsWith( std::string identifier, std::string end ) {
> ```

Const ref params.

---

*[cpp/ycm/ClangCompleter/CompletionData.cpp, line 160 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-xHGYwKPJwdVrhCjk:-Kc-xHGYwKPJwdVrhCjl:b-wffd95) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/ClangCompleter/CompletionData.cpp#L160)):*
> ```cpp
> std::string RemoveTrailingParens( std::string text ) {
>   if ( IdentifierEndsWith( text, "(" ) ) {
>     text = text.substr( 0, text.length() - 1 );
> ```

This does not have the same perf. The old version would mutate `text`, the new version makes a copy.

---

*[cpp/ycm/ClangCompleter/CompletionData.h, line 123 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-wymLks6vHIyNxRFz:-Kc-wymLks6vHIyNxRG-:brw46c9) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/ClangCompleter/CompletionData.h#L123)):*
> ```objc
>                              bool &saw_placeholder );
> 
>   bool IdentifierEndsWith( std::string identifier, std::string end );
> ```

Const ref params.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701)*

bstaletic:
> Sounds like a plan! Needs to work though. :)

I'll get on with `utility.hpp` replacing when everything else is done.

MSVC -std=c++11 flag will be fixed in the next push.

> MSVC does not support C99 yet; I think they're working on improving that in future versions.

The error regarding removal of `boost::assign` happens only on MSVC11. Do we simply drop the support for it?

I figured out why that one test is failing if we switch from `boost::array` to `std::array`.
It is not that it doesn't throw `std::out_of_range`, it's just that its `error.what()` is rather cryptic with the following format:
```
array::at: __n (which is <very long number that can change depending on the actual index>) >= _Nm (which is <another changable numbe which is in ycmd defined as 256>)
```
Since there are changeable numbers, I believe it would be best to match the whole line with a regex.
Would the following regex work: `array::at: __n -?\(which is \\d+\) >= _Nm \(which is \\d+\)`

---

Review status: 15 of 22 files reviewed at latest revision, 12 unresolved discussions.

---

*[cpp/ycm/Candidate.cpp, line 34 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-f6zb8t1IFbDBhsjY:-Kc0-JMPNxmVOP4iCF0f:b-vxz7lf) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/Candidate.cpp#L34)):*

Didn't we loose the locale here? Shouldn't we use [this version](http://en.cppreference.com/w/cpp/locale/isprint) of isprint? So maybe instead of a lambda we could do:

```cpp
namespace detail
{
const locale& loc = std::locale::classic();
bool isprint( char c ) {
  return std::locale::isprint( c, loc );
}
}
...
return all_of( text.begin(), text.end(), detail::isprint );
```

Or if we want to keep the lambda just use `std::locale::isprint( c, std::locale::classic() )`.
</blockquote></details>

I like how readable the `namespace detail` variant is. So I'd like to go with that.
Would difining that in `cpp/ycm/Detail.h` and implementing in `cpp/ycm/Detail.cpp` be okay?

---

*[cpp/ycm/Candidate.cpp, line 81 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-gkVK-nmPez0RX4RG:-Kc0-fhiGnQzyd2TsRCG:b-xqouwo) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/Candidate.cpp#L81)):*

If we go changin the lambda above we could make a proxy function for `islower` in the `detail` namespace as well so this could be a one-liner that is easy to read.
</blockquote></details>

Agreed with going with the namespace option.

---

*[cpp/ycm/exceptions.h, line 34 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-wPVt3Ga-uTWDT2sN:-Kc0-r_B5JpLmPqpw7hT:b-ccz6i0) ([raw file](https://github.com/valloric/ycmd/blob/bacfdcc940423d5bdbb237d2a94b46e3494fdbf0/cpp/ycm/exceptions.h#L34)):*

Hm, not sure is this a safe change. Then again, was never a huge fan of boost::exception so even if we lose some of its functionality, I don't care too much.
</blockquote></details>

I was thinking the same, but it's passing all the tests, so we can leave it for now and we can always write more tests later.

---

*[cpp/ycm/IdentifierDatabase.cpp, line 89 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-hYyvVH6ZoNpw3c80:-Kc00DZzofQddzAYyXNE:bw8kjmq) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/IdentifierDatabase.cpp#L89)):*

I think this is longer than 80 column.
</blockquote></details>

Will be fixed after introducing `isupper` from the `detail` namespace, which is discussed above.

---

*[cpp/ycm/Result.cpp, line 222 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-wdN8yfslwAyEmcud:-Kc01u80Ut52_4l-EC2Y:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/Result.cpp#L222)):*

You probably want to take params by const ref, not by value. The latter introduces copies.

So: `const std::string& text` and same for the other param.
</blockquote></details>

Done.

---

*[cpp/ycm/Result.h, line 52 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-wrGjmGcPRcGr-FOl:-Kc02MLvI-3UdSYhC_17:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/Result.h#L52)):*

const ref (see previous comment)
</blockquote></details>

Done.

---

*[cpp/ycm/ClangCompleter/CompletionData.cpp, line 149 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-wvwcLbePHPJQ6KHu:-Kc03Vraj2EOYVILqsl_:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/ClangCompleter/CompletionData.cpp#L149)):*

Const ref params.
</blockquote></details>

Done.

---

*[cpp/ycm/ClangCompleter/CompletionData.cpp, line 160 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-xHGYwKPJwdVrhCjk:-Kc0AsXSg9y76WoA9Ijs:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/ClangCompleter/CompletionData.cpp#L160)):*

This does not have the same perf. The old version would mutate `text`, the new version makes a copy.
</blockquote></details>

Done.

---

*[cpp/ycm/ClangCompleter/CompletionData.h, line 123 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-wymLks6vHIyNxRFz:-Kc03U1PuYsEuxp8cb2z:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/ClangCompleter/CompletionData.h#L123)):*

Const ref params.
</blockquote></details>

Done.

---

*[cpp/ycm/ClangCompleter/Location.h, line 35 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-lZy47GRq8aD2EFyv:-Kc03x1lFanDxFKKv7pd:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/ClangCompleter/Location.h#L35)):*

this is now longer than 80 columns
</blockquote></details>

Done.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 25 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-lmkso_44r-zstdET:-Kc03zywOLuhgT2PFSU4:b-1bkme2) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L25)):*

I think I'm missing something: what are those header for?
</blockquote></details>

`algorith` is needed for `std::all_of` and `std::any_of`.
`cstdlib` is needed for `std::abs`. Travis on OSX complained about this one.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701#-:-Kc06UTrekqemk8gG4ud:bd8pt7g)*

bstaletic:
Can anyone explain why would compiling libracerd on travis on linux fail?

micbou:
> Can anyone explain why would compiling libracerd on travis on linux fail?

Could be the new version of Rust (1.15.0 was just released). PR #699 may fix that.

---

Review status: 15 of 22 files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701#-:-Kc0VK8A_63I7kKfj65m:be9e4a5)*

bstaletic:
The builds and tests on CI  succeeded. Is there anything else that needs to be done before pushing `boost::array` to `std::array` change?

---

Review status: 12 of 23 files reviewed at latest revision, 6 unresolved discussions.

---

*[cpp/ycm/Result.cpp, line 225 at r4](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc7hoxQ46hHpBO_b7tX:-Kc7hoxQ46hHpBO_b7tY:b-9q1qi2) ([raw file](https://github.com/valloric/ycmd/blob/0661215caad940dddcf8dcf511fcccadfb8464fd/cpp/ycm/Result.cpp#L225)):*
> ```cpp
>                             const std::string &query ) {
>   if ( text.length() < query.length() )
>     return false;
> ```

Am I missing something or will this never happen?

As far as I can see `SetResultFeaturesFromQuery()` won't be called if query doesn't hit anything stored in identifier database. As a result `QueryIsPrefix()` also won't be called.

That would mean this check is redundant and provides nothing except one more line impossible to cover with tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701#-:-Kc7ina-HugA5QiaSz6A:b2zq3fb)*

Valloric:
Review status: 12 of 23 files reviewed at latest revision, 6 unresolved discussions.

---

*[cpp/ycm/Candidate.cpp, line 81 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-gkVK-nmPez0RX4RG:-Kc9ZP2Gm0wIKDxShbaF:bq484q4) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/Candidate.cpp#L81)):*

Agreed with going with the namespace option.
</blockquote></details>

Yeah I would very much prefer an extracted function than passing a lambda.

---

*[cpp/ycm/Result.cpp, line 225 at r4](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc7hoxQ46hHpBO_b7tX:-Kc9YtW7Rp5-W9b2TsDP:b6hf9d) ([raw file](https://github.com/valloric/ycmd/blob/0661215caad940dddcf8dcf511fcccadfb8464fd/cpp/ycm/Result.cpp#L225)):*

Am I missing something or will this never happen?

As far as I can see `SetResultFeaturesFromQuery()` won't be called if query doesn't hit anything stored in identifier database. As a result `QueryIsPrefix()` also won't be called.

That would mean this check is redundant and provides nothing except one more line impossible to cover with tests.
</blockquote></details>

If you're confident it can't be hit, remove it. :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701)*

bstaletic:
Review status: 12 of 25 files reviewed at latest revision, 6 unresolved discussions.

---

*[cpp/ycm/Candidate.cpp, line 81 at r1](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc-gkVK-nmPez0RX4RG:-Kc9qIjhyB5C7tsjgeyH:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/d0e5dde55d2deaa236fa67423b4c4fd6df1aecfd/cpp/ycm/Candidate.cpp#L81)):*

Yeah I would very much prefer an extracted function than passing a lambda.
</blockquote></details>

Done.

---

*[cpp/ycm/Result.cpp, line 225 at r4](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc7hoxQ46hHpBO_b7tX:-Kc9qL4dYFpTPpxzpESQ:bvo3x61) ([raw file](https://github.com/valloric/ycmd/blob/0661215caad940dddcf8dcf511fcccadfb8464fd/cpp/ycm/Result.cpp#L225)):*

If you're confident it can't be hit, remove it. :)
</blockquote></details>

Still looking into it. Identifier match ranking algorithm is really complex. (:
The actual method I'm looking at, which will eventually answer this question, is `Result Candidate::QueryMatchResult( const std::string& text)` implemented in `Candidate.cpp` on line 80.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701)*

bstaletic:
I pushed a commit that replaces `boost::noncopyable` with the actual code that deletes the copy construct.

The only place where `boost::noncopyable` is left is in `cpp/ycm/ycm_core.cpp` when exposing classes to python.
If I'm not missing anything big, `boost::python::class_` depends on `boost::noncopyable` to realise that the class is noncopyable and not to try to access the copy construct.

Feel free to point out where have I gone wrong.

---

Review status: 11 of 34 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701#-:-KcDz1-meFLAhedk2b95:b-qyw5cv)*

bstaletic:
Review status: 11 of 34 files reviewed at latest revision, 6 unresolved discussions.

---

*[cpp/ycm/Result.cpp, line 225 at r4](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc7hoxQ46hHpBO_b7tX:-KcI7bzijTZB7P28ynKt:b-ksn5r0) ([raw file](https://github.com/valloric/ycmd/blob/0661215caad940dddcf8dcf511fcccadfb8464fd/cpp/ycm/Result.cpp#L225)):*

Still looking into it. Identifier match ranking algorithm is really complex. (:
The actual method I'm looking at, which will eventually answer this question, is `Result Candidate::QueryMatchResult( const std::string& text)` implemented in `Candidate.cpp` on line 80.
</blockquote></details>

Now I am confident. That function will only be called if query is a subsequence of text.
"fooooo" is most definitely not a subsequence of "foo", so the check is definitely redundant.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701)*

Valloric:
> Bump ycmd core version (Is this needed?)

To answer this, it's necessary when you change an external-facing API in the C++ code. Such APIs are the ones used by the YCM Python code.

You probably need to bump it now.

With that, I think I'm :lgtm: on this PR now! :)

---

Review status: 11 of 34 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701#-:-KcJZeyWIyK6euZ0obQa:b-zde794)*

vheon:
I have to say I'm not a super fan of an entire Translation Unit just for 3 function 2 of which are used in only one location and the other one in 2. I mean at this point for `isprint` and `islower` a static function like:

```cpp
static bool isprint( char c ) {
  return std::locale::isprint( c, std::locale::classic() );
}

static bool islower( char c ) {
  return std::locale::islower( c, std::locale::classic() );
}
```

would suffice (avoiding a new TU and then the linking). I don't have the strongest opinion on this so is not a deal breaker to merge this for me.

---

Reviewed 1 of 21 files at r1, 3 of 6 files at r2, 2 of 2 files at r4, 14 of 16 files at r5, 1 of 2 files at r6.
Review status: 32 of 35 files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/ycm/Detail.h, line 18 at r5](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcJky2HHJrI87scFjL2:-KcJky2IzGaoOG6c7tyY:b-474qy0) ([raw file](https://github.com/valloric/ycmd/blob/9cf82755d61bbcd6bc5064a2017d990829d6330a/cpp/ycm/Detail.h#L18)):*
> ```objc
> // along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
> 
> #ifndef DETAIL_H
> ```

I don't know if could be a problem but for the rest of the files the defined symbol is postfixed with a random string (I guess Val [used this snippet] (https://github.com/honza/vim-snippets/blob/master/UltiSnips/cpp.snippets#L57-L72) for the other files) to avoid conflict. I don't know how common is `DETAIL_H` though :confused: 

---

*[cpp/ycm/Detail.h, line 27 at r5](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcJjF0lep0T7CfeZR6s:-KcJjF0mKV98KPqqiRKj:b-tbaqtx) ([raw file](https://github.com/valloric/ycmd/blob/9cf82755d61bbcd6bc5064a2017d990829d6330a/cpp/ycm/Detail.h#L27)):*
> ```objc
> namespace detail {
> 
> const std::locale loc( "C" );
> ```

Looks like `std::locale::classic()` is actually a static variable to this exact locale so I guess we can just use that (ref. http://en.cppreference.com/w/cpp/locale/locale/classic)

---

*[cpp/ycm/IdentifierDatabase.cpp, line 87 at r5](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcJhnznPykJeG3iw3RW:-KcJhnzov3jr_xkyAnHm:b-dmwmdl) ([raw file](https://github.com/valloric/ycmd/blob/9cf82755d61bbcd6bc5064a2017d990829d6330a/cpp/ycm/IdentifierDatabase.cpp#L87)):*
> ```cpp
>   bool query_has_uppercase_letters = any_of( query.cbegin(),
>                                              query.cend(),
> 					     detail::isupper );
> ```

I think we have tabs here while re rest looks like is indented with spaces.

---

*[cpp/ycm/PythonSupport.cpp, line 85 at r5](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcJi3iGegCqzuwnxo9J:-KcJi3iHQVVWCQjbM5un:byxsirm) ([raw file](https://github.com/valloric/ycmd/blob/9cf82755d61bbcd6bc5064a2017d990829d6330a/cpp/ycm/PythonSupport.cpp#L85)):*
> ```cpp
>     bool query_has_uppercase_letters = any_of( query.cbegin(),
>                                                query.cend(),
> 					       detail::isupper );
> ```

Also here looks like there is a mixed of tabs and spaces.

---

*[cpp/ycm/ycm_core.cpp, line 225 at r5](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcJlpBin0yqbXQ1CRCw:-KcJlpBin0yqbXQ1CRCx:b-qor2p) ([raw file](https://github.com/valloric/ycmd/blob/bacfdcc940423d5bdbb237d2a94b46e3494fdbf0/cpp/ycm/ycm_core.cpp#L225)):*
> ```cpp
> 
> // Boost.Thread forces us to implement this.
> // We don't use any thread-specific (local) storage so it's fine to implement
> ```

Aren't we still using boost thread? why is safe to remove this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701#-:-KcJk1Z1QtWvIGSY4c3A:b-d77jwu)*

bstaletic:
@vheon
I removed `Detail.h` and `Detail.cpp` in favour of static functions.

---

Review status: 28 of 33 files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/ycm/IdentifierDatabase.cpp, line 87 at r5](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcJhnznPykJeG3iw3RW:-KcJrsISrsrvTIK4JHYW:bup4gbn) ([raw file](https://github.com/valloric/ycmd/blob/9cf82755d61bbcd6bc5064a2017d990829d6330a/cpp/ycm/IdentifierDatabase.cpp#L87)):*

I think we have tabs here while re rest looks like is indented with spaces.
</blockquote></details>

Switched to static functions to not introduce a new translation unit.

---

*[cpp/ycm/PythonSupport.cpp, line 85 at r5](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcJi3iGegCqzuwnxo9J:-KcJru3lMtmK0d0gyPqC:bup4gbn) ([raw file](https://github.com/valloric/ycmd/blob/9cf82755d61bbcd6bc5064a2017d990829d6330a/cpp/ycm/PythonSupport.cpp#L85)):*

Also here looks like there is a mixed of tabs and spaces.
</blockquote></details>

Switched to static functions to not introduce a new translation unit.

---

*[cpp/ycm/ycm_core.cpp, line 225 at r5](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcJlpBin0yqbXQ1CRCw:-KcJn9vzlaGqXQhgMYLe:b-c89cbn) ([raw file](https://github.com/valloric/ycmd/blob/bacfdcc940423d5bdbb237d2a94b46e3494fdbf0/cpp/ycm/ycm_core.cpp#L225)):*

Aren't we still using boost thread? why is safe to remove this?
</blockquote></details>

The PR #697 made a switch from boost thread to C++11 threads.

---

*[cpp/ycm/Detail.h, line 18 at r5](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcJky2HHJrI87scFjL2:-KcJr62gwXnQT5QPGB03:bup4gbn) ([raw file](https://github.com/valloric/ycmd/blob/9cf82755d61bbcd6bc5064a2017d990829d6330a/cpp/ycm/Detail.h#L18)):*

I don't know if could be a problem but for the rest of the files the defined symbol is postfixed with a random string (I guess Val [used this snippet] (https://github.com/honza/vim-snippets/blob/master/UltiSnips/cpp.snippets#L57-L72) for the other files) to avoid conflict. I don't know how common is `DETAIL_H` though :confused: 
</blockquote></details>

Switched to static functions to not introduce a new translation unit.

---

*[cpp/ycm/Detail.h, line 27 at r5](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcJjF0lep0T7CfeZR6s:-KcJrenUvSODe2eb2Hq3:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/9cf82755d61bbcd6bc5064a2017d990829d6330a/cpp/ycm/Detail.h#L27)):*

Looks like `std::locale::classic()` is actually a static variable to this exact locale so I guess we can just use that (ref. http://en.cppreference.com/w/cpp/locale/locale/classic)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701#-:-KcJrw26K-zRpOpdQhoP:b6z0fn0)*

vheon:
Reviewed 1 of 6 files at r2, 1 of 16 files at r5, 4 of 5 files at r7.

---

*[cpp/ycm/ycm_core.cpp, line 225 at r5](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcJlpBin0yqbXQ1CRCw:-KcK9qxhOR968tFqWUt4:bee88m9) ([raw file](https://github.com/valloric/ycmd/blob/bacfdcc940423d5bdbb237d2a94b46e3494fdbf0/cpp/ycm/ycm_core.cpp#L225)):*

The PR #697 made a switch from boost thread to C++11 threads.
</blockquote></details>

I read `thread` and actually thought about `boost::thread` while we only use the `locks` here and the threads are in Python-land (at least I think otherwise I really don't know what I'm talking about :unamused: )

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701)*

bstaletic:
What's up with codecov again?
In some places it is reporting an increase in coverage by 0.06%.
In others it is reporting a drop in coverage even in places I didn't touch.

Also, the last change provides no reason for coverage to change at all.

micbou:
Reviewed 13 of 21 files at r1, 5 of 6 files at r2, 3 of 3 files at r3, 2 of 2 files at r4, 11 of 16 files at r5, 1 of 2 files at r6, 3 of 5 files at r7.

---

*[cpp/ycm/PythonSupport.cpp, line 62 at r7](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcKaJ9SEB-4V4qSOQBl:-KcKaJ9SEB-4V4qSOQBm:bdowsm7) ([raw file](https://github.com/valloric/ycmd/blob/133fa7e8bf060879f7303e98d38879da82209815/cpp/ycm/PythonSupport.cpp#L62)):*
> ```cpp
> static bool IsUpper( char c ) {
>   return std::isupper( c, std::locale::classic() );
> }
> ```

The same function is defined in `IdentifierDatabase.cpp`. I think we should move it to `Utils.h` and include this header (it's already included in `IdentifierDatabase.cpp`).

---

*[cpp/ycm/Result.cpp, line 229 at r7](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcKdEYxni1R6fFNrlAO:-KcKdEYxni1R6fFNrlAP:b-6x10y9) ([raw file](https://github.com/valloric/ycmd/blob/133fa7e8bf060879f7303e98d38879da82209815/cpp/ycm/Result.cpp#L229)):*
> ```cpp
> 
>   return true;
> }
> ```

Missing blank line.

---

*[cpp/ycm/ClangCompleter/CompletionData.cpp, line 156 at r4](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc7xu2-zzr39wsZn6n1:-Kc7xu2-zzr39wsZn6n2:b-3wubw2) ([raw file](https://github.com/valloric/ycmd/blob/0661215caad940dddcf8dcf511fcccadfb8464fd/cpp/ycm/ClangCompleter/CompletionData.cpp#L156)):*
> ```cpp
>                                     end );
>   return false;
> }
> ```

Missing two blank lines.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701)*

bstaletic:
Review status: 29 of 35 files reviewed at latest revision, 6 unresolved discussions.

---

*[cpp/ycm/PythonSupport.cpp, line 62 at r7](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcKaJ9SEB-4V4qSOQBl:-KcKmMMPO2XC3eCjUEFv:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/133fa7e8bf060879f7303e98d38879da82209815/cpp/ycm/PythonSupport.cpp#L62)):*

The same function is defined in `IdentifierDatabase.cpp`. I think we should move it to `Utils.h` and include this header (it's already included in `IdentifierDatabase.cpp`).
</blockquote></details>

Done.

---

*[cpp/ycm/Result.cpp, line 229 at r7](https://reviewable.io:443/reviews/valloric/ycmd/701#-KcKdEYxni1R6fFNrlAO:-KcKmNfS-a5RMJXIbNmV:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/133fa7e8bf060879f7303e98d38879da82209815/cpp/ycm/Result.cpp#L229)):*

Missing blank line.
</blockquote></details>

Done.

---

*[cpp/ycm/ClangCompleter/CompletionData.cpp, line 156 at r4](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kc7xu2-zzr39wsZn6n1:-KcKmRTgYjsGsEHXANSk:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/0661215caad940dddcf8dcf511fcccadfb8464fd/cpp/ycm/ClangCompleter/CompletionData.cpp#L156)):*

Missing two blank lines.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701)*

vheon:
Reviewed 6 of 6 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701)*

micbou:
:lgtm_strong:

---

Reviewed 6 of 6 files at r8.

---

*[cpp/ycm/Utils.cpp, line 56 at r8](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kcs45B3Zy2TACf7cfTn:-Kcs45B3Zy2TACf7cfTo:b-x9h8ou) ([raw file](https://github.com/valloric/ycmd/blob/ead191dcc3f17aade94d45d83bcb0834f6301d2e/cpp/ycm/Utils.cpp#L56)):*
> ```cpp
> 
> bool IsUpper( char c )
> {
> ```

Opening curly brace should be put on the same line.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701#-:-Kcs3_fqGKwxqRwDP9cO:b-jbsf8z)*

bstaletic:
Review status: 34 of 35 files reviewed at latest revision, 3 unresolved discussions.

---

*[cpp/ycm/Utils.cpp, line 56 at r8](https://reviewable.io:443/reviews/valloric/ycmd/701#-Kcs45B3Zy2TACf7cfTn:-KcsNqA1UxiHXthAPJP1:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/ead191dcc3f17aade94d45d83bcb0834f6301d2e/cpp/ycm/Utils.cpp#L56)):*

Opening curly brace should be put on the same line.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701)*

Valloric:
I see 2 LGTM's and open issues resolved, so here we go!

@bstaletic Thanks a ton for the PR!

@homu r+

---

Review status: 34 of 35 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701#-:-Kcsknpvrrc2Vd0_s-iL:b-2yn6ig)*

bstaletic:
@Valloric No problem. But this is not over yet. So you'll be seeing more of me soon. :D

micbou:
@homu retry

---

Reviewed 1 of 1 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701#-:-KctWrrmriKsyOmnc3u1:b8khjrm)*

oblitum:
I was merging upstream with my fork and noticed some strange additions in this pull request, for example, does it make sense that the `IdentifierEndsWith` function is declared as a private member of `CompletionData` but is defined as a free function in an anonymous namespace?

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/701#-:-Kf-erwe6CTk0cb-SSQC:b-jzw0a1)*

PR #704, opened 2017-02-08
Add Geany and GNU nano clients to readme

orsonteodoro:
Please add gycm Geany plugin and my branch of GNU nano.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/704)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,8 @@ Known ycmd clients:
 - [YcmdCompletion][sublime-ycmd]: Sublime client
 - [kak-ycmd][]: Kakoune client.
 - [you-complete-me][vscode-you-complete-me]: VSCode client.
+- [gycm][]: Geany client.
+- [nano-ycmd][]: GNU nano client.
 
 Feel free to send a pull request adding a link to your client here if you've
 built one.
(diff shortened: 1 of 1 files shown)


vheon:
Is not actually relevant with this PR but why on https://github.com/orsonteodoro/nano-ycmd#dependencies are mentioned `jedi`, `jedihttp`, `Clang`? when they are either a submodule of ycmd or something we download automatically (like libclang).

Plus
> Either nettle, openssl, or "libgcrypt with glib" cryptographic library, to mitigate MITM attack between ycmd and nano text editor

at most you can do a passive MITM attack because the requests are signed with HMAC so they cannot be tampered with.

But for this PR I think :lgtm:

orsonteodoro:
On Gentoo Linux we keep the third party as separate packages and try to refrain from bundling (https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies).  When I created separate packages, I forgot to put version checking on them and 0.9.0 was only available in the main portage tree offered by Gentoo, so I created a new package for jedi 0.10.0.

Jedi was working before I added completion commands then it started breaking when I used them on python because I was using 0.9.0.  Also, the completer commands didn't work without specific versions of jedi because of the introduction of follow_imports (https://github.com/vheon/JediHTTP/issues/26#issuecomment-268235744).  

As for clang, I was going to use clang -dumpmachine to obtain the triplet (machine-vendor-operatingsystem) for x86_64-pc-linux-gnu-libgcrypt-config command, but I discovered I could use the host variable produced by the config.guess.  I just found out that I could just use libgcrypt-config instead.

vheon:
> On Gentoo Linux we keep the third party as separate packages and try to refrain from bundling 

Yeah you can do whatever you want but we still search for the JediHTTP that we have in the `third_party` directory and if you have JediHTTP installed we don't care. And that is true for most our dependencies, for sure the one that I mentioned.

Valloric:
Thanks for the PR!

@homu r+

PR #705, opened 2017-02-14
[READY] Switch from boost to std for hashes, arrays and regexes

bstaletic:
This is a continuation of #701. These changes will not be trivial and will need to be discussed before proceeding. At this moment only `std::hash` is working. The other two boost components will be discussed after agreeing that this hash implementation is fine.

- [x] Use `std::hash` instead of `boost::hash`
- [x] Use `std::array` instead of `boost::array`
~~Use `regex_search` and related from `std` namespace.~~

### Regex component will be in another PR

As it stands for now the regex part is the hardest to replce, so, in order not to make this PR big and drag it out, regex component will be in a separate PR.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/705)

### DIFF
--- a/cpp/ycm/ClangCompleter/TranslationUnitStore.cpp
+++ b/cpp/ycm/ClangCompleter/TranslationUnitStore.cpp
@@ -32,17 +32,21 @@ namespace YouCompleteMe {
 namespace {
 
 std::size_t HashForFlags( const std::vector< std::string > &flags ) {
-  /* 
+  /*
    * The way boost::hash implements hash< std::vector<T> > goes after the
    * described way in "Library Extension Technical Report - Issue List"
-   * section 6.18. When (if) it ends up in the STL this whole function could br
+   * section 6.18. This function's body is taken straight (copy-pasted)
+   * from the TR to replicate boost's behaviour.
+   *
+   * When (if) it ends up in the STL this whole function could be
    * replaced with:
    *
-   * return std::hash< std::vecto< std::string > >()( flags );
+   * return std::hash< std::vector< std::string > >()( flags );
    */
   size_t seed = 0;
-  for ( auto flag : flags )
+  for ( auto flag : flags ) {
     seed ^= std::hash< std::string >()( flag ) + ( seed<<6 ) + ( seed>>2 );
+  }
   return seed;
 }
 
--- a/cpp/ycm/ClangCompleter/TranslationUnitStore.cpp
+++ b/cpp/ycm/ClangCompleter/TranslationUnitStore.cpp
@@ -20,7 +20,7 @@
 #include "Utils.h"
 #include "exceptions.h"
 
-#include <boost/functional/hash.hpp>
+#include <functional>
 
 using std::lock_guard;
 using std::shared_ptr;
@@ -32,7 +32,18 @@ namespace YouCompleteMe {
 namespace {
 
 std::size_t HashForFlags( const std::vector< std::string > &flags ) {
-  return boost::hash< std::vector< std::string > >()( flags );
+  /* 
+   * The way boost::hash implements hash< std::vector<T> > goes after the
+   * described way in "Library Extension Technical Report - Issue List"
+   * section 6.18. When (if) it ends up in the STL this whole function could br
+   * replaced with:
+   *
+   * return std::hash< std::vecto< std::string > >()( flags );
+   */
+  size_t seed = 0;
+  for ( auto flag : flags )
+    seed ^= std::hash< std::string >()( flag ) + ( seed<<6 ) + ( seed>>2 );
+  return seed;
 }
 
 }  // unnamed namespace
--- a/cpp/ycm/IdentifierUtils.cpp
+++ b/cpp/ycm/IdentifierUtils.cpp
@@ -61,7 +61,7 @@ struct StringEqualityComparer :
 // static initialization.
 const std::unordered_map < const char *,
       const char *,
-      boost::hash< std::string >,
+      std::hash< std::string >,
       StringEqualityComparer > LANG_TO_FILETYPE = {
         { "Ant"        , "ant"        },
         { "Asm"        , "asm"        },
--- a/cpp/ycm/LetterNodeListMap.h
+++ b/cpp/ycm/LetterNodeListMap.h
@@ -22,7 +22,7 @@
 
 #include <vector>
 #include <memory>
-#include <boost/array.hpp>
+#include <array>
 
 #define NUM_LETTERS 128
 
@@ -71,7 +71,7 @@ public:
   YCM_DLL_EXPORT NearestLetterNodeIndices *ListPointerAt( char letter );
 
 private:
-  typedef boost::array<NearestLetterNodeIndices , NUM_LETTERS>
+  typedef std::array<NearestLetterNodeIndices , NUM_LETTERS>
     NearestLetterNodeArray;
 
   std::unique_ptr< NearestLetterNodeArray > letters_;
--- a/cpp/ycm/PythonSupport.cpp
+++ b/cpp/ycm/PythonSupport.cpp
@@ -25,6 +25,7 @@
 #include <algorithm>
 #include <vector>
 #include <locale>
+#include <utility>
 
 using std::any_of;
 using boost::python::len;
@@ -96,7 +97,7 @@ boost::python::list FilterAndSortCandidates(
 
       if ( result.IsSubsequence() ) {
         ResultAnd< int > result_and_object( result, i );
-        result_and_objects.push_back( boost::move( result_and_object ) );
+        result_and_objects.push_back( std::move( result_and_object ) );
       }
     }
 
--- a/cpp/ycm/tests/LetterNode_test.cpp
+++ b/cpp/ycm/tests/LetterNode_test.cpp
@@ -72,23 +72,6 @@ TEST( LetterNodeTest, ThrowOnNonAsciiCharacters ) {
   //   €: \xe2\x82\xac
   //   𐍈: \xf0\x90\x8d\x88
   ASSERT_THROW( LetterNode root_node( "uni¢𐍈d€" ), std::out_of_range );
-
-  /*
-   * Since what() is not standardised for std::array
-   * we do not test against its string value.
-   *
-   * When it gets standardised this part of code can
-   * be uncommented to check against the what() part
-   * of the error. To make the test valid, once the
-   * string checking is re-enabled, the string literal
-   * that is the argument of StrEq will have to be
-   * changed as well
-   */
-  // try {
-  //   LetterNode root_node( "uni¢𐍈d€" );
-  // } catch ( std::out_of_range error ) {
-  //   EXPECT_THAT( error.what(), StrEq( "array<>: index out of range" ) );
-  // }
 }
 
 } // namespace YouCompleteMe
--- a/cpp/ycm/tests/LetterNode_test.cpp
+++ b/cpp/ycm/tests/LetterNode_test.cpp
@@ -73,11 +73,22 @@ TEST( LetterNodeTest, ThrowOnNonAsciiCharacters ) {
   //   𐍈: \xf0\x90\x8d\x88
   ASSERT_THROW( LetterNode root_node( "uni¢𐍈d€" ), std::out_of_range );
 
-  try {
-    LetterNode root_node( "uni¢𐍈d€" );
-  } catch ( std::out_of_range error ) {
-    EXPECT_THAT( error.what(), StrEq( "array<>: index out of range" ) );
-  }
+  /*
+   * Since what() is not standardised for std::array
+   * we do not test against its string value.
+   *
+   * When it gets standardised this part of code can
+   * be uncommented to check against the what() part
+   * of the error. To make the test valid, once the
+   * string checking is re-enabled, the string literal
+   * that is the argument of StrEq will have to be
+   * changed as well
+   */
+  // try {
+  //   LetterNode root_node( "uni¢𐍈d€" );
+  // } catch ( std::out_of_range error ) {
+  //   EXPECT_THAT( error.what(), StrEq( "array<>: index out of range" ) );
+  // }
 }
 
 } // namespace YouCompleteMe
--- a/cpp/ycm/ycm_core.cpp
+++ b/cpp/ycm/ycm_core.cpp
@@ -78,7 +78,7 @@ BOOST_PYTHON_MODULE(ycm_core)
           &IdentifierCompleter::CandidatesForQueryAndType );
 
   class_< std::vector< std::string >,
-      boost::shared_ptr< std::vector< std::string > > >( "StringVector" )
+      std::shared_ptr< std::vector< std::string > > >( "StringVector" )
     .def( vector_indexing_suite< std::vector< std::string > >() );
 
 #ifdef USE_CLANG_COMPLETER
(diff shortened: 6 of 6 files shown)


Valloric:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[cpp/ycm/ClangCompleter/TranslationUnitStore.cpp, line 41 at r1](https://reviewable.io:443/reviews/valloric/ycmd/705#-Kcz4b1uG0Q6m8Wg3bbH:-Kcz4b1uG0Q6m8Wg3bbI:bcfm6xs) ([raw file](https://github.com/valloric/ycmd/blob/ed29a1ffe2c6402a209755821b35f5e6523824fd/cpp/ycm/ClangCompleter/TranslationUnitStore.cpp#L41)):*
> ```cpp
>    * replaced with:
>    *
>    * return std::hash< std::vecto< std::string > >()( flags );
> ```

Typo: vecto

---

*[cpp/ycm/ClangCompleter/TranslationUnitStore.cpp, line 45 at r1](https://reviewable.io:443/reviews/valloric/ycmd/705#-Kcz4eq07oBT1fYOQdey:-Kcz4eq07oBT1fYOQdez:b-8ml3eg) ([raw file](https://github.com/valloric/ycmd/blob/ed29a1ffe2c6402a209755821b35f5e6523824fd/cpp/ycm/ClangCompleter/TranslationUnitStore.cpp#L45)):*
> ```cpp
>   size_t seed = 0;
>   for ( auto flag : flags )
>     seed ^= std::hash< std::string >()( flag ) + ( seed<<6 ) + ( seed>>2 );
> ```

This seems a bit magical. Could you add some comments? Also braces around `for` block.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/705)*

bstaletic:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[cpp/ycm/ClangCompleter/TranslationUnitStore.cpp, line 45 at r1](https://reviewable.io:443/reviews/valloric/ycmd/705#-Kcz4eq07oBT1fYOQdey:-Kcz5d77DSF69Di3Zd5b:b3h37jy) ([raw file](https://github.com/valloric/ycmd/blob/ed29a1ffe2c6402a209755821b35f5e6523824fd/cpp/ycm/ClangCompleter/TranslationUnitStore.cpp#L45)):*

This seems a bit magical. Could you add some comments? Also braces around `for` block.
</blockquote></details>

I agree that it does seem a bit magical.
There already is a comment from where the whole function comes from. 

Should I change the comment somehow? The least I could do is put a direct link to the TR pdf.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/705)*

Valloric:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[cpp/ycm/ClangCompleter/TranslationUnitStore.cpp, line 45 at r1](https://reviewable.io:443/reviews/valloric/ycmd/705#-Kcz4eq07oBT1fYOQdey:-Kcz9a0SHlCHYYfe0zV5:b-k1an3e) ([raw file](https://github.com/valloric/ycmd/blob/ed29a1ffe2c6402a209755821b35f5e6523824fd/cpp/ycm/ClangCompleter/TranslationUnitStore.cpp#L45)):*

I agree that it does seem a bit magical.
There already is a comment from where the whole function comes from. 

Should I change the comment somehow? The least I could do is put a direct link to the TR pdf.
</blockquote></details>

Ah, it wasn't clear to me from the comment that this was copy-pasted. Could you clarify the comment a bit so that's more obvious?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/705)*

bstaletic:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[cpp/ycm/ClangCompleter/TranslationUnitStore.cpp, line 41 at r1](https://reviewable.io:443/reviews/valloric/ycmd/705#-Kcz4b1uG0Q6m8Wg3bbH:-KczBFnqAjKPHT7lcP6R:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/ed29a1ffe2c6402a209755821b35f5e6523824fd/cpp/ycm/ClangCompleter/TranslationUnitStore.cpp#L41)):*

Typo: vecto
</blockquote></details>

Done.

---

*[cpp/ycm/ClangCompleter/TranslationUnitStore.cpp, line 45 at r1](https://reviewable.io:443/reviews/valloric/ycmd/705#-Kcz4eq07oBT1fYOQdey:-KczBFP-dnQ0wY1dJU7R:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/ed29a1ffe2c6402a209755821b35f5e6523824fd/cpp/ycm/ClangCompleter/TranslationUnitStore.cpp#L45)):*

Ah, it wasn't clear to me from the comment that this was copy-pasted. Could you clarify the comment a bit so that's more obvious?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/705)*

Valloric:
> The other two boost components will be discussed after agreeing that this hash implementation is fine.

Since the hash function comes from the TR, this should be perfectly fine.

bstaletic:
I'll now push the commit that was supposed to replace `boost::array` with `std::array`.

The problem with this one is the `LetterNodeTest.ThrowOnNonAsciiCharacter` test.

I'm not sure what's going on, as the `std::array` does throw a `std::out_of_range` exception, but when the test is ran thte asserion fails with a seemingly valid pointer to a member of the array.

Here's a [gist](https://gist.github.com/bstaletic/b801aeb9a7ebaa4d8f6578ba1be7672b) with the code that the test runs (start with line 65).

The output of the gist above:
```
Try
117
110
105
-62
Catch
array::at: __n (which is 18446744073709551554) >= _Nm (which is 128)
```

The first big number is actuall -62 but represented as unsigned.
The second number is defined within ycmd.

I tried changing the assertion of the test to a regex that matches the "array::at:..." exception, but that didn't work because the test is not actually getting that error.

So, for now, the test is left as is and will fail.

Valloric:
> I'm not sure what's going on, as the std::array does throw a std::out_of_range exception, but when the test is ran thte asserion fails with a seemingly valid pointer to a member of the array.

You might need to investigate that one further; even if we decided to remove that test, we need to understand why the test behavior is changing. Without that understanding, we can't switch to `std::array`.

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/705#-:-Kd9CI40u4heyStktTSY:b-k7p25a)*

bstaletic:
I am fully aware that as it is right now, this PR s not mergable. The thing is, I have poked around a lot and whereever I looked at the code behaved the same. Except during the test. So I would appreciate if someone could help me find out what's wrong here.

On a slightly unrelated note: The regex component is behaving strangely too. But there, I still have a few ideas what to do.

fsimonis:
I just had a look at this and dug a bit into libc++ and cppreference.
`std::array::at()` performs the range check and throws a `std::out_of_range` with the message `"array::at"`.
So the logs from travis make sense as calling `std:exception::what()` returns a `const char*`.

The message does not seem to be standardised, so checking against it might be a bad idea.

Further information:
- `std::array::at` http://llvm.org/svn/llvm-project/libcxx/trunk/include/array
- `__throw_out_of_range` http://llvm.org/svn/llvm-project/libcxx/trunk/include/stdexcept
- http://en.cppreference.com/w/cpp/error/exception/what

bstaletic:
@fsimonis Thanks for looking into this.

I really didn't have the time to dive into the code again. I've been busy with some other projects, so I apologise for the delay.

If I'm getting this correctly, this does work and only the assertion needs to be changed. The problem is that the message differs on different platforms (although it is the same on linux and OSX acording to Travis).

@Valloric How do we proceed? Do we use some `#ifdef`s to be able to use different assertions on different OS's (this approach reminds me Vim source code)?

Once again, I'm still very busy with real life stuff, so I might respond infrequently.

Valloric:
> I've been busy with some other projects, so I apologise for the delay.

No worries, it's OSS. :) Your time is a gift that we appreciate and we wouldn't be so rude to ask for more gifts. As long as there's an intent to proceed with this PR, we can leave it open. If there's no progress after a few months, we can close it (so it doesn't sit in our queue) and re-open later as needed.

> How do we proceed? Do we use some #ifdefs to be able to use different assertions on different OS's (this approach reminds me Vim source code)?

I'd like to avoid the ifdefs it at all possible. :) Do we need to match on the message text? As long as we detect a failure, I'm OK with that.

bstaletic:
The simplest solution was to just comment out the string matching part of the test. The check for `std::out_of_range` exception is still there and is valid.

fsimonis:
@bstaletic The message is implementation-dependant. The message both of us and travis got is the one from [libc++](https://libcxx.llvm.org/). I just looked up the [eastl::array](https://github.com/electronicarts/EASTL/blob/master/include/EASTL/array.h) and they use yet another message.
I think throwing out this part of test would be the easiest (and probably sanest) fix.

@Valloric As you said: the test is valid as long as it detects `std::out_of_range`.

@bstaletic how far are you with the migration to `std::regex`? I am happy to give you a hand.

bstaletic:
@fsimonis Thanks for your generous offer. 
> how far are you with the migration to std::regex?

There seems to be a bug in gcc. `ReadUtf8File()` reads the whole file in one string along with new line characters. (i.e. `'\n'`). When such a string is fed to `std::search_regex()` it fails to find anything.

When the same file is read and analysed line by line `std::search_regex()` is able to process it correctly. It works properly in tha case, but the two tests fail and again I can't figure out why.

So before doing any major refactoring, I'd like to know how to approach the issue.

Valloric:
If regex stuff is more complicated, maybe leave that for a separate PR? It would let you finish the parts you can before going deeper on the harder regex stuff.

bstaletic:
It definitely is much more work than every other boost component until now. Maybe I should make it a separate PR. That would leave this one small and wouldn't drag the merge out for a long time.

Valloric:
> It definitely is much more work than every other boost component until now. Maybe I should make it a separate PR. That would leave this one small and wouldn't drag the merge out for a long time.

My thoughts exactly. :)

bstaletic:
@Valloric In that case this is ready. I'll change the title and update the opening post.

@fsimonis For further discussion about regex component open an issue on my ycmd fork.

vheon:
Just a minor comment but I'm fine either way.

:lgtm:

---

Reviewed 3 of 4 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/ycm/tests/LetterNode_test.cpp, line 91 at r4](https://reviewable.io:443/reviews/valloric/ycmd/705#-KetnFFxNjb_v-36TrvK:-KetnFFxNjb_v-36TrvL:bhs3t) ([raw file](https://github.com/valloric/ycmd/blob/01b0b3fed9689a4fef3565e439f70a867e18ec56/cpp/ycm/tests/LetterNode_test.cpp#L91)):*
> ```cpp
>   // } catch ( std::out_of_range error ) {
>   //   EXPECT_THAT( error.what(), StrEq( "array<>: index out of range" ) );
>   // }
> ```

Do we actually want to leave this here? I'm not even sure why we had it in the first place :confused: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/705#-:-KetnMDrUASYBEL-JqaB:bo6zw96)*

bstaletic:
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/ycm/tests/LetterNode_test.cpp, line 91 at r4](https://reviewable.io:443/reviews/valloric/ycmd/705#-KetnFFxNjb_v-36TrvK:-Ketp5BPDH7dAUb3rdlq:bt4n63e) ([raw file](https://github.com/valloric/ycmd/blob/01b0b3fed9689a4fef3565e439f70a867e18ec56/cpp/ycm/tests/LetterNode_test.cpp#L91)):*

Do we actually want to leave this here? I'm not even sure why we had it in the first place :confused: 
</blockquote></details>

I'm fine with removing it, but I left it there because I though that @micbou had a valid reason to include that part of the test when he wrote #453 .

To be honest, it just looks like checking the for the same thing in a slightly less future proof way.

Like I said, I'm fine with anything regarding this part of the code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/705)*

Valloric:
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/ycm/tests/LetterNode_test.cpp, line 91 at r4](https://reviewable.io:443/reviews/valloric/ycmd/705#-KetnFFxNjb_v-36TrvK:-KeuE1bjUznVkGXaRZOX:bl7iqma) ([raw file](https://github.com/valloric/ycmd/blob/01b0b3fed9689a4fef3565e439f70a867e18ec56/cpp/ycm/tests/LetterNode_test.cpp#L91)):*

I'm fine with removing it, but I left it there because I though that @micbou had a valid reason to include that part of the test when he wrote #453 .

To be honest, it just looks like checking the for the same thing in a slightly less future proof way.

Like I said, I'm fine with anything regarding this part of the code.
</blockquote></details>

Let's just remove it then.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/705)*

Valloric:
:lgtm: from me as well (with the minor comment).

---

Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/705#-:-KeuEMdekHiPdtTmMn36:b-lfo85p)*

bstaletic:
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/ycm/tests/LetterNode_test.cpp, line 91 at r4](https://reviewable.io:443/reviews/valloric/ycmd/705#-KetnFFxNjb_v-36TrvK:-KeuH3WE_V9DAhoUCcKn:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/01b0b3fed9689a4fef3565e439f70a867e18ec56/cpp/ycm/tests/LetterNode_test.cpp#L91)):*

Let's just remove it then.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/705)*

Valloric:
@homu r=vheon

Thanks for the PR!

micbou:
*[cpp/ycm/tests/LetterNode_test.cpp, line 91 at r4](https://reviewable.io:443/reviews/valloric/ycmd/705#-KetnFFxNjb_v-36TrvK:-KeuiwjYg_vJAr5Z_ktr:b1vknp2) ([raw file](https://github.com/valloric/ycmd/blob/01b0b3fed9689a4fef3565e439f70a867e18ec56/cpp/ycm/tests/LetterNode_test.cpp#L91)):*

Done.
</blockquote></details>

I did this because I felt that it was not enough to check for the exception type and I didn't find a way to compare the exception message with `ASSERT_THROW`. Yes, this looks redundant but was needed to handle the case where no exception is raised (without the `ASSERT_THROW`, the test would have passed). Anyway, I don't mind that we removed it (should have left a comment explaining that).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/705)*

Issue #534, opened 2016-06-28
Code Completion for IOS project.


jojojames:
I'm trying to set up ycmd with completions for objective-c/iOS based from an xcode project. Semantic completions don't seem to be working.

I've set up a sample project here.
https://github.com/jojojames/DummyXcode

.ycm_extra_conf located here
https://github.com/jojojames/DummyXcode/blob/master/.ycm_extra_conf.py

compile_commands.json here
https://github.com/jojojames/DummyXcode/blob/master/compile_commands.json

I had a similar problem (using the same sample project as above) described here which was resolved through blacklisting some compile flags.
https://github.com/Andersbakken/rtags/issues/738

Not sure where to get a complete log (down to the clang error)
2016-06-28 10:45:16,497 - ERROR - Exception from semantic completer (using general): Traceback (most recent call last):
  File "/Users/james/.vim/dein/repos/github.com/Valloric/YouCompleteMe/third_party/ycmd/ycmd/../ycmd/handlers.py", line 112, in GetCompletions
    .ComputeCandidates( request_data ) )
  File "/Users/james/.vim/dein/repos/github.com/Valloric/YouCompleteMe/third_party/ycmd/ycmd/../ycmd/completers/completer.py", line 168, in ComputeCandidates
    candidates = self._GetCandidatesFromSubclass( request_data )
  File "/Users/james/.vim/dein/repos/github.com/Valloric/YouCompleteMe/third_party/ycmd/ycmd/../ycmd/completers/completer.py", line 184, in _GetCandidatesFromSubclass
    raw_completions = self.ComputeCandidatesInner( request_data )
  File "/Users/james/.vim/dein/repos/github.com/Valloric/YouCompleteMe/third_party/ycmd/ycmd/../ycmd/completers/cpp/clang_completer.py", line 112, in ComputeCandidatesInner
    raise RuntimeError( NO_COMPLETIONS_MESSAGE )
RuntimeError: No completions found; errors in the file?

Using a more barebones ycm file that only specifies flags gets 70-80% of completions working.
https://github.com/jojojames/.emacs.d/blob/master/lang/objc/ycm_conf.py

puremourning:
Thanks for the detailed report. can you grab the output of :YcmDebugInfo? Also, what is the output of :YcmDiags?

[...]

PR #710, opened 2017-02-20
[ACTUALLY READY] Allow compilation for iOS projects

puremourning:
Historically it hasn't been possible to use ycmd with iOS projects for 2 reasons:

1. Opensource libclang seems to be unable to build the system headers for th iOS platform. It's not known _why_ this is, but it is known that Apple patch clang in Xcode
2. ycmd's flag handling hasn't played well with the myriad switches that are needed to build using iOS system headers

The advice has always been to use `--system-libclang` in such environments. However, this only solves point 1.

For a completely unrelated reason I was today testing something which required using the iOS system headers (from `/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk`) and found that even with Apple's libclang I wasn't able to compile even a simple c program.

The issue is that we scribble flags to point to the MacOS platform on OS X to make "normal" things work. In order to compile for iOS, you need to add `-isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk` to your flags. However, we then go and stick `-isystem /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/blah/blah` in the flags which causes problems.

The solution is to only add the additional mac paths if user doesn't specify a `-isysroot`.

This requires doing it for every flags for file request (rather than once at the start), but this is not a performance critical code path.

Fixes https://github.com/Valloric/ycmd/issues/534

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/710)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -295,11 +295,22 @@ def _CallExtraConfFlagsForFile( module, filename, client_data ):
     return module.FlagsForFile( filename )
 
 
+def _SysRootSpecifedIn( flags ):
+  for flag in flags:
+    if flag == '-isysroot' or flag.startswith( '--sysroot' ):
+      return True
+
+  return False
+
+
 def PrepareFlagsForClang( flags, filename, add_extra_clang_flags = True ):
   flags = _AddLanguageFlagWhenAppropriate( flags )
   flags = _RemoveXclangFlags( flags )
   flags = _RemoveUnusedFlags( flags, filename )
   if add_extra_clang_flags:
+    if OnMac() and not _SysRootSpecifedIn( flags ):
+      for path in _MacIncludePaths():
+        flags.extend( [ '-isystem', path ] )
     flags = _EnableTypoCorrection( flags )
 
   vector = ycm_core.StringVector()
@@ -496,11 +507,13 @@ if OnMac():
   )
 
 
+def _MacIncludePaths():
+  # This method exists for testing only
+  return MAC_INCLUDE_PATHS
+
+
 def _ExtraClangFlags():
   flags = _SpecialClangIncludes()
-  if OnMac():
-    for path in MAC_INCLUDE_PATHS:
-      flags.extend( [ '-isystem', path ] )
   # On Windows, parsing of templates is delayed until instantiation time.
   # This makes GetType and GetParent commands fail to return the expected
   # result when the cursor is in a template.
--- a/ycmd/tests/clang/flags_test.py
+++ b/ycmd/tests/clang/flags_test.py
@@ -31,7 +31,7 @@ from ycmd.tests.test_utils import MacOnly
 from ycmd.responses import NoExtraConfDetected
 from ycmd.tests.clang import TemporaryClangProject, TemporaryClangTestDir
 
-from hamcrest import assert_that, calling, contains, raises
+from hamcrest import assert_that, calling, contains, has_item, not_, raises
 
 
 @patch( 'ycmd.extra_conf_store.ModuleForSourceFile', return_value = Mock() )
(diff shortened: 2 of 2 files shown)


micbou:
Reviewed 1 of 1 files at r1.

---

*[ycmd/completers/cpp/flags.py, line 304 at r1](https://reviewable.io:443/reviews/valloric/ycmd/710#-Kd_WFdSxik6lgeGruf4:-Kd_WFdTTqcdf2dqY028:bc7e8un) ([raw file](https://github.com/valloric/ycmd/blob/eaad7815eed7b871e1172b48e0677c0d39939b61/ycmd/completers/cpp/flags.py#L304)):*
Shouldn't we check for the presence of the `--sysroot` and `--sysroot=` flags too? According to GCC documentation, [`-isysroot` is equivalent to `--sysroot` on OS X](https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html):
> -isysroot dir
    This option is like the --sysroot option, but applies only to header files (except for Darwin targets, where it applies to both header files and libraries). See the --sysroot option for more information. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/710)*

Valloric:

---

*[ycmd/completers/cpp/flags.py, line 304 at r1](https://reviewable.io:443/reviews/valloric/ycmd/710#-Kd_WFdSxik6lgeGruf4:-Kdc-TSMBzJvyXAf0bHe:b-xo5cga) ([raw file](https://github.com/valloric/ycmd/blob/eaad7815eed7b871e1172b48e0677c0d39939b61/ycmd/completers/cpp/flags.py#L304)):*

Shouldn't we check for the presence of the `--sysroot` and `--sysroot=` flags too? According to GCC documentation, [`-isysroot` is equivalent to `--sysroot` on OS X](https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html):
> -isysroot dir
    This option is like the --sysroot option, but applies only to header files (except for Darwin targets, where it applies to both header files and libraries). See the --sysroot option for more information. 
</blockquote></details>

Agreed with @micbou 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/710)*

puremourning:
OK i have (finally) updated this PR and added some tests. It's now [READY]. Sorry for the delay!

bstaletic:
In case the tests are passing this is :lgtm:. Though, I have a minor question.

---

Reviewed 2 of 2 files at r2.

---

*[ycmd/completers/cpp/flags.py, line 512 at r2](https://reviewable.io:443/reviews/valloric/ycmd/710#-KkbESRRebt3XRyUsXOk:-KkbESRRebt3XRyUsXOl:b-bp0gwy) ([raw file](https://github.com/valloric/ycmd/blob/53e26c9a695f00413c312d0a60e7c979d86e9af6/ycmd/completers/cpp/flags.py#L512)):*
> ```Python
> def _MacIncludePaths():
>   # This method exists for testing only
>   return MAC_INCLUDE_PATHS
> ```

Shall we keep this method? I personally find using `MAC_INCLUDE_PATHS` clearer, but I'd also be fine with keeping it this way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/710#-:-KkbFInbM-f1ZLuAMnHi:b-cvaz9b)*

puremourning:

---

*[ycmd/completers/cpp/flags.py, line 512 at r2](https://reviewable.io:443/reviews/valloric/ycmd/710#-KkbESRRebt3XRyUsXOk:-KkbG7ObrrXja3ErQZB8:blszku) ([raw file](https://github.com/valloric/ycmd/blob/53e26c9a695f00413c312d0a60e7c979d86e9af6/ycmd/completers/cpp/flags.py#L512)):*

Shall we keep this method? I personally find using `MAC_INCLUDE_PATHS` clearer, but I'd also be fine with keeping it this way.
</blockquote></details>

As the comment says, it only exists so that I can mock out MAC_INCLUDE_PATHS in the tests. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/710)*

puremourning:
Looks like the tests are due to some connectivity flake with llvm

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/710#-:-KkbHeTIcguKPhCNC-Qy:b-xfhjoe)*

puremourning:
I retried the appveyor and it built fine, so this is good

vheon:
:lgtm: @zzbot r=bstaletic 

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/710#-:-Kkbj-x6ZcrIStDysvHP:b-43p6gy)*

puremourning:
@zzbot retry

PR #711, opened 2017-02-22
[READY] Fix finding node on Debian-based distributions

micbou:
On Debian-based distributions, `node` is installed by default as `nodejs`. See issue https://github.com/Valloric/YouCompleteMe/issues/2532. We fix that by looking for `nodejs` before `node`. We also need to change how we start the `tsserver` script because it contains the line `#!/usr/bin/env node` at the top.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/711)

### DIFF
--- a/build.py
+++ b/build.py
@@ -450,13 +450,13 @@ def BuildRacerd():
 
 
 def SetUpTern():
-  paths = {}
-  for exe in [ 'node', 'npm' ]:
-    path = FindExecutable( exe )
-    if not path:
-      sys.exit( 'ERROR: {0} is required to set up ternjs.'.format( exe ) )
-    else:
-      paths[ exe ] = path
+  # On Debian-based distributions, node is by default installed as nodejs.
+  node = PathToFirstExistingExecutable( [ 'nodejs', 'node' ] )
+  if not node:
+    sys.exit( 'ERROR: node is required to set up Tern.' )
+  npm = FindExecutable( 'npm' )
+  if not npm:
+    sys.exit( 'ERROR: npm is required to set up Tern.' )
 
   # We install Tern into a runtime directory. This allows us to control
   # precisely the version (and/or git commit) that is used by ycmd.  We use a
@@ -477,7 +477,7 @@ def SetUpTern():
   # (third_party/tern_runtime) that defines the packages that we require,
   # including Tern and any plugins which we require as standard.
   os.chdir( p.join( DIR_OF_THIS_SCRIPT, 'third_party', 'tern_runtime' ) )
-  CheckCall( [ paths[ 'npm' ], 'install', '--production' ] )
+  CheckCall( [ npm, 'install', '--production' ] )
 
 
 def WritePythonUsedDuringBuild():
--- a/ycmd/completers/javascript/tern_completer.py
+++ b/ycmd/completers/javascript/tern_completer.py
@@ -49,7 +49,8 @@ PATH_TO_TERN_BINARY = os.path.abspath(
     'bin',
     'tern' ) )
 
-PATH_TO_NODE = utils.PathToFirstExistingExecutable( [ 'node' ] )
+# On Debian-based distributions, node is by default installed as nodejs.
+PATH_TO_NODE = utils.PathToFirstExistingExecutable( [ 'node', 'nodejs' ] )
 
 # host name/address on which the tern server should listen
 # note: we use 127.0.0.1 rather than localhost because on some platforms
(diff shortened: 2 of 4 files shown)


Valloric:
:lgtm:

Seems like reasonable approach. Thanks for the PR!

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/711#-:-Kdc-8P62BQqLV-IiyAl:bkfsxrh)*

puremourning:
Is the travis failure a failure of travis, or did something break?

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[ycmd/completers/javascript/tern_completer.py, line 53 at r2](https://reviewable.io:443/reviews/valloric/ycmd/711#-KdqSP-J0jRmLlYzxB1T:-KdqSP-J0jRmLlYzxB1U:b-5u37fc) ([raw file](https://github.com/valloric/ycmd/blob/01d59104022bd9dfc01017437540f66f194c9d86/ycmd/completers/javascript/tern_completer.py#L53)):*
> ```Python
> 
> # On Debian-based distributions, node is by default installed as nodejs.
> PATH_TO_NODE = utils.PathToFirstExistingExecutable( [ 'node', 'nodejs' ] )
> ```

I feel like these should be the other way round. I think there is a real thing called `node` on Debian-based systems, which is different from the nodejs that we want. I think this will find the path to that `node`, but we actually want to find `nodejs`. In the installer script, we do the check the other way round, so i feel like it should be consistent. Did i miss something?

---

*[ycmd/completers/typescript/typescript_completer.py, line 48 at r2](https://reviewable.io:443/reviews/valloric/ycmd/711#-KdqSvByaH9W311F8uGJ:-KdqSvByaH9W311F8uGK:b16vlhg) ([raw file](https://github.com/valloric/ycmd/blob/01d59104022bd9dfc01017437540f66f194c9d86/ycmd/completers/typescript/typescript_completer.py#L48)):*
> ```Python
> 
> # On Debian-based distributions, node is by default installed as nodejs.
> PATH_TO_NODE = utils.PathToFirstExistingExecutable( [ 'node', 'nodejs' ] )
> ```

Same here, i think we should look for `nodejs` first

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/711#-:-KdqTeeJjuemRwwy_Nl9:bliwuxk)*

puremourning:
Reviewed 2 of 3 files at r1, 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/711)*

micbou:
PR #712 fixed the issue. This one is now ready.

---

Reviewed 3 of 3 files at r1, 3 of 3 files at r2, 2 of 2 files at r3.

---

*[ycmd/completers/javascript/tern_completer.py, line 53 at r2](https://reviewable.io:443/reviews/valloric/ycmd/711#-KdqSP-J0jRmLlYzxB1T:-KdqdawXyRbOyD2NBcQj:b6xu9x) ([raw file](https://github.com/valloric/ycmd/blob/01d59104022bd9dfc01017437540f66f194c9d86/ycmd/completers/javascript/tern_completer.py#L53)):*

I feel like these should be the other way round. I think there is a real thing called `node` on Debian-based systems, which is different from the nodejs that we want. I think this will find the path to that `node`, but we actually want to find `nodejs`. In the installer script, we do the check the other way round, so i feel like it should be consistent. Did i miss something?
</blockquote></details>

No, you are right. Done.

---

*[ycmd/completers/typescript/typescript_completer.py, line 48 at r2](https://reviewable.io:443/reviews/valloric/ycmd/711#-KdqSvByaH9W311F8uGJ:-KdqdlQ-eQYubOlEKtk9:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/01d59104022bd9dfc01017437540f66f194c9d86/ycmd/completers/typescript/typescript_completer.py#L48)):*

Same here, i think we should look for `nodejs` first
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/711#-:-KdqdUrPPVCn9T9baqoo:b-y7pz9v)*

puremourning:
:lgtm:

---

Reviewed 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/711#-:-KdqhgM1zOELhvWOcJhp:bnfp4nl)*

vheon:
:lgtm: @homu r=valloric

---

Reviewed 1 of 3 files at r1, 2 of 3 files at r2, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/711#-:-Kdrc6BA77-EayWAoEix:b8kx2cz)*

PR #716, opened 2017-03-09
[READY] Rename JediCompleter to PythonCompleter

micbou:
The completer name for Python should not be tied to Jedi. For instance, if we decide to replace Jedi with another semantic engine or if we add diagnostics support through flake8, `JediCompleter` would not be a suitable name anymore and we would have to change it.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/716)

### DIFF
--- a/ycmd/completers/python/hook.py
+++ b/ycmd/completers/python/hook.py
@@ -23,8 +23,8 @@ from future import standard_library
 standard_library.install_aliases()
 from builtins import *  # noqa
 
-from ycmd.completers.python.jedi_completer import JediCompleter
+from ycmd.completers.python.python_completer import PythonCompleter
 
 
 def GetCompleter( user_options ):
-  return JediCompleter( user_options )
+  return PythonCompleter( user_options )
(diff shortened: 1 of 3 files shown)


puremourning:
not so sure about this. i always thought the completer was named after the completion engine, so that you could (in theory) have multiple completers for a single language. e.g. clang_completer, tern_completer, etc.

I actually have 2 for java in my language server branch.

Valloric:
@puremourning I do agree with that, but we're not really consistent about it; for example, `CsharpCompleter` instead of `OmniSharpCompleter`. 

Personally, I don't really care too much and I'm fine with both approaches. The idea that the current system supports multiple semantic engines for one language is theoretically true, but in practice we wouldn't really allow that. I don't think we should ever get in the business of maintaining multiple engines per language and I definitely don't want users to have to think about which one to use. We've always tried to make YCM have a great out-of-the-box experience. If there are two semantic engines we _could_ use, let's just use the one that's better for most use-cases and be done with it.

A great argument for not doing this renaming would be that we don't currently have a need for it. I'm not aware of plans to change our Python completer to something else (though I'm not against the idea). And even if we decided to do that, we can perform the renaming then. :)

TL;DR: [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it).

micbou:
Let's rename it when needed then. Closing.

PR #719, opened 2017-03-12
[READY] Remove Boost.Regex in favour of std variant

bstaletic:
The final piece in boost to std transition. Continuation of #705.

To make this work on linux with gcc `ReadUtf8File()` had to be changed to return a string vector. Otherwise no matches were ever found. Seems like a gcc bug.

That being said tests on linux are stll failing. On OSX on the other hand tests are all passing. The tests that are failing are:

- IdentifierCompleterTest.TagsEndToEndWorks
- IdentifierUtilsTest.ExtractIdentifiersFromTagsFileWorks

Even though the tests are failing on linux, reading the `basic.tags` file results in the expected behaviour (i.e. tags are provided for completion).

Note:
This leaves only boost::filesystem and boost::python. So `update_boost.py` should be updated, as well as some parts of `CMakeLists.txt` that mention boost threads that are no longer in use.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/719)

### DIFF
--- a/cpp/ycm/IdentifierUtils.cpp
+++ b/cpp/ycm/IdentifierUtils.cpp
@@ -124,10 +124,8 @@ FiletypeIdentifierMap ExtractIdentifiersFromTagsFile(
   std::smatch matches;
   const std::regex expression( TAG_REGEX );
 
-  for ( auto line : tags_file_contents )
-  {
-    if ( std::regex_search( line, matches, expression ) )
-    {
+  for ( auto line : tags_file_contents ) {
+    if ( std::regex_search( line, matches, expression ) ) {
       std::string language( matches[ 3 ] );
       std::string filetype = FindWithDefault( LANG_TO_FILETYPE,
                                               language.c_str(),
@@ -140,7 +138,8 @@ FiletypeIdentifierMap ExtractIdentifiersFromTagsFile(
       path = fs::absolute( path, path_to_tag_file.parent_path() )
               .make_preferred();
 
-      filetype_identifier_map[ filetype ][ path.string() ].push_back( identifier );
+      filetype_identifier_map[ filetype ][ path.string() ]
+              .push_back( identifier );
     }
   }
 
--- a/cpp/ycm/IdentifierUtils.cpp
+++ b/cpp/ycm/IdentifierUtils.cpp
@@ -19,12 +19,11 @@
 #include "Utils.h"
 
 #include <unordered_map>
-#include <boost/regex.hpp>
-#include <boost/algorithm/string/regex.hpp>
+#include <regex>
 
 namespace YouCompleteMe {
 
-namespace fs = boost::filesystem;
+namespace fs = std::filesystem;
 
 namespace {
 
@@ -37,9 +36,9 @@ const char *const TAG_REGEX =
   // The second field is the path to the file that has the identifier; either
   // absolute or relative to the tags file.
   "([^\\t\\n\\r]+)"
-  "\\t.*?"  // Non-greedy everything
+  "\\t[^\\n]*?"  // Non-greedy everything
   "language:([^\\t\\n\\r]+)"  // We want to capture the language of the file
-  ".*?$";
+  "[^\\n]*?$";
 
 // Only used as the equality comparer for the below unordered_map which stores
 // const char* pointers and not std::string but needs to hash based on string
@@ -125,11 +124,10 @@ FiletypeIdentifierMap ExtractIdentifiersFromTagsFile(
   std::string::const_iterator start = tags_file_contents.begin();
   std::string::const_iterator end   = tags_file_contents.end();
 
-  boost::smatch matches;
-  const boost::regex expression( TAG_REGEX );
-  const boost::match_flag_type options = boost::match_not_dot_newline;
+  std::smatch matches;
+  const std::regex expression( TAG_REGEX );
 
-  while ( boost::regex_search( start, end, matches, expression, options ) ) {
+  while ( std::regex_search( start, end, matches, expression, options ) ) {
     start = matches[ 0 ].second;
 
     std::string language( matches[ 3 ] );
--- a/cpp/ycm/IdentifierUtils.cpp
+++ b/cpp/ycm/IdentifierUtils.cpp
@@ -23,7 +23,7 @@
 
 namespace YouCompleteMe {
 
-namespace fs = std::filesystem;
+namespace fs = boost::filesystem;
 
 namespace {
 
@@ -113,7 +113,7 @@ const char *const NOT_FOUND = "YCMFOOBAR_NOT_FOUND";
 FiletypeIdentifierMap ExtractIdentifiersFromTagsFile(
   const fs::path &path_to_tag_file ) {
   FiletypeIdentifierMap filetype_identifier_map;
-  std::string tags_file_contents;
+  std::vector< std::string > tags_file_contents;
 
   try {
     tags_file_contents = ReadUtf8File( path_to_tag_file );
@@ -121,29 +121,27 @@ FiletypeIdentifierMap ExtractIdentifiersFromTagsFile(
     return filetype_identifier_map;
   }
 
-  std::string::const_iterator start = tags_file_contents.begin();
-  std::string::const_iterator end   = tags_file_contents.end();
-
   std::smatch matches;
   const std::regex expression( TAG_REGEX );
 
-  while ( std::regex_search( start, end, matches, expression, options ) ) {
-    start = matches[ 0 ].second;
+  for ( auto line : tags_file_contents )
+  {
+    if ( std::regex_search( line, matches, expression ) )
+    {
+      std::string language( matches[ 3 ] );
+      std::string filetype = FindWithDefault( LANG_TO_FILETYPE,
+                                              language.c_str(),
+                                              NOT_FOUND );
+      if ( filetype == NOT_FOUND )
+        continue;
 
-    std::string language( matches[ 3 ] );
-    std::string filetype = FindWithDefault( LANG_TO_FILETYPE,
-                                            language.c_str(),
-                                            NOT_FOUND );
+      std::string identifier( matches[ 1 ] );
+      fs::path path( matches[ 2 ].str() );
+      path = fs::absolute( path, path_to_tag_file.parent_path() )
+              .make_preferred();
 
-    if ( filetype == NOT_FOUND )
-      continue;
-
-    std::string identifier( matches[ 1 ] );
-    fs::path path( matches[ 2 ].str() );
-    path = fs::absolute( path, path_to_tag_file.parent_path() )
-           .make_preferred();
-
-    filetype_identifier_map[ filetype ][ path.string() ].push_back( identifier );
+      filetype_identifier_map[ filetype ][ path.string() ].push_back( identifier );
+    }
   }
 
   return filetype_identifier_map;
--- a/cpp/ycm/IdentifierUtils.cpp
+++ b/cpp/ycm/IdentifierUtils.cpp
@@ -31,14 +31,14 @@ namespace {
 // http://ctags.sourceforge.net/FORMAT
 // TL;DR: The only supported format is the one Exuberant Ctags emits.
 const char *const TAG_REGEX =
-  "^([^\\t\\n\\r]+)"  // The first field is the identifier
+  "^([^\\t]+)"  // The first field is the identifier
   "\\t"  // A TAB char is the field separator
   // The second field is the path to the file that has the identifier; either
   // absolute or relative to the tags file.
-  "([^\\t\\n\\r]+)"
-  "\\t[^\\n]*?"  // Non-greedy everything
-  "language:([^\\t\\n\\r]+)"  // We want to capture the language of the file
-  "[^\\n]*?$";
+  "([^\\t]+)"
+  "\\t.*"
+  "language:([^\\t]+)"  // We want to capture the language of the file
+  ".*$";
 
 // Only used as the equality comparer for the below unordered_map which stores
 // const char* pointers and not std::string but needs to hash based on string
@@ -112,34 +112,34 @@ const char *const NOT_FOUND = "YCMFOOBAR_NOT_FOUND";
 
 FiletypeIdentifierMap ExtractIdentifiersFromTagsFile(
   const fs::path &path_to_tag_file ) {
-  FiletypeIdentifierMap filetype_identifier_map;
   std::vector< std::string > tags_file_contents;
 
   try {
     tags_file_contents = ReadUtf8File( path_to_tag_file );
   } catch ( ... ) {
-    return filetype_identifier_map;
+    return {};
   }
 
+  FiletypeIdentifierMap filetype_identifier_map;
   std::smatch matches;
-  const std::regex expression( TAG_REGEX );
+  const std::regex expression( TAG_REGEX, std::regex::optimize );
 
-  for ( auto line : tags_file_contents ) {
+  for ( const auto& line : tags_file_contents ) {
     if ( std::regex_search( line, matches, expression ) ) {
-      std::string language( matches[ 3 ] );
-      std::string filetype = FindWithDefault( LANG_TO_FILETYPE,
+      const std::string language( matches[ 3 ] );
+      const std::string filetype = FindWithDefault( LANG_TO_FILETYPE,
                                               language.c_str(),
                                               NOT_FOUND );
       if ( filetype == NOT_FOUND )
         continue;
 
-      std::string identifier( matches[ 1 ] );
+      const std::string identifier( matches[ 1 ] );
       fs::path path( matches[ 2 ].str() );
       path = fs::absolute( path, path_to_tag_file.parent_path() )
               .make_preferred();
 
       filetype_identifier_map[ filetype ][ path.string() ]
-              .push_back( identifier );
+              .emplace_back( identifier );
     }
   }
 
--- a/cpp/ycm/IdentifierUtils.cpp
+++ b/cpp/ycm/IdentifierUtils.cpp
@@ -128,8 +128,8 @@ FiletypeIdentifierMap ExtractIdentifiersFromTagsFile(
     if ( std::regex_search( line, matches, expression ) ) {
       const std::string language( matches[ 3 ] );
       const std::string filetype = FindWithDefault( LANG_TO_FILETYPE,
-                                              language.c_str(),
-                                              NOT_FOUND );
+                                                    language.c_str(),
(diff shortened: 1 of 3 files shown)


Valloric:
> To make this work on linux with gcc ReadUtf8File() had to be changed to return a string vector. 

Not bothered by that at all. :)

---

Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[cpp/ycm/Utils.cpp, line 41 at r1](https://reviewable.io:443/reviews/valloric/ycmd/719#-Kf2QdcIvPfo-psi8Z5a:-Kf2QdcIvPfo-psi8Z5b:b91xtej) ([raw file](https://github.com/valloric/ycmd/blob/dee1e0008cfa21e94d6cb472fab5bf76068a9603/cpp/ycm/Utils.cpp#L41)):*
> ```cpp
>   while( getline( file, line ) )
>   {
> 	  contents.push_back(line);
> ```

Formatting looks off; seems to have a tab instead of spaces and the brace is in the wrong position.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/719#-:-Kf2QsecQwZTWsJox2Rz:b-abohcx)*

bstaletic:

---

*[cpp/ycm/Utils.cpp, line 41 at r1](https://reviewable.io:443/reviews/valloric/ycmd/719#-Kf2QdcIvPfo-psi8Z5a:-Kf2_3T9KPzQiJ5aArlr:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/dee1e0008cfa21e94d6cb472fab5bf76068a9603/cpp/ycm/Utils.cpp#L41)):*

Formatting looks off; seems to have a tab instead of spaces and the brace is in the wrong position.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/719)*

Valloric:
There seem to be some more formatting issues; could you run our `style_format.sh` script from the repo root? It should take care of issues.

bstaletic:
The script didn't do absolutely anything. So I did my best to manually fix all the styling issues.

As for the linux tests failing, I'm really confused that those fail only on Linux. Does anyone have a clue as to what is going on?

vheon:
Reviewed 1 of 3 files at r1, 2 of 2 files at r3.

---

*[cpp/ycm/Utils.cpp, line 43 at r3](https://reviewable.io:443/reviews/valloric/ycmd/719#-Kf8Qjn3lEqK51imbB-b:-Kf8Qjn42jhJUajiea7x:b-i5isxd) ([raw file](https://github.com/valloric/ycmd/blob/54ebfe848dcbbd000b0e4ae0a7fabf880be32fd2/cpp/ycm/Utils.cpp#L43)):*
> ```cpp
>   }
> 
>   if ( contents.size() == 0 )
> ```

Can't we just return `contents` here as well? if `contents.size()` is equal to `0`, returning it would be the same as `std::vector< std::string >()`, or am I missing something?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/719)*

vheon:
From the error looks like there is some problem with the regex definition 😕  What I'm worried about is that the Travis build uses GCC 4.8.5 where the C++11 support is experimental apparently, so I'm not sure if is a real problem with the code right now or if it is the GCC regex library. I don't have my VM at hand now to check it :/ @bstaletic did you try our Vagrant env?

bstaletic:
It fails on Vagrant env as well. I didn't run the tests on my host, which is Arch with GCC-6.3.1. The reason I didn't try it on host is because, when I quickly tried to set it up, the script `run_tests.py` failed to run.

---

---

*[cpp/ycm/Utils.cpp, line 43 at r3](https://reviewable.io:443/reviews/valloric/ycmd/719#-Kf8Qjn3lEqK51imbB-b:-Kf8rgQsvCCz7r11mUzZ:b-okww8h) ([raw file](https://github.com/valloric/ycmd/blob/54ebfe848dcbbd000b0e4ae0a7fabf880be32fd2/cpp/ycm/Utils.cpp#L43)):*

Can't we just return `contents` here as well? if `contents.size()` is equal to `0`, returning it would be the same as `std::vector< std::string >()`, or am I missing something?
</blockquote></details>

You're right. No need for additional check.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/719#-:-Kf8pVdMF86zHu0feiff:b-y95hk8)*

bstaletic:
Update:
I realised I could follow the full instalation guide and simply target `ycm_core_tests` instead of `ycm_core`. I wish I could say otherwise, but the tests failed on GCC-6.3.1 too.

---

Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/719#-:-Kf9KFoYgMfdjBjWw2iH:bw3hi60)*

bstaletic:
I have some more (good|bad)? news.

I've been playing with ycmd on another branch (my toy branch - labeled move fast and break things). When I had ran the core tests I noticed that all 68 of them are passing. That made me scratch my head. 

I went back to this branch and... again all tests are passing. Made some more tests trying to find the relevant change. Libclang? Bundled boost vs system provided? Since I made no changes to this branch neither made any sense.

So I played with different parameters for `build.py`. Yet whatever I had tried I couldn't get these tests to fail.

TL;DR:
On my machine the tests are passing. At least with GCC-6.3.1. No changes have been made to the branch.

I'll go and try Vagrant now, but it will take a while, as I need to reinitialise the environment.

bstaletic:
First test on Vagrant is done.

With everything set to the default value in Vagrant env (python2.6, donwloading libclang and compiling bundled boost) it fails by throwing "bad regex" exception.

I wanted to test Vagrant on python3.3 (though I don't see python as a possible reason for test failures), but running `./build.py` produced:
```
CMake Error at /usr/share/cmake-2.8/Modules/FindPackageHandleStandardArgs.cmake:108 (message):
  Could NOT find PythonLibs (missing: PYTHON_INCLUDE_DIRS) (found suitable
  version "3.3.0", minimum required is "3.3")
```
after running `pyenv global 3.3.0`. I'll try to make python3 on Vagrant work and see what happens.

bstaletic:
Since this still has a chance of getting merged, I solved the merge conflict introduced after merging #726.

Bu11etmagnet:
Not the best of titles. There will be a `std::variant` in C++17 which is utterly unrelated to any of the regex classes.

vheon:
@bstaletic didn't you reverted the change of #726?

bstaletic:
@vheon Thanks for pointing that out.

Note to self: Don't rebase at 04:00AM.

cpp/ycm/IdentifierUtils.cpp
```diff
@@ -114,38 +113,34 @@ const char *const NOT_FOUND = "YCMFOOBAR_NOT_FOUND";
...
-    std::string identifier( matches[ 1 ] );
-    fs::path path( matches[ 2 ].str() );
-    path = fs::absolute( path, path_to_tag_file.parent_path() )
-           .make_preferred();
-
-    filetype_identifier_map[ filetype ][ path.string() ].push_back( identifier );
+  std::smatch matches;
+  const std::regex expression( TAG_REGEX );
+
+  for ( auto line : tags_file_contents ) {
```
fsimonis: Use `const auto& line` to prevent copy construction of lines.

cpp/ycm/IdentifierUtils.cpp
```diff
@@ -37,9 +36,9 @@ const char *const TAG_REGEX =
   // The second field is the path to the file that has the identifier; either
   // absolute or relative to the tags file.
   "([^\\t\\n\\r]+)"
```
fsimonis: You changed `ReadUtf8File` to return a vector of lines, thus you do not have to check for `\r` and `\n` anymore.

cpp/ycm/IdentifierUtils.cpp
```diff
@@ -114,38 +113,34 @@ const char *const NOT_FOUND = "YCMFOOBAR_NOT_FOUND";
...
-      continue;
-
-    std::string identifier( matches[ 1 ] );
-    fs::path path( matches[ 2 ].str() );
-    path = fs::absolute( path, path_to_tag_file.parent_path() )
-           .make_preferred();
-
-    filetype_identifier_map[ filetype ][ path.string() ].push_back( identifier );
+  std::smatch matches;
+  const std::regex expression( TAG_REGEX );
```
fsimonis: You can pass `std::regex::optimize` to the regex. It might be worth a try as the regex is used multiple times. A one time construction would be ideal of course, especially as `TAG_REGEX`is a `const`.

From [cppreference](http://en.cppreference.com/w/cpp/regex/syntax_option_type)
> Instructs the regular expression engine to make matching faster, with the potential cost of making construction slower. For example, this might mean converting a non-deterministic FSA to a deterministic FSA.

cpp/ycm/IdentifierUtils.cpp
```diff
@@ -114,38 +113,34 @@ const char *const NOT_FOUND = "YCMFOOBAR_NOT_FOUND";
 FiletypeIdentifierMap ExtractIdentifiersFromTagsFile(
   const fs::path &path_to_tag_file ) {
   FiletypeIdentifierMap filetype_identifier_map;
-  std::string tags_file_contents;
+  std::vector< std::string > tags_file_contents;
 
   try {
     tags_file_contents = ReadUtf8File( path_to_tag_file );
   } catch ( ... ) {
     return filetype_identifier_map;
```
fsimonis: You should be able to write `return {};` (C++11) to default construct the result.
This allows you to move the line `FiletypeIdentifierMap filetype_identifier_map;` further down in front of the loop.

cpp/ycm/IdentifierUtils.cpp
```diff
@@ -114,38 +113,34 @@ const char *const NOT_FOUND = "YCMFOOBAR_NOT_FOUND";
...
-    fs::path path( matches[ 2 ].str() );
-    path = fs::absolute( path, path_to_tag_file.parent_path() )
-           .make_preferred();
-
-    filetype_identifier_map[ filetype ][ path.string() ].push_back( identifier );
+  std::smatch matches;
+  const std::regex expression( TAG_REGEX );
+
+  for ( auto line : tags_file_contents ) {
+    if ( std::regex_search( line, matches, expression ) ) {
```
fsimonis: You can define all the `std::string`s in this block as `const`.

cpp/ycm/IdentifierUtils.cpp
```diff
@@ -114,38 +113,34 @@ const char *const NOT_FOUND = "YCMFOOBAR_NOT_FOUND";
...
+      if ( filetype == NOT_FOUND )
+        continue;
+
+      std::string identifier( matches[ 1 ] );
+      fs::path path( matches[ 2 ].str() );
+      path = fs::absolute( path, path_to_tag_file.parent_path() )
+              .make_preferred();
+
+      filetype_identifier_map[ filetype ][ path.string() ]
+              .push_back( identifier );
```
fsimonis: You can use `.emplace_back(std::move(identifier));` as it is the last usage of `identifier`.

bstaletic:
I ran the same benchmark with the changes @fsimonis proposed. They made no difference to the performance of `std::regex`. The only change that I noticed in performance is that 1st run of optimised `std::regex` was now stable at 0.6 seconds instead of varying by plus or minus 0.1 second.

I guess the compiler is already smart enough to incorporate these changes instead of the programmer. Either way the regex itself looks farmore readable after the changes and the rest of the changes seem more than reasonable.

> A one time construction would be ideal of course, especially as TAG_REGEXis a const.

The regex is already constructed one time only and is declared `const`. Am I missing anything?

bstaletic:
Now that #723 is merged, this should pass the tests. Is there a way to rerun the tests?

vheon:
@bstaletic I see you are a `Contributor` don't you have access to Travis? I'll do it without problem but just to know if you know how to do it 😃

vheon:
Anyway if you have access to Travis on the right there is a button saying `Restart build`

vheon:
Ops I think I confused `Contributor` with `Collaborator` 😝  sorry, I'll do it.

bstaletic:
The restarted build didn't pick up on using gcc-4.9 and failed. I pushed the commit with changes @fsimonis proposed and it solved that problem as well.

This should be ready for merge now.

@vheon I've seen the `Restart build` on travis on my fork of ycmd, but can't see it here.

Valloric:
:lgtm: 

Thanks for the PR!

---

Review status: 1 of 3 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/719#-:-KhPKiNRU58MZ97esZd2:b-y8s7n3)*

bstaletic:
Can anyone explain the unexpected coverage changes? I won't mind writing some tests, but I can't make sense of codecov's report.

Valloric:
> Can anyone explain the unexpected coverage changes? I won't mind writing some tests, but I can't make sense of codecov's report.

Good question. Codecov provides a "changes", "patch" and "project" coverage in the github summary and I'm not sure I understand them all. I think the only part that is ultimately relevant is this one:

> codecov/project — 93.32% (+0.22%) compared to 13da3d4

Which says that coverage goes up with your PR, which is all I care about.

Don't forget, you can get coverage info locally; you don't have to rely on understanding codecov's UI. 

What I think codecov is trying to say here (with its terrible UX) is that some other lines in the codebase are now not being covered. Might be worth figuring out why that is so, but honestly I wouldn't worry too much about it.

vheon:
The "patch" coverage should be how much of the code of the PR is covered.

bstaletic:
And the "changes" reports if there are "unexpected changes". For this one the documentation only mentions missing tests i.e. test was there but now isn't.

Codecov complains about two files:

- `Utils.cpp` - line: `while( getline( file, line ) )`
- `IdentifierUtils.cpp` - line: `if ( std::regex_search( line, matches, expression ) ) {`

Both lines are partially covered and I'm not sure how to cover them completely.

---

Review status: 1 of 3 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/719#-:-KhTrWkpfdZZfP2spGbN:b88rtug)*

puremourning:
Also worth noting that the codecov guys on Gitter are really good peeps and usually happy to answer reasonable questions :)

vheon:
Just a minor comment otherwise :lgtm:

---

Reviewed 2 of 2 files at r7.

---

*[cpp/ycm/IdentifierUtils.cpp, line 131 at r7](https://reviewable.io:443/reviews/valloric/ycmd/719#-KhwsYFtpkZEkJ9xnizX:-KhwsYFulESef0nwIO6m:b93lsg1) ([raw file](https://github.com/valloric/ycmd/blob/a8e649b067a0c0f395488e97eab7729bb7f0afb6/cpp/ycm/IdentifierUtils.cpp#L131)):*
> ```cpp
>       const std::string language( matches[ 3 ] );
>       const std::string filetype = FindWithDefault( LANG_TO_FILETYPE,
>                                               language.c_str(),
> ```

Did we messed up the indentation here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/719#-:-Khwsois7xDgyxbHbRAO:bh86k4y)*

bstaletic:
I just heard from Steve from codecov. He asked me to confirm if I want him to review the pull request's coverage data.
Let's see what the latest push reports. If it still reports unexpected changes I'll confirm.

---

Review status: 2 of 3 files reviewed at latest revision, 7 unresolved discussions.

---

*[cpp/ycm/IdentifierUtils.cpp, line 131 at r7](https://reviewable.io:443/reviews/valloric/ycmd/719#-KhwsYFtpkZEkJ9xnizX:-KhxV0_x1BGExyENYBWv:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/a8e649b067a0c0f395488e97eab7729bb7f0afb6/cpp/ycm/IdentifierUtils.cpp#L131)):*

Did we messed up the indentation here?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/719#-:-KhxV5HY0T3jkjq6bWsE:bn5vhxy)*

bstaletic:
Can anyone restart the builds? The errors are spurious.
Travis couldn't download something when running some `apt-get` command.
Appveyor on python2.7 couldn't find CXX compiler.

puremourning:
I kicked tragic but I don't think i can do appveyor

Valloric:
Kicked appveyor.

bstaletic:
I've looked carefully at all the files codecov reports changes on and tried to find an explanation as to why there are changes at all.

- `IdentifierCompleter.cpp` - Specific constructor is now covered, but I have no actual explanation.
- `IdentifierDatabase.cpp` - Range based for statements used to be partially covered. This seems like a nice side effect of using a newer gcc.
- `IdentifierUtils.cpp` - It has one line less to be covered so its relative coverage went down (that one line was covered), but it still only has one partially covered line - only a different one.
- `Result.cpp` - Another partially covered line fully covered. Assuming GCC version.
- `PythonSupport.cpp` - No explanation.
- `ClangUtils.cpp` - Functions that are never used are reported as covered.
- `CompletionData.cpp` - No explanation.
- `Diagnostic.h` - Already was at 0% coverage. No explanation for the diff.
- `TranslationUnitStore.cpp` - No explanation.

In the mean time I asked Steve from codecov to review the PR.

bstaletic:
Oh, forgot about something.

- `utils.py` - Some comments and docstrings are reported as covered. This makes no sense.

vheon:
Reviewed 1 of 1 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/719)*

vheon:
@bstaletic I think that we can safely merge this, if codecov want to check if there are some problem with their product they have all the data here with this PR

bstaletic:
@vheon I'm okay with merging if you guys are. I just don't like seeing failing checks in PR's. On the other hand, I see nothing wrong with the actual code.

Valloric:
Yep, let's merge this. Thanks again!

@homu r+

bstaletic:
> Thanks again!

Don't mention it!

Now we only need to wait for boost to release version 1.64, which should happen today (GMT+0 though). After that I'll post the PR updating the `BoostParts`, `update_boost.py` and `CMakeLists.txt`.

PR #721, opened 2017-03-14
[READY] Update Clang to 4.0.0

micbou:
Two changes were required to build against Clang 4.0.0:
 - renaming `libclang.so.4.0` to `libclang.so.4` on Linux. Otherwise, the dynamic linker will not find libclang when loading `ycm_core` because it's looking for `libclang.so.4` instead of `libclang.so.4.0` (even though we explicitly link against the latter);
 - removing the RPATH fix on macOS: `libclang.dylib` doesn't have the `@executable_path/../lib` LC_RPATH entry anymore. If we don't remove the fix, we get the following error from `install_name_tool`:
   
   ```
   error: install_name_tool: no LC_RPATH load command with path:
   @executable_path/../lib found in: libclang.dylib (for architecture x86_64),
   required for specified option "-delete_rpath @executable_path/../lib"
   ```

No changes were needed to pass the tests.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/721)


micbou:
There is no pre-built Clang 4.0.0 available for 32-bit Linux so we raise an error in this case.

---

Reviewed 1 of 42 files at r1.
Review status: 1 of 41 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/721#-:-KfCgztxYOoGo4W7e4o9:bv4hjeo)*

Valloric:
We should probably bump the ycm_core version so that people recompile. It's the only way we have of notifying people to pick up the new version, because as of this commit, we don't support any version other than 4.0.

---

Review status: 0 of 41 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/721#-:-KfCv5Sr1pkLU66oJv0x:brms9j7)*

micbou:
Done.

---

Reviewed 41 of 42 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/721#-:-KfCwnrzZuNFJKpI9AYZ:b-896fix)*

vheon:
:lgtm:

---

Reviewed 41 of 42 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/721#-:-KfE0zATaVXeo0tQhYFU:bnfp4nl)*

Valloric:
:lgtm:

Thanks for the PR!

@homu r=vheon

PR #722, opened 2017-03-15
[READY] Do not install aliases from future

micbou:
When calling `standard_library.install_aliases`, `python-future` imports [the `test` module](https://docs.python.org/2/library/test.html) on Python 2 to monkey-patch it. This can fail if the current directory has a file named `test.py` and is part of the `PYTHONPATH` environment variable. In that case, `python-future` will attempt to load the `test.py` file and fail if it contains an error. Here's a simple way to reproduce the issue:
```
$ echo "raise RuntimeError()" > test.py
$ export PYTHONPATH=.
$ python2 /path/to/ycmd/folder/ycmd
  File "C:\Python27\lib\runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "C:\Python27\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "C:\Users\micbou\.vim\bundle\YouCompleteMe\third_party\ycmd\ycmd\__main__.py", line 32, in <module>
    standard_library.install_aliases()
  File "C:\Users\micbou\.vim\bundle\YouCompleteMe\third_party\ycmd\third_party\python-future\src\future\standard_library\__init__.py", line 483, in install_aliases
    import test
  File "C:\Users\micbou\tests\ycm-issue-2568\test.py", line 1, in <module>
    raise RuntimeError()
RuntimeError
```
One could argue that people should avoid naming Python files the same as modules from the standard library. I would agree for something like `os.py` but `test.py` is a really common name. In fact, this issue has already been reported multiple times: https://github.com/Valloric/YouCompleteMe/issues/2332, https://github.com/Valloric/YouCompleteMe/issues/2467 and https://github.com/Valloric/YouCompleteMe/issues/2568. So, we should do something about it. Since we only need the aliases for the `urllib` module (in particular, the `urljoin` and `urlparse` functions), the solution is to remove the lines:
```python
from future import standard_library
standard_library.install_aliases()
```
from the top of each file and import `urljoin` and `urlparse` where we need them with:
```python
from future.moves.urllib.parse import urljoin, urlparse
```

Another benefit of this change is that it improves ycmd startup time on Python 2 (about 20%):

<table>
  <tr>
    <th rowspan="2">Platform</th>
    <th colspan="2">First run (ms)</th>
    <th colspan="2">Subsequent runs (ms)</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
    <td>Before</td>
    <td>After</td>
  </tr>
  <tr>    
    <td>Ubuntu 16.04 64-bit</td>
    <td>229</td>
    <td>178</td>
    <td>202</td>
    <td>156</td>
  </tr>
  <tr>
    <td>macOS 10.12</td>
    <td>257</td>
    <td>204</td>
    <td>222</td>
    <td>173</td>
  </tr>
  <tr> 
    <td>Windows 10 64-bit</td>
    <td>406</td>
    <td>315</td>
    <td>298</td>
    <td>245</td>
  </tr>  
</table>

These results were obtained by running [this script](https://github.com/micbou/ycmd/blob/profiling-startup/startup.py) on each platform. They represent the time it takes for the server to be ready.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/722)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,8 +113,7 @@ Here's what you should watch out for:
     from __future__ import unicode_literals
     from __future__ import print_function
     from __future__ import division
-    from future import standard_library
-    standard_library.install_aliases()
+    # Not installing aliases from python-future; it's unreliable and slow.
     from builtins import *  # noqa
     ```
 
@@ -147,6 +146,12 @@ Here's what you should watch out for:
 - If you run tests and get failures on importing ycm_core that mention
   `initycm_core` or `PyInit_ycm_core`, you've built the C++ parts of ycmd for
   py2 and are trying to run tests in py3 (or vice-versa). Rebuild!
+- Use the `urllib` module from `python-future`:
+
+    ```python
+    from future.moves.urllib.parse import urljoin, urlparse
+    ```
+
 
 [build-bots]: https://travis-ci.org/Valloric/ycmd
 [ycmd-users]: https://groups.google.com/forum/?hl=en#!forum/ycmd-users
--- a/run_tests.py
+++ b/run_tests.py
@@ -4,8 +4,7 @@ from __future__ import print_function
 from __future__ import division
 from __future__ import unicode_literals
 from __future__ import absolute_import
-from future import standard_library
-standard_library.install_aliases()
+# Not installing aliases from python-future; it's unreliable and slow.
 from builtins import *  # noqa
 
 import platform
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -4,7 +4,8 @@ flake8       >= 3.0.0; python_version == '2.7' or python_version >  '3.3'
 mock         >= 1.3.0
 nose         >= 1.3.7
 PyHamcrest   >= 1.8.5
-WebTest      >= 2.0.20
+# WebTest 2.0.24 dropped support of Python 2.6
+WebTest      >= 2.0.20, < 2.0.24
 ordereddict  >= 1.1
 nose-exclude >= 0.4.1
 unittest2    >= 1.1.0
--- a/ycmd/__main__.py
+++ b/ycmd/__main__.py
@@ -28,8 +28,7 @@ sys.path.insert( 0, os.path.dirname( os.path.abspath( __file__ ) ) )
 from server_utils import SetUpPythonPath, CompatibleWithCurrentCore
 SetUpPythonPath()
 
-from future import standard_library
-standard_library.install_aliases()
+# Not installing aliases from python-future; it's unreliable and slow.
 from builtins import *  # noqa
 
 import atexit
--- a/ycmd/bottle_utils.py
+++ b/ycmd/bottle_utils.py
@@ -19,8 +19,7 @@ from __future__ import unicode_literals
 from __future__ import print_function
 from __future__ import division
 from __future__ import absolute_import
-from future import standard_library
-standard_library.install_aliases()
+# Not installing aliases from python-future; it's unreliable and slow.
 from builtins import *  # noqa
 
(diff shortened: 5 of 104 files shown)


vheon:
:lgtm:

---

Reviewed 104 of 104 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/722#-:-KfIvC5MlVca0nw3e2ox:bnfp4nl)*

Valloric:
I agree with the idea, but there should be a comment somewhere (and possibly in all the files that do `from builtins import *`) to not call `install_aliases()`. I don't want this knowledge to remain only in our heads. Someone (us a couple of years from now) will look at the way we use `future` and will ask "why don't we call install_aliases? let's add that."

With such a change, I'm :lgtm: .

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/722#-:-KfNvSC49--vdnhWELbu:b-pb5wcl)*

micbou:
> I agree with the idea, but there should be a comment somewhere (and possibly in all the files that do from builtins import *) to not call install_aliases(). I don't want this knowledge to remain only in our heads. Someone (us a couple of years from now) will look at the way we use future and will ask "why don't we call install_aliases? let's add that."

Aren't the changes in `CONTRIBUTING.md` enough?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/722#-:-KfO01GCoZfOm52j50Xy:bkxss18)*

Valloric:
> Aren't the changes in CONTRIBUTING.md enough?

That change also doesn't explain why we don't just use `install_aliases()`. I don't want to rely on us remembering the reason.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/722#-:-KfOG89N6b_WzKTKYkt7:b-iffd5p)*

micbou:
Fine. Let's agree on what we should say. What do you think of:
```python
from __future__ import absolute_import
from __future__ import unicode_literals
from __future__ import print_function
from __future__ import division
# Not installing aliases from python-future; it's unreliable and slow.
from builtins import *  # noqa
```
?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/722#-:-KfOph9TQO-2ZbItVV-T:bikqwww)*

Valloric:
> Fine. Let's agree on what we should say. What do you think of:

LGTM. :)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/722#-:-KfPBQo2EwHTT7XvQcXe:b83basp)*

micbou:
Updated.

---

Reviewed 1 of 104 files at r1, 103 of 103 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/722#-:-KfRHmNwPRYi9q3fbgYi:b-bw8e14)*

Valloric:
Fantastic, thank you!

@homu r=vheon

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/722#-:-KfUNnqcCPXjCjMhVi5j:b-bgbwek)*

puremourning:
:lgtm:

Oh heh :) Beat me to it ;)

---

Reviewed 1 of 104 files at r1, 102 of 103 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/722#-:-KfXmoTdzFX-dD_pj_Ke:b-6vy048)*

PR #723, opened 2017-03-18
[READY] Require GCC 4.9

micbou:
[GCC 4.9 has improved support for C++11; in particular regex support](https://gcc.gnu.org/gcc-4.9/changes.html). This is needed for PR #719 to pass the tests. See [this build](https://travis-ci.org/micbou/ycmd/builds/212329412) and [the corresponding branch](https://github.com/micbou/ycmd/tree/boost_regex_to_std_regex_test).

cc @bstaletic

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/723)


bstaletic:
I was just about to comment on #719 that it needs GCC 4.9. (:

If ycmd is going to require GCC-4.9 then `cpp/CMakeLists.txt` should be updated.

Specifically, [line 118](https://github.com/Valloric/ycmd/blob/master/cpp/CMakeLists.txt#L118) should say:
```CMake
  if ( NOT  CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.9  )
```

micbou:
> If ycmd is going to require GCC-4.9 then cpp/CMakeLists.txt should be updated.

Done.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KfTwVctHew6brG8cAet:b7ftrxs)*

bstaletic:
Vagrant environment shouldn't be left behind as well.
Ubunto 16.04 has GCC 5.3.1 and the [image](https://atlas.hashicorp.com/boxes/search) for Vagrant is out.

Apart from that this is :lgtm: 

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KfU08z6CNlUBE-rpGMb:b-506y29)*

Valloric:
:lgtm: with the Vagrant image update.

Great to see we got to the bottom of why that test was failing!

Thanks for the PR!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KfUNJUUt1KyjZ1k1_Rh:bimr559)*

puremourning:

---

*[cpp/CMakeLists.txt, line 118 at r1](https://reviewable.io:443/reviews/valloric/ycmd/723#-KfXhDQKkScqR21BbxNY:-KfXhDQKkScqR21BbxNZ:b-ntj5cc) ([raw file](https://github.com/valloric/ycmd/blob/a1567e2e559adaf8fff00672b6d5d10bfdd59cf3/cpp/CMakeLists.txt#L118)):*
>   if ( NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.9 )
>     set( CPP11_AVAILABLE true )
>   endif()

I'm not sure i follow this exactly. We're saying that any compiler older than gcc 4.9 doesn't support C++11 ? That just isn't true.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723)*

puremourning:
I don't feel like the benefit of using `std::regex` (vs. boost) outweighs the (IMO) _large_ drawback of requiring gcc 4.9. We know that we have users on RedHat 6 (like me ^_^) and thus far we've recommended using the dev-toolset 2, which contains gcc 4.8.

yes there are later versions of gcc you can get on those OSes, but IMO switching to `std:regex` doesn't add anything significant functionally (and as we can see can be buggy in the standard library) so IMO it's quite hard to justify.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KfXhZq36TQOdTGgEZjq:bnqsk8j)*

bstaletic:

---

*[cpp/CMakeLists.txt, line 118 at r1](https://reviewable.io:443/reviews/valloric/ycmd/723#-KfXhDQKkScqR21BbxNY:-KfY2qIZsp4vZpWYEP7V:b-hgityd) ([raw file](https://github.com/valloric/ycmd/blob/a1567e2e559adaf8fff00672b6d5d10bfdd59cf3/cpp/CMakeLists.txt#L118)):*

>   if ( NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.9 )
>     set( CPP11_AVAILABLE true )
>   endif()

I'm not sure i follow this exactly. We're saying that any compiler older than gcc 4.9 doesn't support C++11 ? That just isn't true.
</blockquote></details>

That line is saying that any GCC compiler older than 4.9 doesn't have (proper) C++11support.

While it's true that there is *some* C++11 support even in GCC-4.6, GCC has full support for C++11 only since 4.9. Regex is not the only C++11 feature not (completely) implemented in GCC prior to 4.9.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723)*

vheon:
> yes there are later versions of gcc you can get on those OSes, but IMO switching to std:regex doesn't add anything significant functionally (and as we can see can be buggy in the standard library) so IMO it's quite hard to justify.

Well that is true about any switch we did from `boost` to `std`. Functionally we didn't change anything. Is the fact that we compile less `boost` code that change.

Valloric:
> I don't feel like the benefit of using std::regex (vs. boost) outweighs the (IMO) large drawback of requiring gcc 4.9. We know that we have users on RedHat 6 (like me ^_^) and thus far we've recommended using the dev-toolset 2, which contains gcc 4.8.
> yes there are later versions of gcc you can get on those OSes, but IMO switching to std:regex doesn't add anything significant functionally (and as we can see can be buggy in the standard library) so IMO it's quite hard to justify.

These are very reasonable points, as can be expected from Ben. :)

The counterpoint is that gcc 4.9 was released on April 22, 2014, which is almost 3 years ago. We switch to latest libclang as soon as it comes out, so that's at worst a 6 month window. We also explicitly support the latest Ubuntu LTS for out-of-the-box support and that's at worst a 2 year window.

RHEL 6 was released on 10 November 2010, a full six and a half years ago. Yes, plenty of corp environments still have that as their default (as does LinkedIn, where I personally work). I still find it difficult to swallow that we can't fully move to C++11 because of that; it just seems _wrong_, no? Especially given that we've never had explicit out-of-the-box support for latest RHEL, and even if we did, that would be RHEL 7, not RHEL 6. And the IT departments that are still on RHEL 6 are exactly the type that will want to stay on it until somebody pries it from their cold, dead hands.

So I'm leaning towards not being too bothered about RHEL 6. At some point we've got to throw off those shackles.

OTOH, your point that we're not really getting much out of dropping boost::regex is also compelling, so the choice isn't clear-cut.

@bstaletic Is boost::regex the only major part of boost ycmd is using? In other words, if we can get rid of that, can we ultimately get rid of _all_ the boost libraries we use? Because if we can eventually drop all of boost if we drop boost::regex, I'll find that to be a compelling enough reason.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KfYYNwTYlVkXAXZUa0O:brj6740)*

bstaletic:
> Is boost::regex the only major part of boost ycmd is using? In other words, if we can get rid of that, can we ultimately get rid of all the boost libraries we use? Because if we can eventually drop all of boost if we drop boost::regex, I'll find that to be a compelling enough reason.

Not quite. Right now the remaining boost components in ycmd are `regex`, `python` and *`filesystem`*. Regex part is (finally) done, but is left for discussion. `Boost.Python` is a completely different beast I wasn't even intending to tackle. So that only leaves `filesystem`. 

In one of branches in my ycmd fork I managed to successfully replace `boost::filesystem`, but that TS (tachnical service) didn't make it into STL until C++17.

Even the currently latest GCC version has an experimental C++17 support, that is actually named C++1z. On top of that, the header is `experimental/filesystem`, which means it's still a draft from one of the previous versions.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KfYcBgUIauNSpNL_dxH:b-fqevxm)*

Valloric:
Yes, fair points. I also can't believe I forgot about boost::python; we're unlikely to ever be able to migrate away from that.

This makes @puremourning's point stronger. 

@micbou @vheon What are your thoughts on this?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KfYujOLcaLC3hNJ2yBV:b-5djfcv)*

micbou:
Another reason to not do the switch is that `std::regex` may be slower than `boost::regex`. See https://lists.gnu.org/archive/html/lmi/2016-07/msg00028.html and http://stackoverflow.com/questions/14205096/c11-regex-slower-than-python (there is a comparison with `boost::regex` in the accepted answer and it doesn't look good for `std::regex`).

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KfaUfF648JNsj7zpDBK:b-a72t1h)*

bstaletic:
How sure are we that performance matters when it comes to regex and YCM? It's just parsing a (potentially big) tag file.

I'll do a benchmark and post results as soon as I'm able to.

bstaletic:
Regex type | Unoptimised 1st run | Unoptimised subsequent runs | Optimised 1st run | Optimised subsequent runs
----------------|----------------------------|------------------------------|------------------------|---------------------------
std::regex | 2.1s | 1.8s | 0.6s | 0.45s
boost::regex | 0.6s | 0.3s | 0.45s | 0.2s

The benchmark has been done on a >26k tags file with only C language tags. To generate it I used the folloing command:
```bash
find . -regextype posix-extended -regex '.*\.(c|h)' | xargs gcc -M $@ -I -I -I/home/bstaletic/eurobot2k17/src/libs/STM32F4xx_HAL_Driver/Inc/ -I/home/bstaletic/eurobot2k17/src/libs/CMSIS/Device/ST/STM32F4xx/Include/ -I/home/bstaletic/eurobot2k17/src/libs/CMSIS/Include/ -I/home/bstaletic/eurobot2k17/src/initialisation -I/home/bstaletic/eurobot2k17/src/Middlewares/Third_Party/FreeRTOS/Source/portable/GCC/ARM_CM4F/ -I/home/bstaletic/eurobot2k17/src/Middlewares/Third_Party/FreeRTOS/Source/include/ -I/home/bstaletic/eurobot2k17/src/Middlewares/Third_Party/FreeRTOS/Source/CMSIS_RTOS/ | sed -e 's/[\\ ]/\n/g' | sed -e '/^$/d' -e '/:/d' | xargs ctags --fields=+l --langmap=c:.c.h
```

Unoptimised builds were done with `-O0` and unoptimised builds were done with `-O2`.
Values in the table are averages made from 5 consecutive runs. All the numbers were pretty stable except the `std::regex` optimised first run. That one was varying by 0.1 second.

[Code](https://gist.github.com/bstaletic/c8acf19ef0f6b9bea0548c24f6f236a8) used to measure this is just a call to `ExtractFileFromTagsFile()`. To make the most of the difference due to namespace switch, `ReadUtf8File()` was used as it is in #719.

Command used to time the execution:
```bash
time ./a.out
```

Valloric:
Thanks for running some benchmarks; looks like perf is in the same ballpark. Either way, I'm not overly concerned about perf since this is used in a feature (tags extraction) that's IMO deprecated and around only for the sake of backwards compatibility.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KfbPbxuBOzRO7qQE76u:b-5bsji2)*

vheon:
> OTOH, your point that we're not really getting much out of dropping boost::regex is also compelling, so the choice isn't clear-cut.

This is still unclear to me; I mean what did we get from switching from `boost::unique_ptr` to `std::unique_ptr`? they are the same but with the former we could support even older compiler I guess, so on one hand we were C++98 compatible on the other we aren't but I don't see then what we gain except dropping a piece of `boost`.

I'm fine either way, I use RHEL 6 myself but I don't use dev-toolset2 to build ycmd.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KfbieClwbNGc_HH-Z9w:b-x626o3)*

Valloric:
I've given this quite a bit of thought, and I'm leaning towards merging this. The following paragraph from one of my previous replies sums up my thoughts well:

> RHEL 6 was released on 10 November 2010, a full six and a half years ago. Yes, plenty of corp environments still have that as their default (as does LinkedIn, where I personally work). I still find it difficult to swallow that we can't fully move to C++11 because of that; it just seems wrong, no? Especially given that we've never had explicit out-of-the-box support for latest RHEL, and even if we did, that would be RHEL 7, not RHEL 6. And the IT departments that are still on RHEL 6 are exactly the type that will want to stay on it until somebody pries it from their cold, dead hands.

Conclusion:

1. We've never claimed out-of-the-box compatibility with latest RHEL as we do with latest Ubuntu LTS, but _even if we did_, that would be RHEL 7, not 6.
2. Corps on RHEL 6 are likely to stay on it for the foreseeable future. It's not as if there's something right around the corner that would make us staying on pre-gcc4.9 a temporary prospect.
3. Ties into the previous point: we need to have a timeline for moving to newer libs, compilers etc. That's why Ubuntu LTS is a nice target; every 2 years, a new one comes out and that's our baseline (our LTS target changes 6 months after the new LTS is out). With "let's support RHEL 6," there's no such timeline for progress. RHEL 7 is already out.
4. It irks me to no end that we shouldn't move forward because some corp IT departments are silly.

This is going to be one of those "damned if we do, damned if we don't" decisions where no option makes everybody happy. So I say let's proceed.

vheon:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KhIqYWFWfoBLRu3cKx7:bnfp4nl)*

puremourning:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KhIwc0Lun0_kBw9NmMt:bnfp4nl)*

Valloric:
:lgtm: from me as well. Thanks everyone for the thoughtful comments & discussion! 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KhJ8nQUB-UFE0eJ5vEr:b-vgay4e)*

micbou:
I've changed the message about the compiler not supporting C++11 so that users won't argue about it. This is now ready.

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723#-:-KhJE-_i3_XNokhMN-qp:b6ttpwd)*

vheon:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/723)*

bstaletic:
@micbou The vagrant image still needs to be changed.

micbou:
I tried to update it and make the tests pass but wasn't successful. I'll let someone else do the work in a separate PR.

bstaletic:
I'm trying to get the new image to work right now.

bstaletic:
Sorry, guys, but I thought I could have a quick look at it and possibly solve it. Instead I messed the Vagrant file enough to have no idea where is the directory that `vagrant ssh` changes to. I'm also in other projects up to my neck this april. So I'll have to leave it to someone else as well.

Valloric:
Let's merge this now and we'll fix the vagrant image later.

@homu r=vheon

PR #724, opened 2017-03-19
[READY] Do not import urljoin and urlparse from python-future

micbou:
PR #722 was supposed to fix the issue where python-future imports any `test.py` file from the Python path. Unfortunately, it now happens on Python 3 instead of Python 2. See https://github.com/Valloric/YouCompleteMe/issues/2577 and https://github.com/PythonCharmers/python-future/issues/268 for an explanation. To avoid that, we reproduce the `urllib` compatibility fix from python-future in `ycmd/utils.py` but only for `urljoin` and `urlparse` since we don't need the other functions. We then import `urljoin` and `urlparse` from `ycmd.utils` instead of `future.moves.urllib.parse`.

Once this is merged, we'll do the same in YCM.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/724)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -146,10 +146,10 @@ Here's what you should watch out for:
 - If you run tests and get failures on importing ycm_core that mention
   `initycm_core` or `PyInit_ycm_core`, you've built the C++ parts of ycmd for
   py2 and are trying to run tests in py3 (or vice-versa). Rebuild!
-- Use the `urllib` module from `python-future`:
+- Import the `urljoin` and `urlparse` functions from `ycmd/utils.py`:
 
     ```python
-    from future.moves.urllib.parse import urljoin, urlparse
+    from ycmd.utils import urljoin, urlparse
     ```
 
 
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -26,7 +26,6 @@ from builtins import *  # noqa
 
 from collections import defaultdict
 from future.utils import itervalues
-from future.moves.urllib.parse import urljoin
 import logging
 import os
 import re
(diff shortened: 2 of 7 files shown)


puremourning:
:lgtm: with minor comment. 

---

Reviewed 7 of 7 files at r1.

---

*[ycmd/utils.py, line 36 at r1](https://reviewable.io:443/reviews/valloric/ycmd/724#-KfaxKHHm_N8tBrXridD:-KfaxKHHm_N8tBrXridE:b-r1a0s9) ([raw file](https://github.com/valloric/ycmd/blob/7c7e5625f9b174922ad2d2baf1800e32d2f22ad1/ycmd/utils.py#L36)):*
> ```Python
> 
> 
> if PY2:
> ```

Let's add a comment here explaining why we're doing this. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/724#-:-KfaxQ9KJlY77YUQl9F3:b-68r0lh)*

Valloric:
:lgtm: with minor comment.

Thanks for the PR!

---

---

*[ycmd/utils.py, line 36 at r1](https://reviewable.io:443/reviews/valloric/ycmd/724#-KfaxKHHm_N8tBrXridD:-KfbPO4WJPc4gKj8mhZG:b-xki1by) ([raw file](https://github.com/valloric/ycmd/blob/7c7e5625f9b174922ad2d2baf1800e32d2f22ad1/ycmd/utils.py#L36)):*

Let's add a comment here explaining why we're doing this. 
</blockquote></details>

Agreed, needs a comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/724#-:-KfbPRd_gfhtMoEnqt57:bio5jzb)*

micbou:
Reviewed 7 of 7 files at r1, 1 of 1 files at r2.

---

*[ycmd/utils.py, line 36 at r1](https://reviewable.io:443/reviews/valloric/ycmd/724#-KfaxKHHm_N8tBrXridD:-KfbSZvrw0qWF9A496Am:bs3mhdg) ([raw file](https://github.com/valloric/ycmd/blob/7c7e5625f9b174922ad2d2baf1800e32d2f22ad1/ycmd/utils.py#L36)):*

Agreed, needs a comment.
</blockquote></details>

Comment added.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/724)*

Valloric:
@homu r=puremourning

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/724#-:-KfeP452cUqG3WQv0hXx:bgppt70)*

PR #729, opened 2017-03-25
Made rust_src_path errors easier to understand.

Emilgardis:
See Valloric/YouCompleteMe#2586

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/729)

### DIFF
--- a/ycmd/completers/rust/rust_completer.py
+++ b/ycmd/completers/rust/rust_completer.py
@@ -56,11 +56,11 @@ BINARY_NOT_FOUND_MESSAGE = (
   'racerd binary not found. Did you build it? '
   'You can do so by running "./build.py --racer-completer".' )
 NON_EXISTING_RUST_SOURCES_PATH_MESSAGE = (
-  'Rust sources path does not exist. Check the value of the rust_src_path '
+  'Rust sources path does not exist. Check the value of the ycm_rust_src_path '
   'option or the RUST_SRC_PATH environment variable.' )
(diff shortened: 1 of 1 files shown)


Emilgardis:
Is this good or would it be good if ycm also had a fallback onto g:rust_src_path used by rust.vim?

micbou:
Thanks but we can't do that as [YCM is not the only client of ycmd](https://github.com/Valloric/ycmd#known-ycmd-clients). Ideally, we would need a mechanism that makes it easy for clients to rename ycmd options in strings to their counterparts (e.g. `rust_src_path` into `g:ycm_rust_src_path` for YCM, `ycmd-rust-src-path` for emacs-ycmd, etc). Honestly, I don't think it's worth the trouble.

> would it be good if ycm also had a fallback onto g:rust_src_path used by rust.vim?

I don't see a `g:rust_src_path` option in rust.vim (there is [a `g:ftplugin_rust_source_path` option](https://github.com/rust-lang/rust.vim/blob/732b5fcb3652f81726d5f0f5b97c9027c01f057a/doc/rust.txt#L84-L88) though).
 

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/729#-:-Kg6BFI8aZkSQo9jZvpp:bed9tam)*

Emilgardis:
Ok, closing this pr for now as the implementation would most likely require huge changes to be made correct.
Thank you for enlightening me :)

PR #742, opened 2017-04-20
cache last flags to use for header file

niuys:
Most of time I use 'GoToInclude' or 'GoToDefination' or other command to jump between header and cpp files, it's ok to use last cpp file flags for current header file. Although there is thread problem, it's very convenient to use only for one. This is just for proposal, maybe there is  other better method to deal this problem

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/742)


Valloric:
Thank you for the PR!

I'm not sure this is the best approach. We already do cache flags in the layers above the extra conf file and I don't think it's a good idea to just return the last computed flags since they could be for completely different files. Also, this has threading issues (as you mentioned).

So thanks again for your efforts, but I don't think we want to go down this path.

PR #743, opened 2017-04-20
improve cmake's lib finding: rename the clang lib only once

pattakosn:
if the expected clang lib names are not found, provide an error
message and exit.

As it was, if one run cmake twice, then configure would fail as
the rename operation would no longer find the original file name.
This fix first checks if the required libclang file name does not
exist. If it does not it looks for the libclang file as named in
the original clang distribution and renames it.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/743)


micbou:
Thanks for the PR but we took a different approach by creating a symlink in ycmd folder instead of renaming the clang library. See #744.

PR #744, opened 2017-04-20
[READY] Create a symlink instead of renaming libclang

micbou:
We currently rename `libclang.so` because, for some reason, the linker looks for `libclang.so.x` instead of `libclang.so.x.y` when loading `ycm_core.so` (though we are explicitly linking it to `libclang.so.x.y`). The issue is that we are renaming it in the Clang distribution while we should rename it inside ycmd folder. In addition, we should create a symlink instead because we don't know if the linker will still look for `libclang.so.x` in future versions (this wasn't the case for prior versions of Clang). Finally, we should extract the version from the library filename to stay compatible with previous versions of Clang.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/744)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -25,16 +25,12 @@ option( USE_SYSTEM_LIBCLANG "Set to ON to use the system libclang library" OFF )
 set( PATH_TO_LLVM_ROOT "" CACHE PATH "Path to the root of a LLVM+Clang binary distribution" )
 set( EXTERNAL_LIBCLANG_PATH "" CACHE PATH "Path to the libclang library to use" )
 
-set( CLANG_MAJOR_VERSION 4 )
-set( CLANG_MINOR_VERSION 0 )
-set( CLANG_PATCH_VERSION 0 )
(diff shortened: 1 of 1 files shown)


Valloric:
:lgtm:

Thanks for the PR!

Curious why codecov is saying coverage is going down. Seems like a bug on their end, possibly.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/744#-:-KiBLOtPs9zi1Bq0H4xH:bl7zf3d)*

micbou:
Probably because the Travis builds failed on OS X. I've updated the PR to make them pass.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/744#-:-KiBLhXvyV554NeDUJth:bcgwmn6)*

vheon:
:lgtm: @homu r=Valloric

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/744#-:-KiBqI4yQC2s8rSBbkSY:b9nem37)*

PR #746, opened 2017-04-21
[READY] Ignore identifiers from comments and strings by filetype

micbou:
When `collect_identifiers_from_comments_and_strings` is `0`, we are ignoring identifiers inside comments and strings from different languages. This is incorrect because we may ignore identifiers from comments that are not actual comments in the current working language. For instance, the `include` identifier will be ignored in the following C++ code:
```cpp
#include <iostream>
```
because we assume it's inside a Python-style comment.

The solution is to ignore identifiers from comments and strings according to filetype similarly to what we do to extract identifiers. For now, I've added support to the languages for which ycmd provides semantic completion. We'll add more with time.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/746)

### DIFF
--- a/ycmd/completers/all/identifier_completer.py
+++ b/ycmd/completers/all/identifier_completer.py
@@ -229,7 +229,7 @@ def _IdentifiersFromBuffer( text,
                             filetype,
                             collect_from_comments_and_strings ):
   if not collect_from_comments_and_strings:
-    text = identifier_utils.RemoveIdentifierFreeText( text )
+    text = identifier_utils.RemoveIdentifierFreeText( text, filetype )
   idents = identifier_utils.ExtractIdentifiersFromText( text, filetype )
   vector = ycm_core.StringVector()
   for ident in idents:
--- a/ycmd/identifier_utils.py
+++ b/ycmd/identifier_utils.py
@@ -26,30 +26,76 @@ from builtins import *  # noqa
 
 import re
 
-COMMENT_AND_STRING_REGEX = re.compile(
-  "//.*?$" # Anything following '//'
-  "|"
-  "#.*?$"  # Anything following '#'
-  "|"
-  "/\*(?:\n|.)*?\*/"  # C-style comments, '/* ... */'
-  "|"
-  # Python-style multi-line single-quote string
-  "'''(?:\n|.)*?'''"
-  "|"
-  # Python-style multi-line double-quote string
-  '"""(?:\n|.)*?"""'
-  "|"
-  # Anything inside single quotes, '...', but mind:
-  #  1. that the starting single quote is not escaped
-  #  2. the escaped slash (\\)
-  #  3. the escaped single quote inside the string
-  r"(?<!\\)'(?:\\\\|\\'|.)*?'"
-  "|"
-  # Anything inside double quotes, "...", but mind:
-  #  1. that the starting double quote is not escaped
-  #  2. the escaped slash (\\)
-  #  3. the escaped double quote inside the string
-  r'(?<!\\)"(?:\\\\|\\"|.)*?"', re.MULTILINE )
+C_STYLE_COMMENT = "/\*(?:\n|.)*?\*/"
+CPP_STYLE_COMMENT = "//.*?$"
+PYTHON_STYLE_COMMENT = "#.*?$"
+# Anything inside single quotes, '...', but mind:
+#  1. that the starting single quote is not escaped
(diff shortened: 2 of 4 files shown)


Valloric:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/identifier_utils.py, line 52 at r1](https://reviewable.io:443/reviews/valloric/ycmd/746#-KiMOrDhrJalrFWk5rjd:-KiMOrDhrJalrFWk5rje:b-llw5u9) ([raw file](https://github.com/valloric/ycmd/blob/f33e9c4c75cd85fa3e2641fb63eef801d5277a8a/ycmd/identifier_utils.py#L52)):*
> ```Python
> MULTILINE_DOUBLE_QUOTE_STRING = '"""(?:\n|.)*?"""'
> 
> DEFAULT_COMMENT_AND_STRING_REGEX = re.compile(
> ```

Maybe make the default be the old regex we used? It would be backwards compatible. The current approach would mean we pick up words from comments in say Bash (which uses Python-style pound comment signs) as identifiers.

It's not pretty, but it won't break anyone and it would still be superior to what we had previously.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/746)*

micbou:
Reviewed 2 of 4 files at r1, 2 of 2 files at r2, 1 of 1 files at r3.

---

*[ycmd/identifier_utils.py, line 52 at r1](https://reviewable.io:443/reviews/valloric/ycmd/746#-KiMOrDhrJalrFWk5rjd:-KiMokL9KFci6O3ZaGvi:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/f33e9c4c75cd85fa3e2641fb63eef801d5277a8a/ycmd/identifier_utils.py#L52)):*

Maybe make the default be the old regex we used? It would be backwards compatible. The current approach would mean we pick up words from comments in say Bash (which uses Python-style pound comment signs) as identifiers.

It's not pretty, but it won't break anyone and it would still be superior to what we had previously.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/746)*

Valloric:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/identifier_utils.py, line 58 at r4](https://reviewable.io:443/reviews/valloric/ycmd/746#-KiScOhjt1Yd14fY2M-_:-KiScOhjt1Yd14fY2M-a:b-9214r4) ([raw file](https://github.com/valloric/ycmd/blob/898899fadfd72e911644f188888ded9ccf21ed94/ycmd/identifier_utils.py#L58)):*
> ```Python
>   MULTILINE_SINGLE_QUOTE_STRING + "|" +
>   MULTILINE_DOUBLE_QUOTE_STRING + "|" +
>   SINGLE_QUOTE_STRING + "|" +
> ```

This seems like it would benefit from `join` as in `"|".join([a, b, c])`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/746)*

Valloric:
:lgtm: 

Minor inline comment, otherwise looking good, thanks!

---

Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/746#-:-KiScfzbEjiclhX8G-qG:bjm6suw)*

micbou:

---

*[ycmd/identifier_utils.py, line 58 at r4](https://reviewable.io:443/reviews/valloric/ycmd/746#-KiScOhjt1Yd14fY2M-_:-KiSwpKIH_OPeoPXFfPZ:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/898899fadfd72e911644f188888ded9ccf21ed94/ycmd/identifier_utils.py#L58)):*

This seems like it would benefit from `join` as in `"|".join([a, b, c])`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/746)*

vheon:
:lgtm: @homu r=Valloric

---

Reviewed 2 of 4 files at r1, 1 of 2 files at r2, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/746#-:-KiVjImNrSGu2fEbdOql:b9nem37)*

PR #747, opened 2017-04-24
build: Fix finding the python library

cryptomilk:
This fixes detection of the python3 libraries on openSUSE. cmake
normally correctly detects python but this custom code doesn't check the
default directories for libraries.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/747)

### DIFF
--- a/build.py
+++ b/build.py
@@ -153,7 +153,7 @@ def GetPossiblePythonLibraryDirectories():
   # On pyenv, there is no Python dynamic library in the directory returned by
   # the LIBPL variable. Such library is located in the parent folder of the
   # standard Python library modules.
-  return [ sysconfig.get_config_var( 'LIBPL' ), library_dir ]
+  return [ sysconfig.get_config_var( 'LIBPL' ), library_dir, '/usr/lib64', '/usr/lib']
 
 
 def FindPythonLibraries():


micbou:
What are the values returned by `sysconfig.get_python_lib( standard_lib = True )` and `sysconfig.get_config_var( 'LIBPL' )` on openSUSE?

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/747#-:-KiVQ-xbj1ucUgJLbckt:b5sz2mm)*

cryptomilk:
```
/usr/lib/python3.6
/usr/lib64/python3.6/config-3.6m-x86_64-linux-gnu
```

```
ll /usr/lib/python3.6/
total 0
drwxr-xr-x 1 root root 1736 Apr 21 14:25 site-packages

ll /usr/lib64/python3.6/config-3.6m-x86_64-linux-gnu
total 156
-rw-r--r-- 1 root root 68870 Mar 31 16:02 Makefile
-rw-r--r-- 1 root root 15296 Mar 31 16:02 Setup
-rw-r--r-- 1 root root   327 Mar 31 16:02 Setup.config
-rw-r--r-- 1 root root    41 Mar 31 16:02 Setup.local
-rw-r--r-- 1 root root  3302 Mar 31 16:02 config.c
-rw-r--r-- 1 root root  1621 Mar 31 16:02 config.c.in
-rw-r--r-- 1 root root  7122 Mar 31 16:02 install-sh
-rw-r--r-- 1 root root  7521 Mar 31 16:02 makesetup
-rw-r--r-- 1 root root  2036 Mar 31 16:02 python-config.py
-rw-r--r-- 1 root root 32928 Mar 31 16:02 python.o

rpm -ql python3-devel | grep ".so"
/usr/include/python3.6m/bytesobject.h
/usr/include/python3.6m/classobject.h
/usr/lib64/libpython3.6m.so
/usr/lib64/libpython3.so
```

cryptomilk:
Ping!

micbou:
Sorry for the last answer. Apparently, openSUSE patches the `get_python_lib` function to return `/usr/lib64` on Python 2 but not on Python 3:
```sh
$ python -c "from distutils import sysconfig; print(sysconfig.get_python_lib(standard_lib=True))"
/usr/lib64/python2.7
$ python3 -c "from distutils import sysconfig; print(sysconfig.get_python_lib(standard_lib=True))"
/usr/lib/python3.6
```
This is why the script fails to find the library on Python 3. However, we can't just add the `/usr/lib` and `/usr/lib64` to the list of paths to search because this may cause the script to compile the ycmd library against a different Python than the one used to run the script and we want to avoid that as this would be confusing. In addition, this will not work for a custom Python installed in `/usr/local`. I've sent PR #754 that implements a different approach.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/747#-:-KiyyjpZtjGAiJ1nUYoY:b-qxep3w)*

PR #750, opened 2017-05-04
[RFC][API] Only split lines on line feed characters (\n)

micbou:
[The `splitlines` method splits on line boundaries like `\v` and `\f` for unicode strings](https://docs.python.org/2/library/stdtypes.html#unicode.splitlines) but we only want it to split on newline characters (`\n`, `\r\n`, and `\r`). Since old Macintosh newlines (`\r`) are obsolete and Windows newlines (`\r\n`) end with a `\n` character, we can simply split on `\n`. Trailing carriage return characters (`\r`) should not be an issue.

This also documents that all lines should end with a line feed character (`\n`).

Fixes https://github.com/Valloric/YouCompleteMe/issues/2296.


Valloric:
Thanks for the PR!

I'm not sure we can ignore Mac (`\r`) newlines even if they are deprecated; since almost all dev tools support all three forms of newlines, I'd be surprised if there weren't some Mac tooling that still generates them or old files with `\r` newlines and nobody noticed.

I'm also wary of leaving behind a trailing `\r` on Windows files; it's one of those things that could bite us in the future.

I think we should fix `\v` and `\f` but leave the other behavior alone.

Happy to hear out differing opinions, as usual.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/750#-:-KjJQcrxvV2eYmEzQALa:b-4j5rr3)*

micbou:
The thing is that splitting lines simultaneously on all three line endings (`\n`, `\r\n`, and `\r`) is wrong because editors don't necessarily do that to represent lines. They may split lines on `\n`, `\r\n`, or `r` but not all at the same time. To give an example, if a file contains `line1\rline2\r\nline3\nline4` and is opened with Vim or Emacs on Linux, it will be shown to the user as:
```
line1^Mline2^M
line3
line4
```
but ycmd will see it as:
```
line1
line2
line3
line4
```
Ideally, we should ask the editor which line ending is used to represent lines and split them accordingly but it's simpler to split lines on `\n` and document it.

> I'm also wary of leaving behind a trailing \r on Windows files; it's one of those things that could bite us in the future.

We currently have the opposite issue: we remove trailing `\r` while they are present in the editor.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/750#-:-KjJe6DEcfkoFhKzaMLT:bfmoq43)*

Valloric:
Fair points, but my assumption is that the file is at least consistent in line endings: it's either all `\n`, `\r` or `\r\n`. We _could_ just state that we want all line endings to be `\n` and I wouldn't be entirely against that idea; we would state our API and it's up to the clients to follow it. It's a reasonable approach.

... but then you would leave this work of handling newline normalization up to every single client. That might be the right call, but I can easily envision problems with normalizing those line endings in YCM instead of in ycmd. We just shuffled the problem into the other codebase, and now every other client has to do the same work too.

So it's not really obvious to me which is better.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/750#-:-KjKbkcv7RPu4MjcTiiT:b-csq50)*

micbou:
> That might be the right call, but I can easily envision problems with normalizing those line endings in YCM instead of in ycmd. We just shuffled the problem into the other codebase, and now every other client has to do the same work too.

[We already do that in YCM](https://github.com/Valloric/YouCompleteMe/blob/e45a409ac99d2e2ff9643da494829a4aec313673/python/ycm/vimsupport.py#L131) but like you said, other clients may need to do it (or at least they should if they want to avoid such issue). The main reason to leave the work to clients is that they know how lines are split while ycmd has no idea unless additional information is given by the clients (e.g. the value of `fileformat` for Vim).

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/750#-:-KjKeQR1R16Jv2VSpyIA:bbuuvt6)*

Valloric:
Like I said, I'm not against the idea of demanding `\n` line endings. I'd be ok with it.

Can we do this from a perspective of backwards compatibility though? At the very least, we'd have to notify all the client authors (that we're aware of).

@puremourning @vheon Thoughts?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/750#-:-KjKwxzVRhYluVmPOdoh:b-3ggbjw)*

micbou:
> Can we do this from a perspective of backwards compatibility though?

Not sure how we can be backward compatible. With an option? A new field in the request?

>  At the very least, we'd have to notify all the client authors (that we're aware of).

Let's notify them now. They may have some input on the matter. Ping @abingham @ptrv @Qusic @LuckyGeck @mawww @richard1122 @jakeanq @orsonteodoro.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/750#-:-KjO8sYbIiZDYcjhfv7v:b-u180oa)*

bstaletic:
What about splitting on every `\r\n`, every `\r` and then every `\n`? That way clients won't have to normalise their line endings. This also doesn't introduce trailing `\r` characters, nor does it introduce extra empty lines in case of splitting on just `\n` and `\r`.

Something like this:
```py
return [ AllLines for WindowsLines in contents.split( '\r\n' )
                    for OldMacLines in WindowsLines.split( '\r' )
                      for AllLines in OldMacLines.split( '\n') ]
```

The order of the last two splits shouldn't matter, but it would be important that the `\r\n` split is done first.

In case we decide to ignore older macs:
```py
return [ lines for WindowsLines in contents.split( '\r\n' ) for AllLines in lines.split( '\n') ]
```

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/750#-:-KkqViUOPztUbFQrHRC0:b-z8ah6y)*

micbou:
As I mentioned above, splitting on all three newlines is incorrect because editors don't necessarily display lines to the user by simultaneously splitting them on `\r\n`, `\n`, and `\r`. If we want the same representation of the buffer lines in the editor and in ycmd, we need a rule on how a client must join the buffer lines before sending the content to the server. The server will then split the content according to this rule. I am suggesting that the rule should be to join/split lines on the `\n` character because, AFAIK, all modern editors split lines on at least this character.

Still, if we decide to split on all three newlines, we should go with your approach. I compared your code with a regex split (`re.split( '\r\n|\r|\n' )`) and yours is more efficient.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/750#-:-KkuKzN0CbU-vJ8WepXx:b-b676xv)*

bstaletic:
Right, having lines represented in ycmd that differs from what the user's editor is showing can lead to problems.

As for efficient python, even faster would be replacing all line ending with one uniform and then splitting on that line ending.

```py
return contents.replace( '\r\n', '\n' ).replace( '\r', '\n' ).split( '\n' ) 
```
Or if we decide to ignore `\r`:
```py
return contents.replace( '\r\n', '\n' ).split( '\n' )
```

These solutions are about twice as fast as my previous proposal.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/750#-:-KkuX0eN8dqpkJWVlUu7:b2m72rb)*

bstaletic:
Should we just blindly split on `\n`? Maybe it would be better to split on the "native" line ending, like `std::getline()` does.

---

Reviewed 5 of 5 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/750#-:-LCm7xPBCbOZzy49hikd:b31m760)*

micbou:
It's been a while but since none of the maintainers of the other clients raised their voice about this change, I think we should go for it. PR is ready.

> Maybe it would be better to split on the "native" line ending, like std::getline() does.

I just tested it and `std::getline` splits only on `\n` by default.

---

Reviewed 5 of 5 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/750#-:-KnpW-w1hNve_kUzKIQQ:bhuobat)*

bstaletic:
Spliting on native line endings could backfire more often that sticking to `\n`.

:lgtm: 

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/750#-:-LCnUXEG507BN1klX2mG:b-rn4npe)*

puremourning:
:lgtm:

---

Reviewed 3 of 5 files at r2, 2 of 2 files at r3.

---

*[README.md, line 72 at r3](https://reviewable.io/reviews/valloric/ycmd/750#-LCs31f9BI5cfgwAvr-B:-LCs31f9BI5cfgwAvr-C:b-ecm6z6) ([raw file](https://github.com/valloric/ycmd/blob/2d608337e5affd81a524f74d0d775bd096ed8abf/README.md#L72)):*
> ```Markdown
> 
> - All strings going into and out of the server are UTF-8 encoded.
> - All lines should end with `\n`.
> ```

don't say "should", it implies it's ok not to. say 'must' or don't say it at all: 'All lines end with \n'

---

*[docs/openapi.yml, line 108 at r3](https://reviewable.io/reviews/valloric/ycmd/750#-LCqqsi41fkDiix93ZIq:-LCs3Ei_EKJvc1KZHgTi:b-qxnycd) ([raw file](https://github.com/valloric/ycmd/blob/2d608337e5affd81a524f74d0d775bd096ed8abf/docs/openapi.yml#L108)):*

Shouldn't a breaking change bump the major number?
</blockquote></details>

we're still in v0 so we can (and sometimes do) change anything.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/750#-:-LCs9W5UA9lMh9wjxNHh:bnfp4nl)*

micbou:
Reviewed 3 of 3 files at r4.

---

*[README.md, line 72 at r3](https://reviewable.io/reviews/valloric/ycmd/750#-LCs31f9BI5cfgwAvr-B:-LCsBNAv1-6A9kT4ue4K:b-ku8h5h) ([raw file](https://github.com/valloric/ycmd/blob/2d608337e5affd81a524f74d0d775bd096ed8abf/README.md#L72)):*

don't say "should", it implies it's ok not to. say 'must' or don't say it at all: 'All lines end with \n'
</blockquote></details>

Right. I picked your second suggestion.

---

*[docs/openapi.yml, line 108 at r3](https://reviewable.io/reviews/valloric/ycmd/750#-LCqqsi41fkDiix93ZIq:-LCsBZmbDPl5W1LzVtVO:bypvrw9) ([raw file](https://github.com/valloric/ycmd/blob/2d608337e5affd81a524f74d0d775bd096ed8abf/docs/openapi.yml#L108)):*

we're still in v0 so we can (and sometimes do) change anything.
</blockquote></details>

What @puremourning said.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/750)*

puremourning:
Thanks. Let's do this @zzbot r+

---

Reviewed 3 of 3 files at r4.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/750#-:-LCsBsSP40HUaoNRtaN5:b-dz6alw)*

micbou:
@zzbot retry

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/750#-:-LCsSc0IDmHv6gMhE7vm:b-kc0oa4)*

PR #753, opened 2017-05-10
fix build error for godef

### LINKED ISSUE #752: Failed to build godef
I'm using go 1.8 on CentOS 7, and `./build.py --gocode-completer` failed with the following log:

```
[100%] Built target ycm_core
/root/src/github.com/Valloric/ycmd/third_party/godef/godef.go:18:2: local import "./go_local/ast" in non-local package
/root/src/github.com/Valloric/ycmd/third_party/godef/godef.go:19:2: local import "./go_local/parser" in non-local package
/root/src/github.com/Valloric/ycmd/third_party/godef/godef.go:20:2: local import "./go_local/printer" in non-local package
/root/src/github.com/Valloric/ycmd/third_party/godef/godef.go:21:2: local import "./go_local/token" in non-local package
/root/src/github.com/Valloric/ycmd/third_party/godef/godef.go:22:2: local import "./go_local/types" in non-local package
```

I encountered this on osx with go1.8 as well.

I think in newer versions of go, `go build` will not work with local import. I haven't tried with older go versions.
But fixing it should be very simple. In `build.py`, use `go build godef.go` instead of just `go build`.

I'll commit a PR for this soon, since there's only one line of code is changed...

inoc603:
fix #752

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/753)

### DIFF
--- a/build.py
+++ b/build.py
@@ -422,7 +422,7 @@ def BuildGoCode():
   os.chdir( p.join( DIR_OF_THIS_SCRIPT, 'third_party', 'gocode' ) )
   CheckCall( [ 'go', 'build' ] )
   os.chdir( p.join( DIR_OF_THIS_SCRIPT, 'third_party', 'godef' ) )
-  CheckCall( [ 'go', 'build' ] )
+  CheckCall( [ 'go', 'build', 'godef.go' ] )
 
 
 def BuildRacerd():


micbou:
Thanks for the PR. @Manishearth may have some comments on this as he is the maintainer of the version of Godef that we use. This could also be a good time to update the dependency. Anyway, the builds are passing so :lgtm: (don't pay attention to the coverage report; it's spurious).

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/753#-:-Kjlxt7drDslN9SmRTcV:b-scfn55)*

Manishearth:
Looks fine to me. I'll try updating the dep; but it's nontrivial since there's a bunch of path rewriting

PR #758, opened 2017-05-14
Return True from /ready request for filetype with no semenatic completer and subserver param

ptrv:
`ready` request should return `true` for a filetype without no semantic completer even if is used for the `subserver` parameter.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/758)

### DIFF
--- a/ycmd/handlers.py
+++ b/ycmd/handlers.py
@@ -159,8 +159,11 @@ def GetReady():
 
 
 def _IsSubserverReady( filetype ):
-  completer = _server_state.GetFiletypeCompleter( [filetype] )
-  return completer.ServerIsReady()
+  if _server_state.FiletypeCompletionAvailable( [ filetype ] ):
+    completer = _server_state.GetFiletypeCompleter( [ filetype ] )
+    return completer.ServerIsReady()
+  else:
+    return True
 
 
 @app.post( '/semantic_completion_available' )
--- a/ycmd/tests/misc_handlers_test.py
+++ b/ycmd/tests/misc_handlers_test.py
@@ -173,3 +173,17 @@ def MiscHandlers_DebugInfo_ExtraConfFoundButNotLoaded_test( app ):
       'completer': None
     } )
   )
+
+
+@SharedYcmd
+def MiscHandlers_Ready_WithFiletypeCompleter_test( app ):
+  with PatchCompleter( DummyCompleter, filetype = 'dummy_filetype' ):
+    assert_that( app.get( '/ready', { 'subserver': "dummy_filetype" } ).json,
+                 equal_to( True ) )
+
+
+@SharedYcmd
+def MiscHandlers_Ready_NoFiletypeCompleter_test( app ):
+  with PatchCompleter( DummyCompleter, filetype = 'dummy_filetype' ):
+    assert_that( app.get( '/ready', { 'subserver': "another_filetype" } ).json,
+                 equal_to( True ) )


vheon:
> ready request should return true for a filetype without no semantic completer even if is used for the subserver parameter.

Should we? I mean why return `True` if we don't have a sub-server as asked?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/758#-:-Kk7vqcELbFhTqhD0ZiZ:b-ue1tdz)*

puremourning:
lgtm with minor nit. 

---

---

*[ycmd/handlers.py, line 162 at r1](https://reviewable.io:443/reviews/valloric/ycmd/758#-Kk7vsHnxANyWI3KZp79:-Kk7vsHnxANyWI3KZp7A:bkov9ey) ([raw file](https://github.com/valloric/ycmd/blob/dbc11d35e5dfdf3a04aa0fe112b017deadf31118/ycmd/handlers.py#L162)):*
> ```Python
> 
> def _IsSubserverReady( filetype ):
>   if _server_state.FiletypeCompletionAvailable( [ filetype] ):
> ```

Space inside square bracket

---

*[ycmd/handlers.py, line 163 at r1](https://reviewable.io:443/reviews/valloric/ycmd/758#-Kk7vxWCaFO1SKgailr1:-Kk7vxWDNyF47WNyG8RY:bokrkoy) ([raw file](https://github.com/valloric/ycmd/blob/dbc11d35e5dfdf3a04aa0fe112b017deadf31118/ycmd/handlers.py#L163)):*
> ```Python
> def _IsSubserverReady( filetype ):
>   if _server_state.FiletypeCompletionAvailable( [ filetype] ):
>     completer = _server_state.GetFiletypeCompleter( [filetype] )
> ```

Here also, though I realise this isn't your code

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/758#-:-Kk7wDZkBJAfSfdhcPsa:b93a6j1)*

ptrv:
*[ycmd/handlers.py, line 162 at r1](https://reviewable.io:443/reviews/valloric/ycmd/758#-Kk7vsHnxANyWI3KZp79:-Kk7z4xvVQKgBxJWw9FL:bb6un1j) ([raw file](https://github.com/valloric/ycmd/blob/dbc11d35e5dfdf3a04aa0fe112b017deadf31118/ycmd/handlers.py#L162)):*

Space inside square bracket
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/758)*

ptrv:
*[ycmd/handlers.py, line 163 at r1](https://reviewable.io:443/reviews/valloric/ycmd/758#-Kk7vxWCaFO1SKgailr1:-Kk7zBd0crm6H9TpwlYL:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/dbc11d35e5dfdf3a04aa0fe112b017deadf31118/ycmd/handlers.py#L163)):*

Here also, though I realise this isn't your code
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/758)*

ptrv:
I think it's fine to return `True` because for a filetype without semantic completer, the server is ready to handle requests.  

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/758#-:-Kk7zQ4sSTPGOkqVW8cM:bbcra09)*

Valloric:
:lgtm:

Thanks for the PR!

@vheon It makes sense that we return true, otherwise clients have to maintain a list of semantic completers that ycmd supports. That shouldn't be necessary.

@zzbot r=puremourning

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/758#-:-Kk83WEQPqORD8xCtjUe:b26pgkk)*

micbou:
The issue is that this will even return `True` when ycmd is unable to load the semantic completer. This is why some tests are failing: they rely on the `/ready` request returning an error instead of `True`. I think it would be a mistake to return `True` or `False` if an error occur; returning the error is far more valuable.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/758#-:-KkCY3IVm0dooKPV5IfT:b-g4yuwt)*

micbou:
> otherwise clients have to maintain a list of semantic completers that ycmd supports

They should use the `semantic_completion_available` endpoint for this.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/758#-:-KkC_bf3HrsbgOwcNKIX:bj8t137)*

ptrv:
After giving it another thought I agree that returning always `True` is wrong especially that we loose the information about reason for not being ready. I was a bit confused by the api because there are tow properties that one can use: `include_subserver` for c-sharp and  also just `subserver` for a specified filetype. I was always thinking `include_server` should include a check for subserver _IF_ it exists. 

> They should use the semantic_completion_available endpoint for this

Yes, that is what I ended up with, sending a `/ready` request without subserver, then a `semantic_completion_available`, if true then another `/ready` request with subserver.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/758#-:-KkCdp1cQL7EANjY2KNW:b-liaa6y)*

ptrv:
I close this PR in favor of keeping the behavior.
Sorry for the noise, but at least we had a nice discussion about this part of the api :smile:

Valloric:
@ptrv Going with what @micbou has said is usually a good idea since he now probably knows more about ycmd & YCM internals than anyone else (myself definitely included). :)

PR #759, opened 2017-05-15
[READY][API] Drop include_subservers parameter in healthy and ready endpoints

micbou:
The `ready` endpoint currently accepts two parameters:
 - `subserver`: return `True` if the subserver for the given filetype is ready;
 - `include_subservers`: return `True` if the C# subserver is ready.

From this, it's obvious that we don't need the `include_subservers` parameter and that we should drop it in favor of `subserver`. We don't use this parameter anywhere in our code base and, AFAIK, no clients rely on it so we can safely remove it.

For consistency, we update the `healthy` endpoint by renaming its `include_subservers` parameter to `subserver` and extending it to any filetype. For the same reason as above, this shouldn't break anything.

Add tests for these two endpoints.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/759)

### DIFF
--- a/examples/example_client.py
+++ b/examples/example_client.py
@@ -102,10 +102,10 @@ class YcmdHandle( object ):
     return returncode is None
 
 
-  def IsReady( self, include_subservers = False ):
+  def IsReady( self, filetype = None ):
     if not self.IsAlive():
       return False
-    params = { 'include_subservers': 1 } if include_subservers else None
+    params = { 'subserver': filetype } if filetype else None
     response = self.GetFromHandler( 'ready', params )
     response.raise_for_status()
     return response.json()
@@ -191,7 +191,7 @@ class YcmdHandle( object ):
     self.PostToHandlerAndLog( EXTRA_CONF_HANDLER, request_json )
 
 
-  def WaitUntilReady( self, include_subservers = False ):
+  def WaitUntilReady( self, filetype = None ):
     total_slept = 0
     time.sleep( 0.5 )
     total_slept += 0.5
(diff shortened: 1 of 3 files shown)


micbou:
In fact, we are using the `include_subservers` parameter in the example client. We update it to use `subserver` instead.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/759#-:-KkD5QqA1ndDfZXHwiqO:bhlvxan)*

bstaletic:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/759#-:-KkDwEXVWB8BJ4uh3BLG:bnfp4nl)*

Valloric:
:lgtm: 

All good. :)

@zzbot r=bstaletic

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/759#-:-KkEaRXA1LR5L9YM81Au:b6c9qkf)*

micbou:
@zzbot retry

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/759#-:-KkEo1OSlqY9-kXUj65a:b-kc0oa4)*

micbou:
Tests are still flaky...

@zzbot retry

micbou:
@zzbot retry

vheon:
I believe that the `include_subservers` was never updated accordingly but its purpose was that when setted to `True` it would return if `ycmd` and *all* the subservers were ready so at this point if Omnisharp, JediHTTP, racerd, TSServer where all ready. But it doesn't make much sense since we would have to loop over them to see which one was actually started so I think we're good :+1:

PR #769, opened 2017-06-02
[READY] Add benchmark infrastructure

micbou:
This PR sets the infrastructure for adding benchmarks through [the Google benchmark library](https://github.com/google/benchmark) and for automatically running them on Travis and AppVeyor. They can also be run locally with the `benchmark.py` script. The library is included in the repository for compilation ease. Benchmarks are run on all platforms because optimizations may be platform-dependent.

For now, there is only one benchmark based on the output of *Program 2* in https://github.com/Valloric/YouCompleteMe/issues/2668. It measures the filter and sort algorithm on a worst-case scenario: all identifiers share a common prefix and the query is part of the prefix. In that case, no identifiers are filtered and since they all have the same weight, the algorithm falls back to lexicographic sorting. This scenario is not uncommon in practice. For instance, C libraries often use a common prefix for naming variables and functions to simulate namespaces.

Here's the output of the benchmark on my configuration:
```
------------------------------------------------------------------------------------------
Benchmark                                       Time           CPU Iterations
------------------------------------------------------------------------------------------
CandidatesWithCommonPrefix_bench/1           1955 ns       1898 ns     345165
CandidatesWithCommonPrefix_bench/2          11563 ns      11681 ns      64102
CandidatesWithCommonPrefix_bench/4          30761 ns      30594 ns      22436
CandidatesWithCommonPrefix_bench/8          69551 ns      69532 ns      11218
CandidatesWithCommonPrefix_bench/16        143963 ns     143924 ns       4986
CandidatesWithCommonPrefix_bench/32        292668 ns     290603 ns       2362
CandidatesWithCommonPrefix_bench/64        862766 ns     869571 ns        897
CandidatesWithCommonPrefix_bench/128      2205099 ns    2191318 ns        299
CandidatesWithCommonPrefix_bench/256      8895499 ns    8840057 ns         90
CandidatesWithCommonPrefix_bench/512     17704787 ns   17680113 ns         45
CandidatesWithCommonPrefix_bench/1024    45564517 ns   45760293 ns         15
CandidatesWithCommonPrefix_bench/2048    96960893 ns   98057771 ns          7
CandidatesWithCommonPrefix_bench/4096   217881085 ns  218401400 ns          3
CandidatesWithCommonPrefix_bench/8192   481444392 ns  483603100 ns          2
CandidatesWithCommonPrefix_bench/16384 1005462405 ns  982806300 ns          1
CandidatesWithCommonPrefix_bench/32768 1805209871 ns 1809611600 ns          1
CandidatesWithCommonPrefix_bench/65536 4215533125 ns 4212027000 ns          1
CandidatesWithCommonPrefix_bench_BigO     3979.06 NlgN    3974.50 NlgN
CandidatesWithCommonPrefix_bench_RMS           10 %          9 %
```
As you can see, performance becomes unacceptable starting from 16000 identifiers which is not a lot. A great feature of Google benchmark is that it can calculate the algorithm complexity. As expected, we have a `O(n log n)` complexity where `n` is the number of candidates (we are using `std::sort` to sort our candidates).

Thanks to this benchmark, I was able to improve the performance on this particular case by a factor of 60. I'll send the changes once this PR is merged.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/769)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,10 @@ install:
   - source ci/travis/travis_install.sh
 compiler:
   - gcc
-script: ./run_tests.py
+script:
+  - ci/travis/travis_script.sh
 after_success:
-  - bash <(curl -s https://codecov.io/bash)
+  - if [ "${COVERAGE}" == "true" ]; then bash <(curl -s https://codecov.io/bash); fi
 env:
   global:
     # Travis can run out of RAM, so we need to be careful here.
@@ -25,6 +26,7 @@ env:
     - USE_CLANG_COMPLETER=true YCMD_PYTHON_VERSION=2.6
     - USE_CLANG_COMPLETER=true YCMD_PYTHON_VERSION=2.7
     - USE_CLANG_COMPLETER=true YCMD_PYTHON_VERSION=3.3
+    - YCM_BENCHMARK=true YCMD_PYTHON_VERSION=3.3 COVERAGE=false
 matrix:
   exclude:
     - os: osx
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,12 +20,17 @@ environment:
   - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
     arch: 64
     python: 36
+  - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+    arch: 64
+    python: 36
+    YCM_BENCHMARK: true
+    COVERAGE: false
 install:
   - ci\appveyor\appveyor_install.bat
 build_script:
   - ci\appveyor\appveyor_build.bat
 after_build:
-  - codecov
+  - if %COVERAGE% == true ( codecov )
 # Disable automatic tests
 test: off
 cache:
--- a/benchmark.py
+++ b/benchmark.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+from __future__ import division
+from __future__ import unicode_literals
+from __future__ import absolute_import
+
+import os
+import os.path as p
+import subprocess
+import sys
+
+DIR_OF_THIS_SCRIPT = p.dirname( p.abspath( __file__ ) )
+DIR_OF_THIRD_PARTY = p.join( DIR_OF_THIS_SCRIPT, 'third_party' )
+
+sys.path.insert( 1, p.abspath( p.join( DIR_OF_THIRD_PARTY, 'argparse' ) ) )
+
+import argparse
+
+
+def ParseArguments():
+  parser = argparse.ArgumentParser()
+  parser.add_argument( '--msvc', type = int, choices = [ 12, 14, 15 ],
+                       default = 15, help = 'Choose the Microsoft Visual '
+                       'Studio version (default: %(default)s).' )
+
+  return parser.parse_known_args()
+
+
+def BuildYcmdLibsAndRunBenchmark( args, extra_args ):
+  build_cmd = [
+    sys.executable,
+    p.join( DIR_OF_THIS_SCRIPT, 'build.py' ),
+    '--clang-completer'
+  ] + extra_args
+
+  os.environ[ 'YCM_BENCHMARK' ] = '1'
+
+  if args.msvc:
+    build_cmd.extend( [ '--msvc', str( args.msvc ) ] )
+
+  subprocess.check_call( build_cmd )
+
+
+def Main():
+  args, extra_args = ParseArguments()
+  BuildYcmdLibsAndRunBenchmark( args, extra_args )
+
+
+if __name__ == "__main__":
+  Main()
--- a/build.py
+++ b/build.py
@@ -330,12 +330,27 @@ def RunYcmdTests( build_dir ):
   if OnWindows():
     # We prepend the folder of the ycm_core_tests executable to the PATH
     # instead of overwriting it so that the executable is able to find the
-    # python35.dll library.
+    # Python library.
     new_env[ 'PATH' ] = DIR_OF_THIS_SCRIPT + ';' + new_env[ 'PATH' ]
   else:
     new_env[ 'LD_LIBRARY_PATH' ] = DIR_OF_THIS_SCRIPT
 
   CheckCall( p.join( tests_dir, 'ycm_core_tests' ), env = new_env )
+
+
+def RunYcmdBenchmarks( build_dir ):
+  benchmarks_dir = p.join( build_dir, 'ycm', 'benchmarks' )
+  new_env = os.environ.copy()
+
+  if OnWindows():
+    # We prepend the folder of the ycm_core_tests executable to the PATH
+    # instead of overwriting it so that the executable is able to find the
+    # Python library.
+    new_env[ 'PATH' ] = DIR_OF_THIS_SCRIPT + ';' + new_env[ 'PATH' ]
+  else:
+    new_env[ 'LD_LIBRARY_PATH' ] = DIR_OF_THIS_SCRIPT
+
+  CheckCall( p.join( benchmarks_dir, 'ycm_core_benchmarks' ), env = new_env )
 
 
 # On Windows, if the ycmd library is in use while building it, a LNK1104
@@ -390,20 +405,27 @@ def BuildYcmdLib( args ):
 
     CheckCall( [ 'cmake' ] + full_cmake_args, exit_message = exit_message )
 
-    build_target = ( 'ycm_core' if 'YCM_TESTRUN' not in os.environ else
-                     'ycm_core_tests' )
+    build_targets = [ 'ycm_core' ]
+    if 'YCM_TESTRUN' in os.environ:
+      build_targets.append( 'ycm_core_tests' )
+    if 'YCM_BENCHMARK' in os.environ:
+      build_targets.append( 'ycm_core_benchmarks' )
 
-    build_command = [ 'cmake', '--build', '.', '--target', build_target ]
     if OnWindows():
       config = 'Debug' if args.enable_debug else 'Release'
-      build_command.extend( [ '--config', config ] )
+      build_config = [ '--config', config ]
     else:
-      build_command.extend( [ '--', '-j', str( NumCores() ) ] )
+      build_config = [ '--', '-j', str( NumCores() ) ]
 
-    CheckCall( build_command, exit_message = exit_message )
+    for target in build_targets:
+      build_command = ( [ 'cmake', '--build', '.', '--target', target ] +
+                        build_config )
+      CheckCall( build_command, exit_message = exit_message )
 
     if 'YCM_TESTRUN' in os.environ:
       RunYcmdTests( build_dir )
+    if 'YCM_BENCHMARK' in os.environ:
+      RunYcmdBenchmarks( build_dir )
   finally:
     os.chdir( DIR_OF_THIS_SCRIPT )
 
--- a/ci/appveyor/appveyor_build.bat
+++ b/ci/appveyor/appveyor_build.bat
@@ -7,4 +7,8 @@ if %msvc% == 2013 (
   set msvc=15
 )
 
-python run_tests.py --msvc %msvc%
+if defined YCM_BENCHMARK (
+  python benchmark.py --msvc %msvc%
+) else (
+  python run_tests.py --msvc %msvc%
+)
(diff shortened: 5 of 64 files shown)


bstaletic:
I have tested this last night and confirmed that it works as expected. I did leave a minor comment, but this is :lgtm: any way.

---

Reviewed 9 of 59 files at r1.
Review status: 9 of 59 files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/ycm/CMakeLists.txt, line 231 at r1](https://reviewable.io:443/reviews/valloric/ycmd/769#-KlbN4YKTSlua_D_jFX_:-KlbN4YKTSlua_D_jFXa:b-5s7gt6) ([raw file](https://github.com/valloric/ycmd/blob/db4c096b7483fc089705b5926b7619d2ad3e830c/cpp/ycm/CMakeLists.txt#L231)):*
> ```CMake
>                              benchmarks/*.h benchmarks/*.cpp
>                              CMakeFiles/*.cpp
>                              *client* )
> ```

`*client* )` could be joined with the line above. Or every glob in this `file()` statement can be on a different line.

No strong opinion, feel free to disregard this comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769#-:-KlbNXmRY94NwXRi6IYb:bcgv599)*

Valloric:
I've used Google's C++ benchmarking library extensively when I worked there; the API seems to have changed a bit, but it's largely the same thing. The library was great back then and I don't expect that to have changed.

@micbou Thanks for setting this up! I never really benchmarked the filter/sort part; it was plenty fast for all the use-cases I could come up with years ago. But if others have found cases where it's slow and we could make it faster, great!

The complexity of the algo being O(N*logN) is indeed expected. Back in 2011 when I was still thinking about starting YCM, I remember coming up with a proof that it's not possible to perform subsequence matching on a list of strings (the strings being known ahead of time, but not which subset of the strings or their order) faster than O(N*logN). That's just Big Oh, so plenty of constant factors to optimize.

:lgtm: 

---

Review status: 9 of 59 files reviewed at latest revision, 2 unresolved discussions.

---

*[cpp/ycm/benchmarks/IdentifierCompleter_bench.cpp, line 28 at r1](https://reviewable.io:443/reviews/valloric/ycmd/769#-KlejwHpLCy6dZ3cB-Vg:-KlejwHpLCy6dZ3cB-Vh:b-3js3k3) ([raw file](https://github.com/valloric/ycmd/blob/db4c096b7483fc089705b5926b7619d2ad3e830c/cpp/ycm/benchmarks/IdentifierCompleter_bench.cpp#L28)):*
> ```cpp
>     benchmark::State& state ) {
> 
>   // Generate a list of candidates of the form a_a_a_[a-z]{5}.
> ```

You might want to put `state.PauseTiming();` before this block and `state.ResumeTiming();` right after it. Otherwise the setup time will be included in your benchmark run. It will get spread over all the iterations, but still.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769#-:-KlekPGknkagTiYPHzmD:bbc2jai)*

micbou:
I added a method to `CandidateRepository` for clearing the candidates. This is needed to isolate the benchmarks (and the `CandidateReposity` tests). If we don't clear the candidates, measurements are influenced by previous runs.

Also, I changed the benchmark to use a query with a lowercase and an uppercase letter. This makes the benchmark more relevant because if we want to improve performance against it, we need a solution that deal with both kinds of case, not just lowercase.

---

Reviewed 59 of 59 files at r1, 1 of 1 files at r2, 3 of 3 files at r3, 1 of 1 files at r4, 3 of 3 files at r5.

---

*[cpp/ycm/CMakeLists.txt, line 231 at r1](https://reviewable.io:443/reviews/valloric/ycmd/769#-KlbN4YKTSlua_D_jFX_:-KliU02KribOn7MJYyl4:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/db4c096b7483fc089705b5926b7619d2ad3e830c/cpp/ycm/CMakeLists.txt#L231)):*

`*client* )` could be joined with the line above. Or every glob in this `file()` statement can be on a different line.

No strong opinion, feel free to disregard this comment.
</blockquote></details>

Done.

---

*[cpp/ycm/benchmarks/IdentifierCompleter_bench.cpp, line 28 at r1](https://reviewable.io:443/reviews/valloric/ycmd/769#-KlejwHpLCy6dZ3cB-Vg:-KliU57WigBl-S9d__QV:b7qbju0) ([raw file](https://github.com/valloric/ycmd/blob/db4c096b7483fc089705b5926b7619d2ad3e830c/cpp/ycm/benchmarks/IdentifierCompleter_bench.cpp#L28)):*

You might want to put `state.PauseTiming();` before this block and `state.ResumeTiming();` right after it. Otherwise the setup time will be included in your benchmark run. It will get spread over all the iterations, but still.
</blockquote></details>

Isn't the measurement only occurring in the `KeepRunning` loop?

---

*[cpp/ycm/benchmarks/IdentifierCompleter_bench.cpp, line 43 at r4](https://reviewable.io:443/reviews/valloric/ycmd/769#-KliVze-1qR-U8_VsaWL:-KliVze-1qR-U8_VsaWM:btgalvc) ([raw file](https://github.com/valloric/ycmd/blob/fa2bb0bdcbe07419c352697c3133729d9e256f36/cpp/ycm/benchmarks/IdentifierCompleter_bench.cpp#L43)):*
> ```cpp
>   }
> 
>   IdentifierCompleter completer( candidates );
> ```

Moved this out of the main loop because we don't want to measure the addition of candidates to the repository.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769#-:-KliXg0DeVXBQJSkWy9J:b-zgnrcy)*

vheon:
I know that is easier for us but are we sure we want to keep the `benchmarks` library in the source tree? I'm only saying this because people already complains about the size of the project when they have to download it and this only increase the burden :confused:  I understand that it would be more fragile but we could clone the library "on demand" on the first invocation of the `benchmark.py` script if the library was not already there.

---

Reviewed 6 of 59 files at r1, 1 of 1 files at r2, 1 of 3 files at r3, 1 of 1 files at r4, 1 of 3 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769#-:-KlmaOalwRPg642jvvX3:b-17b1o1)*

micbou:
The `benchmark` folder only takes 336 KB. That's negligible compared to the ~370 MB that ycmd takes after installation. In addition, like the `gmock` folder, we will never update it so it will only increase Git history once.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769#-:-KlmzINvRkvn89T2PGtL:bgyj235)*

Valloric:
@vheon Complaints about YCM on-disk size have always been silly to me. It's a few hundred MB, who cares. Disks are both huge and cheap.

---

---

*[cpp/ycm/benchmarks/IdentifierCompleter_bench.cpp, line 28 at r1](https://reviewable.io:443/reviews/valloric/ycmd/769#-KlejwHpLCy6dZ3cB-Vg:-Klns9eg2InzUipLxN3x:b57ko80) ([raw file](https://github.com/valloric/ycmd/blob/db4c096b7483fc089705b5926b7619d2ad3e830c/cpp/ycm/benchmarks/IdentifierCompleter_bench.cpp#L28)):*

Isn't the measurement only occurring in the `KeepRunning` loop?
</blockquote></details>

Why would it be? The API has changed since the last time I used it, but unless the first call to KeepRunning mutates some internal state so that measurement starts (quite possible), the whole benchmark time will be measured from the start of the function.

That's the way it worked before at least.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769#-:-Klnrnr-MN_zCwpFIaYe:b-f8delf)*

bstaletic:
I measured YCM/ycmd disk usage today. It's a bit over 200MB. You decide if that is big.

---

Reviewed 1 of 1 files at r2, 2 of 3 files at r3, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769#-:-Klo3AzMMOCA3YVP92H_:b-v6nz9j)*

Valloric:
@bstaletic It's not something we should worry about at all IMO. If it's less than 1GB, I don't care.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769#-:-KloAy7-MH-YbgWXUJtG:bqzmrey)*

puremourning:
Reviewed 1 of 59 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769)*

puremourning:
:lgtm:

---

Reviewed 56 of 59 files at r1, 1 of 1 files at r2, 1 of 3 files at r3, 1 of 1 files at r4, 3 of 3 files at r5.

---

*[benchmark.py, line 25 at r5](https://reviewable.io:443/reviews/valloric/ycmd/769#-KloOYiM0LbgWah6p-1e:-KloOYiM0LbgWah6p-1f:b-ac1j2a) ([raw file](https://github.com/valloric/ycmd/blob/aa1b12cd54467434de049ca274788804f7667622/benchmark.py#L25)):*
> ```Python
>   parser.add_argument( '--msvc', type = int, choices = [ 12, 14, 15 ],
>                        help = 'Choose the Microsoft Visual '
>                        'Studio version. (default: 15).' )
> ```

maybe i'm being thick, but what ensures the default is 15? it might be implicit, but i think i would prefer `default=15` if that's a thing.

---

*[cpp/ycm/benchmarks/IdentifierCompleter_bench.cpp, line 28 at r1](https://reviewable.io:443/reviews/valloric/ycmd/769#-KlejwHpLCy6dZ3cB-Vg:-KloS4m-Kys0J-sCBWQs:b-iwzze) ([raw file](https://github.com/valloric/ycmd/blob/db4c096b7483fc089705b5926b7619d2ad3e830c/cpp/ycm/benchmarks/IdentifierCompleter_bench.cpp#L28)):*

Why would it be? The API has changed since the last time I used it, but unless the first call to KeepRunning mutates some internal state so that measurement starts (quite possible), the whole benchmark time will be measured from the start of the function.

That's the way it worked before at least.
</blockquote></details>

As I recall from using it before, the first call to `KeepRunning` starts the timing for an iteration, and subsequent calls end/restart. Seems to be backed up here: https://github.com/google/benchmark/blob/93bfabc8b89270c00816c3cafb601475eaf85364/include/benchmark/benchmark_api.h#L334

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769#-:-KloSRvLw7B0vUqKretH:bnfp4nl)*

micbou:
Reviewed 2 of 3 files at r6, 1 of 3 files at r8.
Review status: 62 of 64 files reviewed at latest revision, 3 unresolved discussions.

---

*[benchmark.py, line 25 at r5](https://reviewable.io:443/reviews/valloric/ycmd/769#-KloOYiM0LbgWah6p-1e:-KloSfkJAedd9ugK-dZt:b-vif6da) ([raw file](https://github.com/valloric/ycmd/blob/aa1b12cd54467434de049ca274788804f7667622/benchmark.py#L25)):*

maybe i'm being thick, but what ensures the default is 15? it might be implicit, but i think i would prefer `default=15` if that's a thing.
</blockquote></details>

Copy-paste from the `run_tests.py` script. I've changed it in both scripts.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769)*

vheon:
:lgtm:

---

Reviewed 1 of 59 files at r1, 3 of 3 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769#-:-Klor8YFaoI360ONfE0x:bnfp4nl)*

micbou:
I've squashed the commits. PR is ready.

---

Reviewed 1 of 3 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769#-:-KlpC-5RqYziV6HwiRui:bx33bim)*

bstaletic:
@zzbot r+

---

Reviewed 1 of 3 files at r5, 2 of 3 files at r6, 2 of 3 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/769#-:-KlrkrlTeBecr7NYFQtW:bi59kwd)*

PR #772, opened 2017-06-07
[READY] Clang completer - optional completion chunk test

bstaletic:
This should cover the last non trivial and completely non covered function in C++ layer (and ycmd in general) that's not boost poilerplate code.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/772)

### DIFF
--- a/cpp/ycm/tests/ClangCompleter/ClangCompleter_test.cpp
+++ b/cpp/ycm/tests/ClangCompleter/ClangCompleter_test.cpp
@@ -37,14 +37,14 @@ TEST( ClangCompleterTest, CandidatesForLocationInFile ) {
   std::vector< CompletionData > completions_class =
     completer.CandidatesForLocationInFile(
       PathToTestFile( "basic.cpp" ).string(),
-      26,
+      29,
       7,
       std::vector< UnsavedFile >(),
       std::vector< std::string >() );
   std::vector< CompletionData > completions_struct =
     completer.CandidatesForLocationInFile(
       PathToTestFile( "basic.cpp" ).string(),
-      27,
+      30,
       7,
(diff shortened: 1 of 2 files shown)


Valloric:
:lgtm:

A PR with only tests? What's not to love? :) Thank you!

bstaletic:
This is `[READY]`. As an afterthought, [this line](https://github.com/Valloric/ycmd/blob/master/cpp/ycm/ClangCompleter/CompletionData.cpp#L120) in the function I was trying to cover is dead code, considering how we use the function right now. Should I remove it, or leave it "just in case"?

To save you the trouble of diving into the code, here's what's goinng on:

- We take a completion string in thte constructor.
- Count the chunks.
- For each chunk we call `ExtractDataFromChunk()`.
- Only if we know that the chunk is "optional" do we call `OptionalChunkToString()` on that chunk.

So to get to `OptionalChunkToString()`, `completion_string`can't be `NULL`, thus making line 120 unreachable.

Valloric:
@bstaletic Let's leave it just in case. The way we call the function now may not be the way we call it in the future.

bstaletic:
@Valloric It occured to me that "leave it, just in case" may be a good way to go here and that's why I asked first.

vheon:
:lgtm: @zzbot r=Valloric

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/772#-:-KmAkUIYarpPl_T0qIwV:b-c6qkzn)*

vheon:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/772)*

PR #455, opened 2016-04-09
[READY] Fix issues with multi-byte characters

puremourning:
## Summary

This change introduces more general support for non-ASCII characters in buffers handled by YCMD.

In ycmd's public API, all offsets are byte offsets into the UTF-8 encoded buffers. We also assume (because, we have no other choice) that files stored on disk are also UTF-8 encoded. Internally, almost all of ycmd's functionality operates on unicode strings (python 2 `unicode()` and python 3 `str()` objects, transparently via `future`). Many of the downstream completion engines expect unicode code points as the offsets in their APIs. One special case is the `ycm_core` library (identifier completer and clang completer), which requires instances of the _native_ `str` type. All strings used within the c++ using `boost::python` require passing through `ToCppStringCompatible`

Previously, we were largely just assuming that `code point == byte offset` - i.e. all buffers contained only ASCII characters. This worked up to a point, but more by luck than judgement in a number of places.

## References

In combination with a YCM change and PR https://github.com/Valloric/ycmd/pull/453, I hope this:

- fixes https://github.com/Valloric/ycmd/issues/109
- fixes https://github.com/Valloric/YouCompleteMe/issues/2096
- fixes https://github.com/Valloric/YouCompleteMe/issues/2088
- fixes https://github.com/Valloric/YouCompleteMe/issues/2069
- fixes https://github.com/Valloric/YouCompleteMe/issues/2066
- fixes https://github.com/Valloric/YouCompleteMe/issues/1378

## Overview of changes

The changes fall into the following areas:
 
- Providing access to and conversion to/from code points and byte offsets (`request_wrap.py`)
- Changing certain algorithms/features to work entirely in codepoint space when they are trying to operate on logical 'characters' within the buffer (see known issues for why this isn't perfect, but probably most of the way there)
- Changing the completers to convert between the external (on both sides) and internal representations by using the shortcuts provided in `request_wrap.py`
- Adding tests for each of the completers for both completions and subcommands

## Completer-specific notes

Pretty much all of the completers I tested required some changes:
- clang uses utf-8 and byte offsets, but had some bugs with the `GetDoc` parsing stuff
- OmniSharp speaks codepoint offsets
- Tern speaks codepoint offsets
- JediHTTP speaks codepoint offsets
- tsserver speaks codepoint offsets
- gocode speaks byte offsets
- racer i did not test

## Further work / Known issues

- we act blissfully ignorant of the case where a unicode character consumes multiple code points (such as where there is a modifier after the code point)
[...]

PR #774, opened 2017-06-08
[READY] Extend the GetUtf8String function

micbou:
This PR extends the `GetUtf8String` C++ function to support `unicode` strings and the `newstr`/`newbytes` types from python-future on Python 2. This allows us to remove the `deepcopy` and the string conversions (`ToCppStringCompatible` and `ToUnicode`) on the candidates when passing them to the `FilterAndSortCandidates` C++ function. This gives us a huge boost in performance as these operations are slow, especially the `deepcopy`:

filter-and-sort-candidates-profile

This graph was obtained by running the `prof.py` script in [this branch](https://github.com/micbou/ycmd/tree/profile-filter-and-sort-candidates) with the `--visualize` option. You'll need kcachegrind or qcachegrind and [the `pyprof2calltree` module](https://github.com/pwaller/pyprof2calltree) (`pip install pyprof2calltree`) to generate the graph. The script profiles the `FilterAndSortCandidatesWrap` function a number of times (10 by default) on the candidates returned by Clang when completing the `windows.h` header (46847 candidates). These candidates are filtered down to 12784 through the `G` query:
```c++
#include <windows.h>

int main() {
    G// <C-Space> here.
}
```
The script also returns the average execution time. Results on my configuration are ~734ms before the changes and ~143ms after. A significant improvement.

To be as efficient as possible, `GetUtf8String` is implemented differently depending on the Python version.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/774)


vheon:
Reviewed 4 of 4 files at r1.

---

*[cpp/ycm/PythonSupport.cpp, line 146 at r1](https://reviewable.io:443/reviews/valloric/ycmd/774#-KmAhiYl_r8fHbP9p1IF:-KmAhiYl_r8fHbP9p1IG:bnznq72) ([raw file](https://github.com/valloric/ycmd/blob/0642b44f11ef414f33da5bbd2b9bb80a60c7fdb0/cpp/ycm/PythonSupport.cpp#L146)):*
> ```cpp
> #endif
> 
>   return extract< std::string >( str( value ).encode( "utf8" ) );
> ```

We state that this function returns an empty string in case we can't convert the input string. Does this return an empty string if we don't enter any `if` in the Python2 section?

---

*[ycmd/completers/completer_utils.py, line 194 at r1](https://reviewable.io:443/reviews/valloric/ycmd/774#-KmAidDnCpDUmuGeE9Im:-KmAidDoVGQGAZR_4p5t:b-owxh5r) ([raw file](https://github.com/valloric/ycmd/blob/64ddce48d9a098e39540d60da529080fbbbe8a37/ycmd/completers/completer_utils.py#L194)):*
> ```Python
>   # important for those candidates which are *not* returned after the filter, as
>   # they are not converted back to unicode.
>   cpp_compatible_candidates = _ConvertCandidatesToCppCompatible(
> ```

Looks like the `_ConvertCandidatesToCppCompatible` function is only used here so we can remove the function as well.

---

*[ycmd/completers/completer_utils.py, line 207 at r1](https://reviewable.io:443/reviews/valloric/ycmd/774#-KmAiuXquLGUjEffYTJd:-KmAiuXquLGUjEffYTJe:bruje38) ([raw file](https://github.com/valloric/ycmd/blob/64ddce48d9a098e39540d60da529080fbbbe8a37/ycmd/completers/completer_utils.py#L207)):*
> ```Python
>     ToCppStringCompatible( query ) )
> 
>   return _ConvertCandidatesToPythonCompatible( filtered_candidates,
> ```

Looks like the `_ConvertCandidatesToPythonCompatible` function is only used here so we can remove it as well.

---

*[ycmd/completers/completer_utils.py, line 221 at r1](https://reviewable.io:443/reviews/valloric/ycmd/774#-KmAjCMECuNBUx1O1A7m:-KmAjCMECuNBUx1O1A7n:b-sbt5l2) ([raw file](https://github.com/valloric/ycmd/blob/64ddce48d9a098e39540d60da529080fbbbe8a37/ycmd/completers/completer_utils.py#L221)):*
> ```Python
> 
> 
> def _ConvertCandidates( candidates, sort_property, converter ):
> ```

If we remove the other two functions we can remove this as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/774)*

micbou:
Bumped the core version. If users don't recompile the library, they will encounter unicode issues that were fixed in https://github.com/Valloric/ycmd/pull/455.

---

Reviewed 3 of 4 files at r1, 2 of 2 files at r2.

---

*[cpp/ycm/PythonSupport.cpp, line 146 at r1](https://reviewable.io:443/reviews/valloric/ycmd/774#-KmAhiYl_r8fHbP9p1IF:-KmB76kJJpp-RnN9SYYd:bbcsbkl) ([raw file](https://github.com/valloric/ycmd/blob/0642b44f11ef414f33da5bbd2b9bb80a60c7fdb0/cpp/ycm/PythonSupport.cpp#L146)):*
I don't think that's true. According to [the docs](http://www.boost.org/doc/libs/1_64_0/libs/python/doc/html/tutorial/tutorial/object.html#tutorial.object.extracting_c_objects), `extract<T>`raises an exception if it can't extract the Python object:

> if the C++ type can't be extracted, an appropriate exception is thrown.

I'll remove the statement.

---

*[ycmd/completers/completer_utils.py, line 194 at r1](https://reviewable.io:443/reviews/valloric/ycmd/774#-KmAidDnCpDUmuGeE9Im:-KmB4ch6LNDrre20L_mU:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/64ddce48d9a098e39540d60da529080fbbbe8a37/ycmd/completers/completer_utils.py#L194)):*

Looks like the `_ConvertCandidatesToCppCompatible` function is only used here so we can remove the function as well.
</blockquote></details>

Done.

---

*[ycmd/completers/completer_utils.py, line 207 at r1](https://reviewable.io:443/reviews/valloric/ycmd/774#-KmAiuXquLGUjEffYTJd:-KmB4dBL2QCZi-UzC8IJ:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/64ddce48d9a098e39540d60da529080fbbbe8a37/ycmd/completers/completer_utils.py#L207)):*

Looks like the `_ConvertCandidatesToPythonCompatible` function is only used here so we can remove it as well.
</blockquote></details>

Done.

---

*[ycmd/completers/completer_utils.py, line 221 at r1](https://reviewable.io:443/reviews/valloric/ycmd/774#-KmAjCMECuNBUx1O1A7m:-KmB4dc38sFfToWWqxtK:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/64ddce48d9a098e39540d60da529080fbbbe8a37/ycmd/completers/completer_utils.py#L221)):*

If we remove the other two functions we can remove this as well.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/774#-:-KmB5g7MVq-wXEbe-9Ia:bb0uiuz)*

puremourning:
:lgtm: Very nice indeed.

---

Reviewed 2 of 4 files at r1, 2 of 2 files at r2, 1 of 1 files at r3.

---

*[cpp/ycm/PythonSupport.cpp, line 116 at r3](https://reviewable.io:443/reviews/valloric/ycmd/774#-KmIDUJafYxAQr5xBN5G:-KmIDUJafYxAQr5xBN5H:bxng9y3) ([raw file](https://github.com/valloric/ycmd/blob/2fac9be3aeeba899744306d89bcc995ba77cfd26/cpp/ycm/PythonSupport.cpp#L116)):*
> ```cpp
> 
> std::string GetUtf8String( const boost::python::object &value ) {
> #if PY_MAJOR_VERSION >= 3
> ```

i'm not sure what the style guide is for YCM in this regard, but personally, i would personally indent these #ifs as if they were part of the code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/774#-:-KmIDo8OLJTQOQHkTR01:b1iav7l)*

Valloric:
Fantastic work! Thanks a ton!

I think there's a solid chance this might blow up on some edge cases we haven't thought of (our test suite is pretty extensive around unicode these days, but it isn't perfect). So let's be ready for the bug reports once we merge this.

:lgtm: 

---

---

*[cpp/ycm/PythonSupport.cpp, line 116 at r3](https://reviewable.io:443/reviews/valloric/ycmd/774#-KmIDUJafYxAQr5xBN5G:-KmKabsB4luJ-4KKR8RB:bnunmmu) ([raw file](https://github.com/valloric/ycmd/blob/2fac9be3aeeba899744306d89bcc995ba77cfd26/cpp/ycm/PythonSupport.cpp#L116)):*

i'm not sure what the style guide is for YCM in this regard, but personally, i would personally indent these #ifs as if they were part of the code.
</blockquote></details>

I think the current style is to have the always in the zeroth column.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/774#-:-KmKbFdtDZrDc4KsnWWy:b4cj2wi)*

bstaletic:
Is this waiting for something? I see two :LGTM: but @valloric didn't trigger @zzbot. Any reason for that?

---

Reviewed 2 of 4 files at r1, 2 of 2 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/774#-:-Km_nCOZM3zvUErBesiS:bjra59n)*

Valloric:
@bstaletic no special reason
@zzbot r+

---

Review status: all files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/774#-:-KmcorRCoZI91jxkfDId:b-3zm8z6)*

bstaletic:
@zzbot retry

bstaletic:
Let's try this one more time.

@zzbot retry

PR #777, opened 2017-06-11
[READY] Optimize strings comparison when sorting candidates

micbou:
After filtering the candidates, we sort them by following a number of rules and if none of them apply, we fall back to a lexicographic comparison where lower case letters are prioritized over upper case ones, which is the same as comparing them with inverted case. This is done with `std::lexicographical_compare` using the `ChangeCharCase` comparison function, which is called through `std::function`. Letter case is inverted in `ChangeCharCase` using `std::isupper`, `std::tolower`, and `std::toupper` with the locale parameter. This is inefficient:
 - we switch the case using the locale. In addition to be slow, this is incorrect because our strings are supposed to be encoded in UTF-8 (technically, we only support ASCII). For instance, we don't want to convert the byte `\xE9` (`é`) to `\xC9` (`É`) if the user locale is latin1 since these bytes are not valid UTF-8 sequences;
 - we unnecessarily call `std::function` on `ChangeCharCase`;
 - we compare the strings character by character while we could directly compare them with `<`.

We optimize this by storing the case-swapped text (i.e. the text where the lower case letters are converted to upper case ones and vice versa) as a property of the `Candidate` object and compare both candidates on this property with the `<` operator when falling back to lexicographic sorting. We also implement our own utility functions to test and convert character and string properties. This results in a massive performance improvement for the `CandidatesWithCommonPrefix` benchmark.

Before:
```
--------------------------------------------------------------------------------------------
Benchmark                                                     Time           CPU Iterations
--------------------------------------------------------------------------------------------
CandidatesWithCommonPrefix/1                                805 ns        800 ns     897430
CandidatesWithCommonPrefix/2                              10145 ns       9735 ns      64102
CandidatesWithCommonPrefix/4                              28211 ns      27743 ns      23617
CandidatesWithCommonPrefix/8                              65058 ns      65360 ns      11218
CandidatesWithCommonPrefix/16                            139190 ns     137666 ns       4986
CandidatesWithCommonPrefix/32                            289616 ns     277394 ns       2362
CandidatesWithCommonPrefix/64                            855752 ns     855086 ns        748
CandidatesWithCommonPrefix/128                          2235209 ns    2235683 ns        314
CandidatesWithCommonPrefix/256                          5533323 ns    5447654 ns        126
CandidatesWithCommonPrefix/512                         24811977 ns   24600158 ns         52
CandidatesWithCommonPrefix/1024                        29717991 ns   29900192 ns         24
CandidatesWithCommonPrefix/2048                        66555287 ns   65867089 ns          9
CandidatesWithCommonPrefix/4096                       147582755 ns  148200950 ns          4
CandidatesWithCommonPrefix/8192                       324108104 ns  327602100 ns          2
CandidatesWithCommonPrefix/16384                      704005910 ns  702004500 ns          1
CandidatesWithCommonPrefix/32768                     1694146814 ns 1700410900 ns          1
CandidatesWithCommonPrefix/65536                     3731511967 ns 3712823800 ns          1
IdentifierCompleterFixture_BigO                         3515.82 NlgN  3504.03 NlgN
IdentifierCompleterFixture_RMS                                  8 %          8 %             
```

After:
```
--------------------------------------------------------------------------------------------
Benchmark                                                     Time           CPU Iterations
--------------------------------------------------------------------------------------------
CandidatesWithCommonPrefix/1                                659 ns        654 ns    1121788
CandidatesWithCommonPrefix/2                                960 ns        960 ns     747858
CandidatesWithCommonPrefix/4                               1571 ns       1568 ns     407923
CandidatesWithCommonPrefix/8                               2753 ns       2719 ns     263950
CandidatesWithCommonPrefix/16                              5035 ns       5006 ns     112179
CandidatesWithCommonPrefix/32                             11132 ns      10951 ns      64102
CandidatesWithCommonPrefix/64                             21244 ns      21416 ns      32051
CandidatesWithCommonPrefix/128                            52992 ns      53435 ns      15473
CandidatesWithCommonPrefix/256                           124426 ns     125151 ns       7479
CandidatesWithCommonPrefix/512                           200050 ns     200287 ns       2804
CandidatesWithCommonPrefix/1024                          434089 ns     431772 ns       1662
CandidatesWithCommonPrefix/2048                          960146 ns     973485 ns        641
CandidatesWithCommonPrefix/4096                         2220949 ns    2186930 ns        321
CandidatesWithCommonPrefix/8192                         5132477 ns    5014318 ns        112
CandidatesWithCommonPrefix/16384                       11111614 ns   11212572 ns         64
CandidatesWithCommonPrefix/32768                       24127657 ns   24514443 ns         28
CandidatesWithCommonPrefix/65536                       51418317 ns   51480330 ns         10
IdentifierCompleterFixture_BigO                           49.01 NlgN    49.19 NlgN
IdentifierCompleterFixture_RMS                                1 %          2 %             
```
Algorithm is ~72 times faster in this particular scenario. Not bad.

Added a bunch of tests for the character/string utility functions.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/777)


micbou:
I forgot to mention the increase in memory usage for adding the `_case_swapped_text` property. Let `n` be the number of candidates stored in memory. Since we only support text with at most 80 characters and [assuming an overhead of ~40 bytes from the `std::string` container](http://jovislab.com/blog/?p=76), we should expect an increase in memory usage of at most `120n` bytes. This gives us 120 MB for one million candidates. You probably can guess what happens to the ycmd process when storing as much candidates.

---

Reviewed 13 of 13 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/777#-:-KmO23NkV07dQC5-QjsI:bks8s2j)*

Valloric:
> Algorithm is ~72 times faster in this particular scenario. Not bad.

That's some good shit! :D

> We optimize this by storing the case-swapped text 

This is my only concern. As you pointed out in a separate comment, this will increase memory usage. We already have people complaining about ycmd RAM usage, but we've always traded off memory for speed and I'll take a 72x perf increase. RAM is cheap.

---

---

*[cpp/ycm/Utils.cpp, line 77 at r1](https://reviewable.io:443/reviews/valloric/ycmd/777#-KmSLkVhwgeTZVV22biJ:-KmSLkVhwgeTZVV22biK:b40ssy8) ([raw file](https://github.com/valloric/ycmd/blob/de0b54294deb9f1dfb004d405eb54c2d04a7e680/cpp/ycm/Utils.cpp#L77)):*
> ```cpp
> 
> bool IsPrintable( const std::string &text ) {
>   for ( char letter : text )
> ```

This should really have braces IMO.

---

*[cpp/ycm/Utils.cpp, line 100 at r1](https://reviewable.io:443/reviews/valloric/ycmd/777#-KmSZzL1CIbiPl3nU5Ge:-KmSZzL1CIbiPl3nU5Gf:brb8op1) ([raw file](https://github.com/valloric/ycmd/blob/de0b54294deb9f1dfb004d405eb54c2d04a7e680/cpp/ycm/Utils.cpp#L100)):*
> ```cpp
> // uppercase.
> bool IsLowercase( const std::string &text ) {
>   for ( char letter : text )
> ```

Braces

---

*[cpp/ycm/Utils.cpp, line 127 at r1](https://reviewable.io:443/reviews/valloric/ycmd/777#-KmS_0N88ssAUREiN2HT:-KmS_0N88ssAUREiN2HU:brb8op1) ([raw file](https://github.com/valloric/ycmd/blob/de0b54294deb9f1dfb004d405eb54c2d04a7e680/cpp/ycm/Utils.cpp#L127)):*
> ```cpp
> 
> bool HasUppercase( const std::string &text ) {
>   for ( char letter : text )
> ```

Braces

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/777#-:-KmSLKBLrL0NJp-daZ21:bmg731t)*

bstaletic:
What's up with the linux benchmark? It's not seeing nearly as big of a speed up.
```
Linux: IdentifierCompleterFixture/CandidatesWithCommonPrefix/65536 1313244016 - 1.3 s
OSX: IdentifierCompleterFixture/CandidatesWithCommonPrefix/65536  194837720 ns - ~0.2 s
Windows: IdentifierCompleterFixture/CandidatesWithCommonPrefix/65536  107713000 ns - ~0.1s
```

---

Reviewed 13 of 13 files at r1.
Review status: 12 of 13 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[cpp/ycm/Utils.cpp, line 62 at r1](https://reviewable.io:443/reviews/valloric/ycmd/777#-KmTGaGh3p1-O8dxskuP:-KmTGaGh3p1-O8dxskuQ:b-vtndop) ([raw file](https://github.com/valloric/ycmd/blob/de0b54294deb9f1dfb004d405eb54c2d04a7e680/cpp/ycm/Utils.cpp#L62)):*
> ```cpp
> 
> bool IsAscii( char letter ) {
>   return !( letter & 0x80 );
> ```

This may need a comment. It does check that the eight bit in a `char` is 0, which is a correct way to do this, but I had to compile the line in my head (hex to binary), so may cause confusion if we ever decide to refactor these.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/777#-:-KmTInAMobaJSgh4R0Tc:bl6sosv)*

micbou:
Yes, the Linux benchmark is way slower than the OS X one on Travis but this was already the case before these changes:
```
Linux: IdentifierCompleterFixture/CandidatesWithCommonPrefix/65536 8763373879 ns - 8.8s
OS X: IdentifierCompleterFixture/CandidatesWithCommonPrefix/65536 2591216883 ns - 2.6s
```
The benchmark measurements are also a bit unreliable on Travis.

> This is my only concern. As you pointed out in a separate comment, this will increase memory usage. We already have people complaining about ycmd RAM usage, but we've always traded off memory for speed and I'll take a 72x perf increase. RAM is cheap.

I measured the memory usage from the benchmark using [Massif](http://valgrind.org/docs/manual/ms-manual.html) and obtained the following graphs:
 - [massif.out.master](https://gist.github.com/micbou/b13173a21392496d5a5c2e79e0b27383);
 - [massif.out.case-swapped-text](https://gist.github.com/micbou/fa58b3c56af93f4c904cd1252a255ba9).

There is a ~0.68% increase in memory usage. This is negligible.

---

Reviewed 1 of 1 files at r2, 1 of 1 files at r3, 2 of 2 files at r4.

---

*[cpp/ycm/Utils.cpp, line 62 at r1](https://reviewable.io:443/reviews/valloric/ycmd/777#-KmTGaGh3p1-O8dxskuP:-KmVer_YKxwu1SVbtcY4:bs3mhdg) ([raw file](https://github.com/valloric/ycmd/blob/de0b54294deb9f1dfb004d405eb54c2d04a7e680/cpp/ycm/Utils.cpp#L62)):*

This may need a comment. It does check that the eight bit in a `char` is 0, which is a correct way to do this, but I had to compile the line in my head (hex to binary), so may cause confusion if we ever decide to refactor these.
</blockquote></details>

Comment added.

---

*[cpp/ycm/Utils.cpp, line 77 at r1](https://reviewable.io:443/reviews/valloric/ycmd/777#-KmSLkVhwgeTZVV22biJ:-KmTgCem6o9vDHIIG7mU:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/de0b54294deb9f1dfb004d405eb54c2d04a7e680/cpp/ycm/Utils.cpp#L77)):*

This should really have braces IMO.
</blockquote></details>

Done.

---

*[cpp/ycm/Utils.cpp, line 100 at r1](https://reviewable.io:443/reviews/valloric/ycmd/777#-KmSZzL1CIbiPl3nU5Ge:-KmTgD8dpW267hnnlTPd:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/de0b54294deb9f1dfb004d405eb54c2d04a7e680/cpp/ycm/Utils.cpp#L100)):*

Braces
</blockquote></details>

Done.

---

*[cpp/ycm/Utils.cpp, line 127 at r1](https://reviewable.io:443/reviews/valloric/ycmd/777#-KmS_0N88ssAUREiN2HT:-KmTgDb8hJSBc9403bkU:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/de0b54294deb9f1dfb004d405eb54c2d04a7e680/cpp/ycm/Utils.cpp#L127)):*

Braces
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/777#-:-KmVeUM5iNifxYfNpK6d:bf96xtw)*

bstaletic:
:lgtm:

---

Reviewed 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/777#-:-KmWB8asBxsjr_4fwAXq:bnfp4nl)*

Valloric:
Wow, didn't expect memory growth to be so small!

:lgtm: and @zzbot r=bstaletic

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/777#-:-KmXiNxgu-ZccZYSop7z:bjmw4fp)*

micbou:
@zzbot retry

PR #789, opened 2017-07-04
[READY] Enable windows style flags when --driver-mode=cl is found in flags

bstaletic:
Windows (cl) style flags are currently filtered and thus windows users are unable to use a `.ycm_extra_conf.py` such as this one:
```py
def FlagsForFile( filename ):
    return { 'flags': [
                '--driver-mode=cl',
                '/c',
                '/Zi',
                '/nologo',
                '/W3',
                '/WX-',
                '/O2',
                '/GL',
                '/Gm-',
                '/EHsc',
                '/MT',
                '/GS',
                '/fp:precise',
                '/Zc:wchar_t',
                '/Zc:forScope',
                '/Zc:inline',
                '/Gd',
                '/TP',
                '/wd4099',
                '/Tp' ] }
```

If `--driver-mode=cl` is in flags, libclang gladly accepts these flags. This pull request checks for `--driver-mode=cl` before filtering the flags. U used the above `.ycm_extra_conf.py` to confirm if everything works. The resulting `:YcmDebugInfo`:
```
Printing YouCompleteMe debug information...
-- Client logfile: /tmp/ycm_rbjk7yrj.log
-- Server Python interpreter: /usr/sbin/python
-- Server Python version: 3.6.1
-- Server has Clang support compiled in: True
-- Clang version: clang version 4.0.1 (tags/RELEASE_401/final)
-- Extra configuration file found and loaded
-- Extra configuration path: /home/bstaletic/Temp/ycmd/test/.ycm_extra_conf.py
-- C-family completer debug information:
--   Compilation database path: None
--   Flags: ['--driver-mode=cl', '-resource-dir=/home/bstaletic/.vim/pack/minpac/start/YouCompleteMe/third_party/ycmd/ycmd/../clang_includes', '/c', '/Zi', '/nologo', '/W3',
'/WX-', '/O2', '/GL', '/Gm-', '/EHsc', '/MT', '/GS', '/fp:precise', '/Zc:wchar_t', '/Zc:forScope', '/Zc:inline', '/Gd', '/TP', '/wd4099', '/Tp', '-fspell-checking']
-- Server running at: http://127.0.0.1:51101
-- Server process ID: 16027
-- Server logfiles:
--   /tmp/ycmd_51101_stdout_rh2xfnry.log
--   /tmp/ycmd_51101_stderr_7_n_feiy.log
```

Fixes Valloric/YouCompleteMe#2491

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/789)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -38,6 +38,7 @@ from ycmd.responses import NoExtraConfDetected
 INCLUDE_FLAGS = [ '-isystem', '-I', '-iquote', '-isysroot', '--sysroot',
                   '-gcc-toolchain', '-include-pch', '-include', '-iframework',
                   '-F', '-imacros', '-idirafter' ]
+INCLUDE_FLAGS_WIN_STYLE = [ '/I' ]
 PATH_FLAGS =  [ '--sysroot=' ] + INCLUDE_FLAGS
 
 # We need to remove --fcolor-diagnostics because it will cause shell escape
@@ -47,6 +48,8 @@ STATE_FLAGS_TO_SKIP = set( [ '-c',
                              '-MD',
                              '-MMD',
                              '--fcolor-diagnostics' ] )
+
+STATE_FLAGS_TO_SKIP_WIN_STYLE = set( [ '/c' ] )
 
 # The -M* flags spec:
 #   https://gcc.gnu.org/onlinedocs/gcc-4.9.0/gcc/Preprocessor-Options.html
@@ -144,7 +147,8 @@ class Flags( object ):
 
     sanitized_flags = PrepareFlagsForClang( flags,
                                             filename,
-                                            add_extra_clang_flags )
+                                            add_extra_clang_flags,
+                                            _ShouldAllowWinStyleFlags( flags ) )
 
     if results.get( 'do_cache', True ):
       self.flags_for_file[ filename ] = sanitized_flags
@@ -237,6 +241,16 @@ def _ExtractFlagsList( flags_for_file_output ):
   return [ ToUnicode( x ) for x in flags_for_file_output[ 'flags' ] ]
 
 
+def _ShouldAllowWinStyleFlags( flags ):
+  enable_windows_style_flags = False
+  if OnWindows():
+    for flag in flags:
+      if flag.startswith( '--driver-mode' ):
+        enable_windows_style_flags = ( flag == '--driver-mode=cl' )
+
+  return enable_windows_style_flags
+
+
 def _CallExtraConfFlagsForFile( module, filename, client_data ):
   # We want to ensure we pass a native py2 `str` on py2 and a native py3 `str`
   # (unicode) object on py3. That's the API we provide.
@@ -272,10 +286,13 @@ def _SysRootSpecifedIn( flags ):
   return False
 
 
-def PrepareFlagsForClang( flags, filename, add_extra_clang_flags = True ):
-  flags = _AddLanguageFlagWhenAppropriate( flags )
+def PrepareFlagsForClang( flags,
+                          filename,
+                          add_extra_clang_flags = True,
+                          enable_windows_style_flags = False ):
+  flags = _AddLanguageFlagWhenAppropriate( flags, enable_windows_style_flags )
   flags = _RemoveXclangFlags( flags )
-  flags = _RemoveUnusedFlags( flags, filename )
+  flags = _RemoveUnusedFlags( flags, filename, enable_windows_style_flags )
   if add_extra_clang_flags:
     flags = _EnableTypoCorrection( flags )
 
@@ -305,18 +322,21 @@ def _RemoveXclangFlags( flags ):
   return sanitized_flags
 
 
-def _RemoveFlagsPrecedingCompiler( flags ):
-  """Assuming that the flag just before the first flag (which starts with a
-  dash) is the compiler path, removes all flags preceding it."""
+def _RemoveFlagsPrecedingCompiler( flags, enable_windows_style_flags ):
+  """Assuming that the flag just before the first flag (looks like a flag,
+  not like a file path) is the compiler path, removes all flags preceding it."""
 
   for index, flag in enumerate( flags ):
-    if flag.startswith( '-' ):
+    if ( flag.startswith( '-' ) or
+         ( enable_windows_style_flags and
+           flag.startswith( '/' ) and
+           not os.path.exists( flag ) ) ):
       return ( flags[ index - 1: ] if index > 1 else
                flags )
   return flags[ :-1 ]
 
 
-def _AddLanguageFlagWhenAppropriate( flags ):
+def _AddLanguageFlagWhenAppropriate( flags, enable_windows_style_flags ):
   """When flags come from the compile_commands.json file, the flag preceding the
   first flag starting with a dash is usually the path to the compiler that
   should be invoked. Since LibClang does not deduce the language from the
@@ -325,18 +345,36 @@ def _AddLanguageFlagWhenAppropriate( flags ):
   the file extension. This handles the case where the .h extension is used for
   C++ headers."""
 
-  flags = _RemoveFlagsPrecedingCompiler( flags )
+  flags = _RemoveFlagsPrecedingCompiler( flags, enable_windows_style_flags )
 
-  # First flag is now the compiler path or a flag starting with a dash.
+  # First flag is now the compiler path, a flag starting with a dash or
+  # a flag starting with a forward slash if enable_windows_style_flags is True.
   first_flag = flags[ 0 ]
 
+  # NOTE: This is intentionally NOT checking for enable_windows_style_flags.
+  #
+  # Because of _RemoveFlagsPrecedingCompiler called above, irrelevant of
+  # enable_windows_style_flags. the first flag is either the compiler
+  # (path or executable), a Windows style flag or starts with a dash.
+  #
+  # If it doesn't start with a dash, it is either an absolute path,
+  # a Windows style flag or a C++ compiler executable from $PATH.
+  #   If it starts with a forward slash the flag can either be an absolute
+  #   flag or a Windows style flag.
+  #     If it matches the regex, it is safe to assume the flag is a compiler
+  #     path.
+  #     If it does not match the regex, it could still be a Windows style
+  #     path or an absolute path. - This is determined in _RemoveUnusedFlags()
+  #     and cleaned properly.
+  #   If the flag starts with anything else (i.e. not a '-' or a '/'), the flag
+  #   is a stray file path and shall be gotten rid of in _RemoveUnusedFlags().
   if ( not first_flag.startswith( '-' ) and
        CPP_COMPILER_REGEX.search( first_flag ) ):
     return [ first_flag, '-x', 'c++' ] + flags[ 1: ]
   return flags
 
 
-def _RemoveUnusedFlags( flags, filename ):
+def _RemoveUnusedFlags( flags, filename, enable_windows_style_flags ):
   """Given an iterable object that produces strings (flags for Clang), removes
   the '-c' and '-o' options that Clang does not like to see when it's producing
   completions for a file. Same for '-MD' etc.
@@ -354,20 +392,21 @@ def _RemoveUnusedFlags( flags, filename ):
     flags = flags[ 1: ]
 
   skip_next = False
-  previous_flag_is_include = False
-  previous_flag_starts_with_dash = False
-  current_flag_starts_with_dash = False
+  previous_flag = flags[ 0 ]
+  current_flag = flags[ 0 ]
 
   filename = os.path.realpath( filename )
   for flag in flags:
-    previous_flag_starts_with_dash = current_flag_starts_with_dash
-    current_flag_starts_with_dash = flag.startswith( '-' )
+    previous_flag = current_flag
+    current_flag = flag
 
     if skip_next:
       skip_next = False
       continue
 
-    if flag in STATE_FLAGS_TO_SKIP:
+    if ( flag in STATE_FLAGS_TO_SKIP or
+         ( enable_windows_style_flags and
+           flag in STATE_FLAGS_TO_SKIP_WIN_STYLE ) ):
       continue
 
     if flag in FILE_FLAGS_TO_SKIP:
@@ -383,14 +422,38 @@ def _RemoveUnusedFlags( flags, filename ):
     # "foo.cpp" when we are compiling "foo.h" because the comp db doesn't have
     # flags for headers. The returned flags include "foo.cpp" and we need to
     # remove that.
-    if ( not current_flag_starts_with_dash and
-          ( not previous_flag_starts_with_dash or
-            ( not previous_flag_is_include and '/' in flag ) ) ):
+    if _SkipStrayFilenameFlag( flag,
+                               current_flag,
+                               previous_flag,
+                               enable_windows_style_flags ):
       continue
 
     new_flags.append( flag )
-    previous_flag_is_include = flag in INCLUDE_FLAGS
+
   return new_flags
+
+
+def _SkipStrayFilenameFlag( flag,
+                            current_flag,
+                            previous_flag,
+                            enable_windows_style_flags ):
+  current_flag_starts_with_slash = current_flag.startswith( '/' )
+  previous_flag_starts_with_slash = previous_flag.startswith( '/' )
+
+  current_flag_starts_with_dash = current_flag.startswith( '-' )
+  previous_flag_starts_with_dash = previous_flag.startswith( '-' )
+
+  previous_flag_is_include = ( previous_flag in INCLUDE_FLAGS or
+                               ( enable_windows_style_flags and
+                                 flag in INCLUDE_FLAGS_WIN_STYLE ) )
+
+  return ( not ( current_flag_starts_with_dash or
+                 ( enable_windows_style_flags and
+                   current_flag_starts_with_slash ) ) and
+           ( not ( previous_flag_starts_with_dash or
+                   ( enable_windows_style_flags and
+                     previous_flag_starts_with_slash ) ) or
+             ( not previous_flag_is_include and '/' in flag ) ) )
 
 
 # Return the path to the macOS toolchain root directory to use for system
@@ -532,6 +595,9 @@ def _MakeRelativePathsInFlagsAbsolute( flags, working_directory ):
     return list( flags )
   new_flags = []
   make_next_absolute = False
+  path_flags = ( PATH_FLAGS + INCLUDE_FLAGS_WIN_STYLE
+                 if _ShouldAllowWinStyleFlags( flags )
+                 else PATH_FLAGS )
   for flag in flags:
     new_flag = flag
 
@@ -541,7 +607,7 @@ def _MakeRelativePathsInFlagsAbsolute( flags, working_directory ):
         new_flag = os.path.join( working_directory, flag )
       new_flag = os.path.normpath( new_flag )
     else:
-      for path_flag in PATH_FLAGS:
+      for path_flag in path_flags:
         # Single dash argument alone, e.g. -isysroot <path>
         if flag == path_flag:
           make_next_absolute = True
@@ -592,10 +658,11 @@ def _GetCompilationInfoForFile( database, file_name, file_extension ):
 def UserIncludePaths( flags, filename ):
   quoted_include_paths = [ os.path.dirname( filename ) ]
   include_paths = []
-
   if flags:
     quote_flag = '-iquote'
-    path_flags = [ '-isystem', '-I' ]
+    path_flags = ( [ '-isystem', '-I', '/I' ]
+                   if _ShouldAllowWinStyleFlags( flags )
+                   else [ '-isystem', '-I' ] )
 
     try:
       it = iter( flags )
--- a/ycmd/tests/clang/flags_test.py
+++ b/ycmd/tests/clang/flags_test.py
@@ -223,7 +223,7 @@ def FlagsForFile_DoNotAddMacIncludePathsWithSysroot_test():
 
 def RemoveUnusedFlags_Passthrough_test():
   eq_( [ '-foo', '-bar' ],
-       flags._RemoveUnusedFlags( [ '-foo', '-bar' ], 'file' ) )
+       flags._RemoveUnusedFlags( [ '-foo', '-bar' ], 'file', False ) )
 
 
 def RemoveUnusedFlags_RemoveDashC_test():
@@ -232,14 +232,14 @@ def RemoveUnusedFlags_RemoveDashC_test():
   filename = 'file'
 
   eq_( expected,
-       flags._RemoveUnusedFlags( expected + to_remove, filename ) )
+       flags._RemoveUnusedFlags( expected + to_remove, filename, False ) )
 
   eq_( expected,
-       flags._RemoveUnusedFlags( to_remove + expected, filename ) )
+       flags._RemoveUnusedFlags( to_remove + expected, filename, False ) )
 
   eq_( expected,
        flags._RemoveUnusedFlags(
-         expected[ :1 ] + to_remove + expected[ -1: ], filename ) )
+         expected[ :1 ] + to_remove + expected[ -1: ], filename, False ) )
 
 
 def RemoveUnusedFlags_RemoveColor_test():
@@ -248,14 +248,14 @@ def RemoveUnusedFlags_RemoveColor_test():
   filename = 'file'
 
   eq_( expected,
-       flags._RemoveUnusedFlags( expected + to_remove, filename ) )
+       flags._RemoveUnusedFlags( expected + to_remove, filename, False ) )
 
   eq_( expected,
-       flags._RemoveUnusedFlags( to_remove + expected, filename ) )
+       flags._RemoveUnusedFlags( to_remove + expected, filename, False ) )
 
   eq_( expected,
        flags._RemoveUnusedFlags(
-         expected[ :1 ] + to_remove + expected[ -1: ], filename ) )
+         expected[ :1 ] + to_remove + expected[ -1: ], filename, False ) )
 
 
 def RemoveUnusedFlags_RemoveDashO_test():
@@ -264,14 +264,14 @@ def RemoveUnusedFlags_RemoveDashO_test():
   filename = 'file'
 
   eq_( expected,
-       flags._RemoveUnusedFlags( expected + to_remove, filename ) )
+       flags._RemoveUnusedFlags( expected + to_remove, filename, False ) )
 
   eq_( expected,
-       flags._RemoveUnusedFlags( to_remove + expected, filename ) )
+       flags._RemoveUnusedFlags( to_remove + expected, filename, False ) )
 
   eq_( expected,
        flags._RemoveUnusedFlags(
-         expected[ :1 ] + to_remove + expected[ -1: ], filename ) )
+         expected[ :1 ] + to_remove + expected[ -1: ], filename, False ) )
 
 
 def RemoveUnusedFlags_RemoveMP_test():
@@ -280,14 +280,14 @@ def RemoveUnusedFlags_RemoveMP_test():
   filename = 'file'
 
   eq_( expected,
-       flags._RemoveUnusedFlags( expected + to_remove, filename ) )
+       flags._RemoveUnusedFlags( expected + to_remove, filename, False ) )
 
   eq_( expected,
-       flags._RemoveUnusedFlags( to_remove + expected, filename ) )
+       flags._RemoveUnusedFlags( to_remove + expected, filename, False ) )
 
   eq_( expected,
        flags._RemoveUnusedFlags(
-         expected[ :1 ] + to_remove + expected[ -1: ], filename ) )
+         expected[ :1 ] + to_remove + expected[ -1: ], filename, False ) )
 
 
 def RemoveUnusedFlags_RemoveFilename_test():
@@ -296,15 +296,15 @@ def RemoveUnusedFlags_RemoveFilename_test():
   filename = 'file'
 
   eq_( expected,
-       flags._RemoveUnusedFlags( expected + to_remove, filename ) )
+       flags._RemoveUnusedFlags( expected + to_remove, filename, False ) )
 
   eq_( expected,
        flags._RemoveUnusedFlags( expected[ :1 ] + to_remove + expected[ 1: ],
-                                 filename ) )
+                                 filename, False ) )
 
   eq_( expected,
        flags._RemoveUnusedFlags(
-         expected[ :1 ] + to_remove + expected[ -1: ], filename ) )
+         expected[ :1 ] + to_remove + expected[ -1: ], filename, False ) )
 
 
 def RemoveUnusedFlags_RemoveFlagWithoutPrecedingDashFlag_test():
@@ -313,11 +313,50 @@ def RemoveUnusedFlags_RemoveFlagWithoutPrecedingDashFlag_test():
   filename = 'file'
 
   eq_( expected,
-       flags._RemoveUnusedFlags( expected + to_remove, filename ) )
+       flags._RemoveUnusedFlags( expected + to_remove, filename, False ) )
 
   eq_( expected,
        flags._RemoveUnusedFlags( expected[ :1 ] + to_remove + expected[ 1: ],
-                                 filename ) )
+                                 filename, False ) )
+  expected = [ 'g++', '-foo', '--driver-mode=cl', '-xc++', '-bar',
+               'include_dir', '/I', 'include_dir_other' ]
+  to_remove = [ '..' ]
+  filename = 'file'
+
+  eq_( expected,
+       flags._RemoveUnusedFlags( expected + to_remove, filename, True ) )
+
+  eq_( expected,
+       flags._RemoveUnusedFlags( expected[ :1 ] + to_remove + expected[ 1: ],
+                                 filename, True ) )
+
+
+def RemoveUnusedFlags_MultipleDriverModeFlags_test():
+  expected = [ 'g++',
+               '--driver-mode=cl',
+               '/Zi',
+               '-foo',
+               '--driver-mode=gcc',
+               '--driver-mode=cl',
+               'include_dir' ]
+  to_remove = [ 'unrelated_file', '/c' ]
+  filename = 'file'
+
+  eq_( expected,
+       flags._RemoveUnusedFlags( expected + to_remove, filename, True ) )
+  eq_( expected,
+       flags._RemoveUnusedFlags( expected[ :1 ] + to_remove + expected[ 1: ],
+                                 filename, True ) )
+
+  flags_expected = [ '/usr/bin/g++', '--driver-mode=cl', '--driver-mode=gcc' ]
+  flags_all = [ '/usr/bin/g++',
+                '/Zi',
+                '--driver-mode=cl',
+                '/foo',
+                '--driver-mode=gcc' ]
+  filename = 'file'
+
+  eq_( flags_expected, flags._RemoveUnusedFlags( flags_all, filename, False ) )
 
 
 def RemoveUnusedFlags_Depfiles_test():
@@ -337,7 +376,7 @@ def RemoveUnusedFlags_Depfiles_test():
     '-arch', 'armv7',
   ]
 
-  assert_that( flags._RemoveUnusedFlags( full_flags, 'test.m' ),
+  assert_that( flags._RemoveUnusedFlags( full_flags, 'test.m', False ),
                contains( *expected ) )
 
 
@@ -365,15 +404,15 @@ def RemoveUnusedFlags_RemoveFilenameWithoutPrecedingInclude_test():
     expected = [ 'clang', flag, '/foo/bar', '-isystem/zoo/goo' ]
 
     eq_( expected,
-         flags._RemoveUnusedFlags( expected + to_remove, filename ) )
+         flags._RemoveUnusedFlags( expected + to_remove, filename, False ) )
 
     eq_( expected,
          flags._RemoveUnusedFlags( expected[ :1 ] + to_remove + expected[ 1: ],
-                                   filename ) )
+                                   filename, False ) )
 
     eq_( expected + expected[ 1: ],
          flags._RemoveUnusedFlags( expected + to_remove + expected[ 1: ],
-                                   filename ) )
+                                   filename, False ) )
 
   include_flags = [ '-isystem', '-I', '-iquote', '-isysroot', '--sysroot',
                     '-gcc-toolchain', '-include', '-include-pch',
@@ -402,7 +441,11 @@ def RemoveXclangFlags_test():
 
 def AddLanguageFlagWhenAppropriate_Passthrough_test():
   eq_( [ '-foo', '-bar' ],
-       flags._AddLanguageFlagWhenAppropriate( [ '-foo', '-bar' ] ) )
+       flags._AddLanguageFlagWhenAppropriate( [ '-foo', '-bar' ], False ) )
+  eq_( [ '-foo', '-bar', '--driver-mode=cl' ],
+       flags._AddLanguageFlagWhenAppropriate( [ '-foo',
+                                                '-bar',
+                                                '--driver-mode=cl' ], True ) )
 
 
 def _AddLanguageFlagWhenAppropriateTester( compiler, language_flag = [] ):
@@ -416,7 +459,7 @@ def _AddLanguageFlagWhenAppropriateTester( compiler, language_flag = [] ):
   for to_remove in to_removes:
     eq_( [ compiler ] + language_flag + expected,
          flags._AddLanguageFlagWhenAppropriate( to_remove + [ compiler ] +
-                                                expected ) )
+                                                expected, False ) )
 
 
 def AddLanguageFlagWhenAppropriate_CCompiler_test():
--- a/ycmd/tests/clang/get_completions_test.py
+++ b/ycmd/tests/clang/get_completions_test.py
@@ -35,7 +35,7 @@ from ycmd.completers.cpp.clang_completer import NO_COMPLETIONS_MESSAGE
 from ycmd.responses import UnknownExtraConf, NoExtraConfDetected
 from ycmd.tests.clang import IsolatedYcmd, PathToTestFile, SharedYcmd
 from ycmd.tests.test_utils import ( BuildRequest, CompletionEntryMatcher,
-                                    ErrorMatcher, ExpectedFailure )
+                                    ErrorMatcher, ExpectedFailure, WindowsOnly )
 from ycmd.utils import ReadFile
 
 NO_COMPLETIONS_ERROR = ErrorMatcher( RuntimeError, NO_COMPLETIONS_MESSAGE )
@@ -506,6 +506,58 @@ def GetCompletions_Include_ClientDataGivenToExtraConf_test( app ):
     has_item( CompletionEntryMatcher( 'include.hpp',
               extra_menu_info = '[File]' ) )
   )
+
+
+@SharedYcmd
+@WindowsOnly
+def GetCompletions_ClangCLDriver_SimpleCompletion_test( app ):
+  RunTest( app, {
+    'description': 'basic completion with --driver-mode=cl',
+    'extra_conf': [ 'driver_mode_cl', '.ycm_extra_conf.py' ],
+    'request': {
+      'filetype': 'cpp',
+      'filepath': PathToTestFile( 'driver_mode_cl', 'driver_mode_cl.cpp' ),
+      'line_num': 8,
+      'column_num': 18,
+      'force_semantic': True,
+    },
+    'expect': {
+      'response': requests.codes.ok,
+      'data': has_entries( {
+        'completion_start_column': 3,
+        'completions': contains_inanyorder(
+          CompletionEntryMatcher( 'driver_mode_cl_include_func', 'void' ),
+          CompletionEntryMatcher( 'driver_mode_cl_include_int', 'int' ),
+        ),
+        'errors': empty(),
+      } )
+    }
+  } )
+
+
+@SharedYcmd
+@WindowsOnly
+def GetCompletions_ClangCLDriver_IncludeStatementCandidate_test( app ):
+  RunTest( app, {
+    'description': 'Completion inside include statement with CL driver',
+    'extra_conf': [ 'driver_mode_cl', '.ycm_extra_conf.py' ],
+    'request': {
+      'filetype': 'cpp',
+      'filepath': PathToTestFile( 'driver_mode_cl', 'driver_mode_cl.cpp' ),
+      'line_num': 1,
+      'column_num': 34,
+    },
+    'expect': {
+      'response': requests.codes.ok,
+      'data': has_entries( {
+        'completion_start_column': 11,
+        'completions': contains_inanyorder(
+          CompletionEntryMatcher( 'driver_mode_cl_include.h', '[File]' ),
+        ),
+        'errors': empty(),
+      } )
+    }
+  } )
 
 
 @ExpectedFailure( 'Filtering and sorting does not support candidates with '
--- a/ycmd/tests/clang/testdata/driver_mode_cl/.ycm_extra_conf.py
+++ b/ycmd/tests/clang/testdata/driver_mode_cl/.ycm_extra_conf.py
@@ -1,0 +1,7 @@
+import os
+
+DIR_OF_THIS_SCRIPT = os.path.dirname( os.path.abspath( __file__ ) )
+
+def FlagsForFile( filename ):
+  return { 'flags': [ '-xc++', '/c', '--driver-mode=cl', '/I', 'driver_mode_cl_include' ],
+           'include_paths_relative_to_dir': DIR_OF_THIS_SCRIPT }
--- a/ycmd/tests/clang/testdata/driver_mode_cl/driver_mode_cl.cpp
+++ b/ycmd/tests/clang/testdata/driver_mode_cl/driver_mode_cl.cpp
@@ -1,0 +1,10 @@
+#include "driver_mode_cl_include.h"
+
+int main(void)
+{
+  // Position after the last underscore:
+  // line = 8
+  // column = 18
+  driver_mode_cl_
+  return 0;
+}
--- a/ycmd/tests/clang/testdata/driver_mode_cl/driver_mode_cl_include/driver_mode_cl_include.h
+++ b/ycmd/tests/clang/testdata/driver_mode_cl/driver_mode_cl_include/driver_mode_cl_include.h
@@ -1,0 +1,7 @@
+#ifndef DRIVER_MODE_CL_INCLUDE_H
+#define DRIVER_MODE_CL_INCLUDE_H
+
+void driver_mode_cl_include_func(void);
+int driver_mode_cl_include_int;
+
+#endif /* DRIVER_MODE_CL_INCLUDE_H */


micbou:
Thanks for the PR.

---

Reviewed 2 of 2 files at r1.

---

*[ycmd/completers/cpp/flags.py, line 406 at r1](https://reviewable.io:443/reviews/valloric/ycmd/789#-KoDY4KOKjEYq66Gh2CY:-KoDY4KOKjEYq66Gh2CZ:b-e1ljk0) ([raw file](https://github.com/valloric/ycmd/blob/3e61b03244a4d70501cfeef9323629901712d34e/ycmd/completers/cpp/flags.py#L406)):*
> ```Python
> 
>     if flag == '--driver-mode=cl':
>       enable_windows_style_flags = True
> ```

This variable should be set before looping through the flags, otherwise, Windows flags defined before `--driver-mode=cl` will be ignored. We could simply set  `enable_windows_style_flags` like this:
```python
enable_windows_style_flags = '--driver-mode=cl' in flags
```
or if we want to be more correct, we should find the last occurrence of `--driver-mode=` in the list of flags and check if its value is `cl` or not. Probably not worth the trouble.

---

*[ycmd/completers/cpp/flags.py, line 433 at r1](https://reviewable.io:443/reviews/valloric/ycmd/789#-KoDaK49rqW8R1BtGGuP:-KoDaK49rqW8R1BtGGuQ:b-wdsidr) ([raw file](https://github.com/valloric/ycmd/blob/3e61b03244a4d70501cfeef9323629901712d34e/ycmd/completers/cpp/flags.py#L433)):*
> ```Python
> 
>   if enable_windows_style_flags:
>     new_flags += _FilterWindowsFlags( flags_without_preceeding_dash )
> ```

This doesn't keep the order of flags which I think is important. If `enable_windows_style_flags` is set before, this can be done in the loop.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-KoDXi98Oks-6QLZBo9g:b-ek0rgs)*

bstaletic:

---

*[ycmd/completers/cpp/flags.py, line 406 at r1](https://reviewable.io:443/reviews/valloric/ycmd/789#-KoDY4KOKjEYq66Gh2CY:-KoF_sg1dUBKwSQL7G4B:bi1mat8) ([raw file](https://github.com/valloric/ycmd/blob/3e61b03244a4d70501cfeef9323629901712d34e/ycmd/completers/cpp/flags.py#L406)):*

This variable should be set before looping through the flags, otherwise, Windows flags defined before `--driver-mode=cl` will be ignored. We could simply set  `enable_windows_style_flags` like this:
```python
enable_windows_style_flags = '--driver-mode=cl' in flags
```
or if we want to be more correct, we should find the last occurrence of `--driver-mode=` in the list of flags and check if its value is `cl` or not. Probably not worth the trouble.
</blockquote></details>

Done. I included the "more correct" checking as well.

---

*[ycmd/completers/cpp/flags.py, line 433 at r1](https://reviewable.io:443/reviews/valloric/ycmd/789#-KoDaK49rqW8R1BtGGuP:-KoFa1SE4Amo0BDaMJ_Y:bhtpdem) ([raw file](https://github.com/valloric/ycmd/blob/3e61b03244a4d70501cfeef9323629901712d34e/ycmd/completers/cpp/flags.py#L433)):*

This doesn't keep the order of flags which I think is important. If `enable_windows_style_flags` is set before, this can be done in the loop.
</blockquote></details>

Done. It occured to me that it was not preserving order though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

micbou:
Reviewed 1 of 1 files at r2.

---

*[ycmd/completers/cpp/flags.py, line 428 at r2](https://reviewable.io:443/reviews/valloric/ycmd/789#-KoFetFKVxS0s9P9K99o:-KoFetFKVxS0s9P9K99p:b-23vwkp) ([raw file](https://github.com/valloric/ycmd/blob/1581d48eeb2acd4b1471575aefe2b098e20ee6b3/ycmd/completers/cpp/flags.py#L428)):*
> ```Python
>            ( not previous_flag_is_include and '/' in flag ) ) and
>              ( not enable_windows_style_flags or
>                ( enable_windows_style_flags and os.path.exists( flag ) ) ) ):
> ```

This can be simplified to:
```python
not enable_windows_style_flags or os.path.exists( flags )
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

bstaletic:

---

*[ycmd/completers/cpp/flags.py, line 428 at r2](https://reviewable.io:443/reviews/valloric/ycmd/789#-KoFetFKVxS0s9P9K99o:-KoGRzResSPwr1mG-IC2:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/1581d48eeb2acd4b1471575aefe2b098e20ee6b3/ycmd/completers/cpp/flags.py#L428)):*

This can be simplified to:
```python
not enable_windows_style_flags or os.path.exists( flags )
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

Valloric:

---

*[ycmd/completers/cpp/flags.py, line 397 at r3](https://reviewable.io:443/reviews/valloric/ycmd/789#-KoOtRpigl9Q8D9LsdnX:-KoOtRpigl9Q8D9LsdnY:b-gk1zct) ([raw file](https://github.com/valloric/ycmd/blob/9e181f75e011606e8b3c896e4286fed46f9c586a/ycmd/completers/cpp/flags.py#L397)):*
> ```Python
> 
>   for flag in flags:
>     if flag.startswith( '--driver-mode' ):
> ```

Why don't we just use `if flag == ''--driver-mode=cl''`? I see there was some discussion about making sure we look at the last entry, but the current code is functionally identical to just a simple string equality check.

So I don't see the benefit of the `startswith` part.

We _could_ look at the last `--driver-mode` flag, but then this could would still need to be changed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

bstaletic:

---

*[ycmd/completers/cpp/flags.py, line 397 at r3](https://reviewable.io:443/reviews/valloric/ycmd/789#-KoOtRpigl9Q8D9LsdnX:-KoP0QYtc1ykHlnSk1yL:be459w7) ([raw file](https://github.com/valloric/ycmd/blob/9e181f75e011606e8b3c896e4286fed46f9c586a/ycmd/completers/cpp/flags.py#L397)):*

Why don't we just use `if flag == ''--driver-mode=cl''`? I see there was some discussion about making sure we look at the last entry, but the current code is functionally identical to just a simple string equality check.

So I don't see the benefit of the `startswith` part.

We _could_ look at the last `--driver-mode` flag, but then this could would still need to be changed.
</blockquote></details>

If we used `if flag == '--driver-mode=cl'` while looping throught all flags, `enable_windows_style_flags` would end up having the value of the latest comparison.

If we do the `flag == '--driver-mode=cl'` comparison only when `flag.startswith( '--driver-mode' )` then `enable_windows_style_flags` will have the bool value set the last time a flag started with `--driver-mode`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

Valloric:

---

*[ycmd/completers/cpp/flags.py, line 397 at r3](https://reviewable.io:443/reviews/valloric/ycmd/789#-KoOtRpigl9Q8D9LsdnX:-KoT1vOZHCyiRXXHXFmX:b-plu2vx) ([raw file](https://github.com/valloric/ycmd/blob/9e181f75e011606e8b3c896e4286fed46f9c586a/ycmd/completers/cpp/flags.py#L397)):*

If we used `if flag == '--driver-mode=cl'` while looping throught all flags, `enable_windows_style_flags` would end up having the value of the latest comparison.

If we do the `flag == '--driver-mode=cl'` comparison only when `flag.startswith( '--driver-mode' )` then `enable_windows_style_flags` will have the bool value set the last time a flag started with `--driver-mode`.
</blockquote></details>

Ah, you're right, my bad! Make sure to add a test for this; I can totally see someone making the same mistake in the future.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

Valloric:
:lgtm: with a test for the `--driver-mode` flag position stuff.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-KoT24nqqlFhBPxCLAev:b-cdn4b5)*

bstaletic:
I found a bug in the new code. I'll be debugging this tonight.
If anyone is interested, it picks up windows style flags from an index on before the first `--driver-mode` flag,

Once again, sorry for dragging on forever what should have been a quick PR.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-KsmJXLxMQTDU2tiDvgK:b1vq1xy)*

crapp:
Whats the staus of this PR? Any help needed?

bstaletic:
Here's the problem.

`_RemoveFlagsPrecedingCompiler()` currently removes a lot. It assumes that the first flag before the one starting with a dash is the name of the compiler and removes the rest.

If we change the function to:
```python
def _RemoveFlagsPrecedingCompiler( flags ):
  """Assuming that the flag just before the first flag (which starts with a
  dash) is the compiler path, removes all flags preceding it."""

  for index, flag in enumerate( flags ):
    if flag.startswith( '-' ) or ( flag.startswith( '/' ) and not os.path.isfile( flag ) ):
      return ( flags[ index - 1: ] if index > 1 else
               flags )
  return flags[ :-1 ]
```
Then the fuction would not strip too much. So far so good.
Then the real problem begins. This change breaks a ton of tests.
This is because with `--driver-mode=cl` there is no way to distinguish between an absolute path and a flag.

So I gave it a fight two days in a row. And every time I fixed something five other things would break.
At this point it's really annoying and frankly this PR doesn't scratch my own itch. But I don't think it's okay to just close a PR on the account of "being lazy". So now I'm hoping someone would jump in and do the rest of the work.

crapp:
OK I'll have a look at it because for me it would be very helpful to use YouCompleteme with clang-cl driver mode.

bstaletic:
Finally made this work, as far as I can tell.
I wouldn't be surprised if I overlooked some corner case.

---

---

*[ycmd/completers/cpp/flags.py, line 397 at r3](https://reviewable.io:443/reviews/valloric/ycmd/789#-KoOtRpigl9Q8D9LsdnX:-L1CcrIhwcHE1nyRemZZ:b66lk7i) ([raw file](https://github.com/valloric/ycmd/blob/02a294209d149bf0a3e64f02462b85b8acf75459/ycmd/completers/cpp/flags.py#L397)):*

Ah, you're right, my bad! Make sure to add a test for this; I can totally see someone making the same mistake in the future.
</blockquote></details>

I believe the provided tests are enough to prevent this mistake from happening.

---

*[ycmd/completers/cpp/flags.py, line 577 at r4](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1CilJUKgM3T2ZodiDn:-L1CilJUKgM3T2ZodiDo:b7o06oj) ([raw file](https://github.com/valloric/ycmd/blob/fcbbacd1f4c8273a4268d924b6f0721df692cfa9/ycmd/completers/cpp/flags.py#L577)):*
> ```Python
>     else:
>       # TODO: Should we care about `/I` when enable_windows_style_flags
>       for path_flag in PATH_FLAGS:
> ```

My current solution simply does the following:

```python
INCLUDE_FLAGS_WIN_STYLE = [ '/I' ]
PATH_FLAGS =  [ '--sysroot=' ] + INCLUDE_FLAGS + INCLUDE_FLAGS_WIN_STYLE
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L1CcZ9-T_nEEzK32uIj:bx1idur)*

crapp:
@bstaletic Thank you so much. I won't have time to play with this until the 8th but I'll report back.

bstaletic:
@crapp I'll look forward to your review. Might be the first real world test of this.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L1HBg8JMwNzMbGe-GuC:bkd900z)*

puremourning:
I realise this is still RFC, but I feel like there are now a bunch of special cases where we probably need more tests.

---

Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 577 at r4](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1CilJUKgM3T2ZodiDn:-L1IXHqE8Xx5jbV-x44g:b-iw5n6c) ([raw file](https://github.com/valloric/ycmd/blob/fcbbacd1f4c8273a4268d924b6f0721df692cfa9/ycmd/completers/cpp/flags.py#L577)):*

My current solution simply does the following:

```python
INCLUDE_FLAGS_WIN_STYLE = [ '/I' ]
PATH_FLAGS =  [ '--sysroot=' ] + INCLUDE_FLAGS + INCLUDE_FLAGS_WIN_STYLE
```
</blockquote></details>

That seems reasonable, though a more complete solution would be:

```python
  path_flags = [ PATH_FLAGS ]
  if enable_windows_style_flags:
      path_flags.extend( WINDOWS_PATH_FLAGS )
  for path_flag in path_flags:
```

I guess in _theory_ there is a chance that when driver mode is _not_ `cl`, a real path could be `/I`

---

*[ycmd/completers/cpp/flags.py, line 118 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IVv38DWLRi4qDwa5L:-L1IVv38DWLRi4qDwa5M:b-7nj23e) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L118)):*
> ```Python
>     for flag in flags:
>       if flag.startswith( '--driver-mode' ):
>         self.enable_windows_style_flags = ( flag == '--driver-mode=cl' )
> ```

It feels strange that `_ShouldXYZ` has the side effect of _calculating_ `self.XYZ`. Perhaps the method should be called `_CheckForWindowsStyleFlags` ?

Then again, it doesn't look like `self._enable_windows_style_flags` needs to exist in `self` ? could this method not just return True/False ?

---

*[ycmd/completers/cpp/flags.py, line 161 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWWoNw55rpb2s9iME:-L1IWWoNw55rpb2s9iMF:b-mdyrgj) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L161)):*
> ```Python
>                                             filename,
>                                             add_extra_clang_flags,
>                                             self.enable_windows_style_flags )
> ```

why do we store this as a member ?

---

*[ycmd/completers/cpp/flags.py, line 326 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWlhMJw5Xh1RcsDsp:-L1IWlhMJw5Xh1RcsDsq:b-x5k3wf) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L326)):*
> ```Python
> 
> def _RemoveFlagsPrecedingCompiler( flags, enable_windows_style_flags ):
>   """Assuming that the flag just before the first flag (which starts with a
> ```

probably needs updating

---

*[ycmd/completers/cpp/flags.py, line 354 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWsXdxew0Q_U2dsK3:-L1IWsXdxew0Q_U2dsK4:bp9e13j) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L354)):*
> ```Python
>   first_flag = flags[ 0 ]
> 
>   if ( not first_flag.startswith( '-' ) and
> ```

does this need to to check for windows flags too ? e.g. `and not (windows and starts with /)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L1IXoq_i7nwdhhSFJvc:bigm26p)*

bstaletic:
> there are now a bunch of special cases

I agree that the code feels clumsy and isn't very pretty to look at. And I'm being nice to myself and this code I've written.
I'll gladly add more tests, but would like it if I got some hint, as in "test *this* case".

---

Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 577 at r4](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1CilJUKgM3T2ZodiDn:-L1Lkb7tdRWd0fO7DEL_:b-wqgg75) ([raw file](https://github.com/valloric/ycmd/blob/fcbbacd1f4c8273a4268d924b6f0721df692cfa9/ycmd/completers/cpp/flags.py#L577)):*
> I guess in theory there is a chance that when driver mode is not cl, a real path could be /I

That's correct, but I would argue `/I`, as an absolute path in one's project, isn't sane. Yet I have seen `I` deep in te file hierarchy as an include path in some projects.

I was thinking of propagating the above `self.enable_windows_style_flags` as a parameter all the way to `_MakeRelativeePathsInFlagsAbsolute()`, but at the time didn't find an obvious way to do it. I'll take another look soon.

---

*[ycmd/completers/cpp/flags.py, line 118 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IVv38DWLRi4qDwa5L:-L1IZGd4l9XnDae2KZy2:bzbfyz9) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L118)):*

It feels strange that `_ShouldXYZ` has the side effect of _calculating_ `self.XYZ`. Perhaps the method should be called `_CheckForWindowsStyleFlags` ?

Then again, it doesn't look like `self._enable_windows_style_flags` needs to exist in `self` ? could this method not just return True/False ?
</blockquote></details>

Now that I think of it, yes this could just return `True` or `False`.
So should I make it return `True`/`False` and just assign the value?

---

*[ycmd/completers/cpp/flags.py, line 161 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWWoNw55rpb2s9iME:-L1Lh_6Ct6Af6FdbE8LK:b-6wtzwr) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L161)):*

why do we store this as a member ?
</blockquote></details>

Because that's the first thing that came to my mind after thinking "I'll need this boolean in more than one method". If you have a better idea, I'm listening.

---

*[ycmd/completers/cpp/flags.py, line 326 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWlhMJw5Xh1RcsDsp:-L1IYTE6fEJB5ZOpKVaT:bn7201k) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L326)):*

probably needs updating
</blockquote></details>

My idea was:
1. Flags starting with a dash are always accepted.
2. Windows style flags are accepted when `--driver-mode=cl` is present *and* the flag is not a file.
3. So the thing before the first flag (either 1. or 2.) is the compiler.

It did occur to me that this might be wrong now.
That's why I asked on gitter why this assumption is a good on. But I received no answer.

---

*[ycmd/completers/cpp/flags.py, line 354 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWsXdxew0Q_U2dsK3:-L1LiLc2C8ULlFXWdKvr:b-dux90k) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L354)):*

does this need to to check for windows flags too ? e.g. `and not (windows and starts with /)`
</blockquote></details>

I think not. Definitely not how you have written it. Here's why.

- Because of `_RemoveFlagsPrecedingCompiler`, irrelevant of `enable_windows_style_flags`, the frst flag is (at least should be) either the compiler, a Win style flag, or starts with a dash.
- If it doesn't start wth a dash, it's either an absolute path, a win style flag or a C++ compiler from `$PATH`.
  - If it starts with a forward slash it's either a win flag or an absolute path.
    - If it matches the reges (I think) it's safe to assume that's a compiler path.
    - If it doesn't match it could still be a win style flag or some stray absolute path - which is determined later and cleaned properly.
  - If it starts with anything else, it's a stray file path and should be gotten rid of in `_RemoveUnusedFlags()`.

If we do it like you suggested you we would break the case where `/usr/bin/g++` is the first flag, but later we also have `--driver-mode=cl`.
Yes, this is messy...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L1LlzG2bYE2StueopPw:b-6vxkxz)*

puremourning:
There wasn't a _specific_ case in mind. I was just sort of expecting there to be more flag combinations that would cause us to worry. If not, and you're happy that it's hitting the cases, then I'm cool with it.

---

Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 577 at r4](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1CilJUKgM3T2ZodiDn:-L1OKp7IohZdYypy2Bts:b-76uwvk) ([raw file](https://github.com/valloric/ycmd/blob/fcbbacd1f4c8273a4268d924b6f0721df692cfa9/ycmd/completers/cpp/flags.py#L577)):*

> I guess in theory there is a chance that when driver mode is not cl, a real path could be /I

That's correct, but I would argue `/I`, as an absolute path in one's project, isn't sane. Yet I have seen `I` deep in te file hierarchy as an include path in some projects.

I was thinking of propagating the above `self.enable_windows_style_flags` as a parameter all the way to `_MakeRelativeePathsInFlagsAbsolute()`, but at the time didn't find an obvious way to do it. I'll take another look soon.
</blockquote></details>

oh i see the problem. at this point you don't know `enable_windows_style_flags`. I think we can eat bugs if `/I` is a real directory on some real system :)

---

*[ycmd/completers/cpp/flags.py, line 118 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IVv38DWLRi4qDwa5L:-L1OK-DTMHQgpZ96S3IZ:b-1gw9un) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L118)):*

Now that I think of it, yes this could just return `True` or `False`.
So should I make it return `True`/`False` and just assign the value?
</blockquote></details>

I think so, yes.

---

*[ycmd/completers/cpp/flags.py, line 161 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWWoNw55rpb2s9iME:-L1OK3tKaiKn4w0Aj7RW:b9vg7de) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L161)):*

Because that's the first thing that came to my mind after thinking "I'll need this boolean in more than one method". If you have a better idea, I'm listening.
</blockquote></details>

I think we're already just using the member version right here, so it doesn't seem to serve a purpose (yet?). It's YAGNI right now ;)

---

*[ycmd/completers/cpp/flags.py, line 326 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWlhMJw5Xh1RcsDsp:-L1OKERN6dD3uyYumwJI:b5nl1bo) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L326)):*

My idea was:
1. Flags starting with a dash are always accepted.
2. Windows style flags are accepted when `--driver-mode=cl` is present *and* the flag is not a file.
3. So the thing before the first flag (either 1. or 2.) is the compiler.

It did occur to me that this might be wrong now.
That's why I asked on gitter why this assumption is a good on. But I received no answer.
</blockquote></details>

I meant that the comment needs updating (it says starts with a dash). Perhaps just update it to be "looks like a compiler flag" or something.

---

*[ycmd/completers/cpp/flags.py, line 354 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWsXdxew0Q_U2dsK3:-L1OKeqcCrapallIpmft:bp82enc) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L354)):*

I think not. Definitely not how you have written it. Here's why.

- Because of `_RemoveFlagsPrecedingCompiler`, irrelevant of `enable_windows_style_flags`, the frst flag is (at least should be) either the compiler, a Win style flag, or starts with a dash.
- If it doesn't start wth a dash, it's either an absolute path, a win style flag or a C++ compiler from `$PATH`.
  - If it starts with a forward slash it's either a win flag or an absolute path.
    - If it matches the reges (I think) it's safe to assume that's a compiler path.
    - If it doesn't match it could still be a win style flag or some stray absolute path - which is determined later and cleaned properly.
  - If it starts with anything else, it's a stray file path and should be gotten rid of in `_RemoveUnusedFlags()`.

If we do it like you suggested you we would break the case where `/usr/bin/g++` is the first flag, but later we also have `--driver-mode=cl`.
Yes, this is messy...
</blockquote></details>

That's a great explanation, thanks! Let's put that explanation in a block comment right here :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L1OLJCss3YrnZDybqNN:bm9onrm)*

bstaletic:
I tried to think of the weirdest cases that could break the new code.
Yes, there are just a tiny amount of tests, but I can't think of a case different enough to warrant a new test.
That's why I said I'd gladly add tests if someone points out some untested case.

That said, I wouldn't be surprised if our users start to report bugs with this, but to be fair that is happening even without the `--driver-mode=cl` support. I'm not sure if that speaks in favour of this PR or not.

---

Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 577 at r4](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1CilJUKgM3T2ZodiDn:-L1OZhNQ6F_i5qSu9YKC:b-eod9d1) ([raw file](https://github.com/valloric/ycmd/blob/fcbbacd1f4c8273a4268d924b6f0721df692cfa9/ycmd/completers/cpp/flags.py#L577)):*

oh i see the problem. at this point you don't know `enable_windows_style_flags`. I think we can eat bugs if `/I` is a real directory on some real system :)
</blockquote></details>

Well, I can use `enable_windows_style_flags`, but it will always be `False`. :D

---

*[ycmd/completers/cpp/flags.py, line 118 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IVv38DWLRi4qDwa5L:-L1OOEQIrmSTfeJoLjbn:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L118)):*

I think so, yes.
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/flags.py, line 161 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWWoNw55rpb2s9iME:-L1OTdJ637-e3HIa9GzK:bhqtlfo) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L161)):*

I think we're already just using the member version right here, so it doesn't seem to serve a purpose (yet?). It's YAGNI right now ;)
</blockquote></details>

But `PrepareFlagsForClang()` isn't a member function, so, unless I'm doing something really stupid, it can only see the `enable_windows_style_flags` if I pass it as a function parameter.

Once again, I may just be unable to see my mistake.

---

*[ycmd/completers/cpp/flags.py, line 326 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWlhMJw5Xh1RcsDsp:-L1OSi1Vm4sqKLaOvUcq:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L326)):*

I meant that the comment needs updating (it says starts with a dash). Perhaps just update it to be "looks like a compiler flag" or something.
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/flags.py, line 354 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWsXdxew0Q_U2dsK3:-L1OS8oXeeYCmtcTMbeM:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L354)):*

That's a great explanation, thanks! Let's put that explanation in a block comment right here :)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L1OMuZikQMczqn4yOWm:b-g81se8)*

puremourning:
Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 161 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWWoNw55rpb2s9iME:-L1OcOk38r-fiMlD-wUc:b-drixv) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L161)):*

But `PrepareFlagsForClang()` isn't a member function, so, unless I'm doing something really stupid, it can only see the `enable_windows_style_flags` if I pass it as a function parameter.

Once again, I may just be unable to see my mistake.
</blockquote></details>

erm

```python
enable_windows_style_flags = self._ShouldAllowWinStyleFlags( flags )
sanitized_flags = PrepareFlagsForClang (..., enable_windows_style_flags )
```

or perhaps

```python
sanitized_flags = PrepareFlagsForClang (..., self._ShouldAllowWinStyleFlags( flags )  )
```

That seems overly simplistic, so perhaps I'm missing some important subtlety!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

bstaletic:
Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 161 at r5](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1IWWoNw55rpb2s9iME:-L1One7bFu-ksxmi6AZ2:b-k6tal6) ([raw file](https://github.com/valloric/ycmd/blob/0eaf64e72f96dd38e4fa00de521ee70e7c1a8185/ycmd/completers/cpp/flags.py#L161)):*

erm

```python
enable_windows_style_flags = self._ShouldAllowWinStyleFlags( flags )
sanitized_flags = PrepareFlagsForClang (..., enable_windows_style_flags )
```

or perhaps

```python
sanitized_flags = PrepareFlagsForClang (..., self._ShouldAllowWinStyleFlags( flags )  )
```

That seems overly simplistic, so perhaps I'm missing some important subtlety!
</blockquote></details>

Done.

No more unneeded `self.`. The method doesn't rely on `self` either, so it got extracted too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

puremourning:
:lgtm: Thanks!

---

Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L1PR1aHWj4XHjX4hAFO:b-g2a5fb)*

micbou:
Reviewed 1 of 1 files at r3, 1 of 2 files at r4, 1 of 1 files at r6.

---

*[ycmd/completers/cpp/flags.py, line 577 at r4](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1CilJUKgM3T2ZodiDn:-L1U2pkkIqVmbJ5r_Mrk:bhnnypz) ([raw file](https://github.com/valloric/ycmd/blob/fcbbacd1f4c8273a4268d924b6f0721df692cfa9/ycmd/completers/cpp/flags.py#L577)):*

Well, I can use `enable_windows_style_flags`, but it will always be `False`. :D
</blockquote></details>

Let's do that in a separate PR.

---

*[ycmd/completers/cpp/flags.py, line 362 at r6](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1Pc0hyBvjSc0hSMkUN:-L1Pc0hz2vhztkMQ1NkE:b-b2ui52) ([raw file](https://github.com/valloric/ycmd/blob/153ab47f3d92190b78bdd4d971839a621232f658/ycmd/completers/cpp/flags.py#L362)):*
> ```Python
>   # If it doesn't start with a dash, it is either an absolute path,
>   # a Windows style flag or a C++ compiler executable from $PATH.
>   #   If it starts with a forward slash the falg can either be an absolute
> ```

 `flag` instead of `falg`.

---

*[ycmd/completers/cpp/flags.py, line 434 at r6](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1U-og_jM8MzAamBWOU:-L1U-og_jM8MzAamBWOV:bkp7p5x) ([raw file](https://github.com/valloric/ycmd/blob/153ab47f3d92190b78bdd4d971839a621232f658/ycmd/completers/cpp/flags.py#L434)):*
> ```Python
>                  current_flag_starts_with_slash ) ) and
>          ( not previous_flag_starts_with_dash or
>            ( not previous_flag_is_include and '/' in flag ) ) ):
> ```

This doesn't work when a `/I` flag is given in two parts e.g. `[ '/I', '\\path\\to\\include' ]` because in that case, the previous flag doesn't start with a dash but a slash. We need to check `enable_windows_style_flags` in the second part of the condition too:
```python
if ( not ( current_flag_starts_with_dash or                 
           ( enable_windows_style_flags and                 
             current_flag_starts_with_slash ) ) and         
     ( not ( previous_flag_starts_with_dash or              
             ( enable_windows_style_flags and               
               previous_flag_starts_with_slash ) ) or       
       ( not previous_flag_is_include and '/' in flag ) ) ):
  continue                             
```
Since this is starting to be ugly, I would suggest to move the whole condition to a separate function.         

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

bstaletic:
Review status: 1 of 2 files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---

*[ycmd/completers/cpp/flags.py, line 362 at r6](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1Pc0hyBvjSc0hSMkUN:-L1Vho5iizqEemIxaXTc:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/153ab47f3d92190b78bdd4d971839a621232f658/ycmd/completers/cpp/flags.py#L362)):*

 `flag` instead of `falg`.
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/flags.py, line 434 at r6](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1U-og_jM8MzAamBWOU:-L1VhpXPjj_eViNTYLgn:bqm5w04) ([raw file](https://github.com/valloric/ycmd/blob/153ab47f3d92190b78bdd4d971839a621232f658/ycmd/completers/cpp/flags.py#L434)):*

This doesn't work when a `/I` flag is given in two parts e.g. `[ '/I', '\\path\\to\\include' ]` because in that case, the previous flag doesn't start with a dash but a slash. We need to check `enable_windows_style_flags` in the second part of the condition too:
```python
if ( not ( current_flag_starts_with_dash or                 
           ( enable_windows_style_flags and                 
             current_flag_starts_with_slash ) ) and         
     ( not ( previous_flag_starts_with_dash or              
             ( enable_windows_style_flags and               
               previous_flag_starts_with_slash ) ) or       
       ( not previous_flag_is_include and '/' in flag ) ) ):
  continue                             
```
Since this is starting to be ugly, I would suggest to move the whole condition to a separate function.         
</blockquote></details>

I've extracted this in a separate function, but have not implemented the fix because it breaks an already present test.

```
======================================================================
FAIL: ycmd.tests.clang.flags_test.RemoveUnusedFlags_RemoveFlagWithoutPrecedingDashFlag_test
  File "/usr/lib64/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/bstaletic/Temp/ycmd/ycmd/tests/clang/flags_test.py", line 327, in RemoveUnusedFlags_RemoveFlagWithoutPrecedingDashFlag_test
    flags._RemoveUnusedFlags( expected + to_remove, filename, True ) )
AssertionError: ['g++', '-foo', '--driver-mode=cl', '-xc++', '-bar', 'include_dir', '/foo'] != ['g++', '-foo', '--driver-mode=cl', '-xc++', '-bar', 'include_dir', '/foo', '..']

```

As you can see, the `..` path wasn't removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

bstaletic:
Review status: 0 of 2 files reviewed at latest revision, 8 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 434 at r6](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1U-og_jM8MzAamBWOU:-L24U4GnTKEEEymDK9Px:bnuzhks) ([raw file](https://github.com/valloric/ycmd/blob/153ab47f3d92190b78bdd4d971839a621232f658/ycmd/completers/cpp/flags.py#L434)):*

I've extracted this in a separate function, but have not implemented the fix because it breaks an already present test.

```
======================================================================
FAIL: ycmd.tests.clang.flags_test.RemoveUnusedFlags_RemoveFlagWithoutPrecedingDashFlag_test
  File "/usr/lib64/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/bstaletic/Temp/ycmd/ycmd/tests/clang/flags_test.py", line 327, in RemoveUnusedFlags_RemoveFlagWithoutPrecedingDashFlag_test
    flags._RemoveUnusedFlags( expected + to_remove, filename, True ) )
AssertionError: ['g++', '-foo', '--driver-mode=cl', '-xc++', '-bar', 'include_dir', '/foo'] != ['g++', '-foo', '--driver-mode=cl', '-xc++', '-bar', 'include_dir', '/foo', '..']

```

As you can see, the `..` path wasn't removed.
</blockquote></details>

I believe this was my oversight. It wasn't the code that was wrong, but the test that needed updating.

This should be fixed now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

bstaletic:
Review status: 0 of 2 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cpp/flags.py, line 577 at r4](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1CilJUKgM3T2ZodiDn:-L2pa3zD8EdmoXy_vObb:bwgdti6) ([raw file](https://github.com/valloric/ycmd/blob/fcbbacd1f4c8273a4268d924b6f0721df692cfa9/ycmd/completers/cpp/flags.py#L577)):*

Let's do that in a separate PR.
</blockquote></details>

We can do:

```python
path_flags = PATHFLAGS + INCLUDE_FLAGS_WIN_STYLE if _ShouldUseWinStyleFlags( flags ) else PATH_FLAGS
for path_flags in path_flags:
```
With `PATH_FLAGS` changed to its original value.

The negative side of this is that we're iterating over all flags to determine the driver twice.
On the other hand, everything else I've managed to think of gets messy.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

puremourning:
Do any of the tests actually use the windows-style flags to get completions from clang completer? Can we add a test project that has such a ycm extra conf, or compile db? It just occurred to me that is what felt missing :/

---

Review status: 0 of 2 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L2pwAU5wHWdSd1lAQac:bebd903)*

bstaletic:
I'll add the completion test later - completions somewhat work. Having either `/I<dir>` or `/I` and `<dir>` doesn't put the path in user's include path...

Unfortunately, diags as well have some unnecessary warnings that I'm unsure how to silence.

I made the following files:

```
bstaletic@gentoo driver_mode_cl % ls -R `pwd`
/home/bstaletic/Temp/driver_mode_cl:
dash  slash  a.cpp

/home/bstaletic/Temp/driver_mode_cl/dash:
dash.h

/home/bstaletic/Temp/driver_mode_cl/slash:
slash.h
```

```
bstaletic@gentoo driver_mode_cl % cat dash/dash.h
#ifndef DASH
#define DASH

void dash(void);

#endif /* ifndef DASH */
```

```
bstaletic@gentoo driver_mode_cl % cat slash/slash.h
#ifndef SLASH
#define SLASH

void slash(void);

#endif /* ifndef SLASH */
```

And the `a.cpp` simply includes both headers and has an empty `main()`.

Here's the extra conf that is being loaded:

```python
def FlagsForFile( filename ):
  return { 'flags': [ '--driver-mode=cl', '-Idash', '/Islash' ] }
```

And finally the `:YcmDiags`:

```
a.cpp|1 col 1 warning| unknown argument ignored in clang-cl: '-fspell-checking'
a.cpp|1 col 1 warning| unknown argument ignored in clang-cl: '-fallow-editor-placeholders'
```

Note that these diagnostics appear only when the buffer has no associated file. After saving the `a.cpp` buffer to a file and reopening vim, the diags dissapear.

I know we are adding `-fspell-checking` to the flags, but what's even stranger, after commenting out the line that adds that flag, the diag above turns to:

```
a.cpp|1 col 1 warning| unknown argument ignored in clang-cl: '-fno-spell-checking'
a.cpp|1 col 1 warning| unknown argument ignored in clang-cl: '-fallow-editor-placeholders'
```

Yes, once we don't add the flag `clang-cl` is complaining about the default value of the flag.

---

Review status: 0 of 2 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L2q0xfkaNVwlfdpf2Tc:b4ms9hc)*

bstaletic:
Like I said, there's no completion for `slash.h`, but writing `#include "slash.h"` manually doesn't cause any diagnostics to appear.

---

Review status: 0 of 2 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L2qJVx25h44u4JGPKn7:bgh8zpk)*

bstaletic:
Okay, I added a completion test.

But the problems remain.

- When the buffer has no file on the disk there are some weird diagnostics.
- The paths that get added with `/I` don't appear in include path completions.

---

Review status: 0 of 7 files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L2rVkyLnqM4xN75ok4g:b-vop8qx)*

bstaletic:
Current status of the PR:

- MacOS has a heart attack (AST deserialization error) when completing with `--driver-mode=cl`.
- Phantom diagnostics appearing in some very specific cases.
- Libclang looks for system headers in Windows paths. Which works for Windows but is nonsense on *nix and requires manual specification of `-isystem` flags from the user.

@puremourning tried bdebugging the AST deserialization error, not sure how for he managed to get.
There was also a suggestion to enable this only for Windows users, which would cover most, if not all of real world cases, eve though, theoretically, one could use CL driver on Linux.

---

Review status: 0 of 6 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L3n0FNsZIBRO9XTgWat:byabe08)*

bstaletic:
I've solved the conflicts.

Current problems with this:

- MacOS can't parse anything. Always gets an AST deserialization error.
- In CL mode libclang looks for Windows path for headers, so users of other OS's would have to manually add their `-isystem` flags.
- There are some weird diagnostics when a buffer has no file associated with it. (May be unrelated.)

@puremourning Suggested that we enable this only on Windows, where it works the best and is the most likely OS where CL mode would be used. Personally, I agree with him.

---

Review status: 0 of 6 files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L4CoZdpusrtdorV9QhN:bs7iew1)*

crapp:
Windows only would be okay for me :) 
Maybe add a warning for other platforms?

bstaletic:
The tests are finally passing.
CL driver is allowed only on Windows.

Not marked `[READY]` because I should probably squash the two commits.

---

Review status: 0 of 6 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cpp/flags.py, line 577 at r4](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1CilJUKgM3T2ZodiDn:-L4_Tk5hIviCcOM0POob:b9et45o) ([raw file](https://github.com/valloric/ycmd/blob/fcbbacd1f4c8273a4268d924b6f0721df692cfa9/ycmd/completers/cpp/flags.py#L577)):*

We can do:

```python
path_flags = PATHFLAGS + INCLUDE_FLAGS_WIN_STYLE if _ShouldUseWinStyleFlags( flags ) else PATH_FLAGS
for path_flags in path_flags:
```
With `PATH_FLAGS` changed to its original value.

The negative side of this is that we're iterating over all flags to determine the driver twice.
On the other hand, everything else I've managed to think of gets messy.
</blockquote></details>

I did take care of `_ShouldAllowWinStyleFlags( flags )` because I thought it was needed since now we only allow `--driver-mode=cl` on Windows. Same for `path_flags` in `UserIncludePaths()`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L4_UPGuC4gjke7qr0sS:buxvsi5)*

crapp:
I have just merged this PR with my local installation on a Windows Server. I am working on a 2 Million loc codebase and always had problems getting it to compile with clang using default driver. using `--driver-mode=cl` and all the MSVC flags I normally have to use, YouCompleteMe is finally working without flaws and is quite fast in providing completions. Very happy with this pull request so far. Good job @bstaletic

bstaletic:
@crapp That's great to hear. Real world testing beats the testing in this PR by a long shot.

> Good job

Thanks. It was kinda frustrating and took over half a year to get it to a usable state.

bstaletic:
Appveyor flaked, but [coverage seems fine](https://codecov.io/gh/Valloric/ycmd/commit/365cac3c3b425474e847abc6cd9753b7e4d16691).

I'm marking this as `[READY]`.

micbou:
@crapp Thanks for testing. I tried it myself on a small program and it seems to work as expected.

---

Reviewed 6 of 6 files at r7, 2 of 2 files at r8.
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cpp/flags.py, line 434 at r6](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1U-og_jM8MzAamBWOU:-L4f1i0ftU-z0fgc6U5e:b-t9bpe6) ([raw file](https://github.com/valloric/ycmd/blob/153ab47f3d92190b78bdd4d971839a621232f658/ycmd/completers/cpp/flags.py#L434)):*

I believe this was my oversight. It wasn't the code that was wrong, but the test that needed updating.

This should be fixed now.
</blockquote></details>

Sorry to be annoying but would it be possible to introduce two variables `previous_flag` and `current_flag` and to pass them to that function then compute all the variables `previous_flag_starts_with_dash`, `current_flag_starts_with_dash`, etc. inside the function? This would be cleaner.

---

*[ycmd/tests/clang/get_completions_test.py, line 513 at r8](https://reviewable.io:443/reviews/valloric/ycmd/789#-L4f3B20XIr4JstllpFv:-L4f3B20XIr4JstllpFw:b-gj1g1q) ([raw file](https://github.com/valloric/ycmd/blob/365cac3c3b425474e847abc6cd9753b7e4d16691/ycmd/tests/clang/get_completions_test.py#L513)):*
> ```Python
> @SharedYcmd
> @WindowsOnly
> def GetCompletions_ClangCLDriver_SimpleCompletion_test( app, *args ):
> ```

`*args` is not needed.

---

*[ycmd/tests/clang/get_completions_test.py, line 522 at r8](https://reviewable.io:443/reviews/valloric/ycmd/789#-L4fXUEdCr0qor-YqQzD:-L4fXUEdCr0qor-YqQzE:b1p2ex3) ([raw file](https://github.com/valloric/ycmd/blob/365cac3c3b425474e847abc6cd9753b7e4d16691/ycmd/tests/clang/get_completions_test.py#L522)):*
> ```Python
>       'line_num': 8,
>       'column_num': 18,
>       'extra_conf_data': { '&filetype': 'cpp' },
> ```

We don't need to pass any extra data to that `.ycm_extra_conf.py` file.

---

*[ycmd/tests/clang/get_completions_test.py, line 541 at r8](https://reviewable.io:443/reviews/valloric/ycmd/789#-L4fY6JgLWEEp2xVU3nz:-L4fY6JgLWEEp2xVU3o-:b-gj1g1q) ([raw file](https://github.com/valloric/ycmd/blob/365cac3c3b425474e847abc6cd9753b7e4d16691/ycmd/tests/clang/get_completions_test.py#L541)):*
> ```Python
> @SharedYcmd
> @WindowsOnly
> def GetCompletions_ClangCLDriver_IncludeStatementCandidate_test( app, *args ):
> ```

`*args` is not needed.

---

*[ycmd/tests/clang/get_completions_test.py, line 550 at r8](https://reviewable.io:443/reviews/valloric/ycmd/789#-L4fXVXUs4vT5LF0wOyn:-L4fXVXUs4vT5LF0wOyo:b-s3hmk9) ([raw file](https://github.com/valloric/ycmd/blob/365cac3c3b425474e847abc6cd9753b7e4d16691/ycmd/tests/clang/get_completions_test.py#L550)):*
> ```Python
>       'line_num': 1,
>       'column_num': 34,
>       'extra_conf_data': { '&filetype': 'cpp' },
> ```

Not needed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L4f2gCVeFndJ3Z3SGfu:bt7zhhp)*

bstaletic:
Review status: 4 of 6 files reviewed at latest revision, 11 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 434 at r6](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1U-og_jM8MzAamBWOU:-L4fe_ySLPI87iR7XPIl:b-pnmp2b) ([raw file](https://github.com/valloric/ycmd/blob/153ab47f3d92190b78bdd4d971839a621232f658/ycmd/completers/cpp/flags.py#L434)):*

Sorry to be annoying but would it be possible to introduce two variables `previous_flag` and `current_flag` and to pass them to that function then compute all the variables `previous_flag_starts_with_dash`, `current_flag_starts_with_dash`, etc. inside the function? This would be cleaner.
</blockquote></details>

Done.
It definitely is cleaner.

---

*[ycmd/tests/clang/get_completions_test.py, line 513 at r8](https://reviewable.io:443/reviews/valloric/ycmd/789#-L4f3B20XIr4JstllpFv:-L4fimfJL9dBHKPCojgH:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/365cac3c3b425474e847abc6cd9753b7e4d16691/ycmd/tests/clang/get_completions_test.py#L513)):*

`*args` is not needed.
</blockquote></details>

Done.

---

*[ycmd/tests/clang/get_completions_test.py, line 522 at r8](https://reviewable.io:443/reviews/valloric/ycmd/789#-L4fXUEdCr0qor-YqQzD:-L4fchNmyVasNoQzOLN-:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/365cac3c3b425474e847abc6cd9753b7e4d16691/ycmd/tests/clang/get_completions_test.py#L522)):*

We don't need to pass any extra data to that `.ycm_extra_conf.py` file.
</blockquote></details>

Done.

---

*[ycmd/tests/clang/get_completions_test.py, line 541 at r8](https://reviewable.io:443/reviews/valloric/ycmd/789#-L4fY6JgLWEEp2xVU3nz:-L4fifmmXcZOuDR-9ipx:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/365cac3c3b425474e847abc6cd9753b7e4d16691/ycmd/tests/clang/get_completions_test.py#L541)):*

`*args` is not needed.
</blockquote></details>

Done.

---

*[ycmd/tests/clang/get_completions_test.py, line 550 at r8](https://reviewable.io:443/reviews/valloric/ycmd/789#-L4fXVXUs4vT5LF0wOyn:-L4fcgkcfZPTXDYfPwhz:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/365cac3c3b425474e847abc6cd9753b7e4d16691/ycmd/tests/clang/get_completions_test.py#L550)):*

Not needed.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

micbou:
Reviewed 2 of 2 files at r9.

---

*[ycmd/completers/cpp/flags.py, line 434 at r6](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1U-og_jM8MzAamBWOU:-L4fmrmCpm5mOnC2FP9Y:b-f3fvxu) ([raw file](https://github.com/valloric/ycmd/blob/153ab47f3d92190b78bdd4d971839a621232f658/ycmd/completers/cpp/flags.py#L434)):*

Done.
It definitely is cleaner.
</blockquote></details>

`previous_flag_is_include` can be computed inside that function too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

bstaletic:
Review status: 5 of 6 files reviewed at latest revision, 7 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 434 at r6](https://reviewable.io:443/reviews/valloric/ycmd/789#-L1U-og_jM8MzAamBWOU:-L4fpbRRSwYRe2PlChCQ:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/153ab47f3d92190b78bdd4d971839a621232f658/ycmd/completers/cpp/flags.py#L434)):*

`previous_flag_is_include` can be computed inside that function too.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789)*

micbou:
Nice work. :lgtm:

@zzbot r+

---

Reviewed 1 of 1 files at r10.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/789#-:-L4hQCeslXOANJdcJfNu:bunv4yt)*

joelfrederico:
It looks like this is exactly what I need. The only issue is that I'm using CMake + Ninja + clang-cl + (Visual Studio's) link.exe. The easiest way for me to get the right flags would be to use compile_commands.json without specifying a .ycm_extra_conf.py. What's the best way to indicate to ycmd that my compile_commands.json uses the Windows-style flags? I think it's possible for ycmd to figure this out by checking the "command" entry and seeing that the executable is <path>\\clang-cl.exe, but I don't think it's set up to do that automatically?

bstaletic:
@joelfrederico Try the following patch and see if it solves your issue.

```diff
--- a/ycmd/completers/cpp/flags.py        2018-02-14 23:52:30.586456696 +0100
+++ b/ycmd/completers/cpp/flags.py        2018-02-21 00:09:31.394023103 +0100
@@ -248,7 +248,7 @@
       if flag.startswith( '--driver-mode' ):
         enable_windows_style_flags = ( flag == '--driver-mode=cl' )

-  return enable_windows_style_flags
+  return enable_windows_style_flags or 'clang-cl' in flags[ 0 ]


 def _CallExtraConfFlagsForFile( module, filename, client_data ):
```

puremourning:
we should probably update `CPP_COMPILER_REGEX` too if we were going to recognise `clang-cl` as a c++ compiler.

bstaletic:
@joelfrederico Can't you also simply specify `--driver-mode=cl` even though it's redundant?

joelfrederico:
@bstaletic Sorry, taking awhile, I've had distractions. I'm working on checking out the patch.

> @joelfrederico Can't you also simply specify --driver-mode=cl even though it's redundant?

No, not without a hack. The compile_commands.json file is generated by CMake automatically.

puremourning:
would `CXXFLAGS=--driver-mode=cl cmake ....` work ?

bstaletic:
> Sorry, taking awhile, I've had distractions. I'm working on checking out the patch.

No worries, take your time.

> The compile_commands.json file is generated by CMake automatically.

Yes, and you should be able to do `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --driver-mode=cl")`. @puremourning's suggestion should also work.

joelfrederico:
I hacked your patch to just simply return true just to be safe. It... sort of works... The flags look correct now in YcmDebugInfo, but I'm getting "ycm_core.ClangParseError: An AST deserialization error occurred while parsing the translation unit."

puremourning:
Are you using MSVC 2017 ? Current libclang has problems with the latest MSVC standard library headers, leading to this problem.

The only solution at the moment is to build a trunk clang. It might be resolved with clang 6 in the next month or so.

joelfrederico:
@bstaletic It seems like your patch works, for sure. The flags are correct.

@puremourning Ah that might be why. I'm upgrading to MSVC 2017 as we speak.

joelfrederico:
@puremourning Wait, apologies- did you I should use MSVC 2017 to compile ycmd? Or do you mean that my project should use the STL that comes with MSVC 2017?

puremourning:
I mean that if your project uses the standard library in the latest MSVC, you'll get these errors from ycmd (or any other libclang-based parser using the latest released libclang). This is because libclang 5.0.1 (the latest release version) chokes on the MSVS standard library headers for some reason (I don't know why).

It's fixed in upstream clang trunk, and might be fixed in clang 6 (and thus libclang 6). One solution might be to build clang from source (either clang 6 release candidate, or trunk) and use the YCM "full installation guide" to rebuild ycmd pointing at that build of clang.

joelfrederico:
@puremourning Ah. In that case, I'm still using MSVC 14 2015, not 2017. Does that version of the STL have this issue?

micbou:
Do you get the error each time you are trying to complete or it only happens once?

puremourning:
Well, it would appear that libclang is choking on your code or the STL, though I haven't heard it reported against MSVC 2014. But I don't use Windows or MSVC, so it's difficult for me to check.

It's fair to say that the cl driver mode stuff in libclang is _fairly_ new and so some issues are to be expected.

I would recommend trying the upstream libclang trunk, as others have reported it to be more stable in this mode, though again YMMV :/

joelfrederico:
@micbou It looks like it happens every time I try to complete.

@puremourning Hmmm... well... I may have the rest of the afternoon cut out for me then, that sounds involved.

joelfrederico:
@micbou To be clear- when I use a STL vector and try to dot-complete by typing a dot in insert mode in Vim, nothing happens. When I press escape immediately after, I see a red status in vim: "ClangParseError: An AST deserialization error occurred while parsing the translation unit."

puremourning:
Building LLVM/clang isn't _difficult_, just time consuming. Don't forget to `-DCMAKE_BUILD_TYPE=Release` else it will build debug and be _sloooooooow_ :)

joelfrederico:
@puremourning Yeah, I think I'll give it a try. Even when it works, it's... weird. It was complaining about a signature mismatch for a class that definitely wasn't mismatched. The signature involved a vector. I restarted ycmd and got the AST error again. It may take awhile...

bstaletic:
@puremourning I'm not sure if we should recognise `clang-cl` as C++ compiler. Since there is no `clang-cl++`, without specifying `-x<language>` it falls back to determining the language based on extension.

While the above patch works, it's not very efficient. Here's a better one:

```diff
diff --git a/ycmd/completers/cpp/flags.py b/ycmd/completers/cpp/flags.py
index 945a6af..0f1efad 100644
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -244,9 +244,16 @@ def _ExtractFlagsList( flags_for_file_output ):
 def _ShouldAllowWinStyleFlags( flags ):
   enable_windows_style_flags = False
   if OnWindows():
-    for flag in flags:
+    # clang-cl implies --driver-mode=cl
+    enable_windows_style_flags = 'clang-cl' in flags[ 0 ]
+    # Iterate in reverse because we only care
+    # about the last occurance of --driver-mode flag.
+    # Even with clang-cl, if the last --driver-mode is 'gcc'
+    # we shouldn't enable win flags.
+    for flag in reversed( flags ):
       if flag.startswith( '--driver-mode' ):
-        enable_windows_style_flags = ( flag == '--driver-mode=cl' )
+        enable_windows_style_flags = flag == '--driver-mode=cl'
+        break
 
   return enable_windows_style_flags
 
```

joelfrederico:
@bstaletic I'm not set up to do patches, but it looks like you should be able to skip checking the flags for --driver-mode=cl if you've detected we're using clang-cl already. There's no reason to allow somebody using clang-cl to disable windows-style flags, right? That's guaranteed to be broken, because clang-cl doesn't understand non-windows-stye flags. E.g.:

> def _ShouldAllowWinStyleFlags( flags ):
>   enable_windows_style_flags = False
>   if OnWindows():
>       if 'clang-cl' in flags[ 0 ]:
>         return enable_windows_style_flags = True
>       # Iterate in reverse because we only care
>       # about the last occurance of --driver-mode flag.
>       # Even with clang-cl, if the last --driver-mode is 'gcc'
>       # we shouldn't enable win flags.
>       for flag in reversed( flags ):
>         if flag.startswith( '--driver-mode' ):
>           enable_windows_style_flags = flag == '--driver-mode=cl'
>           break`

bstaletic:
@joelfrederico `clang-cl` does understand and allow non Windows flags. What's even more annoying, you can do `clang-cl --driver-mode=cl` and **disable** MSVC flags, essentially making it equivalent to `clang`.

joelfrederico:
@bstaletic Really?!? Wow... I was going off of the clang-cl help, I guess that's not documented there.

joelfrederico:
> What's even more annoying, you can do `clang-cl --driver-mode=cl` and disable MSVC flags, essentially making it equivalent to `clang`.

Wait... if that's true, don't you want to reverse the sense of the test for the flags if `clang-cl` was found? So that enable_windows_style_flags is false if `clang-cl` is found *and* `--driver-mode=cl` was found?

bstaletic:
Not really. Note that you can still do `clang-cl foo.cpp /I include_dir --driver-mode=cl` where the last flag is a no-op.
My patch does the following
- First checks for `clang-cl`.
- Then searches for the last flag starting with `--driver-mode=`.
  - If found and it equals `--driver-mode=cl` we are certain we are in CL enabled mode.
  - If found and it doesn't equal `--driver-mode=cl` we are certain we are *not* in CL enabled mode.
  - If not found it returns `'clang-cl' in flags[ 0 ]` where `flags[ 0 ]` is actually the compiler name.

That matches with the clang behaviour.

joelfrederico:
@bstaletic Do you have a link to the clang behaviour? It seems to me that your example breaks my understanding of flags.

At the very least, there's a logical inconsistency in your last post:

> Note that you can still do clang-cl foo.cpp /I include_dir --driver-mode=cl where the last flag is a no-op.
> My patch does the following
> 
> - First checks for clang-cl.

Your example finds clang-cl.

> - Then searches for the last flag starting with --driver-mode=.

Your example has this flag.

>   - If found and it equals --driver-mode=cl we are certain we are in CL enabled mode.

This condition is met. But you said earlier that the last flag is a no-op. So we aren't in CL-enabled mode, but your algorithm thinks we are.

>   - If found and it doesn't equal --driver-mode=cl we are certain we are not in CL enabled mode.
>   - If not found it returns 'clang-cl' in flags[ 0 ] where flags[ 0 ] is actually the compiler name.

These two lines aren't reached.

I mean, if you don't want to handle this edge-case, that's okay. I'm just a bit confused because this doesn't match the clang behavior you just described?

bstaletic:
> Do you have a link to the clang behaviour?

No, I have a very simple test case which I'm relying on. `main(){}` in `a.c` and playing with `clang` and `clang-cl`, watching out for errors.

You didn't understand what I meant by no-op. If you are using `clang-cl` it implies `--driver-mode=cl`, so by "no-op" I meant there's no reason to enable it *again*.

> So we aren't in CL-enabled mode

This is where you are wrong. In the example above, MSVC flags are enabled, first with `clang-cl`, then with `--driver-mode=cl`.

joelfrederico:
@bstaletic Ah but you said earlier:

> What's even more annoying, you can do `clang-cl --driver-mode=cl` and disable MSVC flags, essentially making it equivalent to `clang`.

But that's not the same as:

> If you are using `clang-cl` it implies `--driver-mode=cl`, so by "no-op" I meant there's no reason to enable it *again*.

Which one of those is true??

bstaletic:
Heh, in the first example I was thinking of `clang-cl --driver-mode=gcc`. This would make it the same as `clang`. I made a quick edit and hoped no one would notice.

`clang-cl --driver-mode=cl` does just enable MSVC flags twice. So the second statement is true.

joelfrederico:
@bstaletic Okay, then everything makes perfect sense! That initial comment threw me off. I'm on board now!

joelfrederico:
@puremourning Okay, so- I built libclang.dll. The documentation regarding linking against an externally-built libclang.dll for Windows seems to be wrong:

- You need to specify the path to llvm's root. That's the `PATH_TO_LLVM_ROOT` CMake variable.
- Building libclang
  - This doesn't create the headers in `${PATH_TO_LLVM_ROOT}/include`
  - You have to build the install target
  - It might be wise to set the install prefix in CMake so you don't overwrite a previous Clang installation

I'm fighting with LNK2001 errors. It seems like the final linking step of ycm_core isn't linking properly against libclang, or libclang didn't properly export symbols it needs.

puremourning:
I'm sorry I have no experience of doing this on Windows, but my general approach is:

* clone llvm/clang/etc. in some dir (`$CLANG_SOURCE`)
* build llvm/clang into some other dir (`$CLANG_BUILD`)
* install llvm/clang into a final dir (`$CLANG_ROOT`)
* build YCM with `env EXTRA_CMAKE_ARGS=-DPATH_TO_LLVM_ROOT=$CLANG_ROOT ./install.py --all --system-libclang`

I realise that isn't an obvious way from the docs, but ultimately it boils down to using `-DPATH_TO_LLVM_ROOT=$CLANG_ROOT` when running ycmd's cmake. the `--system-libclang` just tells `install.py` to not try and find a libclang to use.

hope that helps.

PR #792, opened 2017-07-12
Allow gcc 4.8 to be used

hotgloupi:
gcc 4.8 supports c++11 (tested on centos6)

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/792)

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -115,7 +115,7 @@ endif()
 # Determining the presence of C++11 support in the compiler
 set( CPP11_AVAILABLE false )
 if ( CMAKE_COMPILER_IS_GNUCXX )
-  if ( NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.9 )
+  if ( NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.8 )
     set( CPP11_AVAILABLE true )
   endif()
 elseif( COMPILER_IS_CLANG )


bstaletic:
GCC 4.8 does support almost all C++11 features. Those that it doesn't support it fakes. So GCC 4.8 will compile ycmd, but will break ctag compatibility.

For reference check #723.

hotgloupi:
If I understand correctly, #723 moves the supported compiler from gcc 4.6 to 4.9, in order to be sure that std::regex is there, right ? The <regex> header is already there with gcc4.8 ...
What do you mean by "breaking ctags compatibility" ?

I've tried this change locally, and the C++ completion works correctly (on RH6)...

bstaletic:
> If I understand correctly, #723 moves the supported compiler from gcc 4.6 to 4.9, in order to be sure that std::regex is there, right ?

That is correct.

>  The header is already there with gcc4.8 ...

It is there, that's true. But the tests on our Vagrant image (Ubuntu virtual machine - 14.04 at the time) were failing with GCC 4.8.

> What do you mean by "breaking ctags compatibility" ?

The regex part is used to parse `tags` files. So if `std::regex` fails, support for ctags fails too.

bstaletic:
[Here's the travis report that's showing the issue.](https://travis-ci.org/Valloric/ycmd/builds/214817252)

As you can see, the linux tests were failing. According to the log, the bot was using GCC 4.8.5.

So the question remains, have you tried this with tags file? What is the patch version of your GCC?

micbou:
[`std::regex` is implemented only since GCC 4.9](https://stackoverflow.com/a/12665408). As @bstaletic said, we use it to parse the tags files. That's why we require GCC 4.9. We don't want users to report issues about tags because they compiled ycmd with a too old version of GCC.

hotgloupi:
Ok, thanks, that makes sense.

PR #793, opened 2017-07-15
[READY] Fix compilation with Clang 3.4

micbou:
Compiling with Clang 3.4 results in the following error:
```
cpp/ycm/IdentifierUtils.cpp:120:12: error: chosen constructor is explicit in copy-initialization
    return {};
           ^~
/usr/include/c++/v1/map:838:14: note: constructor declared here
    explicit map(const key_compare& __comp = key_compare())
             ^
1 error generated.
```
See issue https://github.com/Valloric/ycmd/issues/791.

As Valloric suggested, this PR adds a Linux build with Clang 3.4 on Travis. We are supposed to support Clang 3.3 but 3.4 is the oldest version available on Travis.

I'll update the PR with the fix once the Clang build failed.

Fixes #791.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/793)

PR #795, opened 2017-07-17
[READY] Add an option to make relative paths in flags from extra conf absolute

micbou:
Since PR https://github.com/Valloric/ycmd/pull/680, all the infrastructure is here to convert any relative paths in the list of flags from the `.ycm_extra_conf.py` file to absolute ones. We just need to use the `_MakeRelativePathsInFlagsAbsolute` function.

This PR adds a new optional key `working_directory` to the dictionary returned by the `FlagsForFile` function in the `.ycm_extra_conf.py` file. This option allows specifying the working directory i.e. the directory for which the paths in the list of flags are relative to.

Update YCM's own `.ycm_extra_conf.py` file to reflect this change.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/795)

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,10 @@ The return value must be one of the following:
 
   - `flags`: (mandatory) a list of compiler flags.
 
+  - `include_paths_relative_to_dir': (optional) the directory to which the
+    include paths in the list of flags are relative. Defaults to ycmd working
+    directory.
+
   - `do_cache`: (optional) a boolean indicating whether or not the result of
     this call (i.e. the list of flags) should be cached for this file name.
     Defaults to `True`. If unsure, the default is almost always correct.
--- a/cpp/ycm/.ycm_extra_conf.py
+++ b/cpp/ycm/.ycm_extra_conf.py
@@ -106,35 +106,6 @@ def DirectoryOfThisScript():
   return os.path.dirname( os.path.abspath( __file__ ) )
 
 
-def MakeRelativePathsInFlagsAbsolute( flags, working_directory ):
-  if not working_directory:
-    return list( flags )
-  new_flags = []
-  make_next_absolute = False
-  path_flags = [ '-isystem', '-I', '-iquote', '--sysroot=' ]
-  for flag in flags:
-    new_flag = flag
-
-    if make_next_absolute:
-      make_next_absolute = False
-      if not flag.startswith( '/' ):
-        new_flag = os.path.join( working_directory, flag )
-
-    for path_flag in path_flags:
-      if flag == path_flag:
-        make_next_absolute = True
-        break
-
-      if flag.startswith( path_flag ):
-        path = flag[ len( path_flag ): ]
-        new_flag = path_flag + os.path.join( working_directory, path )
-        break
-
-    if new_flag:
-      new_flags.append( new_flag )
-  return new_flags
-
-
 def IsHeaderFile( filename ):
   extension = os.path.splitext( filename )[ 1 ]
   return extension in [ '.h', '.hxx', '.hpp', '.hh' ]
@@ -159,26 +130,29 @@ def GetCompilationInfoForFile( filename ):
 
 
 def FlagsForFile( filename, **kwargs ):
-  if database:
-    # Bear in mind that compilation_info.compiler_flags_ does NOT return a
-    # python list, but a "list-like" StringVec object
-    compilation_info = GetCompilationInfoForFile( filename )
-    if not compilation_info:
-      return None
+  if not database:
+    return {
+      'flags': flags,
+      'include_paths_relative_to_dir': DirectoryOfThisScript()
+    }
 
-    final_flags = MakeRelativePathsInFlagsAbsolute(
-      compilation_info.compiler_flags_,
-      compilation_info.compiler_working_dir_ )
+  compilation_info = GetCompilationInfoForFile( filename )
+  if not compilation_info:
+    return None
 
-    # NOTE: This is just for YouCompleteMe; it's highly likely that your project
-    # does NOT need to remove the stdlib flag. DO NOT USE THIS IN YOUR
-    # ycm_extra_conf IF YOU'RE NOT 100% SURE YOU NEED IT.
-    try:
-      final_flags.remove( '-stdlib=libc++' )
-    except ValueError:
-      pass
-  else:
-    relative_to = DirectoryOfThisScript()
-    final_flags = MakeRelativePathsInFlagsAbsolute( flags, relative_to )
+  # Bear in mind that compilation_info.compiler_flags_ does NOT return a
+  # python list, but a "list-like" StringVec object.
+  final_flags = list( compilation_info.compiler_flags_ )
 
-  return { 'flags': final_flags }
+  # NOTE: This is just for YouCompleteMe; it's highly likely that your project
+  # does NOT need to remove the stdlib flag. DO NOT USE THIS IN YOUR
+  # ycm_extra_conf IF YOU'RE NOT 100% SURE YOU NEED IT.
+  try:
+    final_flags.remove( '-stdlib=libc++' )
+  except ValueError:
+    pass
+
+  return {
+    'flags': final_flags,
+    'include_paths_relative_to_dir': compilation_info.compiler_working_dir_
+  }
--- a/examples/.ycm_extra_conf.py
+++ b/examples/.ycm_extra_conf.py
@@ -70,35 +70,6 @@ def DirectoryOfThisScript():
   return os.path.dirname( os.path.abspath( __file__ ) )
 
 
-def MakeRelativePathsInFlagsAbsolute( flags, working_directory ):
-  if not working_directory:
-    return list( flags )
-  new_flags = []
-  make_next_absolute = False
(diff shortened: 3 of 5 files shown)


micbou:
Looks like Travis recently added the `bin` folder in the home directory. Linux builds are now failing with this error:
```
mkdir: cannot create directory ‘/home/travis/bin’: File exists
```
We need to merge PR #793 first to add the `-p` flag to [the `mkdir` command in our Travis script](https://github.com/Valloric/ycmd/blob/master/ci/travis/travis_install.linux.sh#L6). I can also create a separate PR for that.

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795#-:-KpHIdPaoZR0Dk3aBeU1:b-yv52dz)*

puremourning:
I worry that this is possibly a breaking change by default. By manipulating the paths when the new flag is not specified we are introducing silent change in behaviour, which I worry will break existing (working) installations. I am fairly certain that this would break my setup at work for example, where the relative nature of the include paths is accommodated in the make system integration. 

---

Review status: 3 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795#-:-KpI4i5ZbE6hrEdHnfWu:b-vskvt5)*

micbou:
This only affects relative paths. Are you passing relative paths to libclang? What's the expected behavior in that case? Seems undefined to me.

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795#-:-KpI5eZqn-JmQed1YsCw:bcilya7)*

puremourning:
I'm pretty sure we pass -I. and it is interpreted relative to the TU. Memory could be off in that. Passing relative paths to a compiler is pretty normal behaviour. I suppose I t might be interpreted relative to the wd of ycmd, but (especially in the case of a global extra conf), I can't see a good justification for assuming that the YCM extra conf path as the root of such relative paths.

 The relative paths in a standard cc invocation are interpreted relative to the wd of the compiler when invoked which is (let's say) typically the path of the cc file. In the compilation database case, we set the calls relative to the compiler wd entry in the database, not relative to the compile commands file.  

Ofc every project is different, but that's why I'm nervous about making assumptions about users' project structure or paradigms. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795#-:-KpI731R6bDjjiNvmg5z:b1vydb8)*

Valloric:
I think @puremourning is right; this would break some people. We could perhaps do this without the new default of always resolving relative paths; we'd only do it when `working_directory` is provided.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795#-:-KpIHjzxD5J7EdXZeGPN:b1ssgjf)*

micbou:
> I'm pretty sure we pass -I. and it is interpreted relative to the TU.

Seems really convoluted to write include paths in its `.ycm_extra_conf.py` relative to the TU, i.e. the currently edited file.

> I can't see a good justification for assuming that the YCM extra conf path as the root of such relative paths.

Because that's natural. When writing relative paths in a file, users expect them to be relative to this file.

> I think @puremourning is right; this would break some people. We could perhaps do this without the new default of always resolving relative paths; we'd only do it when working_directory is provided.

This defeats the purpose of this PR which is helping users writing `.ycm_extra_conf.py` files, because they won't know about this option. They will write relative paths in their `.ycm_extra_conf.py` files, of course relative to the `.ycm_extra_conf.py` file, and it won't work.

I'd also like to point out that, in addition to being less than obvious (what is the actual working directory?), the current behavior for relative paths is undocumented. Do we want to go as far as not breaking backward compatibility for undocumented and obscure behavior?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795#-:-KpIA1MTxsd4f-JipHop:b2qqq8e)*

puremourning:
The use case in question is where the flags are not actually writen into the `.ycm_extra_conf.py`, but the python within shells out to the make system to get the flags (in our case, effectively `make COMPILER=clang show_CXXFLAGS`). This returns the compile flags that would be passed for `file_name` when using the normal make system, rather than somehow repeating that logic within the extra conf file. Pretty sure we're not the only people which do this sort of thing, and it's essentially how the `compile_commands.json` works.

In any case, the current behaviour actually relative to the ycmd wd:

Given the following file structure:

```
$> find .
.
./test.h
./path
./path/test.h
./path/test.cc
./.ycm_extra_conf.py
ukfd-ltp2l(benj:MMA_main.git) tmp/relative>
```

with:

./test.h = `int test;`
./path/test.h = `int test;`
./path/test.cc = 

```c++
#include <test.h>
#include <stdio.h>

int main()
{
    printf( "%c", test );
}
```

and `.ycm_extra_conf.py`:

```python
def FlagsForFile( file_name, **kwargs ):
    return { 
        'flags': [
            '-x', 'c++',
            '-I.'
        ]
    }
```

* With Vim wd = ./path: `:YcmCompleter GetType` on `test` returns `char` (not `int`).
* With Vim wd = ./: `YcmCompleter GetType` on `test` returns `int`.

I confess that it (currently) being relative to ycmd working directory is arbitrary. However, I'm still not convinced that making it relative to the `.ycm_extra_conf.py` seems any less wrong. Again, in particular, when using a global `ycm_extra_conf.py` (which would be in any random dir).

If we were to pick a "default", I would suggest making it relative to the `dirname` of the `file_name`. Though this again is somewhat of a breaking change.

To be clear: I'm _totally_ up for adding the `working_dir` item to the dict. This is a good move and I will likely implement it here as `os.path.dirname( file_name )`, but I'm still worried that forcefully changing the paths to an (arbitrarily) different default, is just causing potentially avoidable headaches in the (perhaps unlikely) case that people have become accustomed to this behaviour. I think we can say in our docs that the default is ycmd working dir, as that's what it currently is.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795#-:-KpI9NNeqXvM7hd4jvDh:b-q6hwxh)*

micbou:
You convinced me. I'll update the PR to only make the paths absolute when `working_directory` is specified (is `working_dir` a better name?). That's a little unfortunate for new users. I think they would expect paths to be relative to a local `.ycm_extra_conf.py` by default (as you said, this doesn't make sense for the global extra conf). But we can't break users that rely on the current behavior.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795#-:-KpLECa_FRhAUSraFIK0:bk7rgut)*

puremourning:
Maybe we should straight up call a spoon a spoon and call it `paths_relative_to` as we aren't actually wetting a working dir?

Valloric:
Not at all against a clearer name than `working_dir`; we could call it `include_paths_relative_to_dir` and remove all confusion.

> This defeats the purpose of this PR which is helping users writing .ycm_extra_conf.py files, because they won't know about this option. 

I'd say the proportion of users who learn how to write an extra conf file by taking ycmd's (which we link to from the docs) and then change it to suit them is >95%. :D 

So if we use it, they'll know about the option.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795#-:-KpNFkJLNU_5-ERs41JT:b-chiyna)*

micbou:
Renamed the option to `include_paths_relative_to_dir` and updated the `.ycm_extra_conf.py` file in the `examples` folder too. PR is ready.

---

Reviewed 4 of 4 files at r3, 4 of 4 files at r4, 1 of 1 files at r5, 2 of 2 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795#-:-KqIELoPOYKM_r1CPbfx:b-52eyxz)*

vheon:
Reviewed 2 of 4 files at r4, 1 of 1 files at r5, 2 of 2 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795)*

Valloric:
:lgtm:

Thanks!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795#-:-KqsZImVycEnhbbA1y3S:b9benhj)*

bstaletic:
:lgtm:
Seems to work as expected. Thanks.
Seeing that @puremourning was the last one asking questions, I'll let him trigger @zzbot.

---

Reviewed 2 of 4 files at r4, 1 of 1 files at r5, 2 of 2 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795#-:-Kr-0Nh8j7OcRljIikD6:bqc3fp8)*

puremourning:
:lgtm:

Thanks!

@zzbot r+

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/795#-:-KsdDgrSsQuLz0XXQfmv:b-7vhnyt)*

PR #799, opened 2017-07-19
[READY] Fix virtualenv prefix in build script

### LINKED ISSUE #798: build.py on Linux via pyenv virtualenv broken in ff770b4
I'm trying to build YCM on Ubuntu with pyenv-installed Python 3 (built with `--enable-shared`), in a separate virtualenv (so `libpython3.so` is not located in the prefix directory, but in the parent lib directory).

However, it complains that it can't find dynamic libraries. I've checked, it used to work fine up until and including commit 5ffe1bfdc9c026d289081e9442f07ab03021189f, but is then broken starting from commit ff770b4f47e793a7c7e839f41861e564ef6a7caf ("Search Python library in lib64").

The reason for that being is that both `sys.prefix` and `sys.base_prefix` point to the virtualenv's lib directory, while `distutils.sysconfig.get_python_lib(standard_lib=1)` (the one which was removed in that commit) points to the parent lib directory where the .so file is located. Could it be please added back? (and the submodule bumped in YCM)

micbou:
In a virtualenv, `sys.prefix` points to the virtualenv directory. Use `sys.real_prefix` instead.

Fixes #798.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/799)

### DIFF
--- a/build.py
+++ b/build.py
@@ -146,8 +146,19 @@ def CheckCall( args, **kwargs ):
     sys.exit( error.returncode )
 
 
+def GetGlobalPythonPrefix():
+  # In a virtualenv, sys.real_prefix points to the parent Python prefix.
+  if hasattr( sys, 'real_prefix' ):
+    return sys.real_prefix
+  # In a pyvenv (only available on Python 3), sys.base_prefix points to the
+  # parent Python prefix. Outside a pyvenv, it is equal to sys.prefix.
+  if PY_MAJOR >= 3:
+    return sys.base_prefix
+  return sys.prefix
+
+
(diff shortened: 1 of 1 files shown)


vheon:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/799#-:-KpPowplEqT8KIlVl_zA:bnfp4nl)*

Valloric:
:lgtm:

Thanks for the PR!

@zzbot r=vheon

PR #810, opened 2017-08-04
[READY] Support unknown languages from tags

### LINKED ISSUE #809: Tags for custom languages?
Unless I'm doing something wrong, ycmd doesn't seem to be adding tag completion to my output for moonscript. I'm using  [emacs-ycmd](https://github.com/abingham/emacs-ycmd) for the completion. I'm generating the tag file using `--fields=+l` but it doesn't want to work. It works fine with lua, so I'm not sure what else I need to do. 

This is the generated tag file:
```!_TAG_FILE_FORMAT       2       /extended format; --format=1 will not append ;" to lines/
!_TAG_FILE_SORTED       1       /0=unsorted, 1=sorted, 2=foldcase/
!_TAG_PROGRAM_AUTHOR    Darren Hiebert  /dhiebert@users.sourceforge.net/
!_TAG_PROGRAM_NAME      Exuberant Ctags //
!_TAG_PROGRAM_URL       http://ctags.sourceforge.net    /official site/
!_TAG_PROGRAM_VERSION   5.9~svn20110310 //
API     test.moon     /^        class TEST $/;"       c       language:moonscript
DELETE  test.moon     /^        DELETE:  =>$/;"   m       language:moonscript

### PR DESCRIPTION
**Pull request by micbou**
When a language from a tags file is unknown, set the filetype to its name in lowercase. This allows the use of tags for custom languages.

Fixes #809.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/810)

### DIFF
--- a/cpp/ycm/IdentifierUtils.cpp
+++ b/cpp/ycm/IdentifierUtils.cpp
@@ -106,8 +106,6 @@ const std::unordered_map < const char *,
         { "YACC"       , "yacc"       }
       };
 
-const char *const NOT_FOUND = "YCMFOOBAR_NOT_FOUND";
-
 }  // unnamed namespace
 
 
@@ -135,10 +133,7 @@ FiletypeIdentifierMap ExtractIdentifiersFromTagsFile(
     std::string language( matches[ 3 ] );
     std::string filetype = FindWithDefault( LANG_TO_FILETYPE,
                                             language.c_str(),
-                                            NOT_FOUND );
-
-    if ( filetype == NOT_FOUND )
-      continue;
+                                            Lowercase( language ).c_str() );
 
     std::string identifier( matches[ 1 ] );
     fs::path path( matches[ 2 ].str() );
--- a/cpp/ycm/Utils.cpp
+++ b/cpp/ycm/Utils.cpp
@@ -119,6 +119,12 @@ char Lowercase( char letter ) {
   return letter;
 }
 
+std::string Lowercase( const std::string &text ) {
+  std::string result;
+  for ( char letter : text )
+    result.push_back( Lowercase( letter ) );
+  return result;
+}
 
 char Uppercase( char letter ) {
   if ( IsLowercase( letter ) )
--- a/cpp/ycm/Utils.h
+++ b/cpp/ycm/Utils.h
@@ -87,6 +87,7 @@ YCM_DLL_EXPORT bool IsLowercase( char letter );
 YCM_DLL_EXPORT bool IsLowercase( const std::string &text );
 YCM_DLL_EXPORT bool IsUppercase( char letter );
 YCM_DLL_EXPORT char Lowercase( char letter );
+YCM_DLL_EXPORT std::string Lowercase( const std::string &text );
 YCM_DLL_EXPORT char Uppercase( char letter );
 YCM_DLL_EXPORT bool HasUppercase( const std::string &text );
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by bstaletic** [comment]
:lgtm:
Thanks fr the PR.

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/810#-:-Kqh6HELnsrfqEcVBfAZ:b-nxl7eh)*

**Comment by Valloric** [comment]
:lgtm:

Thanks for the PR!

@zzbot r=bstaletic

**Comment by puremourning** [comment]

---

*[cpp/ycm/Utils.cpp, line 123 at r1](https://reviewable.io:443/reviews/valloric/ycmd/810#-KqmT8M7WurvxUEbLSe5:-KqmT8M7WurvxUEbLSe6:b-tl2o6m) ([raw file](https://github.com/valloric/ycmd/blob/adc1d84fc41b44f4b7e0ab85d90eef43945f774f/cpp/ycm/Utils.cpp#L123)):*
> ```cpp
> 
> std::string Lowercase( const std::string &text ) {
>   std::string result;
> ```

(No big deal) We should reserve the known size here for performance cruft reasons. I also think we should allow this to be inlined. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/810)*

**Comment by micbou** [comment]
*[cpp/ycm/Utils.cpp, line 123 at r1](https://reviewable.io:443/reviews/valloric/ycmd/810#-KqmT8M7WurvxUEbLSe5:-KqmVEIABV5Cro2ID98K:b7na6ew) ([raw file](https://github.com/valloric/ycmd/blob/adc1d84fc41b44f4b7e0ab85d90eef43945f774f/cpp/ycm/Utils.cpp#L123)):*
> We should reserve the known size here for performance cruft reasons

Something like:
```
result.reserve( text.length() )
```
? I suppose the same applies to the `SwapCase` function below?

> I also think we should allow this to be inlined.

I thought that the compiler would automatically inline this kind of function with `-O2` but I've looked at the assembly code of a trivial program that calls this function and that's not the case. I guess we should add the `inline` keyword for all these utility functions (from `IsAscii` to `SwapCase`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/810)*

**Comment by puremourning** [comment]

---

*[cpp/ycm/Utils.cpp, line 123 at r1](https://reviewable.io:443/reviews/valloric/ycmd/810#-KqmT8M7WurvxUEbLSe5:-KqnTii7LHThtgRhe3Kf:b-do6w65) ([raw file](https://github.com/valloric/ycmd/blob/adc1d84fc41b44f4b7e0ab85d90eef43945f774f/cpp/ycm/Utils.cpp#L123)):*

> We should reserve the known size here for performance cruft reasons

Something like:
```
result.reserve( text.length() )
```
? I suppose the same applies to the `SwapCase` function below?

> I also think we should allow this to be inlined.

I thought that the compiler would automatically inline this kind of function with `-O2` but I've looked at the assembly code of a trivial program that calls this function and that's not the case. I guess we should add the `inline` keyword for all these utility functions (from `IsAscii` to `SwapCase`).
</blockquote></details>

It can't inline if it can't see he source. Eg it is in a different TU. just moving to th header should be enough. And yes the reserve call looks right to me. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/810)*

PR #814, opened 2017-08-12
[READY] Resolve symlinks in extra conf glob pattern

micbou:
When a `.ycm_extra_conf.py` file is found, ycmd determines if this file should be white/blacklisted by checking that its path matches one of the patterns from [the `extra_conf_globlist` option](https://github.com/Valloric/YouCompleteMe#the-gycm_extra_conf_globlist-option). This is done by converting the `.ycm_extra_conf.py` path and the patterns to absolute paths through [the `os.path.abspath` function](https://docs.python.org/2/library/os.path.html#os.path.abspath). However, this function doesn't resolve symbolic links. This means that if a pattern in `extra_conf_globlist` contains a symbolic link, any `.ycm_extra_conf.py` file that should be matched by this pattern if the symlink was resolved, won't be white/blacklisted.

This is fixed on UNIX by replacing `os.path.abspath` with [the `os.path.realpath` function](https://docs.python.org/2/library/os.path.html#os.path.realpath). Windows symlinks are not supported because [`os.path.realpath` doesn't resolve symlinks on this platform](https://bugs.python.org/issue9949) and writing our own implementation of `os.path.realpath` supporting Windows symlinks is not worth the trouble.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/814)

### DIFF
--- a/ycmd/extra_conf_store.py
+++ b/ycmd/extra_conf_store.py
@@ -184,10 +184,13 @@ def Load( module_file, force = False ):
 def _MatchesGlobPattern( filename, glob ):
   """Returns true if a filename matches a given pattern. A '~' in glob will be
   expanded to the home directory and checking will be performed using absolute
-  paths. See the documentation of fnmatch for the supported patterns."""
+  paths with symlinks resolved (except on Windows). See the documentation of
+  fnmatch for the supported patterns."""
 
-  abspath = os.path.abspath( filename )
-  return fnmatch( abspath, os.path.abspath( os.path.expanduser( glob ) ) )
+  # NOTE: os.path.realpath does not resolve symlinks on Windows.
+  # See https://bugs.python.org/issue9949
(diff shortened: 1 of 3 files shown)


bstaletic:
:lgtm:
So that's the difference between `abspath()` and `realpath()`! Thanks.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/814#-:-KrNDhm_oxwYXja7mbtC:bw5zjj3)*

Valloric:
A fantastic PR, as is customary from @micbou. :)

Thanks!

@zzbot r=bstaletic

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/814#-:-KrNgISsagIY72onmeM2:b-jzab5d)*

PR #819, opened 2017-08-26
[READY] Filter and sort candidates when query is empty

micbou:
An obvious reason to sort the candidates with our `FilterAndSortCandidates` function even if the query is empty is that returning candidates in lexicographical order is better than in random order. Consider this example in JavaScript:

sort-on-empty-query

Another reason is that some semantic engines (Jedi, TSServer, Gocode, and C#) return the candidates sorted (generally in lexicographical order) while others don't (Clang, Tern, and racer). Thus, the order of candidates depends on the completer when the query is empty. This is inconsistent.

That's not all. In addition to sorting, `FilterAndSortCandidates` filters candidates that are too long (more than 80 characters) or contain non-ASCII characters. Not filtering them when the query is empty leads to confusion. For instance:

empty-query-non-ascii
*Why did the `foø` candidate disappear?!*

Last point (which is an implementation detail) is that this is going to help implementing partial sorting (and limiting the number of candidates returned) because we won't have to handle the case where the query is empty.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/819)


vheon:
:lgtm: 

---

Reviewed 9 of 9 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/819#-:-Ks_Lt5kBpy_DfgnEzOU:bg7igkv)*

Valloric:
:lgtm:

Thanks!

@zzbot r=vheon

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/819#-:-KsfJTcYYQLNEce_9-yp:bqac5o5)*

PR #820, opened 2017-08-27
[READY] Improve LLVM root path search for prebuilt binaries

micbou:
This PR simplifies the way we determine `PATH_TO_LLVM_ROOT` for Clang prebuilt binaries by using a glob pattern.

Fixes https://github.com/Valloric/YouCompleteMe/issues/2756.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/820)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -127,24 +127,21 @@ if ( USE_CLANG_COMPLETER AND
     execute_process( COMMAND tar -xzf ${CLANG_FILENAME} )
   endif()
 
-  # We need to set PATH_TO_LLVM_ROOT. To do that, we first have to find the
-  # folder name the archive produced. It isn't the archive base name.
-  # On Windows, the find command is not available so we directly set the path.
-  if ( WIN32 )
-    # Since 7-zip release version 15.12, files are not anymore extracted in
-    # $_OUTDIR folder but directly to the root. So, we check the existence of
-    # $_OUTDIR to appropriately set PATH_TO_LLVM_ROOT.
-    if ( EXISTS ${CMAKE_CURRENT_BINARY_DIR}/../$_OUTDIR )
-      set( PATH_TO_LLVM_ROOT ${CMAKE_CURRENT_BINARY_DIR}/../$_OUTDIR )
-    else()
-      set( PATH_TO_LLVM_ROOT ${CMAKE_CURRENT_BINARY_DIR}/.. )
-    endif()
-  else()
-    execute_process( COMMAND
-      find ${CMAKE_CURRENT_BINARY_DIR}/.. -maxdepth 1 -type d -name clang*
-      OUTPUT_VARIABLE PATH_TO_LLVM_ROOT
-      OUTPUT_STRIP_TRAILING_WHITESPACE )
+  # We determine PATH_TO_LLVM_ROOT by searching the libclang library path in
+  # CMake build folder.
+  file( GLOB_RECURSE PATH_TO_LIBCLANG ${CMAKE_BINARY_DIR}/libclang.* )
+  if ( NOT PATH_TO_LIBCLANG )
+    message( FATAL_ERROR "Cannot find path to libclang in prebuilt binaries" )
   endif()
+  # file( GLOB_RECURSE ... ) returns a list of files. Take the first one.
+  list( GET PATH_TO_LIBCLANG 0 PATH_TO_LIBCLANG )
+  # We know that LLVM root is parent to the directory containing libclang so we
+  # need to go up two directories:
+  #
+  #   /path/to/llvm/root/lib/libclang.so/../.. = /path/to/llvm/root/
+  #
+  get_filename_component( PATH_TO_LLVM_ROOT "${PATH_TO_LIBCLANG}/../.."
+                          ABSOLUTE )
 endif()
 
 if ( PATH_TO_LLVM_ROOT OR USE_SYSTEM_LIBCLANG OR EXTERNAL_LIBCLANG_PATH )


bstaletic:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/820#-:-KsY-4E8W_Ir8WU_vF9Y:bnfp4nl)*

puremourning:

---

*[cpp/ycm/CMakeLists.txt, line 132 at r1](https://reviewable.io:443/reviews/valloric/ycmd/820#-KsciNLrObw_8wAgfx-e:-KsciNLrObw_8wAgfx-f:bmo62zt) ([raw file](https://github.com/valloric/ycmd/blob/1f645757815910266b8bd93b85ba813c28e3153f/cpp/ycm/CMakeLists.txt#L132)):*
So, while I'm sure this works, I can't find any evidence in the (abysmal) cmake docs that `**` is portably supported in glob expressions. Or at least, whether it has a special meaning. I suspect this is actually representing 2 consecutive `*` matches, which seems redundant.

I quickly checked the man pages on my OS (macOS) for:

* `man -s 3 glob` - the c glob function
* `man bash` - the bash shell expansion
* `man tcsh` - the csh shell expansion

The only indication of special behaviour i found was in `tcsh`:

>        The globstar shell variable can be set to allow `**' or `***' as a file glob pattern that matches any string of characters including `/', recursively traversing any existing sub-directories.  For example, `ls **.c' will list  all
       the  .c  files  in  the  current  directory  tree.   If  used by itself, it will match match zero or more sub-directories (e.g. `ls /usr/include/**/time.h' will list any file named `time.h' in the /usr/include directory tree; `ls
       /usr/include/**time.h' will match any file in the /usr/include directory tree ending in `time.h'; and `ls /usr/include/**time**.h' will match any .h file with `time' either in a subdirectory name or in the filename  itself).   To
       prevent problems with recursion, the `**' glob-pattern will not descend into a symbolic link containing a directory.  To override this, use `***' (+)

So I'm unsure how the `**` is being interpreted here, and I struggled to find out. So the question is, are you confident in this? If so, meh :)

---

*[cpp/ycm/CMakeLists.txt, line 133 at r1](https://reviewable.io:443/reviews/valloric/ycmd/820#-KscgB3LylIntnJSHzz4:-KscgB3LylIntnJSHzz5:b-7ykd28) ([raw file](https://github.com/valloric/ycmd/blob/1f645757815910266b8bd93b85ba813c28e3153f/cpp/ycm/CMakeLists.txt#L133)):*
> ```CMake
>   # CMake build folder.
>   file( GLOB_RECURSE PATH_TO_LLVM_ROOT ${CMAKE_BINARY_DIR}/**/libclang.* )
>   list( GET PATH_TO_LLVM_ROOT 0 PATH_TO_LLVM_ROOT )
> ```

Because I don't grok cmake (and the documentation for cmake is _appalling_), what the behaviour in the case that this list has 0 elements?

---

*[cpp/ycm/CMakeLists.txt, line 134 at r1](https://reviewable.io:443/reviews/valloric/ycmd/820#-KscfoZUHy49QQQx3xoy:-KscfoZUHy49QQQx3xoz:bwp78lh) ([raw file](https://github.com/valloric/ycmd/blob/1f645757815910266b8bd93b85ba813c28e3153f/cpp/ycm/CMakeLists.txt#L134)):*
> ```CMake
>   file( GLOB_RECURSE PATH_TO_LLVM_ROOT ${CMAKE_BINARY_DIR}/**/libclang.* )
>   list( GET PATH_TO_LLVM_ROOT 0 PATH_TO_LLVM_ROOT )
>   get_filename_component( PATH_TO_LLVM_ROOT "${PATH_TO_LLVM_ROOT}/../.."
> ```

It's not super obvious to me why we need `../..` here or why that's (canonically) correct. I'm sure it is (e.g. it represents `usr/lib` or something), but perhaps we need a comment?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/820)*

micbou:
Thanks for the review.

---

---

*[cpp/ycm/CMakeLists.txt, line 132 at r1](https://reviewable.io:443/reviews/valloric/ycmd/820#-KsciNLrObw_8wAgfx-e:-Kscxj7YTiV1BI3lsP8y:b-wlfr3s) ([raw file](https://github.com/valloric/ycmd/blob/1f645757815910266b8bd93b85ba813c28e3153f/cpp/ycm/CMakeLists.txt#L132)):*

So, while I'm sure this works, I can't find any evidence in the (abysmal) cmake docs that `**` is portably supported in glob expressions. Or at least, whether it has a special meaning. I suspect this is actually representing 2 consecutive `*` matches, which seems redundant.

I quickly checked the man pages on my OS (macOS) for:

* `man -s 3 glob` - the c glob function
* `man bash` - the bash shell expansion
* `man tcsh` - the csh shell expansion

The only indication of special behaviour i found was in `tcsh`:

>        The globstar shell variable can be set to allow `**' or `***' as a file glob pattern that matches any string of characters including `/', recursively traversing any existing sub-directories.  For example, `ls **.c' will list  all
       the  .c  files  in  the  current  directory  tree.   If  used by itself, it will match match zero or more sub-directories (e.g. `ls /usr/include/**/time.h' will list any file named `time.h' in the /usr/include directory tree; `ls
       /usr/include/**time.h' will match any file in the /usr/include directory tree ending in `time.h'; and `ls /usr/include/**time**.h' will match any .h file with `time' either in a subdirectory name or in the filename  itself).   To
       prevent problems with recursion, the `**' glob-pattern will not descend into a symbolic link containing a directory.  To override this, use `***' (+)

So I'm unsure how the `**` is being interpreted here, and I struggled to find out. So the question is, are you confident in this? If so, meh :)
</blockquote></details>

For some reason, I thought I could use `**` like in Zsh. In fact, it's totally useless when using `GLOB_RECURSE` (see the `/dir/*.py` example in the docs). I've modified the pattern.

---

*[cpp/ycm/CMakeLists.txt, line 133 at r1](https://reviewable.io:443/reviews/valloric/ycmd/820#-KscgB3LylIntnJSHzz4:-KscuppuVRRy2VPjBi6e:b-1eepdz) ([raw file](https://github.com/valloric/ycmd/blob/1f645757815910266b8bd93b85ba813c28e3153f/cpp/ycm/CMakeLists.txt#L133)):*

Because I don't grok cmake (and the documentation for cmake is _appalling_), what the behaviour in the case that this list has 0 elements?
</blockquote></details>

If the list is empty, CMake returns the following error:
```
CMake Error at CMakeLists.txt:7 (list):
  list GET given empty list
```
but continues to execute the script. This results in `PATH_TO_LLVM_ROOT` being set to `/` in the next line. Probably harmless but still better to check if the list is empty and abort the script if so.

---

*[cpp/ycm/CMakeLists.txt, line 134 at r1](https://reviewable.io:443/reviews/valloric/ycmd/820#-KscfoZUHy49QQQx3xoy:-KscyOm_t0btsfC2sdmX:b-r5dgqg) ([raw file](https://github.com/valloric/ycmd/blob/1f645757815910266b8bd93b85ba813c28e3153f/cpp/ycm/CMakeLists.txt#L134)):*

It's not super obvious to me why we need `../..` here or why that's (canonically) correct. I'm sure it is (e.g. it represents `usr/lib` or something), but perhaps we need a comment?
</blockquote></details>

Added a comment to clarify the `/../..` part.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/820#-:-Ksd5jJO4OjRMjf6EQl2:b-9qdoyq)*

puremourning:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/820#-:-Ksd8OfbUx0FijzMdHmL:bnfp4nl)*

Valloric:
Thanks for the PR!

@zzbot r=bstaletic

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/820#-:-KsfJzFbq_UPTE6yMiks:b-g4o1ba)*

PR #826, opened 2017-09-02
[NOT READY] Hide C++ symbols in the ycm_core.so by default

bstaletic:
This PR makes GCC and Clang behave similarly to MSVC when it comes to exposing symbols inside a shared object. That means by default all symbols are hidden and only explicitly exposed symbols are visible.

Further more, making symbols visible is only done when the `ycm_core.so` shall be linked for either benchmarks or tests. Otherwise all symbols remain hidden. This still works for importing `ycm_core` inside python as we have `Boost.Python` taking care of that.

As a result the `ycm_core.so` is significantly smaller, making it easier (faster) to load and improves performance.

Logic that was in `cpp/ycm/DLLDefines.h` is now in `cpp/CMakeFiles.txt` because the value of `YCM_EXPORT` (formerly `YCM_DLL_EXPORT`) depends on wether `ycm_core.so` will be linked against or not.

`YCM_DLL_EXPORT` has been renamed to `YCM_EXPORT` because `*.dll` is a Windows thing.

Marked as WIP because I'd like to have everyone report that no performance regressions are found.

My benchmarks are [here](https://gist.github.com/21825e0121f68436b9999559f01e6f2c).
`ycm_core.so` size dropped from 2.8MB to 2.1MB, which is 25%.

@puremourning [posted his results](https://gist.github.com/puremourning/a754914cca4f48699a752f2dc75d384d).

Note: Windows performance and file size is unaffected in tests or benchmarks.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/826)


Valloric:
:lgtm:

Thanks for the PR!

---

Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/CMakeLists.txt, line 108 at r1](https://reviewable.io:443/reviews/valloric/ycmd/826#-Kt-NoKcQyj7tocE3978:-Kt-NoKcQyj7tocE3979:b-tuuc00) ([raw file](https://github.com/valloric/ycmd/blob/7df5487e4fbadbd3a6f54e29d709a267c47c6736/cpp/CMakeLists.txt#L108)):*
> ```CMake
> #############################################################################
> 
> # MSVC has symbols hidden by default. On GCC and Clang we need to explicitly
> ```

You should state in this comment _why_ we want to hide symbols. The PR description has it, but that's not readily available when reading the code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/826#-:-Kt-O1Nf54wOzrf8wTfX:b-19kv2l)*

davits:
*[cpp/CMakeLists.txt, line 253 at r2](https://reviewable.io:443/reviews/valloric/ycmd/826#-Kt0sCGAE5qtOvuC5rmA:-Kt0sCGAE5qtOvuC5rmB:b-qu1vt2) ([raw file](https://github.com/valloric/ycmd/blob/157dee7224255e5c2d4837654331616a82dded9f/cpp/CMakeLists.txt#L253)):*
> ```CMake
> 
> option( USE_SYSTEM_BOOST "Set to ON to use the system boost libraries" OFF )
> if ( DEFINED ENV{YCM_BENCHMARK} OR DEFINED ENV{YCM_TESTRUN})
> ```

Why is it set only for benchmark and tests? What about normal use-model.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/826)*

bstaletic:
Review status: 0 of 13 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[cpp/CMakeLists.txt, line 108 at r1](https://reviewable.io:443/reviews/valloric/ycmd/826#-Kt-NoKcQyj7tocE3978:-Kt0MMV5Zu7ZsD4Si8z7:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/7df5487e4fbadbd3a6f54e29d709a267c47c6736/cpp/CMakeLists.txt#L108)):*

You should state in this comment _why_ we want to hide symbols. The PR description has it, but that's not readily available when reading the code.
</blockquote></details>

Done.

---

*[cpp/CMakeLists.txt, line 253 at r2](https://reviewable.io:443/reviews/valloric/ycmd/826#-Kt0sCGAE5qtOvuC5rmA:-Kt0svWcRY90BN2B0awH:b3j224w) ([raw file](https://github.com/valloric/ycmd/blob/157dee7224255e5c2d4837654331616a82dded9f/cpp/CMakeLists.txt#L253)):*
<details><summary><i>Previously, davits (Davit Samvelyan) wrote…</i></summary><blockquote>

Why is it set only for benchmark and tests? What about normal use-model.
</blockquote></details>

The idea is that the normal use model doesn't need explicit exposing. This is because Boost forces "default" visibility for the module init. So even though it can't be tested normal use case should get the biggest improvements.

The normal use case should have `add_definitions( -DYCM_EXPORT )`, just like how it was defined before in `DLLDefines.h`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/826)*

davits:
*[cpp/CMakeLists.txt, line 253 at r2](https://reviewable.io:443/reviews/valloric/ycmd/826#-Kt0sCGAE5qtOvuC5rmA:-Kt11j3I6Nhab-hhwTCk:bmviwy0) ([raw file](https://github.com/valloric/ycmd/blob/157dee7224255e5c2d4837654331616a82dded9f/cpp/CMakeLists.txt#L253)):*

The idea is that the normal use model doesn't need explicit exposing. This is because Boost forces "default" visibility for the module init. So even though it can't be tested normal use case should get the biggest improvements.

The normal use case should have `add_definitions( -DYCM_EXPORT )`, just like how it was defined before in `DLLDefines.h`.
</blockquote></details>

Oh sorry, my thoughts were going into the direction of dlopen and dlsym, I forgot that we are registering our functions in the module init.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/826)*

PR #827, opened 2017-09-02
[READY] Downgrade boost

bstaletic:
The inaility to import `ycm_core` after including #826 is due to a boost bug. Namely boostorg/python#131. As a workaround we can downgrade boost to 1.63 until boost fixes the bug.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/827)


bstaletic:
Dropped the commit that ended up in the branch by mistake and caused the conflict.

PR #829, opened 2017-09-03
[READY] Run flake8 and nose as modules

micbou:
If a user has Python 2 and 3 installed on its system with flake8 and nose available in the PATH for both Pythons, `run_tests.py` may execute flake8 or nose from a different Python that the one used to run the script and build ycmd. For instance, if the script is executed in Python 3 but nose from Python 2 is found in the PATH, the following error occur when running the tests:
```
Failure: ImportError (dynamic module does not define init function (initycm_core)) ... ERROR
```
The solution is to run flake8 and nose as modules using `sys.executable` with [the `-m` option](https://docs.python.org/2.7/using/cmdline.html#cmdoption-m). Due to [a bug (feature?) in Python 2](http://bugs.python.org/issue2751), we need to specifically call the `__main__` function.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/829)

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -42,7 +42,9 @@ import argparse
 def RunFlake8():
   print( 'Running flake8' )
   subprocess.check_call( [
-    'flake8',
+    sys.executable,
+    # __main__ is required on Python 2.6.
+    '-m', 'flake8.__main__',
     p.join( DIR_OF_THIS_SCRIPT, 'ycmd' )
   ] )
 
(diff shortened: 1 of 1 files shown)


Valloric:
:lgtm:

Thanks!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/829#-:-Kt7X0aMM3Q0RJjrgUfu:b9benhj)*

vheon:
:lgtm:

@zzbot r=Valloric

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/829#-:-Kt7d028Pw-IfEK0_x7G:b-hgv6jp)*

PR #831, opened 2017-09-07
[READY] Fix multiline comments and strings issues

micbou:
When `collect_identifiers_from_comments_and_strings` is `0`, we are replacing comments and strings with empty strings before extracting the identifiers. If one of these comments or strings span multiple lines, the current line number `line_num` may become invalid and, in that case, the identifier completer will fail to add the previous identifier on the `CurrentIdentifierFinished` event:

ignore-comments-current-identifier-finished
As you can see, the `test` identifier is properly extracted before the comment but not after.

Another issue with multiline comments and strings is that, when adding an identifier under the cursor on the `InsertLeave` event, we only remove comments and strings on the current line. This is incorrect if the current line is in the middle of a multiline comment or string:

ignore-comments-insert-leave
The `test` identifier is extracted even though it's inside a comment.

For now, I am updating the tests to showcase both issues. I'll update the PR with the fixes once the builds failed.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/831)

### DIFF
--- a/ycmd/completers/all/identifier_completer.py
+++ b/ycmd/completers/all/identifier_completer.py
@@ -229,10 +229,13 @@ def _RemoveSmallCandidates( candidates, min_num_candidate_size_chars ):
 
 def _GetCursorIdentifier( collect_from_comments_and_strings,
                           request_data ):
-  line = request_data[ 'line_value' ]
+  filepath = request_data[ 'filepath' ]
+  contents = request_data[ 'file_data' ][ filepath ][ 'contents' ]
   filetype = request_data[ 'first_filetype' ]
   if not collect_from_comments_and_strings:
-    line = identifier_utils.RemoveIdentifierFreeText( line, filetype )
+    contents = identifier_utils.RemoveIdentifierFreeText( contents, filetype )
+  contents_per_line = SplitLines( contents )
+  line = contents_per_line[ request_data[ 'line_num' ] - 1 ]
   return identifier_utils.IdentifierAtIndex(
       line,
       request_data[ 'column_codepoint' ] - 1,
--- a/ycmd/identifier_utils.py
+++ b/ycmd/identifier_utils.py
@@ -25,6 +25,7 @@ from __future__ import absolute_import
 from builtins import *  # noqa
 
 import re
+from ycmd.utils import SplitLines
 
 C_STYLE_COMMENT = "/\*(?:\n|.)*?\*/"
 CPP_STYLE_COMMENT = "//.*?$"
@@ -170,8 +171,13 @@ def IdentifierRegexForFiletype( filetype ):
   return FILETYPE_TO_IDENTIFIER_REGEX.get( filetype, DEFAULT_IDENTIFIER_REGEX )
 
 
+def ReplaceWithEmptyLines( regex_match ):
+  return '\n' * ( len( SplitLines( regex_match.group( 0 ) ) ) - 1 )
+
+
 def RemoveIdentifierFreeText( text, filetype = None ):
-  return CommentAndStringRegexForFiletype( filetype ).sub( '', text )
+  return CommentAndStringRegexForFiletype( filetype ).sub(
+    ReplaceWithEmptyLines, text )
 
 
 def ExtractIdentifiersFromText( text, filetype = None ):
--- a/ycmd/tests/identifier_completer_test.py
+++ b/ycmd/tests/identifier_completer_test.py
@@ -92,11 +92,19 @@ def GetCursorIdentifier_LineEmpty_test():
 
 def GetCursorIdentifier_IgnoreIdentifierFromCommentsAndStrings_test():
   eq_( '', ic._GetCursorIdentifier( False, BuildRequestWrap( '"foobar"', 4 ) ) )
+  eq_( '', ic._GetCursorIdentifier( False,
+                                    BuildRequestWrap( '/*\n'
+                                                      ' * foobar\n'
+                                                      ' */', 5, 2 ) ) )
 
 
 def GetCursorIdentifier_CollectIdentifierFromCommentsAndStrings_test():
   eq_( 'foobar', ic._GetCursorIdentifier( True,
                                           BuildRequestWrap( '"foobar"', 4 ) ) )
+  eq_( 'foobar', ic._GetCursorIdentifier( True,
+                                          BuildRequestWrap( '/*\n'
+                                                            ' * foobar\n'
+                                                            ' */', 5, 2 ) ) )
 
 
 def PreviousIdentifier_Simple_test():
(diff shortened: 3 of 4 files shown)


micbou:
Fix the tests by replacing comments and strings with empty lines instead of empty strings. Also, replace them on the whole content and not just the current line when adding an identifier under the cursor.

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/831#-:-KtPDZAbky9ooPnHHGSE:bh9grni)*

puremourning:
Reviewed 1 of 4 files at r1, 1 of 1 files at r2.

---

*[ycmd/identifier_utils.py, line 179 at r2](https://reviewable.io:443/reviews/valloric/ycmd/831#-KtQHij4nbUtgD4UFKx4:-KtQHij4nbUtgD4UFKx5:b1kopec) ([raw file](https://github.com/valloric/ycmd/blob/84df3e54a24526c9c707fb4199186ea7a26c95b7/ycmd/identifier_utils.py#L179)):*
> ```Python
> 
> def RemoveIdentifierFreeText( text, filetype = None ):
>   return CommentAndStringRegexForFiletype( filetype ).sub(
> ```

What happens with something like this:

```
/* test */ realIdentifier
```

Does it kill realIdentifier?

Point is... does this over remove if the comment is not a whole-line comment?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/831)*

micbou:
Reviewed 1 of 1 files at r2.

---

*[ycmd/identifier_utils.py, line 179 at r2](https://reviewable.io:443/reviews/valloric/ycmd/831#-KtQHij4nbUtgD4UFKx4:-KtR4BCtWMPPaw7nsa1x:b-c89r8e) ([raw file](https://github.com/valloric/ycmd/blob/84df3e54a24526c9c707fb4199186ea7a26c95b7/ycmd/identifier_utils.py#L179)):*

What happens with something like this:

```
/* test */ realIdentifier
```

Does it kill realIdentifier?

Point is... does this over remove if the comment is not a whole-line comment?
</blockquote></details>

No, `/* test */` will be matched and `ReplaceWithEmptyLines` will return ``. See [the `RemoveIdentifierFreeText_CstyleComments` test](https://github.com/Valloric/ycmd/blob/dfcd0282df168cad364d62a49b455c688d4bcae4/ycmd/tests/identifier_utils_test.py#L43-L47).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/831)*

Valloric:
That's a fantastic RB description, thank you! Made it very easy to understand the problem.

---

---

*[ycmd/identifier_utils.py, line 174 at r2](https://reviewable.io:443/reviews/valloric/ycmd/831#-KtT5Y_NAGe6BmJkOTwd:-KtT5Y_NAGe6BmJkOTwe:bzgakzn) ([raw file](https://github.com/valloric/ycmd/blob/84df3e54a24526c9c707fb4199186ea7a26c95b7/ycmd/identifier_utils.py#L174)):*
> ```Python
> 
> 
> def ReplaceWithEmptyLines( match ):
> ```

Maybe `regex_match` to emphasize it's the match object from `re`?

---

*[ycmd/identifier_utils.py, line 179 at r2](https://reviewable.io:443/reviews/valloric/ycmd/831#-KtQHij4nbUtgD4UFKx4:-KtT5tsxdhn9ps_OE5RO:b-iyloot) ([raw file](https://github.com/valloric/ycmd/blob/84df3e54a24526c9c707fb4199186ea7a26c95b7/ycmd/identifier_utils.py#L179)):*

No, `/* test */` will be matched and `ReplaceWithEmptyLines` will return ``. See [the `RemoveIdentifierFreeText_CstyleComments` test](https://github.com/Valloric/ycmd/blob/dfcd0282df168cad364d62a49b455c688d4bcae4/ycmd/tests/identifier_utils_test.py#L43-L47).
</blockquote></details>

Let's make sure we have a test for that though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/831#-:-KtT5MsdOM7tdUOI4u7g:b-r8dk4r)*

micbou:
Reviewed 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.

---

*[ycmd/identifier_utils.py, line 174 at r2](https://reviewable.io:443/reviews/valloric/ycmd/831#-KtT5Y_NAGe6BmJkOTwd:-KtTAEJ12y2-M3wwccBN:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/84df3e54a24526c9c707fb4199186ea7a26c95b7/ycmd/identifier_utils.py#L174)):*

Maybe `regex_match` to emphasize it's the match object from `re`?
</blockquote></details>

Done.

---

*[ycmd/identifier_utils.py, line 179 at r2](https://reviewable.io:443/reviews/valloric/ycmd/831#-KtQHij4nbUtgD4UFKx4:-KtTA-5JYQDZBfuiaOiR:bwnavuy) ([raw file](https://github.com/valloric/ycmd/blob/84df3e54a24526c9c707fb4199186ea7a26c95b7/ycmd/identifier_utils.py#L179)):*

Let's make sure we have a test for that though.
</blockquote></details>

Added a test for this case (identifier after a C-style comment).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/831)*

puremourning:

---

*[ycmd/identifier_utils.py, line 179 at r2](https://reviewable.io:443/reviews/valloric/ycmd/831#-KtQHij4nbUtgD4UFKx4:-KtTPZSB_P15Ks3peGUU:b-gf69mu) ([raw file](https://github.com/valloric/ycmd/blob/84df3e54a24526c9c707fb4199186ea7a26c95b7/ycmd/identifier_utils.py#L179)):*

Added a test for this case (identifier after a C-style comment).
</blockquote></details>

Ah I think my mobile broke what I was trying to put. On reflecting on the code I'm pretty sure it is fine, but the example I meant was 

```
/*
 */ anIdentifier
```

Like i said, I'm pretty sure it works, 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/831)*

puremourning:
:lgtm: assuming th case I mentioned works as expected. Thanks!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/831#-:-KtTQrEJQtReCvmtf8TL:bfhkazc)*

micbou:
Reviewed 1 of 1 files at r6.

---

*[ycmd/identifier_utils.py, line 179 at r2](https://reviewable.io:443/reviews/valloric/ycmd/831#-KtQHij4nbUtgD4UFKx4:-KtTeca0kACqYSimlh9L:bijkiir) ([raw file](https://github.com/valloric/ycmd/blob/84df3e54a24526c9c707fb4199186ea7a26c95b7/ycmd/identifier_utils.py#L179)):*

Ah I think my mobile broke what I was trying to put. On reflecting on the code I'm pretty sure it is fine, but the example I meant was 

```
/*
 */ anIdentifier
```

Like i said, I'm pretty sure it works, 
</blockquote></details>

Updated the test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/831)*

Valloric:
:lgtm:

Thanks for the PR!

@zzbot r=puremourning

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/831#-:-KtZ-lie0LMD2UYFqTrz:b-uud7tw)*

PR #835, opened 2017-09-17
[READY] Do not return canonical type if identical to type

micbou:
The `GetType` subcommand from the Clang completer is supposed to return the canonical type of a symbol if it differs from its type. However, in some cases, both types are displayed even if identical. For instance, `GetType` on `bar` in the following code:
```c++
char *foo
auto bar = foo
```
returns `char * => char *` instead of simply `char *`.

This happens because Clang returns that the canonical type is different from the actual type even though they result in the same string. The solution is to only return the canonical type if the strings are different.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/835)

### DIFF
--- a/cpp/ycm/ClangCompleter/TranslationUnit.cpp
+++ b/cpp/ycm/ClangCompleter/TranslationUnit.cpp
@@ -291,9 +291,14 @@ std::string TranslationUnit::GetTypeAtLocation(
   CXType canonical_type = clang_getCanonicalType( type );
 
   if ( !clang_equalTypes( type, canonical_type ) ) {
-    type_description += " => ";
-    type_description += CXStringToString(
-                          clang_getTypeSpelling( canonical_type ) );
+    std::string canonical_type_description = CXStringToString(
+      clang_getTypeSpelling( canonical_type ) );
+
+    // Clang may return that the canonical type of a symbol is distinct from its
+    // type even though they result in the same string. Only append the
+    // canonical type if the strings are different.
+    if ( type_description != canonical_type_description )
+      type_description += " => " + canonical_type_description;
   }
 
   return type_description;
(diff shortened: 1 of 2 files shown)


puremourning:
Whoop. I've been meaning to do this for ages! Thanks! :lgtm: with minor comment. 

---

Reviewed 2 of 2 files at r1.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 303 at r1](https://reviewable.io:443/reviews/valloric/ycmd/835#-KuHKI7IGTK_Hx04EUyv:-KuHKI7JQqdWoBosNxDA:b2vfyvr) ([raw file](https://github.com/valloric/ycmd/blob/eae24f1e85c15a2da3ad50d22327ea835fb353bf/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L303)):*
> ```cpp
>       type_description += " => ";
>       type_description += CXStringToString(
>                             clang_getTypeSpelling( canonical_type ) );
> ```

We should skip doing this a second time and just append `canonical_type_description`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/835#-:-KuHKUCE9HJNXUwZFK42:b-o658z8)*

micbou:
Reviewed 2 of 2 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 303 at r1](https://reviewable.io:443/reviews/valloric/ycmd/835#-KuHKI7IGTK_Hx04EUyv:-KuHOw0uFwbnOToJs_aB:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/eae24f1e85c15a2da3ad50d22327ea835fb353bf/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L303)):*

We should skip doing this a second time and just append `canonical_type_description`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/835)*

bstaletic:
Nice one, thanks.

:lgtm: 
@zzbot r=puremourning

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/835#-:-KuIkyIIC-DnxlxqBxeb:b7989w0)*

PR #841, opened 2017-09-23
[READY] Close options file before starting server

micbou:
When running the shutdown tests on Windows, ycmd is unable to remove the options file that contains the HMAC key. This is because the file is kept open by the tests while the server is started and a file cannot be removed on Windows if another process is holding it open. We fix this by closing the options file as soon as possible.

Also, we don't need to manually flush the file as it is automatically flushed when closing it.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/841)

### DIFF
--- a/examples/example_client.py
+++ b/examples/example_client.py
@@ -21,6 +21,7 @@ if sys.version_info[ 0 ] < 3:
             platform.python_version() )
 
 from base64 import b64encode, b64decode
+from tempfile import NamedTemporaryFile
 import collections
 import hashlib
 import hmac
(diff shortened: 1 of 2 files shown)


bstaletic:
:lgtm:
Looks good.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/841#-:-Kuj-98boEjVXcEr48MS:bq9jdl8)*

Valloric:
:lgtm:

Thanks for the PR!

@zzbot r=bstaletic

PR #843, opened 2017-09-25
[READY] Improve completion of include statements in C-family languages

micbou:
This PR fixes the issue where completion is interrupted after inserting a non-identifier character in include statements. See issues https://github.com/Valloric/YouCompleteMe/issues/281 and https://github.com/Valloric/YouCompleteMe/issues/1553.

This is done by moving the include completion logic from the filename completer to the Clang one and by setting the start column to the right position thanks to PR https://github.com/Valloric/ycmd/pull/681. A benefit of moving the logic to the Clang completer is that `<C-Space>` now works in include statements.

Here's a demo before the changes:

include-statement-before

and after:

include-statement-after

You'll notice that no error is shown to the user after inserting the dot.

Fixes https://github.com/Valloric/YouCompleteMe/issues/281.
Fixes https://github.com/Valloric/YouCompleteMe/issues/1553.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/843)

### DIFF
--- a/ycmd/completers/completer_utils.py
+++ b/ycmd/completers/completer_utils.py
@@ -212,38 +212,6 @@ PREPARED_DEFAULT_FILETYPE_TRIGGERS = _FiletypeTriggerDictFromSpec(
     DEFAULT_FILETYPE_TRIGGERS )
 
 
-INCLUDE_REGEX = re.compile( '\s*#\s*(?:include|import)\s*("|<)' )
-
-
-def AtIncludeStatementStart( line ):
-  match = INCLUDE_REGEX.match( line )
-  if not match:
-    return False
-  # Check if the whole string matches the regex
-  return match.end() == len( line )
-
-
-def GetIncludeStatementValue( line, check_closing = True ):
-  """Returns include statement value and boolean indicating whether
-     include statement is quoted.
-     If check_closing is True the string is scanned for statement closing
-     character (" or >) and substring between opening and closing characters is
-     returned. The whole string after opening character is returned otherwise"""
-  match = INCLUDE_REGEX.match( line )
-  include_value = None
-  quoted_include = False
-  if match:
-    quoted_include = ( match.group( 1 ) == '"' )
-    if not check_closing:
-      include_value = line[ match.end(): ]
-    else:
-      close_char = '"' if quoted_include else '>'
-      close_char_pos = line.find( close_char, match.end() )
-      if close_char_pos != -1:
-        include_value = line[ match.end() : close_char_pos ]
-  return include_value, quoted_include
-
-
 def GetFileContents( request_data, filename ):
   """Returns the contents of the absolute path |filename| as a unicode
   string. If the file contents exist in |request_data| (i.e. it is open and
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -25,6 +25,7 @@ from builtins import *  # noqa
 
 from collections import defaultdict
 from future.utils import iteritems
+import logging
 import os.path
 import re
 import textwrap
@@ -34,10 +35,12 @@ import ycm_core
 from ycmd import responses
 from ycmd.utils import ToCppStringCompatible, ToUnicode
 from ycmd.completers.completer import Completer
-from ycmd.completers.completer_utils import GetIncludeStatementValue
 from ycmd.completers.cpp.flags import ( Flags, PrepareFlagsForClang,
-                                        NoCompilationDatabase )
+                                        NoCompilationDatabase,
+                                        UserIncludePaths )
 from ycmd.completers.cpp.ephemeral_values_set import EphemeralValuesSet
+from ycmd.completers.general.filename_completer import (
+    GenerateCandidatesForPaths )
 from ycmd.responses import NoExtraConfDetected, UnknownExtraConf
 
 CLANG_FILETYPES = set( [ 'c', 'cpp', 'objc', 'objcpp' ] )
@@ -49,6 +52,7 @@ NO_DIAGNOSTIC_MESSAGE = 'No diagnostic for current line!'
 PRAGMA_DIAG_TEXT_TO_IGNORE = '#pragma once in main file'
 TOO_MANY_ERRORS_DIAG_TEXT_TO_IGNORE = 'too many errors emitted, stopping now'
 NO_DOCUMENTATION_MESSAGE = 'No documentation available for current context'
+INCLUDE_REGEX = re.compile( '(\s*#\s*(?:include|import)\s*)(:?"[^"]*|<[^>]*)' )
 
 
 class ClangCompleter( Completer ):
@@ -60,6 +64,7 @@ class ClangCompleter( Completer ):
     self._flags = Flags()
     self._diagnostic_store = None
     self._files_being_compiled = EphemeralValuesSet()
+    self._logger = logging.getLogger( __name__ )
 
 
   def SupportedFiletypes( self ):
@@ -85,10 +90,61 @@ class ClangCompleter( Completer ):
     return files
 
 
+  def ShouldCompleteIncludeStatement( self, request_data ):
+    column_codepoint = request_data[ 'column_codepoint' ] - 1
+    current_line = request_data[ 'line_value' ]
+    return INCLUDE_REGEX.match( current_line[ : column_codepoint ] )
+
+
+  def ShouldUseNowInner( self, request_data ):
+    if self.ShouldCompleteIncludeStatement( request_data ):
+      return True
+    return super( ClangCompleter, self ).ShouldUseNowInner( request_data )
+
+
+  def GetIncludePaths( self, request_data ):
+    column_codepoint = request_data[ 'column_codepoint' ] - 1
+    current_line = request_data[ 'line_value' ]
+    line = current_line[ : column_codepoint ]
+    path_dir, quoted_include, start_codepoint = (
+        GetIncludeStatementValue( line, check_closing = False ) )
+    if start_codepoint is None:
+      return None
+
+    request_data[ 'start_codepoint' ] = start_codepoint
+
+    # We do what GCC does for <> versus "":
+    # http://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html
+    flags = self._FlagsForRequest( request_data )
+    filepath = request_data[ 'filepath' ]
+    quoted_include_paths, include_paths = UserIncludePaths( flags, filepath )
+    if quoted_include:
+      include_paths.extend( quoted_include_paths )
+
+    paths = []
+    for include_path in include_paths:
+      unicode_path = ToUnicode( os.path.join( include_path, path_dir ) )
+      try:
+        # We need to pass a unicode string to get unicode strings out of
+        # listdir.
+        relative_paths = os.listdir( unicode_path )
+      except Exception:
+        self._logger.exception( 'Error while listing %s folder.', unicode_path )
+        relative_paths = []
+
+      paths.extend( os.path.join( include_path, path_dir, relative_path ) for
+                    relative_path in relative_paths  )
+    return paths
+
+
   def ComputeCandidatesInner( self, request_data ):
     filename = request_data[ 'filepath' ]
     if not filename:
       return
+
+    paths = self.GetIncludePaths( request_data )
+    if paths is not None:
+      return GenerateCandidatesForPaths( paths )
 
     if self._completer.UpdatingTranslationUnit(
         ToCppStringCompatible( filename ) ):
@@ -220,18 +276,19 @@ class ClangCompleter( Completer ):
 
   def _ResponseForInclude( self, request_data ):
     """Returns response for include file location if cursor is on the
-       include statement, None otherwise.
-       Throws RuntimeError if cursor is on include statement and corresponding
-       include file not found."""
+    include statement, None otherwise.
+    Throws RuntimeError if cursor is on include statement and corresponding
+    include file not found."""
     current_line = request_data[ 'line_value' ]
-    include_file_name, quoted_include = GetIncludeStatementValue( current_line )
+    include_file_name, quoted_include, _ = GetIncludeStatementValue(
+        current_line )
     if not include_file_name:
       return None
 
+    flags = self._FlagsForRequest( request_data )
     current_file_path = request_data[ 'filepath' ]
-    client_data = request_data.get( 'extra_conf_data', None )
-    quoted_include_paths, include_paths = (
-            self._flags.UserIncludePaths( current_file_path, client_data ) )
+    quoted_include_paths, include_paths = UserIncludePaths( flags,
+                                                            current_file_path )
     if quoted_include:
       include_file_path = _GetAbsolutePath( include_file_name,
                                             quoted_include_paths )
@@ -429,10 +486,6 @@ def ClangAvailableForFiletypes( filetypes ):
   return any( [ filetype in CLANG_FILETYPES for filetype in filetypes ] )
 
 
-def InCFamilyFile( filetypes ):
-  return ClangAvailableForFiletypes( filetypes )
-
-
 def _FilterDiagnostics( diagnostics ):
   # Clang has an annoying warning that shows up when we try to compile header
   # files if the header has "#pragma once" inside it. The error is not
@@ -523,3 +576,40 @@ def _GetAbsolutePath( include_file_name, include_paths ):
     if os.path.isfile( include_file_path ):
       return include_file_path
   return None
+
+
+def GetIncludeStatementValue( line, check_closing = True ):
+  """Returns the tuple |include_value|, |quoted_include|, |start_codepoint|
+  where:
+  - |include_value| is the string from the opening quote or bracket of the
+    include statement in |line|. If |check_closing| is True, the whole string
+    between the quotes or brackets is returned. None if no include statement is
+    found or if |check_closing| is True and there is no closing quote or
+    bracket;
+  - |quoted_include| is True if the statement is a quoted include, False
+    otherwise;
+  - |start_column| is the 1-based column where the completion should start (i.e.
+    at the last path separator '/' or at the opening quote or bracket). None if
+    no include statement is matched or |check_closing| is True."""
+  match = INCLUDE_REGEX.match( line )
+  include_value = None
+  quoted_include = False
+  start_codepoint = None
+  if match:
+    include_start = match.end( 1 ) + 1
+    quoted_include = ( line[ include_start - 1 ] == '"' )
+    if not check_closing:
+      separator_char = '/'
+      separator_char_pos = line.rfind( separator_char, match.end( 1 ) )
+      if separator_char_pos != -1:
+        include_value = line[ include_start : separator_char_pos + 1 ]
+        start_codepoint = separator_char_pos + 2
+      else:
+        include_value = ''
+        start_codepoint = include_start + 1
+    else:
+      close_char = '"' if quoted_include else '>'
+      close_char_pos = line.find( close_char, match.end() )
+      if close_char_pos != -1:
+        include_value = line[ include_start : close_char_pos ]
+  return include_value, quoted_include, start_codepoint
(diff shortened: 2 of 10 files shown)


micbou:
Had to move the code adding the macOS include paths outside of `PrepareFlagsForClang` because it was screwing the tests.

---

Reviewed 7 of 7 files at r1, 2 of 2 files at r2, 3 of 3 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843#-:-KutHhgYf3vJDFa6gnHi:byf0v2z)*

micbou:
Added a filename completer test to fix the coverage decrease.

---

Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843#-:-KuuF556EFfVkP7qbTxw:b-z235x7)*

puremourning:
Thanks! It’s really great to fix this. Will make a lot of people happy :)

Gave it a really light review. Will review fully ASAP. Generally approach looks good. 

---

Review status: 7 of 9 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cpp/clang_completer.py, line 107 at r5](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvUhhrVzBTrZTXkBoi:-KuvUhhrVzBTrZTXkBoj:b-8j2748) ([raw file](https://github.com/valloric/ycmd/blob/c9e27d591b5223f70231bcf1954cc6520601d4a1/ycmd/completers/cpp/clang_completer.py#L107)):*
> ```Python
>   def GetIncludePaths( self, request_data ):
>     current_line = request_data[ 'line_value' ]
>     column_num = request_data[ 'column_num' ]
> ```

I realise this might be moved code, but Shouldn’t this be a column_codepoint? current_line is Unicode not bytes, so we’re indexing a Unicode string with a byte offset, which I think is wrong. 

---

*[ycmd/completers/cpp/clang_completer.py, line 607 at r5](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvWYjysoV1sCmmP5G2:-KuvWYjysoV1sCmmP5G3:b-mki4fz) ([raw file](https://github.com/valloric/ycmd/blob/c9e27d591b5223f70231bcf1954cc6520601d4a1/ycmd/completers/cpp/clang_completer.py#L607)):*
> ```Python
>       close_char_pos = line.find( close_char, match.end() )
>       if close_char_pos != -1:
>         include_value = line[ include_start : close_char_pos ]
> ```

I’m this branch we set `include_value` but not `start_codepoint`. The calling code assumes that start codepoint is set if include value is set. 

---

*[ycmd/completers/general/filename_completer.py, line 106 at r5](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvXLmNrTZ5OdqOIlBT:-KuvXLmNrTZ5OdqOIlBU:bsxpewa) ([raw file](https://github.com/valloric/ycmd/blob/c9e27d591b5223f70231bcf1954cc6520601d4a1/ycmd/completers/general/filename_completer.py#L106)):*
> ```Python
> 
>     return GenerateCandidatesForPaths(
>       _GetPathsStandardCase(
> ```

Optional: I guess we could rename his now. There’s only one case now (I think)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843#-:-KuvXttsCgb36hwCnWg3:b-k4htrs)*

puremourning:
Review status: 7 of 9 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvYFHyXIA5q6-P9qnW:-KuvYFHzeo628HvConAb:bnj17ka) (no related file):*
I didn’t really follow why the mac flags thing had to change for this. I seem to recall there was a reason it was where it was, but I confess I don’t remember and if I didn’t comment it well enough then that’s on me!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843)*

micbou:
Reviewed 2 of 2 files at r5, 2 of 2 files at r6.

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvYFHyXIA5q6-P9qnW:-KuxQ5ATNMuAZ3HPO592:b-hrzzr7) (no related file):*

I didn’t really follow why the mac flags thing had to change for this. I seem to recall there was a reason it was where it was, but I confess I don’t remember and if I didn’t comment it well enough then that’s on me!
</blockquote></details>

Adding the mac include paths makes it impossible to check for the exact list of expected include paths in the tests because of all the system headers. I changed it so that when we specify the flags through the `compilation_flags` field in the request (which we often do in the tests), we don't add those include paths.

---

*[ycmd/completers/cpp/clang_completer.py, line 107 at r5](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvUhhrVzBTrZTXkBoi:-KuxRiAmrA0m582DnmQV:bq7h4ng) ([raw file](https://github.com/valloric/ycmd/blob/c9e27d591b5223f70231bcf1954cc6520601d4a1/ycmd/completers/cpp/clang_completer.py#L107)):*

I realise this might be moved code, but Shouldn’t this be a column_codepoint? current_line is Unicode not bytes, so we’re indexing a Unicode string with a byte offset, which I think is wrong. 
</blockquote></details>

Mistake on my part. This is new code. Fixed.

---

*[ycmd/completers/cpp/clang_completer.py, line 607 at r5](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvWYjysoV1sCmmP5G2:-KuxQxfDRo3uqTmznNSu:b-f0k7jz) ([raw file](https://github.com/valloric/ycmd/blob/c9e27d591b5223f70231bcf1954cc6520601d4a1/ycmd/completers/cpp/clang_completer.py#L607)):*

I’m this branch we set `include_value` but not `start_codepoint`. The calling code assumes that start codepoint is set if include value is set. 
</blockquote></details>

The calling code now checks that `start_codepoint` is set. Also, `start_codepoint` is now 1-based.

---

*[ycmd/completers/general/filename_completer.py, line 106 at r5](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvXLmNrTZ5OdqOIlBT:-KuxQpYsoSVIDMsrBhWs:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/c9e27d591b5223f70231bcf1954cc6520601d4a1/ycmd/completers/general/filename_completer.py#L106)):*

Optional: I guess we could rename his now. There’s only one case now (I think)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843)*

bstaletic:
I gave this a quick test, but I didn't have time to review it properly.

The issue I have noticed is that include path completion inside angled brackets doesn't seem to work any more. I just get no suggestions.

---

Reviewed 1 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843#-:-KuxvioL66Uu3anx0Izc:bcmtb2s)*

bstaletic:
Sorry for the hasty comment, the problem was caused by bad extra conf from when I tried to reproduce the gcc-7.2 bug.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843#-:-KuxyUoWqz3KPICynu7q:b-vhbro3)*

bstaletic:
Reviewed 2 of 7 files at r1, 1 of 3 files at r3, 1 of 1 files at r4, 2 of 2 files at r5, 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks broke.

---

*[ycmd/completers/cpp/clang_completer.py, line 607 at r5](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvWYjysoV1sCmmP5G2:-KuzA-NTcYa3mc7yRih7:b-rro8d7) ([raw file](https://github.com/valloric/ycmd/blob/c9e27d591b5223f70231bcf1954cc6520601d4a1/ycmd/completers/cpp/clang_completer.py#L607)):*

The calling code now checks that `start_codepoint` is set. Also, `start_codepoint` is now 1-based.
</blockquote></details>

I think we should either set `start_codepoint` in this branch or have a comment that would explain why is it not set.

---

*[ycmd/completers/cpp/clang_completer.py, line 582 at r6](https://reviewable.io:443/reviews/valloric/ycmd/843#-Kuz9M98pPRD4L3ouf97:-Kuz9M99oBYrOp4oJVw0:b-q1i8xq) ([raw file](https://github.com/valloric/ycmd/blob/82220efd38601dbb0939b6a1fec2152ff82dd55b/ycmd/completers/cpp/clang_completer.py#L582)):*
> ```Python
> 
> def GetIncludeStatementValue( line, check_closing = True ):
>   """Returns include statement value and boolean indicating whether include
> ```

This comment suggest only two values are returned, but the function returns thee.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843)*

micbou:
Reviewed 1 of 1 files at r7, 1 of 1 files at r8.

---

*[ycmd/completers/cpp/clang_completer.py, line 607 at r5](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvWYjysoV1sCmmP5G2:-KuzsD73fkcYmKlVRJ6U:bsal4j8) ([raw file](https://github.com/valloric/ycmd/blob/c9e27d591b5223f70231bcf1954cc6520601d4a1/ycmd/completers/cpp/clang_completer.py#L607)):*

I think we should either set `start_codepoint` in this branch or have a comment that would explain why is it not set.
</blockquote></details>

It's mentioned in the new docstring.

---

*[ycmd/completers/cpp/clang_completer.py, line 582 at r6](https://reviewable.io:443/reviews/valloric/ycmd/843#-Kuz9M98pPRD4L3ouf97:-Kuzq82Wmkwfbl1CAFCq:byew2g6) ([raw file](https://github.com/valloric/ycmd/blob/82220efd38601dbb0939b6a1fec2152ff82dd55b/ycmd/completers/cpp/clang_completer.py#L582)):*

This comment suggest only two values are returned, but the function returns thee.
</blockquote></details>

I've rewritten the docstring. I think it's now clearer what the function does.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843)*

bstaletic:
Reviewed 1 of 1 files at r9.

---

*[ycmd/completers/cpp/clang_completer.py, line 582 at r6](https://reviewable.io:443/reviews/valloric/ycmd/843#-Kuz9M98pPRD4L3ouf97:-Kv-7hZuYncVD_IVZVSd:by0a2ji) ([raw file](https://github.com/valloric/ycmd/blob/82220efd38601dbb0939b6a1fec2152ff82dd55b/ycmd/completers/cpp/clang_completer.py#L582)):*

I've rewritten the docstring. I think it's now clearer what the function does.
</blockquote></details>

It's much better now. Thanks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843)*

Valloric:
This is pretty damn sweet! Thank you!

I gave this a light review; looks good, though at-a-glance it appears like it might benefit from some more tests. I saw lots of new non-trivial code but not quite as many tests as I'd expect. Let's try to hit really solid test coverage on the new code. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843#-:-Kv4e8KNyq8DV1DeiChy:bt1eah0)*

micbou:
Added 7 new tests for a total of 14 tests. In particular, separate tests for the `-I`, `-isystem`, and `-iquote` flags and for directories containing spaces.

---

Reviewed 1 of 1 files at r9, 3 of 3 files at r10.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843#-:-Kv8dsoPwEdwyDXOip03:b-kdvekb)*

bstaletic:
:lgtm:
This has good enough coverage in my opinion.

What is not covered is exception throwing in [`clang_completer.py`](https://codecov.io/gh/Valloric/ycmd/compare/aeeee206ce0e8a3464afa941997b28b25ac83f48...5d1cc7f9dc9e3a4e4ec97cf87b4c3fa24c6e46e5/diff#D1-131...133) as well as in [`flags.py`](https://codecov.io/gh/Valloric/ycmd/compare/aeeee206ce0e8a3464afa941997b28b25ac83f48...5d1cc7f9dc9e3a4e4ec97cf87b4c3fa24c6e46e5/diff#D6-615...616). I'm fine with both.

---

Reviewed 3 of 3 files at r10.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843#-:-Kvbef1EKNjXXLUKZJiV:b3ya42d)*

Valloric:

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvYFHyXIA5q6-P9qnW:-Kvclg8ZYFuePrL04I4j:b1ql4jn) (no related file):*

Adding the mac include paths makes it impossible to check for the exact list of expected include paths in the tests because of all the system headers. I changed it so that when we specify the flags through the `compilation_flags` field in the request (which we often do in the tests), we don't add those include paths.
</blockquote></details>

:lgtm: , thanks!

Not triggering zzbot because I see @puremourning did a review pass but hasn't provided an LGTM yet.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843)*

puremourning:
LGTM

---

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvYFHyXIA5q6-P9qnW:-Kv-eylPV2vFUn0oQFQP:bp1l1yr) (no related file):*

Adding the mac include paths makes it impossible to check for the exact list of expected include paths in the tests because of all the system headers. I changed it so that when we specify the flags through the `compilation_flags` field in the request (which we often do in the tests), we don't add those include paths.
</blockquote></details>

Ah ok that seems reasonable. I doubt it is a regression for YCM, but we perhaps should check that no other clients use his part of the API

---

*[ycmd/completers/cpp/clang_completer.py, line 607 at r5](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvWYjysoV1sCmmP5G2:-KviXKwXtl_lefHadWKR:bfwkzqj) ([raw file](https://github.com/valloric/ycmd/blob/c9e27d591b5223f70231bcf1954cc6520601d4a1/ycmd/completers/cpp/clang_completer.py#L607)):*

It's mentioned in the new docstring.
</blockquote></details>

I do feel that the API for this method is kinda wonky. Maybe we should just split it into two functions instead of having an argument that sort of fundamentally changes the outputs. Anyway, not really urgent/important and not a blocker.

---

*[ycmd/tests/clang/flags_test.py, line 445 at r10](https://reviewable.io:443/reviews/valloric/ycmd/843#-Kvi_WZmaTnc_UwRHbg-:-Kvi_WZmaTnc_UwRHbg0:b-x1fz1y) ([raw file](https://github.com/valloric/ycmd/blob/5d1cc7f9dc9e3a4e4ec97cf87b4c3fa24c6e46e5/ycmd/tests/clang/flags_test.py#L445)):*
> ```Python
> def AddMacIncludePaths_NoSysroot_test():
>   assert_that(
>     flags._AddMacIncludePaths( [ '-test', '--test1', '--test2=test' ] ),
> ```

I'm less convinced these tests are really testing that we really add mac flags anymore. I mean yes, we're testing that the function that add's mac include paths actually adds a path, but we're not really testing that the flags class ever actually calls that... unless we're proving that somewhere else ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843#-:-Kvi_rrKjGNmHe1XNxyd:bb6ywmx)*

micbou:
Reviewed 2 of 2 files at r11.

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvYFHyXIA5q6-P9qnW:-KvqEmWesFa2zpSgvxhq:b-jyxhvb) (no related file):*

Ah ok that seems reasonable. I doubt it is a regression for YCM, but we perhaps should check that no other clients use his part of the API
</blockquote></details>

I don't think any clients use that API because it's not documented and relying on it means that the `-resource-dir` flag is not added to the list of flags so it won't work properly.

---

*[ycmd/completers/cpp/clang_completer.py, line 607 at r5](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvWYjysoV1sCmmP5G2:-KvqFZ6SfG5Y3ELm6jRh:bq14yyi) ([raw file](https://github.com/valloric/ycmd/blob/c9e27d591b5223f70231bcf1954cc6520601d4a1/ycmd/completers/cpp/clang_completer.py#L607)):*

I do feel that the API for this method is kinda wonky. Maybe we should just split it into two functions instead of having an argument that sort of fundamentally changes the outputs. Anyway, not really urgent/important and not a blocker.
</blockquote></details>

Split into two functions: `GetIncompleteIncludeValue` and `GetFullIncludeValue`.

---

*[ycmd/tests/clang/flags_test.py, line 445 at r10](https://reviewable.io:443/reviews/valloric/ycmd/843#-Kvi_WZmaTnc_UwRHbg-:-KvqXXAr1WFm2xTouggO:bf4j1b2) ([raw file](https://github.com/valloric/ycmd/blob/5d1cc7f9dc9e3a4e4ec97cf87b4c3fa24c6e46e5/ycmd/tests/clang/flags_test.py#L445)):*

I'm less convinced these tests are really testing that we really add mac flags anymore. I mean yes, we're testing that the function that add's mac include paths actually adds a path, but we're not really testing that the flags class ever actually calls that... unless we're proving that somewhere else ?
</blockquote></details>

Rewritten as `FlagsForFile` tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843)*

puremourning:
Reviewed 2 of 7 files at r1, 1 of 1 files at r4, 1 of 2 files at r5, 1 of 2 files at r6, 3 of 3 files at r10, 1 of 2 files at r11, 1 of 1 files at r12.

---

*[a discussion](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvYFHyXIA5q6-P9qnW:-Kvr5CDr12cf2yd04Izj:bnfp4nl) (no related file):*

I don't think any clients use that API because it's not documented and relying on it means that the `-resource-dir` flag is not added to the list of flags so it won't work properly.
</blockquote></details>

:lgtm:

---

*[ycmd/completers/cpp/clang_completer.py, line 607 at r5](https://reviewable.io:443/reviews/valloric/ycmd/843#-KuvWYjysoV1sCmmP5G2:-Kvr4hbCQZWphUKuZx3B:bmpcat2) ([raw file](https://github.com/valloric/ycmd/blob/c9e27d591b5223f70231bcf1954cc6520601d4a1/ycmd/completers/cpp/clang_completer.py#L607)):*

Split into two functions: `GetIncompleteIncludeValue` and `GetFullIncludeValue`.
</blockquote></details>

Yeah! I like this a lot more, thanks so much.

---

*[ycmd/tests/clang/flags_test.py, line 445 at r10](https://reviewable.io:443/reviews/valloric/ycmd/843#-Kvi_WZmaTnc_UwRHbg-:-Kvr5ASLIlEZvqqDBqvM:b-eugxf8) ([raw file](https://github.com/valloric/ycmd/blob/5d1cc7f9dc9e3a4e4ec97cf87b4c3fa24c6e46e5/ycmd/tests/clang/flags_test.py#L445)):*

Rewritten as `FlagsForFile` tests.
</blockquote></details>

Nice! awesome.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843)*

micbou:
Added a `GoToInclude` subcommand test to fully cover the `GetFullIncludeValue` function.

---

Reviewed 1 of 1 files at r12, 1 of 1 files at r13.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843#-:-KvtGb7K38jDYSDLhjXP:b-86jqdl)*

micbou:
I think it's time to merge. Thanks for the reviews.

@zzbot r+

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/843#-:-Kw0ZVtujhmpUL4Nk05F:b6qle2j)*

PR #848, opened 2017-09-28
[READY] Hide C++ symbols by default

bstaletic:
I accidentally closed #826 (force push gone wrong and confused github).
This PR implements the same change but also #827 at the same time.

The boost downgrade is for this change to work properly because of [a bug in the new boost that disallows hidden C++ symbols (i.e. Boost doesn't force the ycm_core to be visible).

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/848)


micbou:
I think it's overkill to create a header file for one definition. This is what I did for `YCM_DLL_EXPORT` and it was a mistake. It would be more reasonable to define `YCM_EXPORT` in `Utils.h`.

---

Reviewed 14 of 14 files at r1.
Review status: 13 of 369 files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/CMakeLists.txt, line 266 at r1](https://reviewable.io:443/reviews/valloric/ycmd/848#-Kv6SXgTu_9FO3Ro_SNn:-Kv6SXgUOfbVSH6-xgb0:btm1uhq) ([raw file](https://github.com/valloric/ycmd/blob/a76bb221e6ff2c299b55d6e02baa385e946c3e29/cpp/CMakeLists.txt#L266)):*
> ```CMake
>       -DYCM_EXPORT=__attribute__\(\(visibility\(\"default\"\)\)\) )
>   endif()
> endif()
> ```

This block should be with the one that sets the `-fvisibility=hidden` flag.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/848#-:-Kv6T5yePsBQ1aOVeSL8:b-6qu9iq)*

bstaletic:
I moved the definition to `Utils.h`.

---

Review status: 2 of 370 files reviewed at latest revision, 1 unresolved discussion.

---

*[cpp/CMakeLists.txt, line 266 at r1](https://reviewable.io:443/reviews/valloric/ycmd/848#-Kv6SXgTu_9FO3Ro_SNn:-Kv6aENLH6FcwbEelD0H:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/a76bb221e6ff2c299b55d6e02baa385e946c3e29/cpp/CMakeLists.txt#L266)):*

This block should be with the one that sets the `-fvisibility=hidden` flag.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/848#-:-Kv6aGX5afIQHoKD4gLh:brz4yh5)*

bstaletic:
The latest commit updates `cpp/ycm/.ycm_extra_conf.py` to reflect this PR's change.

Is it just me or is codecov reporting nonsense?

Valloric:
:lgtm:

Thanks!

---

Review status: 6 of 370 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/848#-:-KvckF8nAVYSD2-6A1d-:b9benhj)*

bstaletic:
I have just force-pushed the update that should make this work with the latest boost.
But I'd advise to wait for the bots to say all the tests have passed.

micbou:
Some minor comments but :lgtm:.

---

Reviewed 1 of 14 files at r1, 8 of 8 files at r2, 1 of 356 files at r4.

---

*[cpp/CMakeLists.txt, line 148 at r3](https://reviewable.io:443/reviews/valloric/ycmd/848#-Kv8gDwcpZ59_69OoMEq:-Kv8gDwcpZ59_69OoMEr:b-6x10y9) ([raw file](https://github.com/valloric/ycmd/blob/48528c93b54ed0f3cb914dcd9dc163c7e52e0c18/cpp/CMakeLists.txt#L148)):*
> ```CMake
> else()
>   add_definitions( -DYCM_EXPORT= )
> endif()
> ```

Missing blank line.

---

*[cpp/ycm/exceptions.h, line 30 at r4](https://reviewable.io:443/reviews/valloric/ycmd/848#-L11roJUpsbFMq_wOggA:-L11roJUpsbFMq_wOggB:bvon34o) ([raw file](https://github.com/valloric/ycmd/blob/9c3b4802e955ce10e4d70eae2bcc17aa2d080909/cpp/ycm/exceptions.h#L30)):*
> ```objc
>  * Thrown when a file does not exist.
>  */
> struct YCM_EXPORT ClangParseError : virtual std::exception {};
> ```

Is there a reason to put `YCM_EXPORT` after `struct` and not before like we do for classes?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/848#-:-L11sOu-vSrDUS23XSGu:b-fh1flj)*

bstaletic:
Review status: 13 of 14 files reviewed at latest revision, 2 unresolved discussions.

---

*[cpp/CMakeLists.txt, line 148 at r3](https://reviewable.io:443/reviews/valloric/ycmd/848#-Kv8gDwcpZ59_69OoMEq:-L11u5YwO6kowdMokQDA:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/48528c93b54ed0f3cb914dcd9dc163c7e52e0c18/cpp/CMakeLists.txt#L148)):*

Missing blank line.
</blockquote></details>

Done.

---

*[cpp/ycm/exceptions.h, line 30 at r4](https://reviewable.io:443/reviews/valloric/ycmd/848#-L11roJUpsbFMq_wOggA:-L11sndQ4utr5GJlPFEF:b10yk3a) ([raw file](https://github.com/valloric/ycmd/blob/9c3b4802e955ce10e4d70eae2bcc17aa2d080909/cpp/ycm/exceptions.h#L30)):*

Is there a reason to put `YCM_EXPORT` after `struct` and not before like we do for classes?
</blockquote></details>

There actually is, and you've found the reason last time.

If we do it the logical way compilers\* will ignore it for whatever reason. And if it gets ignored, the exception is simply not thrown when calling `throw`.

\* I'm certain GCC will ignore it, not sure about clang.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/848)*

micbou:
Let's merge.

@zzbot r+

---

Reviewed 1 of 1 files at r5.

---

*[cpp/ycm/exceptions.h, line 30 at r4](https://reviewable.io:443/reviews/valloric/ycmd/848#-L11roJUpsbFMq_wOggA:-L12-92c7F7pzKt1GD_x:bm4xnc4) ([raw file](https://github.com/valloric/ycmd/blob/9c3b4802e955ce10e4d70eae2bcc17aa2d080909/cpp/ycm/exceptions.h#L30)):*

There actually is, and you've found the reason last time.

If we do it the logical way compilers\* will ignore it for whatever reason. And if it gets ignored, the exception is simply not thrown when calling `throw`.

\* I'm certain GCC will ignore it, not sure about clang.
</blockquote></details>

Right. I vaguely remember now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/848#-:-L121DS745r4ZmghIqeA:b-ptzv6q)*

PR #851, opened 2017-10-02
[READY] Rename completer options

micbou:
It's more intuitive for users if completer options are named after the language and not after the semantic engine e.g. `--rust-completer` instead of `--racer-completer`. To preserve backward compatibility, we keep the old options but deprecate them by hiding them from the `--help` output. We don't rename `--clang-completer` since `clang` is really well known and an appropriate name to designate C-family languages.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/851)

### DIFF
--- a/build.py
+++ b/build.py
@@ -254,25 +254,24 @@ def GetGenerator( args ):
 def ParseArguments():
   parser = argparse.ArgumentParser()
   parser.add_argument( '--clang-completer', action = 'store_true',
-                       help = 'Build C-family semantic completion engine.' )
-  parser.add_argument( '--system-libclang', action = 'store_true',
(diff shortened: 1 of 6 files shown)


bstaletic:
This looks good. :lgtm: 

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/851#-:-KvXB3K6UpFwwOh3bl-n:b-onofyc)*

Valloric:
Heh, this has been bothering me for literally _years_, I just never found it pressing enough to fix. :D
:lgtm:

Thanks for the PR!

@zzbot r=bstaletic

PR #853, opened 2017-10-06
[RFC] Add system header directories to user flags

### LINKED ISSUE #844: Xcode 9 Support
Xcode 9 modifies system header include paths. ycmd/completers/cpp/flags.py no longer produces a proper system include path sequence on systems that have upgraded to/installed Xcode 9.

On my system, `clang -v -E -x c++ -` produces the following:
```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1,
/usr/local/include,
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/9.0.0/include,
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include,
/usr/include,
/System/Library/Frameworks,
/Library/Frameworks
```

micbou:
ycmd uses its own engine to complete and jump in include statements since libclang does not return completions in that scenario and [the `clang_getIncludedFile` function](https://clang.llvm.org/doxygen/group__CINDEX__CURSOR__MANIP.html#gaf61979977343e39f21d6ea0b22167514) behaves in a strange way (it returns nothing after the opening quote or angle bracket). This is done by looking at the include flags from the list of user flags. The issue is that system includes are generally not part of these flags and thus completion and jumping are not available for system headers. A workaround is to call Clang as follows:
```
clang -x c++ -E -v -
```
and copy the paths under the `#include <...> search starts here:` line to the list of flags. Unfortunately, this can't be done programmatically because Clang may not be available on the user system. But what if we provide our own executable that behaves like Clang when passing the `-E` and `-v` flags?

This is what the PR essentially does. A small program called  `ycm_fake_clang` (name subject to change) is built along the YCM library. This program takes a list of flags and a filename as its arguments and create the corresponding translation unit. When retrieving the user flags, ycmd calls it like this:
```
ycm_fake_clang -E -v [flag ...] ycm_dummy_file[.ext]
```
where `[flag ...]` is a list of flags relevant to the system includes used to parse the translation unit and `[.ext]` is the extension of the current file (libclang deduces the language from the file extension when the `-x` flag is not given). ycmd then extracts the system header directories from the output of this command (as if Clang was called with the `-E` and `-v` flags) and adds them to the list of user flags.

This works great on Linux and Windows but is not enough on macOS. Additional include paths are needed on this platform:
 - `clang_includes/include`: this folder is not automatically added with the `-resource-dir` flag;
 - `mac_includes/include`: this one is from the Clang prebuilt binaries.

Fixes #844 
Fixes https://github.com/Valloric/YouCompleteMe/issues/2536
Fixes https://github.com/Valloric/YouCompleteMe/issues/2790

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/853)


puremourning:
On first glance I think this will mean that it isn’t possible to use the iPhone platform and tool chain anymore?

micbou:
The `--sysroot` flag is passed to the fake clang executable so it should still be possible to use a different toolchain.

bstaletic:
On first glance, this looks like a good idea, but since different toolchains are important to me, I'll wait for the weekend to properly test this.

---

Review status: 0 of 186 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/853#-:-Kvk_PyJThaOTPUGRX_q:b-emny8r)*

micbou:
So, it doesn't work when compiled with system libclang. In that case, `ycm_fake_clang` outputs:
```
./ycm_fake_clang -resource-dir=/Users/micbou/projects/YouCompleteMe/third_party/ycmd/clang_includes -x c++ -E -v test.cpp
...
ignoring nonexistent directory "/usr/include/c++/v1"
#include "..." search starts here:
#include <...> search starts here:
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1
 /usr/local/include
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/9.0.0/include
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include
 /usr/include
 /System/Library/Frameworks (framework directory)
 /Library/Frameworks (framework directory)
```
which looks good (the `-resource-dir` flag seems to be ignored). However, this breaks if we prepend `mac_includes/include/c++/v1` and `clang_includes/include` to that list. Now, without `--system-libclang`, `ycm_fake_clang` returns:
```
./ycm_fake_clang -resource-dir=/Users/micbou/projects/YouCompleteMe/third_party/ycmd/clang_includes -x c++ -E -v test.cpp
...
ignoring nonexistent directory "/Users/micbou/projects/YouCompleteMe/third_party/include/c++/v1"
ignoring nonexistent directory "/usr/include/c++/v1"
ignoring nonexistent directory "/Users/micbou/projects/YouCompleteMe/third_party/ycmd/clang/5.0.0/include"
#include "..." search starts here:
#include <...> search starts here:
 /usr/local/include
 /usr/include
 /System/Library/Frameworks (framework directory)
 /Library/Frameworks (framework directory)
End of search list.
```
See the two nonexistent directories `include/c++/v1` and `clang/5.0.0/include` (`-resource-dir` is still ignored)? That's why we add `mac_includes/include/c++/v1` and `clang_includes/include` in the first place.

If we compare to system Clang with `-resource-dir`:
```
echo | clang -resource-dir=/Users/micbou/projects/YouCompleteMe/third_party/ycmd/clang_includes -x c++ -E -v -
...
ignoring nonexistent directory "/Users/micbou/projects/YouCompleteMe/include/c++/v1"
ignoring nonexistent directory "/usr/include/c++/v1"
#include "..." search starts here:
#include <...> search starts here:
 /usr/local/include
 /Users/micbou/projects/YouCompleteMe/third_party/ycmd/clang_includes/include
 /usr/include
 /System/Library/Frameworks (framework directory)
 /Library/Frameworks (framework directory)
```
Surprisingly,`-resource-dir` is not ignored. However, it won't work because clang didn't find `include/c++/v1`. And finally without `-resource-dir`:
```
echo | clang -x c++ -E -v -
...
ignoring nonexistent directory "/usr/include/c++/v1"
#include "..." search starts here:
#include <...> search starts here:
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1
 /usr/local/include
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/9.0.0/include
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include
 /usr/include
 /System/Library/Frameworks (framework directory)
 /Library/Frameworks (framework directory)
```
Same result as `ycm_fake_clang` if compiled with `--system-libclang`.

Given these results, I am tempted to use system Clang (without `-resource-dir`) instead of `ycm_fake_clang` on macOS (or maybe use Clang in priority then fall back to `ycm_fake_clang` if Clang cannot be found) and to not add the `mac_includes/include/c++/v1` and `clang_includes/include` directories. Thoughts?

---

Reviewed 7 of 186 files at r1.
Review status: 7 of 186 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/853#-:-Kvl-_Qk52BzpOeVLdkP:b4fcv9v)*

bstaletic:
Here are my first test results:

`./build.py --system-boost --system-clang --clang-completer`
```
'/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include/g++-v7',
'/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include/g++-v7/x86_64-pc-linux-gnu',
'/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include/g++-v7/backward',
'/home/bstaletic/.vim/pack/bundle/start/YouCompleteMe/third_party/ycmd/clang_includes/include',
```

`./build.py --system-boost --clang-completer`
```

'/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include/g++-v7',
'/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include/g++-v7/x86_64-pc-linux-gnu',
'/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include/g++-v7/backward',
'/home/bstaletic/.vim/pack/bundle/start/YouCompleteMe/third_party/ycmd/clang_includes/include',
```

`echo | g++ -E -v -xc++`
```
/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include/g++-v7
/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include/g++-v7/x86_64-pc-linux-gnu
/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include/g++-v7/backward
/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include
/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include-fixed
/usr/include
```

`echo | clang++ -E -v -xc++`
```
/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include/g++-v7
/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include/g++-v7/x86_64-pc-linux-gnu
/usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/include/g++-v7/backward
/usr/lib64/llvm/5/bin/../../../../lib/clang/5.0.0/include
/usr/include
```

Now can anyone explain why is `clang++` looking inside `/usr/lib/gcc`?

---

Reviewed 186 of 186 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/853#-:-KvmmY9OV6J0dWHBqOMc:bdygy0s)*

micbou:
That's probably because it looks for the GCC standard library. Try to add the `-stdlib=libc++` flag.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/853#-:-KvnFNhGLWeRGhpo_EKL:bnaqdl5)*

micbou:
Other than that, your results looks fine. Did you check that `:YcmDiags` returned no errors in the file?

---

Review status: 185 of 186 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/853#-:-KvnLxIMTBEwYmg_a4c1:b-uesfhw)*

bstaletic:
Yes, I checked `:YcmDiags` everything's fine. At least as long as netive comiling is ocncerned.

With arm-none-eabi things.

- Adding only `--sysroot` resulted in clang++ and ycm_fake_clang finding only one out of five paths.
- Either just adding `--sysroot` or additionally adding `--isystem` doesn't prevent ycmd from appending whatever "system flags" it finds.

More info about the first point:

What I'm expecting to be found for C:
```
 /home/bstaletic/arm-none-eabi/bin/../lib/gcc/arm-none-eabi/5.4.1/include
 /home/bstaletic/arm-none-eabi/bin/../lib/gcc/arm-none-eabi/5.4.1/include-fixed
 /home/bstaletic/arm-none-eabi/bin/../lib/gcc/arm-none-eabi/5.4.1/../../../../arm-none-eabi/include
```

What I'm expecting to be found for C++:
```
 /home/bstaletic/arm-none-eabi/bin/../lib/gcc/arm-none-eabi/5.4.1/../../../../arm-none-eabi/include/c++/5.4.1
 /home/bstaletic/arm-none-eabi/bin/../lib/gcc/arm-none-eabi/5.4.1/../../../../arm-none-eabi/include/c++/5.4.1/arm-none-eabi
 /home/bstaletic/arm-none-eabi/bin/../lib/gcc/arm-none-eabi/5.4.1/../../../../arm-none-eabi/include/c++/5.4.1/backward
 /home/bstaletic/arm-none-eabi/bin/../lib/gcc/arm-none-eabi/5.4.1/include
 /home/bstaletic/arm-none-eabi/bin/../lib/gcc/arm-none-eabi/5.4.1/include-fixed
 /home/bstaletic/arm-none-eabi/bin/../lib/gcc/arm-none-eabi/5.4.1/../../../../arm-none-eabi/include
```

What is clang and ycmd finding with `--sysroot=/home/bstaletic/arm-none-eabi/arm-none-eabi`? Nothing. At least not in the `/home/bstaletic/arm-none-eabi`.

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/853#-:-KvnPsS8OV9M1RtR1wn0:bt6978r)*

bstaletic:
For the record, old behaviour was better for cross toolchains as it didn't require `--systemroot` to stop ycmd from including wrong flags.

puremourning:
So I took a look. If I’m honest, my initial reaction is that i don’t really love it. I’m not super keen on shipping an arbitrary copy of the standard library with YCM and the fake clang driver + parsing the result still requires us to ship a (essentially) fake toolchain.

I realise that we have to hack this irritating macOS standard library path thing once a year with each new macOS/Xcode. But i’m convinced there is a canonical way to determine what we need to use programatically from the installed Xcode or command line tools.

For a start, I think we should ask Xcode-select which toolchain to use, then just use that one. That’s allows the user to determine whether to use Xcode or command line tools and avoids any conflict. 

The second part of the problem (which I thought i fixed last year) is that the Xcode installation is fiddly and requires some logic that’s built in to the binrary installation location combined with an -isysroot flag for the particular platform target (macOS, iOS, etc.). 

FWIW i did try YCM with high sierra when it was GM and it worked fine, and still does for me. So I’m a little perplexed by the current slew of issues.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/853#-:-KvqzvjYX-_thC3Y7wUi:bqa5k43)*

micbou:
@bstaletic

> For the record, old behaviour was better for cross toolchains as it didn't require --systemroot to stop ycmd from including wrong flags.

Are you currently adding the `arm-none-eabi` flags to your `.ycm_extra_conf.py` file? If that's the case then libclang is actually adding the wrong flags but internally so you don't see them in `:YcmDebugInfo` but I can assure you they are added. It works because your include flags take precedence.

@puremourning

> I’m not super keen on shipping an arbitrary copy of the standard library with YCM and the fake clang driver + parsing the result still requires us to ship a (essentially) fake toolchain.

Yes, I don't like shipping the standard library either. That's why I am thinking of using system Clang on macOS.

> For a start, I think we should ask Xcode-select which toolchain to use, then just use that one. That’s allows the user to determine whether to use Xcode or command line tools and avoids any conflict.

Do you mean running `xcode-select` with the `-p` option? If so, I don't see why we wouldn't run `clang` with the `-E` and `-v` options to get the system header paths instead.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/853#-:-Kvr4hkj2xWoAQIvzW0Y:b-1fovmd)*

bstaletic:
Yes, I am currently adding `-isystem` flags to make cross compiled projects work properly. So the current solution is not better, jumt less ugly? Then it's fine, but I'd still like not to see the wrong flags in `:YcmDebugInfo`.

As for bundling a standard library, I am not a fan of that either, but I don't know much about macOS and how fiddly it is.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/853#-:-Kvna0kxJGfAKogv9X9z:b-p705g0)*

Valloric:
@micbou Your heart is in the right place, but I don't think this is the best approach to the problem. :(

I'd be for this extra complexity if it would consistently improve the user experience on all OS's we support. But that doesn't appear to be the case.

---

Review status: 7 of 12 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/853#-:-Kx4vIVGZYcfi3p6-zU3:b-hx5adn)*

bstaletic:
Since we were talking about reviving this PR, how about doing what this PR suggests only on Windows and Linux?
For MacOS we can stick to the current heuristics and thus avoid bundling the standard library.

---

Reviewed 4 of 179 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.

---

*[ycmd/completers/cpp/flags.py, line 548 at r5](https://reviewable.io:443/reviews/valloric/ycmd/853#-L86mEJDSaDNWJQATmRR:-L86mEJDSaDNWJQATmRS:bbpjh4n) ([raw file](https://github.com/valloric/ycmd/blob/f62045cc7d842dc9414008c5f2ec7cd930fd5dcd/ycmd/completers/cpp/flags.py#L548)):*
> ```Python
>     iter_flags = iter( flags )
>     for flag in iter_flags:
>       if flag in [ '-x', '--sysroot', '-gcc-toolchain' ]:
> ```

`-xc++` is also a valid flag, so we would need it in the `elif` branch too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/853#-:-L86mPKj-rA8pJKi-l_D:bv4sgwk)*

midchildan:
___
*[ycmd/completers/cpp/flags.py, line 673 at r11](https://reviewable.io/reviews/valloric/ycmd/853#-LKxnnfr48vSYgibLy4P:-LKyd8ar-L5ryaxvGyoy:b-17i7wc) ([raw file](https://github.com/valloric/ycmd/blob/e416e23c9c1a739016bc2fd2410cb42a8f2dd7c4/ycmd/completers/cpp/flags.py#L673)):*
> Here we are talking about discovering default system include paths. Ones you don't specify explicitly, but would be added with -isystem, so replacing -xcuda with -xc++ is not right.

I figured that replacing `-xcuda` with `-xc++` was a bad idea.

> Is it possible to get clang to parse a file with -nocudainc, but without -nocudalib?

Only if clang finds the CUDA library location. Otherwise, clang refuses to do anything without `-nocudalib`.

midchildan:
___
*[ycmd/completers/cpp/flags.py, line 673 at r11](https://reviewable.io/reviews/valloric/ycmd/853#-LKxnnfr48vSYgibLy4P:-LL2Mnz1A-uLCsyXr8At:b6efk66) ([raw file](https://github.com/valloric/ycmd/blob/e416e23c9c1a739016bc2fd2410cb42a8f2dd7c4/ycmd/completers/cpp/flags.py#L673)):*

Are you sure? I'm asking because I can get clang to print some paths if I pass `-nocudalib` and `-nostdlib` and I certainly did not install a CUDA library. Besides, @midchildan also said clang can work without `--cuda-path` if `-nocudalib` and `-nocudainc` are supplied.
</blockquote></details>

`--cuda-path` can be omitted if both `--nocudalib` and `-nocudainc` is specified. This is especially useful when using the `nvidia-cuda-toolkit` package from Debian. Since Debian's version of CUDA change file locations to comply with FHS, supplying `--cuda-path` won't work.

midchildan:
___
*[ycmd/completers/cpp/flags.py, line 673 at r11](https://reviewable.io/reviews/valloric/ycmd/853#-LKxnnfr48vSYgibLy4P:-LL7RBcGBH6J46p5dfgw:bikr9fo) ([raw file](https://github.com/valloric/ycmd/blob/e416e23c9c1a739016bc2fd2410cb42a8f2dd7c4/ycmd/completers/cpp/flags.py#L673)):*

@midchildan Could you give this PR a test and see if ycmd correctly discovers system include paths for valid CUDA flag configurations?
</blockquote></details>

I can confirm it works on:

1. Ubuntu with the `nvidia-cuda-toolkit` package
2. macOS

joelfrederico:
I'd like to request the ability to specify and/or deduce Clang. I have a non-standard clang installed by macports. It isn't available on the command line as anything intelligible. If I can specify clang, the includes will be correct for the compiler I'm actually using. Alternatively, clang can be deduced from my compile_commands.json and then the includes will be correct too.

Thoughts?

PR #854, opened 2017-10-06
When list duplicated c++/v1 using isystem, include_next will go crazy

4pao:
If both "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain" and "/Library/Developer/CommandLineTools" exsit, there will be two "c++/v1" directories listed sequentially. The include_next in one c++/v1, stdlib.h for example, will go for stdlib.h in next c++/v1 which should be /usr/include/stdlib.h. Then compilation system will complain dozens of 'unknown type name'.

This bug can be workaround by adding "/usr/include" to ycm_extra_conf.py, which might not the case when using with YCM-generator and cumbersome.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/854)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -438,12 +438,17 @@ def _RemoveUnusedFlags( flags, filename ):
 #
 # Most users have xcode installed, but in order to be as compatible as
 # possible we consider both possible installation locations
-MAC_CLANG_TOOLCHAIN_DIRS = [
-  '/Applications/Xcode.app/Contents/Developer/Toolchains/'
-    'XcodeDefault.xctoolchain',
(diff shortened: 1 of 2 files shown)


puremourning:
Thanks for contributing!

As it happens, I've actually submitted a sort of competing PR: https://github.com/Valloric/ycmd/pull/855

I think my version is more efficient as it only performs max 2 additional filesystem operations and only does it when running on a mac.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/854#-:-KvsL3Z51LAyJJ0OxARv:b-sfmg5b)*

4pao:
Thanks.

PR #860, opened 2017-10-20
[READY] Fix empty location extent after missing semicolon

### LINKED ISSUE #859: clang diagnostics location_extent zeros when source has a certain type of error
ycmd commit is a12a0a3a85 on mac, but also happens with older 2e625157631

This might be a libclang bug, but with a source file with errors such as this, the `location_extent` values for all diagnostics are zeros.

```c++
class A {
    A() // my 'missing semicolon' location_extent is wrong
};

i_ruin_everything // i'm the stinker line that causes all the location_extent problems

A::A() {
}

another_error // my location_extent is wrong, too
```

diagnostics look like:

```json
[
    {
        "kind": "ERROR",
        "text": "expected ';' at end of declaration list",
        "ranges": [],
        "location": {
            "filepath": "source.cc",
            "column_num": 8,
            "line_num": 2
        },
        "location_extent": {
            "start": {
                "filepath": "",
                "column_num": 0,
                "line_num": 0
            },
            "end": {
                "filepath": "",
                "column_num": 0,
                "line_num": 0
            }
        },
        "fixit_available": true
    },
    {
        "kind": "ERROR",
        "text": "unknown type name 'i_ruin_everything'",
        "ranges": [],
        "location": {
            "filepath": "source.cc",
            "column_num": 1,
            "line_num": 5
        },
        "location_extent": {
            "start": {
                "filepath": "",
                "column_num": 0,
                "line_num": 0
            },
            "end": {
                "filepath": "",
                "column_num": 0,
                "line_num": 0
            }
        },
        "fixit_available": false
    },
    {
        "kind": "ERROR",
        "text": "constructor cannot have a return type",
        "ranges": [
            {
                "start": {
                    "filepath": "",
                    "column_num": 0,
                    "line_num": 0
                },
                "end": {
                    "filepath": "",
                    "column_num": 0,
                    "line_num": 0
                }
            },
            {
                "start": {
                    "filepath": "source.cc",
                    "column_num": 4,
                    "line_num": 7
                },
                "end": {
                    "filepath": "source.cc",
                    "column_num": 5,
                    "line_num": 7
                }
            }
        ],
        "location": {
            "filepath": "source.cc",
            "column_num": 4,
            "line_num": 7
        },
        "location_extent": {
            "start": {
                "filepath": "",
                "column_num": 0,
                "line_num": 0
            },
            "end": {
                "filepath": "",
                "column_num": 0,
                "line_num": 0
            }
        },
        "fixit_available": false
    },
    {
        "kind": "ERROR",
        "text": "unknown type name 'another_error'",
        "ranges": [],
        "location": {
            "filepath": "source.cc",
            "column_num": 1,
            "line_num": 10
        },
        "location_extent": {
            "start": {
                "filepath": "",
                "column_num": 0,
                "line_num": 0
            },
            "end": {
                "filepath": "",
                "column_num": 0,
                "line_num": 0
            }
        },
        "fixit_available": false
    },
    {
        "kind": "ERROR",
        "text": "expected unqualified-id",
        "ranges": [],
        "location": {
            "filepath": "source.cc",
            "column_num": 50,
            "line_num": 10
        },
        "location_extent": {
            "start": {
                "filepath": "",
                "column_num": 0,
                "line_num": 0
            },
            "end": {
                "filepath": "",
                "column_num": 0,
                "line_num": 0
            }
        },
        "fixit_available": false
    }
]
```

micbou:
We currently use the following (convoluted) method to compute the location extent of a diagnostic i.e. the range of the identifier where the diagnostic occurred:
 - obtain a range from [the `clang_getCursorExtent` function](https://clang.llvm.org/doxygen/group__CINDEX__CURSOR__SOURCE.html#ga79f6544534ab73c78a8494c4c0bc2840), which is supposed to contain our identifier;
 - extract all the tokens from that range with [`clang_tokenize`](https://clang.llvm.org/doxygen/group__CINDEX__LEX.html#ga6b315a71102d4f6c95eb68894a3bda8a);
 - find the token with the same starting position as the diagnostic;
 - use the length of that token to get the ending position;
 - compute the range.

Unfortunately, this doesn't always work as shown in issue https://github.com/Valloric/ycmd/issues/859, In that case, `clang_getCursorExtent` returns an empty range for all diagnostics after the missing semicolon.

This PR fixes that issue by extracting the tokens from the diagnostic position to the immediate next character then using `clang_getTokenExtent` to get the range of the identifier (multiline identifiers are supported).

Fixes #859.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/860)

### DIFF
--- a/cpp/ycm/ClangCompleter/ClangHelpers.cpp
+++ b/cpp/ycm/ClangCompleter/ClangHelpers.cpp
@@ -167,7 +167,7 @@ Range GetLocationExtent( CXSourceLocation source_location,
   clang_tokenize( translation_unit, range, &tokens, &num_tokens );
 
   Location location( source_location );
-  Range final_range;
+  Range final_range( range );
 
   for ( size_t i = 0; i < num_tokens; ++i ) {
     CXToken token = tokens[ i ];
--- a/cpp/ycm/ClangCompleter/ClangHelpers.cpp
+++ b/cpp/ycm/ClangCompleter/ClangHelpers.cpp
@@ -161,8 +161,7 @@ Range GetLocationExtent( CXSourceLocation source_location,
   // I've tried many simpler ways of doing this and they all fail in various
   // situations.
 
-  CXSourceRange range = clang_getCursorExtent(
-                          clang_getCursor( translation_unit, source_location ) );
+  CXSourceRange range = clang_getRange( source_location, source_location );
   CXToken *tokens;
   unsigned int num_tokens;
   clang_tokenize( translation_unit, range, &tokens, &num_tokens );
@@ -171,15 +170,12 @@ Range GetLocationExtent( CXSourceLocation source_location,
   Range final_range;
 
   for ( size_t i = 0; i < num_tokens; ++i ) {
+    CXToken token = tokens[ i ];
     Location token_location( clang_getTokenLocation( translation_unit,
-                                                     tokens[ i ] ) );
+                                                     token ) );
 
     if ( token_location == location ) {
-      std::string name = CXStringToString(
-                           clang_getTokenSpelling( translation_unit, tokens[ i ] ) );
-      Location end_location = location;
-      end_location.column_number_ += name.length();
-      final_range = Range( location, end_location );
+      final_range = Range( clang_getTokenExtent( translation_unit, token ) );
       break;
     }
   }
--- a/cpp/ycm/ClangCompleter/ClangHelpers.cpp
+++ b/cpp/ycm/ClangCompleter/ClangHelpers.cpp
@@ -171,6 +171,9 @@ Range GetLocationExtent( CXSourceLocation source_location,
 
   for ( size_t i = 0; i < num_tokens; ++i ) {
     CXToken token = tokens[ i ];
+    if ( clang_getTokenKind( token ) == CXToken_Comment )
+      continue;
+
     Location token_location( clang_getTokenLocation( translation_unit,
                                                      token ) );
 
--- a/ycmd/tests/clang/diagnostics_test.py
+++ b/ycmd/tests/clang/diagnostics_test.py
@@ -249,3 +249,117 @@ def Diagnostics_MultipleMissingIncludes_test( app ):
       'fixit_available': False
     } ),
   ) )
+
+
+@IsolatedYcmd()
+def Diagnostics_LocationExtent_MissingSemicolon_test( app ):
+  contents = ReadFile( PathToTestFile( 'location_extent.cc' ) )
+
+  event_data = BuildRequest( contents = contents,
+                             event_name = 'FileReadyToParse',
+                             filetype = 'cpp',
+                             compilation_flags = [ '-x', 'c++' ] )
+
+  response = app.post_json( '/event_notification', event_data ).json
+
+  pprint( response )
+
+  assert_that( response, contains(
+    has_entries( {
+      'kind': equal_to( 'ERROR' ),
+      'location': has_entries( {
+        'line_num': 2,
+        'column_num': 9,
+        'filepath': '/foo'
+      } ),
+      # NOTE: we always return a location extent even if we can't find one since
+      # clients expect this field to always exist.
+      'location_extent': has_entries( {
+        'start': has_entries( {
+          'line_num': 0,
+          'column_num': 0,
+          'filepath': ''
+        } ),
+        'end': has_entries( {
+          'line_num': 0,
+          'column_num': 0,
+          'filepath': ''
+        } ),
+      } ),
+      'ranges': empty(),
+      'text': equal_to( "expected ';' at end of declaration list" ),
+      'fixit_available': True
+    } ),
+    has_entries( {
+      'kind': equal_to( 'ERROR' ),
+      'location': has_entries( {
+        'line_num': 5,
+        'column_num': 1,
+        'filepath': '/foo'
+      } ),
+      'location_extent': has_entries( {
+        'start': has_entries( {
+          'line_num': 5,
+          'column_num': 1,
+          'filepath': '/foo'
+        } ),
+        'end': has_entries( {
+          'line_num': 6,
+          'column_num': 11,
+          'filepath': '/foo'
+        } )
+      } ),
+      'ranges': empty(),
+      'text': equal_to( "unknown type name 'multiline_identifier'" ),
+      'fixit_available': False
+    } ),
+    has_entries( {
+      'kind': equal_to( 'ERROR' ),
+      'location': has_entries( {
+        'line_num': 8,
+        'column_num': 7,
+        'filepath': '/foo'
+      } ),
+      'location_extent': has_entries( {
+        'start': has_entries( {
+          'line_num': 8,
+          'column_num': 7,
+          'filepath': '/foo'
+        } ),
+        'end': has_entries( {
+          'line_num': 8,
+          'column_num': 11,
+          'filepath': '/foo'
+        } )
+      } ),
+      'ranges': contains(
+        # FIXME: empty ranges from libclang should be ignored.
+        has_entries( {
+          'start': has_entries( {
+            'line_num': 0,
+            'column_num': 0,
+            'filepath': ''
+          } ),
+          'end': has_entries( {
+            'line_num': 0,
+            'column_num': 0,
+            'filepath': ''
+          } )
+        } ),
+        has_entries( {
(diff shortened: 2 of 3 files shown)


micbou:
In fact, we don't need to compute the location of the next character for the end of the range. We can use the same position for the start and the end.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/860#-:-Kwvg4PmYJTh66jm2W3G:b9folh8)*

joelhock:
Thank you for the quick response, I can verify that this does indeed fix the issue I reported in https://github.com/Valloric/ycmd/issues/859

I hesitate to complicate this further with a much more minor nit, but perhaps there's a quick fix since you're already thinking about this: both with and without this PR, ycmd will take this code: 

```c++
class A {
    A()// extent of missing semicolon is this whole comment if no space before it
};
```

and, if there is no space before the comment, as shown, ycmd will report the entire comment length as the extent for the missing semicolon diagnostic.

```json
[
    {
        "kind": "ERROR",
        "text": "expected ';' at end of declaration list",
        "ranges": [],
        "location": {
            "filepath": "mysource.cpp",
            "column_num": 8,
            "line_num": 2
        },
        "location_extent": {
            "start": {
                "filepath": "mysource.cpp",
                "column_num": 8,
                "line_num": 2
            },
            "end": {
                "filepath": "mysource.cpp",
                "column_num": 82,
                "line_num": 2
            }
        },
        "fixit_available": true
    }
]
```

micbou:
Comments are now ignored when computing the location extent. I think this makes sense as comments are not a functional part of the code.

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/860#-:-Kx3hiCNR_jFERu9CjQg:b-ufp4jc)*

puremourning:

---

*[cpp/ycm/ClangCompleter/ClangHelpers.cpp, line 174 at r2](https://reviewable.io:443/reviews/valloric/ycmd/860#-Kx3oZbizUElFuYjouug:-Kx3oZbjJRXrk4786Th1:b-e7hav1) ([raw file](https://github.com/valloric/ycmd/blob/74fe24ff290eaf1dbbb648efc35c6d5afc641ca6/cpp/ycm/ClangCompleter/ClangHelpers.cpp#L174)):*
> ```cpp
>   for ( size_t i = 0; i < num_tokens; ++i ) {
>     CXToken token = tokens[ i ];
>     if ( clang_getTokenKind( token ) == CXToken_Comment )
> ```

IIRC there are parser errors that can happen within comments. For example, I feel like clang can diagnose `/* within comment` (in fact, i'm certain it can):



Also with `-pedantic` you can get errors about multi-line `//` comments (yep, seriously):



Does this therefore make that not work, or work less well?

---

*[ycmd/tests/clang/diagnostics_test.py, line 275 at r2](https://reviewable.io:443/reviews/valloric/ycmd/860#-Kx3nX8M16VoebGpYg1O:-Kx3nX8M16VoebGpYg1P:bspveij) ([raw file](https://github.com/valloric/ycmd/blob/74fe24ff290eaf1dbbb648efc35c6d5afc641ca6/ycmd/tests/clang/diagnostics_test.py#L275)):*
> ```Python
>         'filepath': '/foo'
>       } ),
>       # NOTE: we always return a location extent even if we can't find one since
> ```

In this scenario, i feel like the right thing to do is to put the squiggle under the character that is the "location" of the diagnostic. Is that what happens ? If not, would it be possible to return a range which was just that one character? In the case of missing semi-colon, the squiggle should be the place where the semi-colon would be written (which i think is the position returned in the `location` entry?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/860)*

micbou:
Reviewed 2 of 2 files at r3.

---

*[cpp/ycm/ClangCompleter/ClangHelpers.cpp, line 174 at r2](https://reviewable.io:443/reviews/valloric/ycmd/860#-Kx3oZbizUElFuYjouug:-Kx3yMo2dbK6kk28DPOC:b-vlmuds) ([raw file](https://github.com/valloric/ycmd/blob/74fe24ff290eaf1dbbb648efc35c6d5afc641ca6/cpp/ycm/ClangCompleter/ClangHelpers.cpp#L174)):*

IIRC there are parser errors that can happen within comments. For example, I feel like clang can diagnose `/* within comment` (in fact, i'm certain it can):



Also with `-pedantic` you can get errors about multi-line `//` comments (yep, seriously):



Does this therefore make that not work, or work less well?

</blockquote></details>

No, the behavior is identical because, with or without these changes, there is no location extent and so the location itself is used (at least in YCM). However, if we don't ignore the comments, we get



which doesn't feel right.

---

*[ycmd/tests/clang/diagnostics_test.py, line 275 at r2](https://reviewable.io:443/reviews/valloric/ycmd/860#-Kx3nX8M16VoebGpYg1O:-Kx41aqYW10udTQRaMx2:b-18brq) ([raw file](https://github.com/valloric/ycmd/blob/74fe24ff290eaf1dbbb648efc35c6d5afc641ca6/ycmd/tests/clang/diagnostics_test.py#L275)):*

In this scenario, i feel like the right thing to do is to put the squiggle under the character that is the "location" of the diagnostic. Is that what happens ? If not, would it be possible to return a range which was just that one character? In the case of missing semi-colon, the squiggle should be the place where the semi-colon would be written (which i think is the position returned in the `location` entry?
</blockquote></details>

If there is no location extent, YCM falls back to the location itself but that may not be the case for other clients so I think your suggestion is sound. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/860)*

puremourning:
:lgtm: nice work!

---

---

*[cpp/ycm/ClangCompleter/ClangHelpers.cpp, line 174 at r2](https://reviewable.io:443/reviews/valloric/ycmd/860#-Kx3oZbizUElFuYjouug:-Kx44LLHhCI2cWdrjo4K:bmkudps) ([raw file](https://github.com/valloric/ycmd/blob/74fe24ff290eaf1dbbb648efc35c6d5afc641ca6/cpp/ycm/ClangCompleter/ClangHelpers.cpp#L174)):*

No, the behavior is identical because, with or without these changes, there is no location extent and so the location itself is used (at least in YCM). However, if we don't ignore the comments, we get



which doesn't feel right.
</blockquote></details>

OK makes sense, thanks!

---

*[ycmd/tests/clang/diagnostics_test.py, line 275 at r2](https://reviewable.io:443/reviews/valloric/ycmd/860#-Kx3nX8M16VoebGpYg1O:-Kx44OibmIA01IAYzFbe:btojbfd) ([raw file](https://github.com/valloric/ycmd/blob/74fe24ff290eaf1dbbb648efc35c6d5afc641ca6/ycmd/tests/clang/diagnostics_test.py#L275)):*

If there is no location extent, YCM falls back to the location itself but that may not be the case for other clients so I think your suggestion is sound. Done.
</blockquote></details>

Great!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/860#-:-Kx44WDReM1L95iERK2W:b-kbynj8)*

puremourning:
Reviewed 1 of 2 files at r2, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/860)*

micbou:

---

*[cpp/ycm/ClangCompleter/ClangHelpers.cpp, line 174 at r2](https://reviewable.io:443/reviews/valloric/ycmd/860#-Kx3oZbizUElFuYjouug:-Kx453KKYlouQYKWdikn:bxi96b2) ([raw file](https://github.com/valloric/ycmd/blob/74fe24ff290eaf1dbbb648efc35c6d5afc641ca6/cpp/ycm/ClangCompleter/ClangHelpers.cpp#L174)):*

OK makes sense, thanks!
</blockquote></details>

Wait, the second example in my screenshot still happens if we ignore comments. Looks like libclang returns `multiple` as an identifier token. Well, I guess we can't do much about it. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/860)*

Valloric:
Thanks for the PR!

@zzbot r=puremourning

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/860#-:-Kx4eTodtYVnPNMtVHQM:b-vzdkt4)*

PR #861, opened 2017-10-20
[READY] Translate libclang error codes to exceptions

micbou:
This PR translates [libclang error codes](https://clang.llvm.org/doxygen/CXErrorCode_8h.html#adba17f287f8184fc266f2db4e669bf0f) returned by the `clang_parseTranslationUnit2FullArgv` and `clang_reparseTranslationUnit` functions to `std::runtime_error` exceptions with the corresponding error message; replacing the `ClangParseError` exception. Unlike `ClangParseError`, we don't silence them as they can help understand why the Clang completer is not working.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/861)


Valloric:
:lgtm:

Thanks for the PR!

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861#-:-KwwJheaKLMSM4qPS4A9:b-19kv2l)*

micbou:
Reviewed 5 of 7 files at r1, 3 of 3 files at r2.

---

*[cpp/ycm/ClangCompleter/ClangCompleter.cpp, line 86 at r2](https://reviewable.io:443/reviews/valloric/ycmd/861#-KwwcORuZY1A_upIsI39:-KwwcORuZY1A_upIsI3A:bbg4gf) ([raw file](https://github.com/valloric/ycmd/blob/5b77cdf42245de8c6dbe48a9fbec1f7737a52c0e/cpp/ycm/ClangCompleter/ClangCompleter.cpp#L86)):*
> ```cpp
>                                          translation_unit_created );
> 
>   if ( !unit )
> ```

We don't need this check and the others below anymore since the `TranslationUnit` pointer should always be set in `GetOrCreate` (an exception is thrown otherwise).

---

*[cpp/ycm/ClangCompleter/ClangCompleter.cpp, line 100 at r2](https://reviewable.io:443/reviews/valloric/ycmd/861#-KwwevDIKstXCOoStYW4:-KwwevDJb1ZC-tD8Inj7:bfir9gh) ([raw file](https://github.com/valloric/ycmd/blob/5b77cdf42245de8c6dbe48a9fbec1f7737a52c0e/cpp/ycm/ClangCompleter/ClangCompleter.cpp#L100)):*
> ```cpp
>   }
> 
>   return std::vector< Diagnostic >();
> ```

Cannot be reached.

---

*[cpp/ycm/tests/ClangCompleter/ClangCompleter_test.cpp, line 259 at r2](https://reviewable.io:443/reviews/valloric/ycmd/861#-Kwwe1OKuImim2yJBc6x:-Kwwe1OLhSMD8-LZJYAA:b4i0wlu) ([raw file](https://github.com/valloric/ycmd/blob/5b77cdf42245de8c6dbe48a9fbec1f7737a52c0e/cpp/ycm/tests/ClangCompleter/ClangCompleter_test.cpp#L259)):*
> ```cpp
>                                                  1,
>                                                  unsaved_files,
>                                                  flags ) );
> ```

These tests are obsolete; they all raise the same `std::runtime_error` exception.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

puremourning:

---

*[cpp/ycm/exceptions.h, line 0 at r2](https://reviewable.io:443/reviews/valloric/ycmd/861#-Kx5HETlM54VL1RARt1q:-Kx5HETlM54VL1RARt1r:b3imei) ([raw file](https://github.com/valloric/ycmd/blob/9fd92f3b757bdc3b515852ea8824bd787678ec16/cpp/ycm/exceptions.h#L0)):*
What was the reason for replacing the _specific_ exception class `ClangParseError` with the _generic_ exception type `std::runtime_error`? Bikeshedding again, but `std:runtime_error` does not feel like the correct thing to use:

> Defines a type of object to be thrown as exception. It reports errors that are due to events beyond the scope of the program and can not be easily predicted.

I think we can easily predict that the errors are due to parsing the code, because that's what the API tells us.

The CPP Core Guidelines  [say explicitly not to do this](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#note-261), so I think we should derive a new exception type.

---

*[cpp/ycm/ClangCompleter/ClangCompleter.cpp, line 87 at r2](https://reviewable.io:443/reviews/valloric/ycmd/861#-Kx5EiPnYgLwAKz_nQHK:-Kx5EiPnYgLwAKz_nQHL:box7927) ([raw file](https://github.com/valloric/ycmd/blob/9fd92f3b757bdc3b515852ea8824bd787678ec16/cpp/ycm/ClangCompleter/ClangCompleter.cpp#L87)):*
> ```cpp
>   try {
>     return unit->Reparse( unsaved_files );
>   } catch ( std::runtime_error & ) {
> ```

should probably catch by const reference

---

*[cpp/ycm/ClangCompleter/ClangUtils.cpp, line 46 at r2](https://reviewable.io:443/reviews/valloric/ycmd/861#-Kx5Fbw2ZeosLpWUlc9g:-Kx5Fbw2ZeosLpWUlc9h:bsl4gpn) ([raw file](https://github.com/valloric/ycmd/blob/9fd92f3b757bdc3b515852ea8824bd787678ec16/cpp/ycm/ClangCompleter/ClangUtils.cpp#L46)):*
Given that we don't actually need to use this as a `string`, I think we should return `const char *` from this method. That way there's no need to perform a heap allocation into the `std:string`, followed by a copy into the `std::runtime_error`.

According to the docs:

> Because copying std::exception is not permitted to throw exceptions, this message is typically stored internally as a separately-allocated reference-counted string. This is also why there is no constructor taking std::string&&: it would have to copy the content anyway.

This means the line `throw std::runtime_error( CXErrorCodeToString( code ) )` requires:
 - a ton of compiler-generated stack unwind nonesense
 - a heap allocation and memcpy to create the temporary std::string
 - a heap allocation and memcpy to create the "internal reference counted string"
 - deallocation of the std::string temporary

OTOH a `const char *` return requires only:
 - a ton of compiler-generated stack unwind nonesense
 - a heap allocation and memcpy to create the "internal reference counted string"

I realise this is pedantic bikeshedding. So feel free to ignore.

---

*[cpp/ycm/ClangCompleter/ClangUtils.cpp, line 60 at r2](https://reviewable.io:443/reviews/valloric/ycmd/861#-Kx5EyeCNKjCGvuqCD2I:-Kx5EyeCNKjCGvuqCD2J:b7fili4) ([raw file](https://github.com/valloric/ycmd/blob/9fd92f3b757bdc3b515852ea8824bd787678ec16/cpp/ycm/ClangCompleter/ClangUtils.cpp#L60)):*
> ```cpp
>              "while parsing the translation unit.";
>     default:
>       return "Unknown error while parsing the translation unit.";
> ```

Is this reachable? I tend to prefer to _not_ have a `default` clause and have the compiler tell us when we haven't covered all of the enum values. Adding a default clause prevents that. For example, if a new code is added in libclang, and we upgrade this default clause would get exercised, but otherwise it's not possible. OTOH if we don't have the default clause, we would get a nice compiler warning that there is a missing `case` statement.

More bikeshedding.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

puremourning:

---

*[cpp/ycm/ClangCompleter/ClangUtils.cpp, line 46 at r2](https://reviewable.io:443/reviews/valloric/ycmd/861#-Kx5Fbw2ZeosLpWUlc9g:-Kx5UCuH5xZW_x9h3dp_:b-glylzc) ([raw file](https://github.com/valloric/ycmd/blob/9fd92f3b757bdc3b515852ea8824bd787678ec16/cpp/ycm/ClangCompleter/ClangUtils.cpp#L46)):*

Given that we don't actually need to use this as a `string`, I think we should return `const char *` from this method. That way there's no need to perform a heap allocation into the `std:string`, followed by a copy into the `std::runtime_error`.

According to the docs:

> Because copying std::exception is not permitted to throw exceptions, this message is typically stored internally as a separately-allocated reference-counted string. This is also why there is no constructor taking std::string&&: it would have to copy the content anyway.

This means the line `throw std::runtime_error( CXErrorCodeToString( code ) )` requires:
 - a ton of compiler-generated stack unwind nonesense
 - a heap allocation and memcpy to create the temporary std::string
 - a heap allocation and memcpy to create the "internal reference counted string"
 - deallocation of the std::string temporary

OTOH a `const char *` return requires only:
 - a ton of compiler-generated stack unwind nonesense
 - a heap allocation and memcpy to create the "internal reference counted string"

I realise this is pedantic bikeshedding. So feel free to ignore.
</blockquote></details>

FWIW https://godbolt.org/g/vku1G3

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

bstaletic:
Reviewed 7 of 7 files at r1, 1 of 3 files at r2, 5 of 8 files at r3.
Review status: 7 of 10 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[cpp/ycm/ClangCompleter/ClangUtils.cpp, line 60 at r2](https://reviewable.io:443/reviews/valloric/ycmd/861#-Kx5EyeCNKjCGvuqCD2I:-KxK3GzU26vTzknPDVZT:bcptjbl) ([raw file](https://github.com/valloric/ycmd/blob/9fd92f3b757bdc3b515852ea8824bd787678ec16/cpp/ycm/ClangCompleter/ClangUtils.cpp#L60)):*

Is this reachable? I tend to prefer to _not_ have a `default` clause and have the compiler tell us when we haven't covered all of the enum values. Adding a default clause prevents that. For example, if a new code is added in libclang, and we upgrade this default clause would get exercised, but otherwise it's not possible. OTOH if we don't have the default clause, we would get a nice compiler warning that there is a missing `case` statement.

More bikeshedding.
</blockquote></details>

Not having a `default` bit me once, but that's a story for some other place, not a PR review.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

micbou:
Reviewed 7 of 8 files at r3, 2 of 2 files at r4, 4 of 4 files at r5.

---

*[cpp/ycm/exceptions.h, line  at r2](https://reviewable.io:443/reviews/valloric/ycmd/861#-Kx5HETlM54VL1RARt1q:-Kx7zI2LEywSRA08V2sv:brz7cea) ([raw file](https://github.com/valloric/ycmd/blob/9fd92f3b757bdc3b515852ea8824bd787678ec16/cpp/ycm/exceptions.h#L)):*

What was the reason for replacing the _specific_ exception class `ClangParseError` with the _generic_ exception type `std::runtime_error`? Bikeshedding again, but `std:runtime_error` does not feel like the correct thing to use:

> Defines a type of object to be thrown as exception. It reports errors that are due to events beyond the scope of the program and can not be easily predicted.

I think we can easily predict that the errors are due to parsing the code, because that's what the API tells us.

The CPP Core Guidelines  [say explicitly not to do this](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#note-261), so I think we should derive a new exception type.
</blockquote></details>

Reasons were:
 - Boost automatically translates C++ exceptions to `RuntimeError` in Python;
 - translating a custom C++ exception to its Python counterpart needs extra work;
 - `std::runtime_error` accepts a message.

I updated the PR to use a custom exception. In addition to translate `ClangParseError` to its Python equivalent, we need to expose the exception to the Python layer (done through the `CreatePythonException` function). Otherwise, we can't catch the exception in Python.

---

*[cpp/ycm/ClangCompleter/ClangCompleter.cpp, line 87 at r2](https://reviewable.io:443/reviews/valloric/ycmd/861#-Kx5EiPnYgLwAKz_nQHK:-KxKw2WL5j5YsX7YzCly:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/9fd92f3b757bdc3b515852ea8824bd787678ec16/cpp/ycm/ClangCompleter/ClangCompleter.cpp#L87)):*

should probably catch by const reference
</blockquote></details>

Done.

---

*[cpp/ycm/ClangCompleter/ClangUtils.cpp, line 46 at r2](https://reviewable.io:443/reviews/valloric/ycmd/861#-Kx5Fbw2ZeosLpWUlc9g:-KxL4sQmeBbJvz85OX4v:bc0x3ju) ([raw file](https://github.com/valloric/ycmd/blob/9fd92f3b757bdc3b515852ea8824bd787678ec16/cpp/ycm/ClangCompleter/ClangUtils.cpp#L46)):*

FWIW https://godbolt.org/g/vku1G3
</blockquote></details>

That's convincing. Done.

---

*[cpp/ycm/ClangCompleter/ClangUtils.cpp, line 60 at r2](https://reviewable.io:443/reviews/valloric/ycmd/861#-Kx5EyeCNKjCGvuqCD2I:-KxL5Gyz1MMxUXFqRXpc:bjtlnwf) ([raw file](https://github.com/valloric/ycmd/blob/9fd92f3b757bdc3b515852ea8824bd787678ec16/cpp/ycm/ClangCompleter/ClangUtils.cpp#L60)):*

Not having a `default` bit me once, but that's a story for some other place, not a PR review.
</blockquote></details>

The code won't compile in debug mode if there is a missing `case` and no `default` (because of `-Wall` implying `-Wswitch`, and `-Werror`) so it's safe to remove the `default` clause. However, the function still needs to return a string in other cases or GCC will complain with the following error:
```
ycmd/cpp/ycm/ClangCompleter/ClangUtils.cpp:60:1: error: control reaches end of non-void function [-Werror=return-type]
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

puremourning:
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[cpp/ycm/ycm_core.cpp, line 58 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-KxypsUj-LcEOoMhXpNf:-KxypsUj-LcEOoMhXpNg:b-awne6k) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ycm_core.cpp#L58)):*
> ```cpp
> 
> 
> PyObject* CreatePythonException( const char* name,
> ```

I'm kinda unsure what this is doing and why. I'm not super familiar with the python binding stuff here, but it seems to be injecting an exception object into a scope? Sorry, can't brain today and struggling to get it!

If our exception inherits from `std::exception` then i would have thought that this would work exactly the same as when it threw `std:;exception` ?

---

*[cpp/ycm/ycm_core.cpp, line 62 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-KxynHc137aSHLXL3bWp:-KxynHc137aSHLXL3bWq:bbtlwz6) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ycm_core.cpp#L62)):*
> ```cpp
>   std::string scope_name = extract< std::string >( scope().attr( "__name__" ) );
>   std::string qualified_name = scope_name + "." + name;
>   char *raw_name = const_cast< char * >( qualified_name.c_str() );
> ```

on the surface, this looks pretty evil.

Why the need for the `const_cast` here? In particular, we're passing this (presumably as `char *`) into the `PyErr_NewException`, but we don't own it and we can't modify it.

I'm expecting that `PyErr_NewException` doesn't really want to modify the string pointed to by `raw_name` and is just using a c interface, so it is probably safe, but maybe a quick note to that effect?

I

---

*[cpp/ycm/ycm_core.cpp, line 70 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-KxynuwsPJu1DNeh_9EB:-KxynuwsPJu1DNeh_9EC:b-j3c0on) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ycm_core.cpp#L70)):*
> ```cpp
> 
> #ifdef USE_CLANG_COMPLETER
> PyObject* PyExc_ClangParseError = NULL;
> ```

This global with no storage class is irksome. Do we really need global variable here? Should it at least be in this file's unnamed namespace? That would at least give it TU scope, rather than global.

I must confess, I'm a little perplexed why `TranslateClangParseError` sets this, and how it is used. Probably at least needs a comment.

---

*[cpp/ycm/ClangCompleter/ClangUtils.h, line 46 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-KxyokrWyRyp_LlZiuiE:-KxyokrXNp1WQL49oIPI:bi6ytlc) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ClangCompleter/ClangUtils.h#L46)):*
> ```objc
>  * Thrown when libclang fails to parse (or reparse) the translation unit.
>  */
> struct ClangParseError : virtual std::runtime_error {
> ```

woah. [virtual inheritance](https://isocpp.org/wiki/faq/multiple-inheritance#virtual-inheritance-where)... that's _unusual_. This is like an interview question in a code review !

Fortunately, unless i'm very much mistaken, this is all single-inheritance, and there is no _dreaded diamond_, so we don't actually need to virtually inherit here. Pretty sure `std::runtime_error` just inherits from `std::exception` so we're all good with:

```
struct ClangParseError : public std::runtime_error {
...
};
```

And then none of us has to explain how RTTI works and the stable layout for a CatDog which is a Cat and is a Dog and both Cat and Dog are Animals and we can pretend that we _just know_, and it _will all be fine_.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 105 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-Kxyr-caV1aYFwthukoB:-Kxyr-caV1aYFwthukoC:bs061pr) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L105)):*
> ```cpp
>                           EditingOptions(),
>                           &clang_translation_unit_ );
>   if ( failure )
> ```

I actually prefer to use the explicit check on `CXError_Success`. It's clearer and how the API is intended (otherwise they'd return int). Implicit narrowing casts (such as this one to bool) are IMO a pretty regrettable language feature.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 372 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-KxyrqhuckRNSY6WlPJ3:-KxyrqhuckRNSY6WlPJ4:bn7fge4) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L372)):*
> ```cpp
>   }
> 
>   if ( failure ) {
> ```

I'd prefer to compare with a value in the enum.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

micbou:
Reviewed 4 of 4 files at r6.

---

*[cpp/ycm/ycm_core.cpp, line 58 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-KxypsUj-LcEOoMhXpNf:-Ky8aJ6qs5tKvbPEnL6Z:br8am53) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ycm_core.cpp#L58)):*
Moved the code to the new `PythonException` templated class.

> If our exception inherits from std::exception then i would have thought that this would work exactly the same as when it threw std:;exception ?

By default, Boost.Python translates all C++ exceptions to `RuntimeError` in Python. So, we need to use [the `register_exception_translator` function](http://www.boost.org/doc/libs/1_65_1/libs/python/doc/html/reference/high_level_components/boost_python_exception_translato.html) to translate `ClangParseError` into its Python equivalent. However, this doesn't allow us to catch the exception in Python. For that, we need to add it to the `ycm_core` module as a Python object so that we can catch it like this:
```python
try {
  ...
} except ycm_core.ClangParseError {
  ...
}
```

---

*[cpp/ycm/ycm_core.cpp, line 62 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-KxynHc137aSHLXL3bWp:-Ky7kcMDMPUWfsyCe1Hb:b-dm6sxo) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ycm_core.cpp#L62)):*

on the surface, this looks pretty evil.

Why the need for the `const_cast` here? In particular, we're passing this (presumably as `char *`) into the `PyErr_NewException`, but we don't own it and we can't modify it.

I'm expecting that `PyErr_NewException` doesn't really want to modify the string pointed to by `raw_name` and is just using a c interface, so it is probably safe, but maybe a quick note to that effect?

I
</blockquote></details>

The signature of `PyErr_NewException` is `char *name, PyObject *base, PyObject *dict` in Python 2 (but `const char *name, PyObject *base, PyObject *dict` in Python 3) so we have to cast away the constness (for Python 2). Added a comment in `PythonException`.

---

*[cpp/ycm/ycm_core.cpp, line 70 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-KxynuwsPJu1DNeh_9EB:-Ky8ddyvzjYkffUUdnf6:bd5vtzj) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ycm_core.cpp#L70)):*

This global with no storage class is irksome. Do we really need global variable here? Should it at least be in this file's unnamed namespace? That would at least give it TU scope, rather than global.

I must confess, I'm a little perplexed why `TranslateClangParseError` sets this, and how it is used. Probably at least needs a comment.
</blockquote></details>

Now a private member of `PythonException`.

---

*[cpp/ycm/ClangCompleter/ClangUtils.h, line 46 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-KxyokrWyRyp_LlZiuiE:-Ky8dkBlOf8Ew3ljLiu9:bw3jqa0) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ClangCompleter/ClangUtils.h#L46)):*

woah. [virtual inheritance](https://isocpp.org/wiki/faq/multiple-inheritance#virtual-inheritance-where)... that's _unusual_. This is like an interview question in a code review !

Fortunately, unless i'm very much mistaken, this is all single-inheritance, and there is no _dreaded diamond_, so we don't actually need to virtually inherit here. Pretty sure `std::runtime_error` just inherits from `std::exception` so we're all good with:

```
struct ClangParseError : public std::runtime_error {
...
};
```

And then none of us has to explain how RTTI works and the stable layout for a CatDog which is a Cat and is a Dog and both Cat and Dog are Animals and we can pretend that we _just know_, and it _will all be fine_.

</blockquote></details>

That's what happens when you copy paste code without understanding it. Thanks for the link. Given that a `struct` has public inheritance by default, I simply removed the `virtual` keyword.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 105 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-Kxyr-caV1aYFwthukoB:-Ky8edkskmEslPFiy5jA:b39r5cr) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L105)):*

I actually prefer to use the explicit check on `CXError_Success`. It's clearer and how the API is intended (otherwise they'd return int). Implicit narrowing casts (such as this one to bool) are IMO a pretty regrettable language feature.
</blockquote></details>

At least, I learned it was working this way. Done.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 372 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-KxyrqhuckRNSY6WlPJ3:-Ky8emdWruFBHAiSp7wN:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L372)):*

I'd prefer to compare with a value in the enum.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

bstaletic:
To be frank, I'm not a fan of using python's C API directly. Yet it looks like we don't have too much choice if we want to keep using exceptions like we alwas had.

So, after having explained what @puremourning asked about the C API, this is :lgtm:.

---

Reviewed 1 of 2 files at r4, 3 of 4 files at r5, 4 of 4 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861#-:-KyCRpyjtbuolCqYXgZN:bhuuzzx)*

puremourning:

---

*[cpp/ycm/PythonSupport.h, line 65 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDatbVpQFj6_Enzx2C:-KyDatbVpQFj6_Enzx2D:b-cgv3mv) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/PythonSupport.h#L65)):*
> ```objc
> 
>     // Add the Python exception to the current Boost.Python module.
>     boost::python::scope().attr( name ) = boost::python::handle<>(
> ```

OK so I read up on the python API and boost python as best I could to understand this.

What I was trying to work out was this: we create this `python_exception_` thing, but who frees it, and when?

I came to the conclusion that I was not sure whether or not we need to decrement the reference count for the the exception object in the destructor of this class. It's frustratingly difficult to be sure. 

I was leaning towards thinking that perhaps the member object should be a `boost::python::handle<PyObject>` instance ? Then the assignment into the module would just be just a copy of the `handle` and the destructor of this class would decrement the reference count, and there's no chance that it gets freed before we call our `operator()` and try and use it.

I honestly don't know if that's likely/possible, but this line seems odd in that it allocates and _immediately_ deallocates a smart pointer object wrapping `python_exception_` ; that seems odd as it will increment and then (almost) immediately decrement the reference count.

Sorry that's waffly, but I'm really not that familiar with this API and the docs are _categorically_ abysmal.

---

*[cpp/ycm/ycm_core.cpp, line 58 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-KxypsUj-LcEOoMhXpNf:-KyDXnIhCZXHlWDtjIGD:b-lo31nq) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ycm_core.cpp#L58)):*

Moved the code to the new `PythonException` templated class.

> If our exception inherits from std::exception then i would have thought that this would work exactly the same as when it threw std:;exception ?

By default, Boost.Python translates all C++ exceptions to `RuntimeError` in Python. So, we need to use [the `register_exception_translator` function](http://www.boost.org/doc/libs/1_65_1/libs/python/doc/html/reference/high_level_components/boost_python_exception_translato.html) to translate `ClangParseError` into its Python equivalent. However, this doesn't allow us to catch the exception in Python. For that, we need to add it to the `ycm_core` module as a Python object so that we can catch it like this:
```python
try {
  ...
} except ycm_core.ClangParseError {
  ...
}
```
</blockquote></details>

Got it, thanks!

P.S. Tee hee. The example syntax in your comment is like the scary lovechild of c++ and python!

---

*[cpp/ycm/ycm_core.cpp, line 55 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDXNClaROnmZhK8JPs:-KyDXNClaROnmZhK8JPt:b-bah9aa) ([raw file](https://github.com/valloric/ycmd/blob/0c2a614f09acc408625f45a76ba67c205a000bb2/cpp/ycm/ycm_core.cpp#L55)):*
> The signature of PyErr_NewException is char *name, PyObject *base, PyObject *dict in Python 2 (but const char 

---

*[cpp/ycm/ClangCompleter/ClangUtils.cpp, line 55 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDaAg2cgrUQ_4AOvSR:-KyDaAg2cgrUQ_4AOvSS:bj2fhyt) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/ClangCompleter/ClangUtils.cpp#L55)):*
> ```cpp
>       return "Libclang crashed while parsing the translation unit.";
>     case CXError_InvalidArguments:
>       return "Parsing the translation unit with invalid arguments.";
> ```

This message doesn't seem to make sense in isolation. Perhaps it should say "Invalid compiler flags supplied for the translation unit." ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

puremourning:

---

*[cpp/ycm/PythonSupport.h, line 76 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDbmGlOYOJ1FCqFNCc:-KyDbmGlOYOJ1FCqFNCd:bb6klr3) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/PythonSupport.h#L76)):*
> ```objc
> 
> private:
>   PyObject* python_exception_;
> ```

initialise to nullptr?

```
PyObject* python_execption_{ nullptr };
````

Shrug.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

**Comment by puremourning** [comment]
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[cpp/ycm/PythonSupport.h, line 65 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDatbVpQFj6_Enzx2C:-KyDcyG_GaEGQlZ7u99D:b-py3wyb) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/PythonSupport.h#L65)):*

OK so I read up on the python API and boost python as best I could to understand this.

What I was trying to work out was this: we create this `python_exception_` thing, but who frees it, and when?

I came to the conclusion that I was not sure whether or not we need to decrement the reference count for the the exception object in the destructor of this class. It's frustratingly difficult to be sure. 

I was leaning towards thinking that perhaps the member object should be a `boost::python::handle<PyObject>` instance ? Then the assignment into the module would just be just a copy of the `handle` and the destructor of this class would decrement the reference count, and there's no chance that it gets freed before we call our `operator()` and try and use it.

I honestly don't know if that's likely/possible, but this line seems odd in that it allocates and _immediately_ deallocates a smart pointer object wrapping `python_exception_` ; that seems odd as it will increment and then (almost) immediately decrement the reference count.

Sorry that's waffly, but I'm really not that familiar with this API and the docs are _categorically_ abysmal.
</blockquote></details>

No I think i'm talking rubbish.

---

*[cpp/ycm/ycm_core.cpp, line 98 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDdGxd4Fw8f_yCPxNq:-KyDdGxd4Fw8f_yCPxNr:b-l0t6vb) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/ycm_core.cpp#L98)):*
> ```cpp
> 
> #ifdef USE_CLANG_COMPLETER
>   PythonException< ClangParseError >( "ClangParseError" );
> ```

This call instantiates the class `PythonException` but that instance goes out of scope at the end of this function. 
Does that mean that the callback ends up pointing at garbage after we leave this function ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

**Comment by micbou** [comment]

---

*[cpp/ycm/PythonSupport.h, line 76 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDbmGlOYOJ1FCqFNCc:-KyN52xiB345h-NehTuN:b-is76im) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/PythonSupport.h#L76)):*

initialise to nullptr?

```
PyObject* python_execption_{ nullptr };
````

Shrug.
</blockquote></details>

Isn't the variable initialized in the constructor?

---

*[cpp/ycm/ycm_core.cpp, line 58 at r5](https://reviewable.io:443/reviews/valloric/ycmd/861#-KxypsUj-LcEOoMhXpNf:-KyN5nD4MTrPzOfrAhx4:b-lj62fz) ([raw file](https://github.com/valloric/ycmd/blob/45b843813d266279c899aa49173f26c87d60d51e/cpp/ycm/ycm_core.cpp#L58)):*

Got it, thanks!

P.S. Tee hee. The example syntax in your comment is like the scary lovechild of c++ and python!
</blockquote></details>

Can't believe I wrote that.

---

*[cpp/ycm/ycm_core.cpp, line 98 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDdGxd4Fw8f_yCPxNq:-KyN8QvW158KQn1-7BdL:bhxxzr2) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/ycm_core.cpp#L98)):*

This call instantiates the class `PythonException` but that instance goes out of scope at the end of this function. 
Does that mean that the callback ends up pointing at garbage after we leave this function ?
</blockquote></details>

Yes but the instance is copied when passed to `boost::python::register_exception_translator` so it's fine.

---

*[cpp/ycm/ClangCompleter/ClangUtils.cpp, line 55 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDaAg2cgrUQ_4AOvSR:-KyN7-7Ml6H90zbh49mA:b-ivnmy8) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/ClangCompleter/ClangUtils.cpp#L55)):*
The thing is that it's not specific to the compiler flags. This can be any argument. The error is documented as:
> The function detected that the arguments violate the function contract.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

puremourning:

---

*[cpp/ycm/PythonSupport.h, line 76 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDbmGlOYOJ1FCqFNCc:-KzFE1CIJICOMZxwWITj:bl29vg9) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/PythonSupport.h#L76)):*

Isn't the variable initialized in the constructor?
</blockquote></details>

Oh yeah.

---

*[cpp/ycm/ycm_core.cpp, line 98 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDdGxd4Fw8f_yCPxNq:-KzFEy6Jzs-G2JYFPX1g:b-xvt10u) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/ycm_core.cpp#L98)):*

Yes but the instance is copied when passed to `boost::python::register_exception_translator` so it's fine.
</blockquote></details>

Hmmmm. Is it safe to call a copy constructor within the constructor of an object? I've been doing c++ for > 15 years and i'm honestly not sure! I can guess that it is safe in the sense that it will make a straight memcpy in this case, but it feels a little hairy. I'd prefer this to more obvious, personally.

---

*[cpp/ycm/ClangCompleter/ClangUtils.cpp, line 55 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDaAg2cgrUQ_4AOvSR:-KzFEeNDKRNQTYP-dYlo:bwt2a1n) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/ClangCompleter/ClangUtils.cpp#L55)):*

The thing is that it's not specific to the compiler flags. This can be any argument. The error is documented as:
> The function detected that the arguments violate the function contract.

</blockquote></details>

How about "Invalid arguments (e.g. compile flags) supplied when parsing the translation unit".

I would argue that as written it feels like a fragment of a sentence.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

micbou:
Reviewed 1 of 2 files at r7, 2 of 2 files at r8.

---

*[cpp/ycm/ycm_core.cpp, line 98 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDdGxd4Fw8f_yCPxNq:-Kzf72hKHCcWZC7jfKGx:b-xi56i0) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/ycm_core.cpp#L98)):*

Hmmmm. Is it safe to call a copy constructor within the constructor of an object? I've been doing c++ for > 15 years and i'm honestly not sure! I can guess that it is safe in the sense that it will make a straight memcpy in this case, but it feels a little hairy. I'd prefer this to more obvious, personally.
</blockquote></details>

I find it really convenient that we only have to instantiate an object to translate a C++ exception into Python. It won't be as clean if we add a method for that.

---

*[cpp/ycm/ClangCompleter/ClangUtils.cpp, line 55 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDaAg2cgrUQ_4AOvSR:-KzKtLJI9yDjhyOxlz1l:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/ClangCompleter/ClangUtils.cpp#L55)):*

How about "Invalid arguments (e.g. compile flags) supplied when parsing the translation unit".

I would argue that as written it feels like a fragment of a sentence.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

micbou:
Fixed the conflicts.

---

Reviewed 2 of 2 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861#-:-L1E7J6QXMWeP-Z1t_IO:b-74ttwz)*

puremourning:

---

*[cpp/ycm/PythonSupport.h, line 60 at r9](https://reviewable.io:443/reviews/valloric/ycmd/861#-L1IUzheNlsKAPTj8W7L:-L1IUzheNlsKAPTj8W7M:bb3zetx) ([raw file](https://github.com/valloric/ycmd/blob/cb91c5d41c68fb209285f777acf49c623dcfb0de/cpp/ycm/PythonSupport.h#L60)):*
> ```objc
>     // cast away constness.
>     char *raw_name = const_cast< char * >( fully_qualified_name.c_str() );
>     python_exception_ = PyErr_NewException( raw_name, base_exception, NULL );
> ```

I'm still not sure where/if this gets freed. I still feel like this is New'ing a thing, and where the Delete'ing of the thing is not clear.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

Valloric:

---

*[cpp/ycm/PythonSupport.h, line 60 at r9](https://reviewable.io:443/reviews/valloric/ycmd/861#-L1IUzheNlsKAPTj8W7L:-L1JjBD0uKHXqXHLrK_T:b9llzq7) ([raw file](https://github.com/valloric/ycmd/blob/cb91c5d41c68fb209285f777acf49c623dcfb0de/cpp/ycm/PythonSupport.h#L60)):*

I'm still not sure where/if this gets freed. I still feel like this is New'ing a thing, and where the Delete'ing of the thing is not clear.
</blockquote></details>

I know very little about direct use of the Python C API (which is why I'm a fan of Boost.Python), but I do know Python objects (which this exception is) get reference counted and thus automatically deleted when that ref count goes to zero.

So there shouldn't be any need for explicit delete calls, no? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

puremourning:

---

*[cpp/ycm/PythonSupport.h, line 60 at r9](https://reviewable.io:443/reviews/valloric/ycmd/861#-L1IUzheNlsKAPTj8W7L:-L1KKsdFXNDezAt8LY2J:b-nurttp) ([raw file](https://github.com/valloric/ycmd/blob/cb91c5d41c68fb209285f777acf49c623dcfb0de/cpp/ycm/PythonSupport.h#L60)):*

I know very little about direct use of the Python C API (which is why I'm a fan of Boost.Python), but I do know Python objects (which this exception is) get reference counted and thus automatically deleted when that ref count goes to zero.

So there shouldn't be any need for explicit delete calls, no? 
</blockquote></details>

I guess that's my point. `python_exception_` is a type of `PyObject*` (a raw pointer). It is instantiated here (via `PyErr_NewException`). We _retain_ a _reference_ (ahem, pointer) to it, but how does Python know that? I see that below, we assign a _wrapped_ version of the same pointer to `name` in the current (python) scope, which I _assume_ means that it lives as long as that scope, which should easily be long enough but we do end up with a messy dangling pointer. It just feels... imprecise, and I don't feel I understand it well enough to know if it is right/wrong/ok/irrelevant.

It's the kind of thing that wouldn't be possible in Rust ;)

My spider sense tingles when i see raw pointers new'd and i don't know where they are deleted. I suspect that we should maintain a wrapped pointer of some description, but I'm really not sure, and would rather know than assume.

Anyway, I don't want to sound like too much of a stickler here, so if everyone is cool, I'm cool. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

bstaletic:
Reviewed 2 of 2 files at r8, 1 of 2 files at r9.

---

*[cpp/ycm/PythonSupport.h, line 60 at r9](https://reviewable.io:443/reviews/valloric/ycmd/861#-L1IUzheNlsKAPTj8W7L:-L1LxOVvQddLXxeKYsUo:b9g53e2) ([raw file](https://github.com/valloric/ycmd/blob/cb91c5d41c68fb209285f777acf49c623dcfb0de/cpp/ycm/PythonSupport.h#L60)):*

I guess that's my point. `python_exception_` is a type of `PyObject*` (a raw pointer). It is instantiated here (via `PyErr_NewException`). We _retain_ a _reference_ (ahem, pointer) to it, but how does Python know that? I see that below, we assign a _wrapped_ version of the same pointer to `name` in the current (python) scope, which I _assume_ means that it lives as long as that scope, which should easily be long enough but we do end up with a messy dangling pointer. It just feels... imprecise, and I don't feel I understand it well enough to know if it is right/wrong/ok/irrelevant.

It's the kind of thing that wouldn't be possible in Rust ;)

My spider sense tingles when i see raw pointers new'd and i don't know where they are deleted. I suspect that we should maintain a wrapped pointer of some description, but I'm really not sure, and would rather know than assume.

Anyway, I don't want to sound like too much of a stickler here, so if everyone is cool, I'm cool. 
</blockquote></details>

To be honest, @puremourning made me unsure how okay this is.

As far as I know, when using python's C API, the programmer himself is responsible for reference counting.
That's why there are `Py_INCREF` and `Py_DECREF` (and others).
Here's a link: https://docs.python.org/2/c-api/refcounting.html

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

micbou:
Reviewed 2 of 2 files at r10, 1 of 1 files at r11, 3 of 3 files at r12.

---

*[cpp/ycm/PythonSupport.h, line 60 at r9](https://reviewable.io:443/reviews/valloric/ycmd/861#-L1IUzheNlsKAPTj8W7L:-L1JtTHiDIYffBHB-ys4:b-hbv441) ([raw file](https://github.com/valloric/ycmd/blob/cb91c5d41c68fb209285f777acf49c623dcfb0de/cpp/ycm/PythonSupport.h#L60)):*

To be honest, @puremourning made me unsure how okay this is.

As far as I know, when using python's C API, the programmer himself is responsible for reference counting.
That's why there are `Py_INCREF` and `Py_DECREF` (and others).
Here's a link: https://docs.python.org/2/c-api/refcounting.html
</blockquote></details>

I removed the call to `boost::python::borrowed` when adding the exception to the module. This was probably resulting in a memory leak as that call was incrementing the reference count which in turn was never decremented.  See https://wiki.python.org/moin/boost.python/handle. I am still not sure there is no memory leak but at least I know that if there is one, it's not a leak that leads to a continuous increase in memory usage when repeatedly raising and catching the exception. Therefore, I am marking the PR as ready.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

bstaletic:
Reviewed 1 of 2 files at r9, 2 of 2 files at r10, 1 of 1 files at r11, 3 of 3 files at r12.

---

*[cpp/ycm/PythonSupport.h, line 60 at r9](https://reviewable.io:443/reviews/valloric/ycmd/861#-L1IUzheNlsKAPTj8W7L:-L3itIPCQL8zkJpkAk9n:b-5pizmp) ([raw file](https://github.com/valloric/ycmd/blob/cb91c5d41c68fb209285f777acf49c623dcfb0de/cpp/ycm/PythonSupport.h#L60)):*

I removed the call to `boost::python::borrowed` when adding the exception to the module. This was probably resulting in a memory leak as that call was incrementing the reference count which in turn was never decremented.  See https://wiki.python.org/moin/boost.python/handle. I am still not sure there is no memory leak but at least I know that if there is one, it's not a leak that leads to a continuous increase in memory usage when repeatedly raising and catching the exception. Therefore, I am marking the PR as ready.
</blockquote></details>

Since @puremourning recently confirmed this is working, this does get a green light from me. :lgtm:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861)*

micbou:
Improved coverage and added comments in the tests explaining why libclang fails to parse (or reparse) the translation unit.

---

Reviewed 6 of 6 files at r13, 3 of 3 files at r14.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861#-:-L4B3oM__XjsaRWQQJLR:b-xfhyg5)*

bstaletic:
:lgtm_strong: I believe we can merge this now.

---

Reviewed 4 of 6 files at r13, 3 of 3 files at r14.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861#-:-L4Byx-MuS7zqMcXfh4W:b-dwzzcq)*

puremourning:
:lgtm: 

Thanks. I did confirm this works well 

@zzbot r+

---

Reviewed 1 of 7 files at r1, 1 of 8 files at r3, 1 of 4 files at r6, 1 of 2 files at r9, 1 of 1 files at r11, 3 of 3 files at r12, 3 of 6 files at r13, 3 of 3 files at r14.

---

*[cpp/ycm/ycm_core.cpp, line 98 at r6](https://reviewable.io:443/reviews/valloric/ycmd/861#-KyDdGxd4Fw8f_yCPxNq:-L4I7OYNZzblRfgYHHeG:b32wtd) ([raw file](https://github.com/valloric/ycmd/blob/ed3b5042c60eae997e5e563a9944bfbbc65558ce/cpp/ycm/ycm_core.cpp#L98)):*

I find it really convenient that we only have to instantiate an object to translate a C++ exception into Python. It won't be as clean if we add a method for that.
</blockquote></details>

ok

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/861#-:-L4I7jm5g9Ey85Yt4yeK:b-jp95gi)*

PR #863, opened 2017-10-23
[READY] Fix bare except statements

micbou:
Latest version of Flake8 reports the following errors:
```
ycmd\completers\all\identifier_completer.py:126:7: E722 do not use bare except'
ycmd\completers\cpp\clang_completer.py:556:3: E722 do not use bare except'
ycmd\completers\general\filename_completer.py:149:3: E722 do not use bare except'
ycmd\completers\javascript\tern_completer.py:250:5: E722 do not use bare except'
```
We fix them by catching `Exception` and logging the exception when appropriate.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/863)

### DIFF
--- a/ycmd/completers/all/identifier_completer.py
+++ b/ycmd/completers/all/identifier_completer.py
@@ -123,7 +123,9 @@ class IdentifierCompleter( GeneralCompleter ):
     for tag_file in tag_files:
       try:
         current_mtime = os.path.getmtime( tag_file )
-      except:
+      except Exception:
+        self._logger.exception(
+            'Error while getting %s last modification time.', tag_file )
         continue
       last_mtime = self._tags_file_last_mtime[ tag_file ]
 
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -30,6 +30,12 @@ import os.path
 import re
 import textwrap
 import xml.etree.ElementTree
+# Python 2.6 raises a different exception when an error occurs while parsing XML
+# from a string (e.g. an empty string).
+try:
+  from xml.etree.ElementTree import ParseError as XmlParseError
+except ImportError:
+  from xml.parsers.expat import ExpatError as XmlParseError
 
 import ycm_core
 from ycmd import responses
@@ -553,7 +559,7 @@ def _BuildGetDocResponse( doc_data ):
   # future, we can use this XML for more interesting things.
   try:
     root = xml.etree.ElementTree.fromstring( doc_data.comment_xml )
-  except:
+  except XmlParseError:
     raise ValueError( NO_DOCUMENTATION_MESSAGE )
 
   # Note: declaration is False-y if it has no child elements, hence the below
--- a/ycmd/completers/general/filename_completer.py
+++ b/ycmd/completers/general/filename_completer.py
@@ -23,6 +23,7 @@ from __future__ import absolute_import
 # Not installing aliases from python-future; it's unreliable and slow.
 from builtins import *  # noqa
 
+import logging
 import os
 import re
 from collections import defaultdict
@@ -32,6 +33,8 @@ from ycmd.utils import GetCurrentDirectory, OnWindows, ToUnicode
 from ycmd import responses
 
 EXTRA_INFO_MAP = { 1 : '[File]', 2 : '[Dir]', 3 : '[File&Dir]' }
+
+_logger = logging.getLogger( __name__ )
 
 
 class FilenameCompleter( Completer ):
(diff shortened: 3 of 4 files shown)


puremourning:
:lgtm:

Thanks. Catching exception is that different from catching nothing afaik, but at least this will help somewhat debugging. I suspect where we're re-throwing, we _probably_ don't need to log, but I don't really mind either way.

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/completers/cpp/clang_completer.py, line 558 at r1](https://reviewable.io:443/reviews/valloric/ycmd/863#-Kx9cVrZTBgNnA9AtT6A:-Kx9cVrZTBgNnA9AtT6B:b2sky9q) ([raw file](https://github.com/valloric/ycmd/blob/ff684fac90e9a07d5c08c7301e66f812cf073276/ycmd/completers/cpp/clang_completer.py#L558)):*
> ```Python
>     root = xml.etree.ElementTree.fromstring( doc_data.comment_xml )
>   except Exception:
>     _logger.exception( 'Error while parsing XML documentation.' )
> ```

I think this might be spurious (log file noise). I think this can happen simply because there is no docs (i.e. the xml is empty). I realise this is entirely my fault, but perhaps we should catch a proper exception here... Then again looking at the docs for ElementTree, it doesn't say if/what exceptions it raises, so perhaps this is catching an error where `comment_xml` isn't a member of `doc_data` ? TBH I don't remember, and I unreservedly apologise for this lame code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/863#-:-Kx9dFi4yGZaPlzXovOs:bt0espu)*

micbou:
Reviewed 4 of 4 files at r1.
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/completers/cpp/clang_completer.py, line 558 at r1](https://reviewable.io:443/reviews/valloric/ycmd/863#-Kx9cVrZTBgNnA9AtT6A:-Kx9iBsOf5vYHTYsGvsR:boc2alq) ([raw file](https://github.com/valloric/ycmd/blob/ff684fac90e9a07d5c08c7301e66f812cf073276/ycmd/completers/cpp/clang_completer.py#L558)):*

I think this might be spurious (log file noise). I think this can happen simply because there is no docs (i.e. the xml is empty). I realise this is entirely my fault, but perhaps we should catch a proper exception here... Then again looking at the docs for ElementTree, it doesn't say if/what exceptions it raises, so perhaps this is catching an error where `comment_xml` isn't a member of `doc_data` ? TBH I don't remember, and I unreservedly apologise for this lame code.
</blockquote></details>

It returns `xml.etree.ElementTree.ParseError` for an empty string so I think we should only catch this exception.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/863)*

bstaletic:
This is :lgtm: , but I'll let @puremourning trigger @zzbot, as I'm not sure if he saw the revised changes in this PR.

---

Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/863#-:-Kx9li_DMZ1ybJOWIJKz:b-aqxhuq)*

bstaletic:

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/863)*

puremourning:
:lgtm:

@zzbot r+

---

---

*[ycmd/completers/cpp/clang_completer.py, line 558 at r1](https://reviewable.io:443/reviews/valloric/ycmd/863#-Kx9cVrZTBgNnA9AtT6A:-Kx9z16oANrtzB4Cipns:btojbfd) ([raw file](https://github.com/valloric/ycmd/blob/ff684fac90e9a07d5c08c7301e66f812cf073276/ycmd/completers/cpp/clang_completer.py#L558)):*

It returns `xml.etree.ElementTree.ParseError` for an empty string so I think we should only catch this exception.
</blockquote></details>

Great!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/863#-:-Kx9z4hj8hWrnJ8YYcal:b-a2cmdb)*

micbou:
Of course, Python 2.6 raises a different exception when parsing XML so we take care of that.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/863#-:-KxCod54yuzchgg-rw0m:b-vjk0mw)*

bstaletic:
:lgtm:
@zzbot retry

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/863#-:-KxJyVufSRqxbt6OvuVt:b39btyg)*

micbou:
There was a change so `retry` doesn't work.

@zzbot r+

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/863#-:-KxKB4GeAz9VtyXsUAY-:bwevzqe)*

PR #868, opened 2017-11-01
Ignore output from tsserver that doesn't look like HTTP headers

bryanforbes:
Fixes Valloric/YouCompleteMe#2818

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/868)


micbou:
There is also a change in TSServer response when no definition is found. It now returns an empty list. See PR https://github.com/Valloric/ycmd/pull/869.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/868#-:-KxtyfBc3QrYMJmmlBM_:beqoj8g)*

micbou:
Implemented in PR #869.

PR #869, opened 2017-11-01
[READY] Support TypeScript 2.6.1

micbou:
TypeScript 2.6.1 introduced a bug in TSServer where NPM warnings are printed on stdout at startup:
```
npm WARN 2.6 No description
npm WARN 2.6 No repository field.
npm WARN 2.6 No license field.
```
See issue https://github.com/Microsoft/TypeScript/issues/19660. The TypeScript completer chokes on this output as it always expects a HTTP-like response from TSServer and assumes that the server died if not. Though it's a TSServer bug, we can handle it by catching the `ValueError` exception raised when parsing invalid headers and by interrupting the reader loop only if the server is not running.

Also, this new version of TypeScript returns an empty list when no definition is found. Update the `_GoToDefinition` and `_GoToType` functions accordingly.

Fixes https://github.com/Valloric/YouCompleteMe/issues/2818.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/869)

### DIFF
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -258,9 +258,10 @@ class TypeScriptCompleter( Completer ):
 
       try:
         message = self._ReadMessage()
-      except RuntimeError:
-        _logger.exception( SERVER_NOT_RUNNING_MESSAGE )
-        self._tsserver_is_running.clear()
+      except ( RuntimeError, ValueError ):
+        _logger.exception( 'Error while reading message from server' )
+        if not self._ServerIsRunning():
+          self._tsserver_is_running.clear()
         continue
 
       # We ignore events for now since we don't have a use for them.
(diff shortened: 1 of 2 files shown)


puremourning:
TBH i'd rather just restrict the version of tsserver that we use and point people having the issue at the bug, but we would at least have to fix the other part (about goto).

Are we certain that the npm messages will only appear during header parsing ? The code for parsing message body isn't very robust either, but is it possible for the npm message to appear after a valid `content-length: ...\n\n` but before an actual message ?

I would be surprised if it is, or if we should even worry about it, but thought i'd ask anyway.

:lgtm: assuming that and that the tests finally pass :)

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/869#-:-Kxyjw2ztvFIFWyPhujH:b-hynamb)*

bstaletic:
> Are we certain that the npm messages will only appear during header parsing ?

This is my concern too. If we can't be certain we better restrict tsserver version.
But if we can be certain, this is :lgtm: 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/869#-:-Ky11mh_BjfDqjF-NbNs:b-ds2gbe)*

micbou:
> is it possible for the npm message to appear after a valid content-length: ...\n\n but before an actual message ?

Looking at TSServer code, [the whole response is written at once on stdout](https://github.com/Microsoft/TypeScript/blob/f67a9ba96e26c0d142934dc73d55d727bd238ae8/src/server/server.ts#L320) (see [`formatMessage`](https://github.com/Microsoft/TypeScript/blob/cf2bdf8250efa062ec8d9df740e49f1778183707/src/server/session.ts#L138)) so it shouldn't be possible.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/869#-:-Ky1lz5J2ouFs2_2souk:b-4xgo6g)*

puremourning:
Shall we ship it then? Fire when ready!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/869#-:-Ky2IXlslRBDT02GYCYa:b-y68ljy)*

micbou:
@zzbot r+

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/869#-:-Ky2JGHF92PLlaMczxGt:bi59kwd)*

PR #873, opened 2017-11-08
[WIP] Cache includes from system include paths.

davits:
Caches include entries extracted from include paths specified by -isystem flag.
On corporate environments where everything is on NFS drives and there are lot of include paths and headers it reduces include completion from several seconds to instantaneous.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/873)

### DIFF
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -184,6 +184,8 @@ class ClangCompleter( Completer ):
          self._GoToInclude( request_data ) ),
       'ClearCompilationFlagCache': ( lambda self, request_data, args:
          self._ClearCompilationFlagCache() ),
+      'ClearIncludeCache'        : ( lambda self, request_data, args:
+         self._ClearIncludeCache() ),
       'GetType'                  : ( lambda self, request_data, args:
          self._GetSemanticInfo( request_data, func = 'GetTypeAtLocation' ) ),
       'GetTypeImprecise'         : ( lambda self, request_data, args:
@@ -341,6 +343,7 @@ class ClangCompleter( Completer ):
 
     return response_builder( message )
 
+
   def _ClearCompilationFlagCache( self ):
     self._flags.Clear()
 
@@ -369,6 +372,11 @@ class ClangCompleter( Completer ):
     # in a nice way
 
     return responses.BuildFixItResponse( fixits )
+
+
+  def _ClearIncludeCache( self ):
+    self._include_cache.Clear()
+
 
   def OnFileReadyToParse( self, request_data ):
     filename = request_data[ 'filepath' ]
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -45,8 +45,7 @@ from ycmd.completers.cpp.flags import ( Flags, PrepareFlagsForClang,
                                         NoCompilationDatabase,
                                         UserIncludePaths )
 from ycmd.completers.cpp.ephemeral_values_set import EphemeralValuesSet
-from ycmd.completers.general.filename_completer import (
-    GenerateCandidatesForPaths )
+from ycmd.completers.cpp.include_cache import IncludeCache, IncludeList
 from ycmd.responses import NoExtraConfDetected, UnknownExtraConf
 
 CLANG_FILETYPES = set( [ 'c', 'cpp', 'objc', 'objcpp' ] )
@@ -68,6 +67,7 @@ class ClangCompleter( Completer ):
       'max_diagnostics_to_display' ]
     self._completer = ycm_core.ClangCompleter()
     self._flags = Flags()
+    self._include_cache = IncludeCache()
     self._diagnostic_store = None
     self._files_being_compiled = EphemeralValuesSet()
     self._logger = logging.getLogger( __name__ )
@@ -127,20 +127,13 @@ class ClangCompleter( Completer ):
     if quoted_include:
       include_paths.extend( quoted_include_paths )
 
-    paths = []
-    for include_path in include_paths:
-      unicode_path = ToUnicode( os.path.join( include_path, path_dir ) )
-      try:
-        # We need to pass a unicode string to get unicode strings out of
-        # listdir.
-        relative_paths = os.listdir( unicode_path )
-      except Exception:
-        self._logger.exception( 'Error while listing %s folder.', unicode_path )
-        relative_paths = []
+    includes = IncludeList()
+    for include in include_paths:
+      unicode_path = ToUnicode( os.path.join( include[ 0 ], path_dir ) )
+      includes.AddIncludes(
+        self._include_cache.GetIncludes( unicode_path, include[ 1 ] ) )
 
-      paths.extend( os.path.join( include_path, path_dir, relative_path ) for
-                    relative_path in relative_paths  )
-    return paths
+    return includes.GetIncludes()
 
 
   def ComputeCandidatesInner( self, request_data ):
@@ -148,9 +141,9 @@ class ClangCompleter( Completer ):
     if not filename:
       return
 
-    paths = self.GetIncludePaths( request_data )
-    if paths is not None:
-      return GenerateCandidatesForPaths( paths )
+    includes = self.GetIncludePaths( request_data )
+    if includes is not None:
+      return includes
 
     if self._completer.UpdatingTranslationUnit(
         ToCppStringCompatible( filename ) ):
@@ -575,9 +568,9 @@ def _BuildGetDocResponse( doc_data ):
       ToUnicode( _FormatRawComment( doc_data.raw_comment ) ) ) )
 
 
-def _GetAbsolutePath( include_file_name, include_paths ):
-  for path in include_paths:
-    include_file_path = os.path.join( path, include_file_name )
+def _GetAbsolutePath( include_file_name, includes ):
+  for include in includes:
+    include_file_path = os.path.join( include[ 0 ], include_file_name )
     if os.path.isfile( include_file_path ):
       return include_file_path
   return None
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -589,35 +589,29 @@ def _GetCompilationInfoForFile( database, file_name, file_extension ):
   return None
 
 
-def UserIncludePaths( flags, filename ):
-  quoted_include_paths = [ os.path.dirname( filename ) ]
+def UserIncludePaths( user_flags, filename ):
+  quoted_include_paths = [ ( ToUnicode( os.path.dirname( filename ) ), False ) ]
   include_paths = []
 
-  if flags:
-    quote_flag = '-iquote'
-    path_flags = [ '-isystem', '-I' ]
+  if user_flags:
+    include_flags = { '-iquote':  ( quoted_include_paths, False ),
+                      '-I':       ( include_paths, False ),
+                      '-isystem': ( include_paths, True ) }
 
     try:
-      it = iter( flags )
-      for flag in it:
-        flag_len = len( flag )
-        if flag.startswith( quote_flag ):
-          quote_flag_len = len( quote_flag )
-          # Add next flag to the include paths if current flag equals to
-          # '-iquote', or add remaining string otherwise.
-          quoted_include_path = ( next( it ) if flag_len == quote_flag_len else
-                                  flag[ quote_flag_len: ] )
-          if quoted_include_path:
-            quoted_include_paths.append( ToUnicode( quoted_include_path ) )
-        else:
-          for path_flag in path_flags:
-            if flag.startswith( path_flag ):
-              path_flag_len = len( path_flag )
-              include_path = ( next( it ) if flag_len == path_flag_len else
-                               flag[ path_flag_len: ] )
-              if include_path:
-                include_paths.append( ToUnicode( include_path ) )
-              break
+      it = iter( user_flags )
+      for user_flag in it:
+        user_flag_len = len( user_flag )
+        for flag in include_flags:
+          if user_flag.startswith( flag ):
+            flag_len = len( flag )
+            include_path = ( next( it ) if user_flag_len == flag_len else
+                             user_flag[ flag_len: ] )
+            if include_path:
+              container = include_flags[ flag ]
+              container[ 0 ].append( ( ToUnicode( include_path ),
+                                       container[ 1 ] ) )
+            break
     except StopIteration:
       pass
 
--- a/ycmd/completers/cpp/include_cache.py
+++ b/ycmd/completers/cpp/include_cache.py
@@ -1,0 +1,92 @@
+# Copyright (C) 2017 Davit Samvelyan davitsamvelyan@gmail.com
+#                    Synopsys.
+#
+# This file is part of ycmd.
+#
+# ycmd is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ycmd is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+# Not installing aliases from python-future; it's unreliable and slow.
+from builtins import *  # noqa
+
+from future.utils import iteritems
+
+import os
+
+from collections import defaultdict
+from ycmd import responses
+
+EXTRA_INFO_MAP = { 1 : '[File]', 2 : '[Dir]', 3 : '[File&Dir]' }
+
+
+class IncludeEntry( object ):
+
+  def __init__( self, name, entry_type ):
+    self.name = name
+    self.entry_type = entry_type
+
+
+class IncludeList( object ):
+
+  def __init__( self ):
+    self._includes = defaultdict( int )
+
+
+  def AddIncludes( self, includes ):
+    for include in includes:
+      self._includes[ include.name ] |= include.entry_type
+
+
+  def GetIncludes( self ):
+    includes = []
+    for name, include_type in iteritems( self._includes ):
+      includes.append( responses.BuildCompletionData(
+        name, EXTRA_INFO_MAP[ include_type ] ) )
+    return includes
+
+
+class IncludeCache( object ):
+
+  def __init__( self ):
+    self._cache = {}
+
+
+  def GetIncludes( self, path, cache ):
+    includes = None
+    if cache:
+      includes = self._cache.get( path )
+
+    if includes is None:
+      includes = self._GetIncludes( path )
+      if cache:
+        self._cache[ path ] = includes
+    return includes
+
+
+  def _GetIncludes( self, path ):
+    try:
+      names = os.listdir( path )
+    except Exception:
+      names = []
+
+    includes = []
+    for name in names:
+      inc_path = os.path.join(path, name)
+      entry_type = 2 if os.path.isdir( inc_path ) else 1
+      includes.append( IncludeEntry( name, entry_type ) )
+
+    return includes
--- a/ycmd/completers/cpp/include_cache.py
+++ b/ycmd/completers/cpp/include_cache.py
@@ -29,11 +29,15 @@ import os
 
 from collections import defaultdict
 from ycmd import responses
-
-EXTRA_INFO_MAP = { 1 : '[File]', 2 : '[Dir]', 3 : '[File&Dir]' }
+from ycmd.completers.general.filename_completer import ( GetPathType,
+                                                         GetPathTypeName )
 
 
 class IncludeEntry( object ):
+  """ Represents single include completion candidate.
+  name is the name/string of the completion candidate,
+  entry_type is an integer indicating whether the candidate is a
+  'File', 'Dir' or both (See EXTRA_INFO_MAP in filename_completer). """
 
   def __init__( self, name, entry_type ):
     self.name = name
@@ -41,6 +45,8 @@ class IncludeEntry( object ):
 
 
 class IncludeList( object ):
+  """ Helper class fo combining include completion candidates from
+  several include paths. """
 
   def __init__( self ):
     self._includes = defaultdict( int )
@@ -55,7 +61,7 @@ class IncludeList( object ):
     includes = []
     for name, include_type in iteritems( self._includes ):
       includes.append( responses.BuildCompletionData(
-        name, EXTRA_INFO_MAP[ include_type ] ) )
+        name, GetPathTypeName( include_type ) ) )
     return includes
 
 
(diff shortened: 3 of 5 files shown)


puremourning:
Haven’t looked in detail yet, but fist thanks for looking at this! Filesystem metadata perf is a problem for me too. 

Second I was wondering how/if the cache is invalidated. I often have Vim running for days and thing s like overnight builds etc will change the dependencies. I wonder if we can be smart and c the mtime for the dir?

bstaletic:
@puremourning As it is now, the cache is never invalidated, but only `-isystem` flags are cached. Those shouldn't change over night.

@davits Tomorrow I should be able to test this over a remote sshfs. Should be a fun exercise.

---

Reviewed 3 of 3 files at r1.

---

*[ycmd/completers/cpp/include_cache.py, line 33 at r1](https://reviewable.io:443/reviews/valloric/ycmd/873#-KyU9glMmysgedoSPc-8:-KyU9glNOXMBValnhvIV:b-z4atif) ([raw file](https://github.com/valloric/ycmd/blob/c6b85fd4bb515f665b1f4832a64e8274bb72dae1/ycmd/completers/cpp/include_cache.py#L33)):*
> ```Python
> from ycmd import responses
> 
> EXTRA_INFO_MAP = { 1 : '[File]', 2 : '[Dir]', 3 : '[File&Dir]' }
> ```

Why do we need the third option? Something is either a file or a directory. Even in the case of a symlink to a directory, python's `os.path` module treats it as a directory.

I understand that the `self._includes[ include.name ] |= include.entry_type` can end up with a value of 1, 2 or , but I don't see a reason not to use `=` instead of `|=`.

---

*[ycmd/completers/cpp/include_cache.py, line 39 at r1](https://reviewable.io:443/reviews/valloric/ycmd/873#-KyUfRxnGMIAXncxo056:-KyUfRxofMNHnmqtw2fe:b-lsbluo) ([raw file](https://github.com/valloric/ycmd/blob/c6b85fd4bb515f665b1f4832a64e8274bb72dae1/ycmd/completers/cpp/include_cache.py#L39)):*
> ```Python
> 
>   def __init__( self, name, entry_type ):
>     self.name = name
> ```

I think all the classes here should have some kind of comment/docstring, but `name` in particular should be explained.

Is it a full path to a file inside include path?
Is it a relative path?
Or is it whatever `GetInclompleteIncludeValue()` returns as `include_value` + a file/directory in `include_value` directory (which is what a user would have in his/her `#include` statement)?
Or perhaps it is a path from user's `-I`, `-isystem` or `-iquote` flag + `GetInclompleteIncludeValue()`'s `include_value` + something from that directory?

---

*[ycmd/completers/cpp/include_cache.py, line 84 at r1](https://reviewable.io:443/reviews/valloric/ycmd/873#-KyUYVRcFe4m_usqb_aM:-KyUYVRdY66Op_hzwFY2:bie6ked) ([raw file](https://github.com/valloric/ycmd/blob/c6b85fd4bb515f665b1f4832a64e8274bb72dae1/ycmd/completers/cpp/include_cache.py#L84)):*
> ```Python
>       names = os.listdir( path )
>     except Exception:
>       names = []
> ```

If an exception is raised, `names` will be [] and the `for` loop will iterate over an empty list.
So this can be an early `return []`, but I'm not sure if that would hurt readability.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/873#-:-KyUTLg6-3JeZTOwxQow:b-7o483z)*

davits:
Good idea! That didn't go though my mind since I restart my vim sessions pretty often :)

---

---

*[ycmd/completers/cpp/include_cache.py, line 33 at r1](https://reviewable.io:443/reviews/valloric/ycmd/873#-KyU9glMmysgedoSPc-8:-KyUpGO73kJJL69Pobkj:b-8e68r8) ([raw file](https://github.com/valloric/ycmd/blob/c6b85fd4bb515f665b1f4832a64e8274bb72dae1/ycmd/completers/cpp/include_cache.py#L33)):*

Why do we need the third option? Something is either a file or a directory. Even in the case of a symlink to a directory, python's `os.path` module treats it as a directory.

I understand that the `self._includes[ include.name ] |= include.entry_type` can end up with a value of 1, 2 or , but I don't see a reason not to use `=` instead of `|=`.
</blockquote></details>

This logic was introduced long time ago by me and the reason was Qt header structure and the way we specify Qt include paths.
If both `qt/include` and `qt/include/QtCore` are added to the include paths, when one types "QtCore" it is both file and directory.
But it can be removed from the path completion logic. I'll see to it.

---

*[ycmd/completers/cpp/include_cache.py, line 39 at r1](https://reviewable.io:443/reviews/valloric/ycmd/873#-KyUfRxnGMIAXncxo056:-KyUtVLeq2U7UdkX7N6A:bvg3zc6) ([raw file](https://github.com/valloric/ycmd/blob/c6b85fd4bb515f665b1f4832a64e8274bb72dae1/ycmd/completers/cpp/include_cache.py#L39)):*

I think all the classes here should have some kind of comment/docstring, but `name` in particular should be explained.

Is it a full path to a file inside include path?
Is it a relative path?
Or is it whatever `GetInclompleteIncludeValue()` returns as `include_value` + a file/directory in `include_value` directory (which is what a user would have in his/her `#include` statement)?
Or perhaps it is a path from user's `-I`, `-isystem` or `-iquote` flag + `GetInclompleteIncludeValue()`'s `include_value` + something from that directory?
</blockquote></details>

Include entry is a single file/dir from include path, it basically becomes single completion candidate.
Docs aren't my strongest side, I'll try to add some documentation.

---

*[ycmd/completers/cpp/include_cache.py, line 84 at r1](https://reviewable.io:443/reviews/valloric/ycmd/873#-KyUYVRcFe4m_usqb_aM:-KyUsfEn-U0E-6t5RW17:b32w0x) ([raw file](https://github.com/valloric/ycmd/blob/c6b85fd4bb515f665b1f4832a64e8274bb72dae1/ycmd/completers/cpp/include_cache.py#L84)):*

If an exception is raised, `names` will be [] and the `for` loop will iterate over an empty list.
So this can be an early `return []`, but I'm not sure if that would hurt readability.
</blockquote></details>

OK

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/873#-:-KyUrIDzqtzuz46wLFUv:bclfxx2)*

bstaletic:

---

*[ycmd/completers/cpp/include_cache.py, line 33 at r1](https://reviewable.io:443/reviews/valloric/ycmd/873#-KyU9glMmysgedoSPc-8:-KyUzvxsNOokTHnFr4uR:bulfub0) ([raw file](https://github.com/valloric/ycmd/blob/c6b85fd4bb515f665b1f4832a64e8274bb72dae1/ycmd/completers/cpp/include_cache.py#L33)):*
<details><summary><i>Previously, davits (Davit Samvelyan) wrote…</i></summary><blockquote>

This logic was introduced long time ago by me and the reason was Qt header structure and the way we specify Qt include paths.
If both `qt/include` and `qt/include/QtCore` are added to the include paths, when one types "QtCore" it is both file and directory.
But it can be removed from the path completion logic. I'll see to it.
</blockquote></details>

I've never worked with Qt, so I didn't know about its file hierarchy. Since there is a case where the same looking completion can represent a file and a directory, I'm fine with leaving this as is.

Alternative would probably mean having two completion entries with the same text that should be inserted, which would be worse, as it would pollute the completion menu.

---

*[ycmd/completers/cpp/include_cache.py, line 39 at r1](https://reviewable.io:443/reviews/valloric/ycmd/873#-KyUfRxnGMIAXncxo056:-KyV-VUGpunauIyor3xt:b1hecn2) ([raw file](https://github.com/valloric/ycmd/blob/c6b85fd4bb515f665b1f4832a64e8274bb72dae1/ycmd/completers/cpp/include_cache.py#L39)):*
<details><summary><i>Previously, davits (Davit Samvelyan) wrote…</i></summary><blockquote>

Include entry is a single file/dir from include path, it basically becomes single completion candidate.
Docs aren't my strongest side, I'll try to add some documentation.
</blockquote></details>

I managed to understand on my own what `name` and thus `IncludeEntry` is. I was just trying to point out the need for docs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/873)*

bstaletic:
For the record, I did give this a test, but specifying `-isystem` flags to point to my laptop a couple of blocks away over sshfs didn't slow ycmd's header completion enough for me to notice anything.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/873#-:-KzA4KgsbXssxwnskXkh:br522cw)*

davits:
@bstaletic, thanks for the test. I suppose it takes couple of several other factors like hard disc and/or network load to slow down header completion.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/873#-:-KzD1YXddEPl3itrGXdW:b-356c06)*

davits:
"couple of several" :smile:  sorry for my English.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/873#-:-KzD2NGMX9Ao2bHHZYeI:b-1po8zk)*

bstaletic:
I'm mostly satisfied with this right now, but I've left a few more comments.

Unfortunately I still have no way of testing this anywhere where include path completion would be slow, so I wonder what will @puremourning say.

---

Reviewed 4 of 4 files at r2.

---

*[ycmd/completers/cpp/clang_completer.py, line 188 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L-4-61VNYroPa3pSIwS:-L-3UWom_LVpp0YpIfEA:b-hxw590) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/clang_completer.py#L188)):*
> ```Python
>          self._ClearCompilationFlagCache() ),
>       'ClearIncludeCache'        : ( lambda self, request_data, args:
>          self._ClearIncludeCache() ),
> ```

Do we need the new subcommand with the `mtime` logic in place?

---

*[ycmd/completers/cpp/include_cache.py, line 71 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L-4-_plpCaBg2GziMgD:-L-4-_plpCaBg2GziMgE:bh5yn6h) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/include_cache.py#L71)):*
> ```Python
> 
>   def __init__( self ):
>     self._cache = {}
> ```

It's not quite obvious what does this dictionary actually contain.
It contains `{ 'dir': ( mtime, [ IncludeEntry_list_appropriate_for_the_dir ] ) }`.
A comment about that would be helpful. Reading the code I had to go back to this a lot of times to remind myself what thte dictionary looks like.

---

*[ycmd/completers/general/filename_completer.py, line 170 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L-3yijpo8SoN2xkeLPY:-L-3yijq-mxldd5QOaRC:bo2w9g7) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/general/filename_completer.py#L170)):*
> ```Python
>       pass
> 
>   return entries
> ```

Is there a reason this is now returning a list instead of a generator object?

If not then I think we should kee[ the original:
```python
  return ( ( relative_path, GetPathType( absolute_path ) )
           for relative_path in relative_paths )
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/873#-:-L-41PkozA_IEDjjyBtL:b8ta16j)*

davits:
Thanks for the review. Waiting for @puremourning to test this out.

---

---

*[ycmd/completers/cpp/clang_completer.py, line 188 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L-4-61VNYroPa3pSIwS:-L-FzEifnW3lqE7ue-mU:b-6sbmov) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/clang_completer.py#L188)):*

Do we need the new subcommand with the `mtime` logic in place?
</blockquote></details>

There might be some corner cases when system include folders are not re-created during nightly builds but, files added/removed to them, which will not change mtime.

---

*[ycmd/completers/general/filename_completer.py, line 170 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L-3yijpo8SoN2xkeLPY:-L-FxzDcYaL_vslLFVYb:by9mmg) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/general/filename_completer.py#L170)):*

Is there a reason this is now returning a list instead of a generator object?

If not then I think we should kee[ the original:
```python
  return ( ( relative_path, GetPathType( absolute_path ) )
           for relative_path in relative_paths )
```
</blockquote></details>

My coding/thinking style is somewhat C++-ish :)
What happens when there is an exception during generation? I met such case when LANG wasn't set to UTF-8 on my system.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/873#-:-L-FzYnFUfQ-g-3ZBhQ8:b-ykcze0)*

bstaletic:

---

*[ycmd/completers/general/filename_completer.py, line 170 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L-3yijpo8SoN2xkeLPY:-L-JTX_Kv89FMRY3PcrX:b-vxl94d) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/general/filename_completer.py#L170)):*
<details><summary><i>Previously, davits (Davit Samvelyan) wrote…</i></summary><blockquote>

My coding/thinking style is somewhat C++-ish :)
What happens when there is an exception during generation? I met such case when LANG wasn't set to UTF-8 on my system.
</blockquote></details>

You can also do something like:

```python
try:
    return expected_object
except Exception:
    return default_because_there_was_an_exception
```
But with this change, I think your solution is more readable and thus favourable. So feel free to implement it either way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/873)*

Valloric:
@puremourning This PR appears to be blocked on you taking a look. :)

puremourning:
Oh Sorry, I saw `[WIP]` in the title and somewhat, err, filed it under 'to be looked at later'

I will add it to the 'to be looked at asap' pile in my brain :)

IIRC my comment was about invalidating the cache, but as this only looks at _system_ headers (`-isystem`) that seems reasonable to essentially "never" invalidate the cache.

puremourning:
OK I took a look. Thanks a _bunch_ for working on this and thanks _even more_ for your saint-like patience.

I guess there are two points in addition to some nits:

1. Now that we are invalidating the cache based on the directory mtime, is there any reason not to go whole hog and cache _every_ include dir? That would be _fantastic_ for me, who has _huge_ dependencies which are included by `-I` and on remote SAN

2. I feel like the PR needs more tests; there's quite a lot of new code and new behaviours and I am sceptical that the updated tests really cover it functionally.

---

Review status: all files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cpp/clang_completer.py, line 132 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1KZl-q0fm8rM1yZqvN:-L1KZl-q0fm8rM1yZqvO:bmb3zu8) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/clang_completer.py#L132)):*
> ```Python
>     includes = IncludeList()
>     for include in include_paths:
>       unicode_path = ToUnicode( os.path.join( include[ 0 ], path_dir ) )
> ```

Is there any way here to avoid naked use of `[ 0 ]` and `[ 1 ]` ? It's not self-documenting what these things are. Can we either use a dict, object or simply assign some local vars in the loop like:

```python
thing_one = include[ 0 ]
other_thing = include[ 1 ]

... use them
```

---

*[ycmd/completers/cpp/clang_completer.py, line 581 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1K_fplI5yEdRtiCYmC:-L1K_fplI5yEdRtiCYmD:b-eovtn0) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/clang_completer.py#L581)):*
> ```Python
> def _GetAbsolutePath( include_file_name, includes ):
>   for include in includes:
>     include_file_path = os.path.join( include[ 0 ], include_file_name )
> ```

same here, could we give `include[ 0 ]` a name ?

---

*[ycmd/completers/cpp/flags.py, line 592 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1K_rdst6w8yY95oEbY:-L1K_rdst6w8yY95oEbZ:b-s6r1o8) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/flags.py#L592)):*
> ```Python
> 
> 
> def UserIncludePaths( user_flags, filename ):
> ```

Can we write a doctoring which describes what this returns now? Previously it was simply a list of paths. Now it is al list of tuples. Tuples are a pain because they contain untagged arbitrary data. We need to document what the tuple contains.

---

*[ycmd/completers/cpp/flags.py, line 597 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1KaiAXsx0FKvWi9fIP:-L1KaiAXsx0FKvWi9fIQ:b-z6tpo) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/flags.py#L597)):*
> ```Python
> 
>   if user_flags:
>     include_flags = { '-iquote':  ( quoted_include_paths, False ),
> ```

this looks like yet-another-place where we have a list of flags which are "include flags". There are others at the top of this file. Should would consider consolidating them? If it makes the rest of the code unnecessarily complex then perhaps not, but I worry that we might forget to update this list when, for example adding support for `driver-mode=cl`

---

*[ycmd/completers/cpp/flags.py, line 605 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1KaN4dJeqLiP8sNBSb:-L1KaN4dJeqLiP8sNBSc:b-4tyq27) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/flags.py#L605)):*
> ```Python
>       for user_flag in it:
>         user_flag_len = len( user_flag )
>         for flag in include_flags:
> ```

shouldn't this use `iterkeys` ?

---

*[ycmd/completers/cpp/include_cache.py, line 48 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1Kblz7lvaNRubbqkkU:-L1Kblz8P67peawuiF_I:b-bgxys4) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/include_cache.py#L48)):*
> ```Python
> 
> class IncludeList( object ):
>   """ Helper class fo combining include completion candidates from
> ```

fo -> for

---

*[ycmd/completers/cpp/include_cache.py, line 78 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1Kc9a_zEXiGl1axVrG:-L1Kc9a_zEXiGl1axVrH:b6pm7bn) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/include_cache.py#L78)):*
> ```Python
> 
> 
>   def GetIncludes( self, path, cache ):
> ```

should `cache` really be called `use_cache` ? And if so, is there merit in defaulting to `True` ?

---

*[ycmd/completers/cpp/include_cache.py, line 102 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1KcifPLYE8J_gSKE_r:-L1KcifPLYE8J_gSKE_s:bm1cqh3) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/include_cache.py#L102)):*
> ```Python
>       mtime = _GetModificationTime( path )
>       if mtime > cache_entry[ 0 ]:
>         del self._cache[ path ]
> ```

any way we can just update the cache ? This saves on one call to `_GetModificationTime` which is a filesystem metadata call, so it isn't free.

---

*[ycmd/completers/cpp/include_cache.py, line 112 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1Kd1o3dJenHdDcokVs:-L1Kd1o3dJenHdDcokVt:b7cnh18) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/include_cache.py#L112)):*
> ```Python
>     try:
>       names = os.listdir( path )
>     except Exception:
> ```

can we catch the proper exceptions by name, rather than bare `Exception` ?

---

*[ycmd/completers/cpp/include_cache.py, line 117 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1KdCF7CKmVa0B2fWkD:-L1KdCF7CKmVa0B2fWkE:b-yh9bgw) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/include_cache.py#L117)):*
> ```Python
>     includes = []
>     for name in names:
>       inc_path = os.path.join(path, name)
> ```

spaces inside `(` and `)`

---

*[ycmd/completers/cpp/include_cache.py, line 121 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1Kd6uX9yBNG9HuT3y8:-L1Kd6uX9yBNG9HuT3y9:b9cxq6q) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/include_cache.py#L121)):*
> ```Python
>         entry_type = GetPathType( inc_path )
>         includes.append( IncludeEntry( name, entry_type ) )
>       except Exception:
> ```

Prefer actual exception type

---

*[ycmd/completers/cpp/include_cache.py, line 130 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1KcOEap8wqJKwrYgmG:-L1KcOEap8wqJKwrYgmH:b-k9gmwt) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/include_cache.py#L130)):*
> ```Python
>   try:
>     return os.path.getmtime( path )
>   except Exception:
> ```

we strongly avoid catching bare `Exception`. If `os.path.getmtime` throws `OSError`, then catch that explicitly and let any other random exception bubble.

---

*[ycmd/completers/general/filename_completer.py, line 37 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1KdOlThs5npVGVbSji:-L1KdOlThs5npVGVbSjj:bb6zqix) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/general/filename_completer.py#L37)):*
> ```Python
> DIR = 2
> # This mapping is also used for the #include completion.
> # We have option N3, because when using Qt with specific include paths
> ```

N3 ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/873#-:-L1KdW0AUFYEnA353U_8:b-80ovcm)*

bstaletic:
> I feel like the PR needs more tests;

Now I'm wondering how would we even test if something is cached properly.

---

Review status: all files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cpp/include_cache.py, line 112 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1Kd1o3dJenHdDcokVs:-L1LqWXngL51m-mMfFc3:b-58epy8) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/cpp/include_cache.py#L112)):*

can we catch the proper exceptions by name, rather than bare `Exception` ?
</blockquote></details>

This should be `OSError`, if I'm not mistaken.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/873#-:-L1LpW7RA1H_6r3HEy3g:byvikmj)*

bstaletic:
Review status: all files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/general/filename_completer.py, line 37 at r2](https://reviewable.io:443/reviews/valloric/ycmd/873#-L1KdOlThs5npVGVbSji:-L1QamkI0qHFAUKXJGVV:b-74yofl) ([raw file](https://github.com/valloric/ycmd/blob/054cf68c8fdab2366506fc6f4fc6fa611f61840f/ycmd/completers/general/filename_completer.py#L37)):*

N3 ?
</blockquote></details>

`option 3` I believe. Two lines below it says `3 : '[File&Dir]'`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/873)*

davits:
Sorry guys I can't make up time to work on this :(
If anyone is interested, you can take it over from here.

PR #875, opened 2017-11-20
[READY] Allow switching to a different JavaScript project with RestartServer

micbou:
Contrarily to what the YCM docs say, issuing the `:YcmCompleter RestartServer` command to switch to a different JavaScript project doesn't work. The reason is that we don't set the working directory when starting the Tern server. Same issue if the user starts the client outside a project and then open a file in that project. It's even worse in that case because no warning will be displayed to the user.

I considered passing the `request_data` object when initializing the completer but that doesn't work because the request is not always available. I also considered adding support for multiple projects similarly to what the C# completer does but I felt it was too much work. Finally, I went for starting the server on the `FileReadyToParse` event.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/875)

### DIFF
--- a/ycmd/completers/javascript/tern_completer.py
+++ b/ycmd/completers/javascript/tern_completer.py
@@ -90,10 +90,8 @@ def GlobalConfigExists( tern_config ):
 
 
 def FindTernProjectFile( starting_directory ):
-  """Finds the path to either a Tern project file or the user's global Tern
-  configuration file. If found, a tuple is returned containing the path and a
-  boolean indicating if the path is to a .tern-project file. If not found,
-  returns (None, False)."""
+  """Returns the path to either a Tern project file or the user's global Tern
+  configuration file."""
   for folder in utils.PathsToAllParentFolders( starting_directory ):
     tern_project = os.path.join( folder, '.tern-project' )
     if os.path.exists( tern_project ):
@@ -105,7 +103,7 @@ def FindTernProjectFile( starting_directory ):
   # don't warn if we find one. The point is that if the user has a .tern-config
   # set up, then she has deliberately done so and a ycmd warning is unlikely
   # to be anything other than annoying.
-  tern_config = os.path.expanduser( '~/.tern-config' )
+  tern_config = os.path.join( os.path.expanduser( '~' ), '.tern-config' )
   if GlobalConfigExists( tern_config ):
     return tern_config
 
@@ -439,9 +437,9 @@ class TernCompleter( Completer ):
             cwd = self._server_working_dir )
 
       if self._ServerIsRunning():
-        _logger.info( 'Tern Server started with pid: %s listening on port %s',
+        _logger.info( 'Tern Server started with pid %d listening on port %d',
                       self._server_handle.pid, self._server_port )
-        _logger.info( 'Tern Server log files are: %s and %s',
+        _logger.info( 'Tern Server log files are %s and %s',
                       self._server_stdout, self._server_stderr )
 
         self._do_tern_project_check = True
@@ -473,6 +471,8 @@ class TernCompleter( Completer ):
 
   def _CleanUp( self ):
     utils.CloseStandardStreams( self._server_handle )
+
+    self._do_tern_project_check = False
 
     self._server_handle = None
     self._server_port = None
--- a/ycmd/completers/javascript/tern_completer.py
+++ b/ycmd/completers/javascript/tern_completer.py
@@ -97,7 +97,7 @@ def FindTernProjectFile( starting_directory ):
   for folder in utils.PathsToAllParentFolders( starting_directory ):
     tern_project = os.path.join( folder, '.tern-project' )
     if os.path.exists( tern_project ):
-      return ( tern_project, True )
+      return tern_project
 
   # As described here: http://ternjs.net/doc/manual.html#server a global
   # .tern-config file is also supported for the Tern server. This can provide
@@ -107,9 +107,9 @@ def FindTernProjectFile( starting_directory ):
   # to be anything other than annoying.
   tern_config = os.path.expanduser( '~/.tern-config' )
   if GlobalConfigExists( tern_config ):
-    return ( tern_config, False )
+    return tern_config
 
-  return ( None, False )
+  return None
 
 
 class TernCompleter( Completer ):
@@ -127,21 +127,17 @@ class TernCompleter( Completer ):
 
     self._do_tern_project_check = False
 
-    # Used to determine the absolute path of files returned by the tern server.
-    # When a .tern_project file exists, paths are returned relative to it.
-    # Otherwise, they are returned relative to the working directory of the tern
-    # server.
-    self._server_paths_relative_to = None
-
     self._server_handle = None
     self._server_port = None
     self._server_stdout = None
     self._server_stderr = None
 
-    self._StartServer()
+    self._server_started = False
+    self._server_working_dir = None
+    self._server_project_file = None
 
 
-  def _WarnIfMissingTernProject( self ):
+  def _WarnIfMissingTernProject( self, request_data ):
     # The Tern server will operate without a .tern-project file. However, it
     # does not operate optimally, and will likely lead to issues reported that
     # JavaScript completion is not working properly. So we raise a warning if we
@@ -150,34 +146,19 @@ class TernCompleter( Completer ):
     # We do this check after the server has started because the server does
     # have nonzero use without a project file, however limited. We only do this
     # check once, though because the server can only handle one project at a
-    # time. This doesn't catch opening a file which is not part of the project
-    # or any of those things, but we can only do so much. We'd like to enhance
-    # ycmd to handle this better, but that is a FIXME for now.
-    if self._ServerIsRunning() and self._do_tern_project_check:
-      self._do_tern_project_check = False
+    # time.
+    if not self._ServerIsRunning() or not self._do_tern_project_check:
+      return
 
-      current_dir = utils.GetCurrentDirectory()
-      ( tern_project, is_project ) = FindTernProjectFile( current_dir )
-      if not tern_project:
-        _logger.warning( 'No .tern-project file detected: ' + current_dir )
-        raise RuntimeError( 'Warning: Unable to detect a .tern-project file '
-                            'in the hierarchy before ' + current_dir +
-                            ' and no global .tern-config file was found. '
-                            'This is required for accurate JavaScript '
-                            'completion. Please see the User Guide for '
-                            'details.' )
-      else:
-        _logger.info( 'Detected Tern configuration file at: ' + tern_project )
-
-        # Paths are relative to the project file if it exists, otherwise they
-        # are relative to the working directory of Tern server (which is the
-        # same as the working directory of ycmd).
-        self._server_paths_relative_to = (
-          os.path.dirname( tern_project ) if is_project else current_dir )
-
-        _logger.info( 'Tern paths are relative to: '
-                      + self._server_paths_relative_to )
-
+    self._do_tern_project_check = False
+    filepath = request_data[ 'filepath' ]
+    if not FindTernProjectFile( filepath ):
+      raise RuntimeError( 'Warning: Unable to detect a .tern-project file '
+                          'in the hierarchy before ' + filepath +
+                          ' and no global .tern-config file was found. '
+                          'This is required for accurate JavaScript '
+                          'completion. Please see the User Guide for '
+                          'details.' )
 
 
   def _GetServerAddress( self ):
@@ -241,7 +222,9 @@ class TernCompleter( Completer ):
 
 
   def OnFileReadyToParse( self, request_data ):
-    self._WarnIfMissingTernProject()
+    self._StartServer( request_data )
+
+    self._WarnIfMissingTernProject( request_data )
 
     # Keep tern server up to date with the file data. We do this by sending an
     # empty request just containing the file data
@@ -257,7 +240,7 @@ class TernCompleter( Completer ):
   def GetSubcommandsMap( self ):
     return {
       'RestartServer':  ( lambda self, request_data, args:
-                                         self._RestartServer() ),
+                                         self._RestartServer( request_data ) ),
       'StopServer':     ( lambda self, request_data, args:
                                          self._StopServer() ),
       'GoToDefinition': ( lambda self, request_data, args:
@@ -281,13 +264,19 @@ class TernCompleter( Completer ):
 
   def DebugInfo( self, request_data ):
     with self._server_state_mutex:
+      extras = [
+        responses.DebugInfoItem( key = 'configuration file',
+                                 value = self._server_project_file )
+      ]
+
       tern_server = responses.DebugInfoServer(
         name = 'Tern',
         handle = self._server_handle,
         executable = PATH_TO_TERN_BINARY,
         address = SERVER_HOST,
         port = self._server_port,
-        logfiles = [ self._server_stdout, self._server_stderr ] )
+        logfiles = [ self._server_stdout, self._server_stderr ],
+        extras = extras )
 
       return responses.BuildDebugInfoResponse( name = 'JavaScript',
                                                servers = [ tern_server ] )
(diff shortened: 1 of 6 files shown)


micbou:
I did some refactoring in the tests by introducing an utility function to start the Tern server. Also, similarly to the recent changes in PR https://github.com/Valloric/ycmd/pull/857, I slightly changed the logic so that we only try to start the Tern server once. PR is ready.

---

Reviewed 6 of 6 files at r1, 4 of 4 files at r2, 1 of 1 files at r3, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/875#-:-KzjVOHK9-OIz6tuOaqz:btltcon)*

bstaletic:
Makes me wonder if we should start all semantic completers on `FileReadyToParse`.
:lgtm: 

---

Reviewed 2 of 6 files at r1, 3 of 4 files at r2, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/875#-:-KzkicwKIi49Km0PObz8:b47si36)*

Valloric:
:lgtm:

Thanks!

---

---

*[ycmd/completers/javascript/tern_completer.py, line 391 at r4](https://reviewable.io:443/reviews/valloric/ycmd/875#-KztKM4qaXarPFBP9fo9:-KztKM4qaXarPFBP9foA:bd23w3l) ([raw file](https://github.com/valloric/ycmd/blob/e587501bcc51c28186cb5285a9017b874a280926/ycmd/completers/javascript/tern_completer.py#L391)):*
> ```Python
>       _logger.info( 'Starting Tern server...' )
> 
>       filepath = request_data[ 'filepath' ]
> ```

This PR adds 10+ lines of code to a function with a TODO of "this function is too long". :(

Doesn't block this PR, but we should consider refactoring this sooner rather than later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/875#-:-KztL4NunhLnflQI_Fkm:b9benhj)*

bstaletic:
> Doesn't block this PR, but we should consider refactoring this sooner rather than later.

So is it time to merge this?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/875#-:-KzxMNWsie9LEP9j6BBi:bxld3i0)*

micbou:
Reviewed 1 of 1 files at r5.

---

*[ycmd/completers/javascript/tern_completer.py, line 391 at r4](https://reviewable.io:443/reviews/valloric/ycmd/875#-KztKM4qaXarPFBP9fo9:-KzzDxErxf3z4vvgYNxP:buddlwo) ([raw file](https://github.com/valloric/ycmd/blob/e587501bcc51c28186cb5285a9017b874a280926/ycmd/completers/javascript/tern_completer.py#L391)):*

This PR adds 10+ lines of code to a function with a TODO of "this function is too long". :(

Doesn't block this PR, but we should consider refactoring this sooner rather than later.
</blockquote></details>

Did some simplifications and moved the new code to a separate function. It's now 54 lines instead of 74.

---

*[ycmd/completers/javascript/tern_completer.py, line 442 at r4](https://reviewable.io:443/reviews/valloric/ycmd/875#-KzzEt6gM8u6epnq4Hkr:-KzzEt6jzPGpjjeCQV-T:b-j6g1z7) ([raw file](https://github.com/valloric/ycmd/blob/e587501bcc51c28186cb5285a9017b874a280926/ycmd/completers/javascript/tern_completer.py#L442)):*
> ```Python
>       except Exception:
>         _logger.exception( 'Unable to start Tern server' )
>         self._CleanUp()
> ```

Removed this `try/except` block as it was unlikely to ever catch something (`CreateLogFile`, `OpenForStdHandle`, and `SafePopen` are never supposed to fail).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/875)*

puremourning:

---

*[ycmd/completers/javascript/tern_completer.py, line 157 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-Kzz_Ke1OdjQX2k8gOAN:-Kzz_Ke2C2PUG8F1ka96:b-6a6fnm) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/completers/javascript/tern_completer.py#L157)):*
> ```Python
>     if not FindTernProjectFile( filepath ):
>       raise RuntimeError( 'Warning: Unable to detect a .tern-project file '
>                           'in the hierarchy before ' + filepath +
> ```

should this be a dir name rather than file name? totally optional. i'm ambivalent.

---

*[ycmd/completers/javascript/tern_completer.py, line 268 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-Kzz_q-9qIbWjxavywkY:-Kzz_q-9qIbWjxavywkZ:bn1el97) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/completers/javascript/tern_completer.py#L268)):*
> ```Python
>     with self._server_state_mutex:
>       extras = [
>         responses.DebugInfoItem( key = 'configuration file',
> ```

should we call it "Tern project file", as that is a little clearer that it is specific to Tern rather than YCM ? I theory at least editors and ycmd itself might have "configuration files". Again, optional.

---

*[ycmd/completers/javascript/tern_completer.py, line 386 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-KzzaJoEvQh4vpZD730d:-KzzaJoEvQh4vpZD730e:b5k095a) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/completers/javascript/tern_completer.py#L386)):*
> ```Python
>     if not self._server_project_file:
>       _logger.warning( 'No .tern-project file detected: %s', filepath )
>       self._server_working_dir = os.path.dirname( filepath )
> ```

I think this might be a slight regression, but I'm not sure.

Previously in this scenario we would use Vim's working dir, which _might_ turn out a better choice than the file's directory.

Example might be:

 -  a user is used to using the javascript completion without a tern project file
 - they typically `cd /some/path` 
 - the structure is like `/some/path/lib_one/*.js`, `/some/path/lib_two/*.js`
 - they use `vim lib_one/file.js`, which contains `var x = require( 'lib_two/another_file.js' )`

I don't know for certain if this would or would not have previously just about worked based on the wd of the term server being `/some/path`. But I'm pretty sure that it wouldn't with the wd of `/some/path/lib_one`. 

I guess the key question for discussion is: if we can't find a tern project file, should we really use the dirname of the file, or revert to the previous behaviour of using the WD of the client. I suspect that we should do the latter, for safety unless there's a good reason not to... or I totally missed the point, which is of course also possible!

---

*[ycmd/completers/javascript/tern_completer.py, line 442 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-KzzgS00LBBkXtbsd0Gi:-KzzgS01FYPlvjT3_n2d:b-83qm92) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/completers/javascript/tern_completer.py#L442)):*
> ```Python
> 
>       if self._ServerIsRunning():
>         _logger.info( 'Tern Server started with pid: %s listening on port %s',
> ```

should these be %d ? just being pedantic, you can ignore.

---

*[ycmd/completers/javascript/tern_completer.py, line 489 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-Kzzgi7-fXeHv8H_FI6A:-Kzzgi7-fXeHv8H_FI6B:b-qmrp1r) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/completers/javascript/tern_completer.py#L489)):*
> ```Python
>     self._server_started = False
>     self._server_working_dir = None
>     self._server_project_file = None
> ```

should we reset `do_tern_project_check` here ? I think it's not a new issue, but after restarting the server in some other directory, we probably want to re-scan for the tern project, and also re-warn for the "new" project.

---

*[ycmd/tests/javascript/event_notification_test.py, line 160 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-KzzhHD-nUV0YlGCYTA_:-KzzhHD-nUV0YlGCYTAa:bxtgcxc) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/tests/javascript/event_notification_test.py#L160)):*
> ```Python
> 
>   # Finally, restart the server in a folder containing a .tern-project file. We
>   # expect no error in that case.
> ```

should we test that if we restart it in some other directory (which doesn't have a .tern-project) that we _do_ get another error (as mentioned, I think we don't, but i'm not 100% sure).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/875)*

micbou:
Reviewed 1 of 1 files at r6.

---

*[ycmd/completers/javascript/tern_completer.py, line 157 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-Kzz_Ke1OdjQX2k8gOAN:-L-27_CK22ENA8oZLFqZ:bvcm8hs) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/completers/javascript/tern_completer.py#L157)):*

should this be a dir name rather than file name? totally optional. i'm ambivalent.
</blockquote></details>

It's simpler to put the filepath here.

---

*[ycmd/completers/javascript/tern_completer.py, line 268 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-Kzz_q-9qIbWjxavywkY:-KzzrUREx2eFcu3zd0BD:bb0ky0l) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/completers/javascript/tern_completer.py#L268)):*

should we call it "Tern project file", as that is a little clearer that it is specific to Tern rather than YCM ? I theory at least editors and ycmd itself might have "configuration files". Again, optional.
</blockquote></details>

No need because this item is added in the `extras` field of the Tern `DebugInfoServer` object. It will appear as `Tern configuration file` in the output of `:YcmDebugInfo`.

---

*[ycmd/completers/javascript/tern_completer.py, line 386 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-KzzaJoEvQh4vpZD730d:-L-1UqqW63LJm23K3v4n:b-55k077) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/completers/javascript/tern_completer.py#L386)):*

I think this might be a slight regression, but I'm not sure.

Previously in this scenario we would use Vim's working dir, which _might_ turn out a better choice than the file's directory.

Example might be:

 -  a user is used to using the javascript completion without a tern project file
 - they typically `cd /some/path` 
 - the structure is like `/some/path/lib_one/*.js`, `/some/path/lib_two/*.js`
 - they use `vim lib_one/file.js`, which contains `var x = require( 'lib_two/another_file.js' )`

I don't know for certain if this would or would not have previously just about worked based on the wd of the term server being `/some/path`. But I'm pretty sure that it wouldn't with the wd of `/some/path/lib_one`. 

I guess the key question for discussion is: if we can't find a tern project file, should we really use the dirname of the file, or revert to the previous behaviour of using the WD of the client. I suspect that we should do the latter, for safety unless there's a good reason not to... or I totally missed the point, which is of course also possible!
</blockquote></details>

I did some tests and Tern doesn't seem to rely at all on the working directory to resolve modules path (I tried with `/tmp` as the working directory and it works) so I don't think it matters. By the way, how's your example supposed to work without a `.tern-project` or `.tern-config` file (the `node` plugin is needed for module resolution)?

---

*[ycmd/completers/javascript/tern_completer.py, line 442 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-KzzgS00LBBkXtbsd0Gi:-L-1UpKKDofG3oHS74FY:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/completers/javascript/tern_completer.py#L442)):*

should these be %d ? just being pedantic, you can ignore.
</blockquote></details>

Done.

---

*[ycmd/completers/javascript/tern_completer.py, line 489 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-Kzzgi7-fXeHv8H_FI6A:-L-1SmYJQdyrmledmzGp:b5nidw9) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/completers/javascript/tern_completer.py#L489)):*

should we reset `do_tern_project_check` here ? I think it's not a new issue, but after restarting the server in some other directory, we probably want to re-scan for the tern project, and also re-warn for the "new" project.
</blockquote></details>

When restarting the server, the `do_tern_project_check` variable is set to `True` so we are going to warn the user again if there is no configuration file. However, this doesn't hurt to reset the variable here.

---

*[ycmd/tests/javascript/event_notification_test.py, line 160 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-KzzhHD-nUV0YlGCYTA_:-L-1SD9d1uj1GBh7ytmx:b7ubfxz) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/tests/javascript/event_notification_test.py#L160)):*
This is tested just above:
> Restart the server and check that it raises it again.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/875)*

puremourning:
:lgtm:

---

Reviewed 2 of 6 files at r1, 3 of 4 files at r2, 1 of 1 files at r5.
Review status: 4 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/completers/javascript/tern_completer.py, line 386 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-KzzaJoEvQh4vpZD730d:-L-3MUcUzQiiISZOsbtv:bneezu1) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/completers/javascript/tern_completer.py#L386)):*

I did some tests and Tern doesn't seem to rely at all on the working directory to resolve modules path (I tried with `/tmp` as the working directory and it works) so I don't think it matters. By the way, how's your example supposed to work without a `.tern-project` or `.tern-config` file (the `node` plugin is needed for module resolution)?
</blockquote></details>

Like i said i wasn't sure if it would work before, or if there are other Tern behaviours that use relative paths in some sense.

I should have said "something like `require( 'lib_two/....' )`" .

Thanks for testing anyway, let's assume this is grand. In any event the solution is to add a `.tern-project`.

---

*[ycmd/completers/javascript/tern_completer.py, line 489 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-Kzzgi7-fXeHv8H_FI6A:-L-3MxUEqCrQXxvlJSB-:b-6mhywn) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/completers/javascript/tern_completer.py#L489)):*

When restarting the server, the `do_tern_project_check` variable is set to `True` so we are going to warn the user again if there is no configuration file. However, this doesn't hurt to reset the variable here.
</blockquote></details>

Great, thanks.

---

*[ycmd/tests/javascript/event_notification_test.py, line 160 at r5](https://reviewable.io:443/reviews/valloric/ycmd/875#-KzzhHD-nUV0YlGCYTA_:-L-3MzycZPkt8IacSVYK:bdwkvgc) ([raw file](https://github.com/valloric/ycmd/blob/7a1e2362b18d506a7167d79b9b3fe6975b985522/ycmd/tests/javascript/event_notification_test.py#L160)):*

This is tested just above:
> Restart the server and check that it raises it again.

</blockquote></details>

Heh ok sorry :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/875#-:-L-3N18sOyrx9b3Q1Cy5:bnfp4nl)*

puremourning:
Reviewed 2 of 2 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/875)*

micbou:
Let's merge.

@zzbot r+

---

Reviewed 2 of 2 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/875#-:-L-4-l_PVqUDemoocZvl:b-ptzv6q)*

PR #876, opened 2017-11-21
[READY] Improve coverage configuration

micbou:
When running the tests locally with coverage enabled, nose will not include the files from subprocesses coverage in its own report and in the HTML report. This is fixed by adding the `--cover-inclusive` option and by specifying the `--ignore-files` option so that nose does not ignore the `__main__.py` file (files starting with a underscore are ignored by default). We also have to exclude the `ycmd/tests/python/testdata` directory because its base name starts with `test` (nose will collect tests in that directory) and it contains Python files.

Another improvement is that we enable subprocesses coverage on AppVeyor.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/876)

### DIFF
--- a/ci/appveyor/appveyor_install.bat
+++ b/ci/appveyor/appveyor_install.bat
@@ -34,6 +34,10 @@ if %errorlevel% neq 0 exit /b %errorlevel%
 pip install codecov
 if %errorlevel% neq 0 exit /b %errorlevel%
 
+:: Enable coverage for Python subprocesses. See:
+:: http://coverage.readthedocs.io/en/latest/subprocess.html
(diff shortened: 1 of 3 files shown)


puremourning:
:lgtm: thanks a lot!

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/876#-:-KzVhg30x4QhT5vftSF7:b-e8wozt)*

bstaletic:
:lgtm:
Thanks.
@zzbot r=puremourning

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/876#-:-KzZtS4VIuPK6-B1fTLP:bya74wh)*

PR #785, opened 2017-06-28
[READY] Let ycmd check rustup configured path as the last means to finding rust src path

bstaletic:
In Valloric/YouCompleteMe#2379 there was talk about enabling ycmd to check rust src path provided by rustup.

This allows users to have rust completion with zero ycmd configuration.

Tests are not included because tests themselves should behave differently based on how rustup was used and I'm not sure how to test that.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/785)

PR #885, opened 2017-12-07
[READY] Omnisharp roslyn

mispencer:
_See https://github.com/Valloric/ycmd/pull/213 for prior pull request._

Allow using omnisharp-roslyn. See https://github.com/Valloric/ycmd/pull/209#issuecomment-136160670.

This pull request does not make roslyn the default (yet). Rather, it must be switched to the roslyn implementation by running the `UseRoslynOmnisharp` completer subcommand. This is per-solution, and pre-runtime.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/885)


mispencer:
Current status:
* Windows is failing the Shutdown_test.FromHandlerWithSubservers_test and I am not sure why
* Travis hasn't been running the Mac tests, so I'm not sure if they are passing
* There are a lot of skipped tests, some which should be fixed
* Fixit in new backend isn't working, and I believe it is implementable
* The libuv check in build.py should be improved
* Dotnet Roslyn version is in the build, but isn't supported in the runtime

bstaletic:
Thanks for once again investing the time into this.

You're probably aware that we don't like the idea of supporting multiple engines for a single language.
So what is stopping us from ditching the legacy engine?
Sorry if this has been asked in the previous attempt.

---

Reviewed 10 of 15 files at r1, 2 of 2 files at r2.
Review status: 12 of 17 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[build.py, line 499 at r3](https://reviewable.io:443/reviews/valloric/ycmd/885#-L25OaPZLmpPfKEVS6ze:-L25OaP_A8p1tpkBBRNk:bsk41j6) ([raw file](https://github.com/valloric/ycmd/blob/70012ae1e83afe8dfe5a39e0fa411d386ec653f2/build.py#L499)):*
> ```Python
>           '/v', 'version' ] )
>       dotnet_46_pattern = re.compile( 'version\s*REG_SZ\s*4.[67].\d*' )
>       
> ```

Trailing spaces.

---

*[ycmd/tests/cs/debug_info_test.py, line 59 at r3](https://reviewable.io:443/reviews/valloric/ycmd/885#-L25POtb_WGY34VcZS9V:-L25POtb_WGY34VcZS9W:b-yfn59l) ([raw file](https://github.com/valloric/ycmd/blob/70012ae1e83afe8dfe5a39e0fa411d386ec653f2/ycmd/tests/cs/debug_info_test.py#L59)):*
> ```Python
>         'address': instance_of( str ),
>         'port': instance_of( int ),
>         'logfiles': empty(),
> ```

Does roslyn actually have no logs?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/885#-:-L2F0pyKb7xTuI6PjrhA:benh7o5)*

mispencer:
Roslyn Omnisharp lacks features which were supported in the legacy Omnisharp. In my personal usage, I never use the Roslyn version unless the project I am working on is too new for the legacy Omnisharp to work.

I have resolved merge conflicts, and have hopefully fixed the Windows issue with the last pushed commit (waiting on AppVeyor to confirm this). I have already confirm that Linux and Mac builds were passing.

Issues that remain:
- There are a lot of skipped tests, some which should be fixed
- Fixit in new backend isn't working, and I believe it is implementable
- The libuv check in build.py should be improved
- Dotnet Roslyn version is in the build, but isn't supported in the runtime

---

Review status: 10 of 16 files reviewed at latest revision, 2 unresolved discussions.

---

*[build.py, line 499 at r3](https://reviewable.io:443/reviews/valloric/ycmd/885#-L25OaPZLmpPfKEVS6ze:-L2JFoDEbRBTbIA9OGmI:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/70012ae1e83afe8dfe5a39e0fa411d386ec653f2/build.py#L499)):*

Trailing spaces.
</blockquote></details>

Done.

---

*[ycmd/tests/cs/debug_info_test.py, line 59 at r3](https://reviewable.io:443/reviews/valloric/ycmd/885#-L25POtb_WGY34VcZS9V:-L2JFs3LjhmtUQ5ITGI9:b7i3bqh) ([raw file](https://github.com/valloric/ycmd/blob/70012ae1e83afe8dfe5a39e0fa411d386ec653f2/ycmd/tests/cs/debug_info_test.py#L59)):*

Does roslyn actually have no logs?
</blockquote></details>

As part of this change, I had piped the log output into ycmd logs. However, I believe this is the cause of the Window's failures, so I have just reverted this back to logging to separate log files.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/885#-:-L2JGIi7qU3UFnaN-W_S:b-ew17x0)*

mispencer:
The window issues should be resolved. Roslyn only works in Windows Python 2 if you execute it though the shell. I don't know why and this is not preferred, but I don't know any other way to resolve, so that's what I am doing.

Roslyn Omnisharp doesn't support Fixit yet, so that will remain skipped until this is implemented in Roslyn.

I don't think force completion or imported type will ever be supported in Roslyn, so these tests are now only ran for legacy Omnisharp. When legacy Omnisharp is removed, this logic can be removed in entirely.

I have added support for dotnet flavor of roslyn, but I don't have a machine to test this on, so maybe it works? I don't expect this to be a common usage case, so this is probably fine.

Since Roslyn Omnisharp does not support the complete-as-you-type completion mode, I have adjusted the triggers of the C# completer to trigger in more limited cases. The implemented behavior is, additionally to normal trigger characters, it will trigger on completion of identifiers. The effect of this is that it will always preempt the identifier completer. I have implemented it this way because Omnisharp is fast enough that performance is not a concern, and the language completer provides more relevant identifier completions than the identifier completer, so we should use it wherever performance permits.

I have found certain projects which Omnisharp Roslyn does not function, so I don't think we are ready to remove the legacy Omnisharp yet.

**The CI failures are flakes and I have completed all todo items, so this is ready for review.**

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/885#-:-L2QXXjSSAA50C0oYAeS:bml1iwu)*

mispencer:
Merge conflicts have been resolved.

bstaletic:
First, thank you for your patience and your effort.

Here's what we all have been thinking about this.

Supporting two C# servers would be a burden. Especially considering none of the maintainers is a C# developer.
I vaguely remember some troubles with Omnisharp just because none of us regularly uses C#.

Maintaining two servers shall be hard. I'm also not sure what I think about the approach of trying and seeing which server works - using `try`/`except` everywhere just because of two backends.

On the other hand Roslyn is supposed to have a LSP server capabilities and that should be fairly easy to use now that #857 has been merged. @micbou @puremourning and I have all been able to hack a random LSP server in a short amount of time.

@puremourning @mispencer @valloric @vheon 
Thoughts?

---

Reviewed 1 of 15 files at r1, 2 of 3 files at r5, 1 of 2 files at r6, 3 of 4 files at r7, 8 of 8 files at r9.
Review status: 16 of 17 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cs/cs_completer.py, line 36 at r9](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3Xucn4bGIgamLT4YmX:-L3Xucn4bGIgamLT4YmY:b-kj558f) ([raw file](https://github.com/valloric/ycmd/blob/1ec8c7baa224e7383d9d4a43563e674c43eb2ce3/ycmd/completers/cs/cs_completer.py#L36)):*
> ```Python
> import requests
> import threading
> from subprocess import ( PIPE )
> ```

Parentheses aren't needed here.

---

*[ycmd/completers/cs/cs_completer.py, line 118 at r9](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3XvpCwbgD_rg65MJm2:-L3XvpCxcC8LVPTV-BZZ:bktz5vk) ([raw file](https://github.com/valloric/ycmd/blob/1ec8c7baa224e7383d9d4a43563e674c43eb2ce3/ycmd/completers/cs/cs_completer.py#L118)):*
> ```Python
> 
>   def ShouldUseNowInner( self, request_data ):
>     """ Preempt the identity completer always, since the C# completer is fast
> ```

Are we sure this is a good idea?
How fast is omnisharp for really big projects?

---

*[ycmd/tests/cs/__init__.py, line 65 at r9](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3XtVgLME_Bff03Tgln:-L3XtVgLME_Bff03Tglo:b-rch4ob) ([raw file](https://github.com/valloric/ycmd/blob/1ec8c7baa224e7383d9d4a43563e674c43eb2ce3/ycmd/tests/cs/__init__.py#L65)):*
> ```Python
> 
> def WaitUntilOmniSharpServerReady( app, filepath ):
>   retries = 2000
> ```

Why does this need so many tries?

---

*[ycmd/tests/cs/debug_info_test.py, line 59 at r3](https://reviewable.io:443/reviews/valloric/ycmd/885#-L25POtb_WGY34VcZS9V:-L2KI4IhR5CU1WFuw6JX:b2uazx1) ([raw file](https://github.com/valloric/ycmd/blob/70012ae1e83afe8dfe5a39e0fa411d386ec653f2/ycmd/tests/cs/debug_info_test.py#L59)):*
<details><summary><i>Previously, mispencer wrote…</i></summary><blockquote>

As part of this change, I had piped the log output into ycmd logs. However, I believe this is the cause of the Window's failures, so I have just reverted this back to logging to separate log files.
</blockquote></details>

I think it's preferable to have separate files for separate logs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/885#-:-L3XwffDUJ76BCyUzVWt:b-vbk070)*

mispencer:
The double backend is needed for me because there were some legacy solutions, which I am actively working on, that I could not get Roslyn Omnisharp to work properly with, and there were some other newer solutions, which I am also actively working on, that the legacy Omnisharp could not open. Thus, to support all my use cases, YCMD needed to support both the legacy and Roslyn Omnisharp. Furthermore, there are still some features from the legacy Omnisharp which I would miss that are not supported in Roslyn Omnisharp.

However, having said all that, I have noticed [a issue](https://github.com/OmniSharp/omnisharp-roslyn/issues/1077) which could have caused my issues with legacy solutions and it is supposedly fixed in the latest version (1.28). I will have to re-test. If bumping the Roslyn Omnisharp version resolves my issue, I would (reluctantly) consent to a single backend with Roslyn Omnisharp.

LSP is still [an open issue](https://github.com/OmniSharp/omnisharp-roslyn/issues/968) in Roslyn Omnisharp. I know the initial implementation was included in the latest beta of Roslyn Omnisharp, but I wasn't under the impression that it was stable yet.

Though LSP may be a simple implementation, it would still a decent implementation effort, and I'm not sure what we would gain. The existing Omnisharp interface is not deprecated, and I don't believe it will be any time soon.

---

Review status: 16 of 17 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/cs/cs_completer.py, line 36 at r9](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3Xucn4bGIgamLT4YmX:-L3ZiSRzD1DYDrOTesz6:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/1ec8c7baa224e7383d9d4a43563e674c43eb2ce3/ycmd/completers/cs/cs_completer.py#L36)):*

Parentheses aren't needed here.
</blockquote></details>

Done.

---

*[ycmd/completers/cs/cs_completer.py, line 118 at r9](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3XvpCwbgD_rg65MJm2:-L3ZiTbwR2NxD6EyMoY-:bfp8b5d) ([raw file](https://github.com/valloric/ycmd/blob/1ec8c7baa224e7383d9d4a43563e674c43eb2ce3/ycmd/completers/cs/cs_completer.py#L118)):*

Are we sure this is a good idea?
How fast is omnisharp for really big projects?
</blockquote></details>

Fast enough. It normally shows up by the time I type 3-4 characters, which is about the minimum needed to be typed to select the correct result anyway.

Note, please, that this is the way that the C# completer *currently works*, so this would be expected behavior by anyone using the C# completion. See https://github.com/Valloric/ycmd/pull/116 for justification why it currently functions this way.

---

*[ycmd/tests/cs/__init__.py, line 65 at r9](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3XtVgLME_Bff03Tgln:-L3ZjeNOy2fHcFF00Y-7:b4i94l) ([raw file](https://github.com/valloric/ycmd/blob/1ec8c7baa224e7383d9d4a43563e674c43eb2ce3/ycmd/tests/cs/__init__.py#L65)):*

Why does this need so many tries?
</blockquote></details>

It probably doesn't need to so high. I'm pretty sure I just got frustrated by CI failures from timeout and set it to a excessively high number.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/885#-:-L3ZbtR6kzqlpayojJSt:b-jfa9v7)*

bstaletic:
> Roslyn Omnisharp doesn't support Fixit yet, so that will remain skipped until this is implemented in Roslyn.

Apart from us not wanting to support two backends, this is the only concern left. I am not sure if we want to take features away from the users. Is there an effort to make FixIts work with Roslyn?

> LSP is still an open issue in Roslyn Omnisharp. I know the initial implementation was included in the latest beta of Roslyn Omnisharp, but I wasn't under the impression that it was stable yet.

> Though LSP may be a simple implementation, it would still a decent implementation effort, and I'm not sure what we would gain. The existing Omnisharp interface is not deprecated, and I don't believe it will be any time soon.

Alright, sounds good to me. I'm fine with sticking to this API.

---

Review status: 16 of 17 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cs/cs_completer.py, line 118 at r9](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3XvpCwbgD_rg65MJm2:-L3c8TmDlf5q24IEsPMK:b-6e9i1v) ([raw file](https://github.com/valloric/ycmd/blob/1ec8c7baa224e7383d9d4a43563e674c43eb2ce3/ycmd/completers/cs/cs_completer.py#L118)):*
<details><summary><i>Previously, mispencer wrote…</i></summary><blockquote>

Fast enough. It normally shows up by the time I type 3-4 characters, which is about the minimum needed to be typed to select the correct result anyway.

Note, please, that this is the way that the C# completer *currently works*, so this would be expected behavior by anyone using the C# completion. See https://github.com/Valloric/ycmd/pull/116 for justification why it currently functions this way.
</blockquote></details>

Alright, then I guess this is fine.

---

*[ycmd/tests/cs/__init__.py, line 65 at r9](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3XtVgLME_Bff03Tgln:-L3c8_jBKNR-XbYZ8XIZ:b2xl1yi) ([raw file](https://github.com/valloric/ycmd/blob/1ec8c7baa224e7383d9d4a43563e674c43eb2ce3/ycmd/tests/cs/__init__.py#L65)):*
<details><summary><i>Previously, mispencer wrote…</i></summary><blockquote>

It probably doesn't need to so high. I'm pretty sure I just got frustrated by CI failures from timeout and set it to a excessively high number.
</blockquote></details>

Okay. Can you add a comment stating this is because of the CI timing out?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/885#-:-L3c95W1cbKRPGHVPsi9:bds7cyx)*

mispencer:
I have opened a issue in Omnisharp-Roslyn (https://github.com/OmniSharp/omnisharp-roslyn/issues/1098) about the fixits (code issues in Omnisharp jargon). We'll see what they say.

Should I be concerned about the coverage failures? 

---

Review status: 15 of 17 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[ycmd/tests/cs/__init__.py, line 65 at r9](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3XtVgLME_Bff03Tgln:-L3k_2mH1h65OY_Mf821:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/1ec8c7baa224e7383d9d4a43563e674c43eb2ce3/ycmd/tests/cs/__init__.py#L65)):*

Okay. Can you add a comment stating this is because of the CI timing out?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/885#-:-L3kZo-B9vnB2IXL22Kc:bhnl57g)*

bstaletic:
Thanks for opening the issue.

> Cygwin

@puremourning @micbou @valloric What are your opinions on this?
I personally feel that if cygwin needs this much "hand holding", it should be tested. And as far as Iknow appveyor can run cygwin builds too.

On the other hand, testing on Cygwin means we actually support it, which hasn't been the case before.

> Coverage

Codecov has been reporting nonsense lately and we should probably start investigating.
But codecov still reports sensible "patch". So getting that part of codecov checks to pass would be great, especially for a code this complex. 
Generally, @valloric likes to see coverage over 95% and @micbou is usually happy with 90% coverage.

---

Reviewed 1 of 2 files at r10.
Review status: 16 of 17 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cs/cs_completer.py, line 691 at r10](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3leiDdChroxtZwxPQJ:-L3leiDdChroxtZwxPQK:b-59ofpk) ([raw file](https://github.com/valloric/ycmd/blob/4c06e0fbb95f291414e698e915c7965b69f55bf6/ycmd/completers/cs/cs_completer.py#L691)):*
> ```Python
>     response = requests.post( target,
>                                    json = parameters,
>                                    timeout = timeout )
> ```

Last two parameters are misaligned.

---

*[ycmd/completers/cs/cs_completer.py, line 805 at r10](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3lfH9qzSt00fgJZ0qI:-L3lfH9qzSt00fgJZ0qJ:bglm6ug) ([raw file](https://github.com/valloric/ycmd/blob/4c06e0fbb95f291414e698e915c7965b69f55bf6/ycmd/completers/cs/cs_completer.py#L805)):*
> ```Python
> 
> 
> def _ConvertFilenameForCygwin( filename, direction ):
> ```

Why is the second parameter called direction?

---

*[ycmd/completers/cs/cs_completer.py, line 819 at r10](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3lfW-A7gmSML6jg7fB:-L3lfW-A7gmSML6jg7fC:b-u100di) ([raw file](https://github.com/valloric/ycmd/blob/4c06e0fbb95f291414e698e915c7965b69f55bf6/ycmd/completers/cs/cs_completer.py#L819)):*
> ```Python
>         dir_arg = '-w' if direction else '-u'
>         command = [ 'cygpath', '-f', '-', dir_arg ]
>         cygpath = utils.SafePopen( command , stdout = PIPE, stdin = PIPE )
> ```

@micbou Were there any caveats/gotchas in using `SafePopen` on Windows?
I remember there was something whe I made the rustup PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/885#-:-L3lae-NBDHwamGtq9v0:b-2fsutd)*

mispencer:
> I personally feel that if cygwin needs this much "hand holding", it should be tested. And as far as Iknow appveyor can run cygwin builds too.
> On the other hand, testing on Cygwin means we actually support it, which hasn't been the case before.

If there is any interest in this, I can look into it.

>Codecov has been reporting nonsense lately and we should probably start investigating.
>But codecov still reports sensible "patch". So getting that part of codecov checks to pass would be great, especially for a code this complex.
>Generally, @Valloric likes to see coverage over 95% and @micbou is usually happy with 90% coverage.

I have made some changes to increase the coverage number, though I kinda feel I'm maximizing a measurement, rather than code quality. 

---

Review status: 14 of 17 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/cs/cs_completer.py, line 691 at r10](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3leiDdChroxtZwxPQJ:-L3nA7O615YJTviGMr89:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/4c06e0fbb95f291414e698e915c7965b69f55bf6/ycmd/completers/cs/cs_completer.py#L691)):*

Last two parameters are misaligned.
</blockquote></details>

Done.

---

*[ycmd/completers/cs/cs_completer.py, line 805 at r10](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3lfH9qzSt00fgJZ0qI:-L3nA7netk3XMFZqSq0w:b-owwb0r) ([raw file](https://github.com/valloric/ycmd/blob/4c06e0fbb95f291414e698e915c7965b69f55bf6/ycmd/completers/cs/cs_completer.py#L805)):*

Why is the second parameter called direction?
</blockquote></details>

It's the direction of conversion (window-to-cygwin or cygwin-to-window).

---

*[ycmd/completers/cs/cs_completer.py, line 819 at r10](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3lfW-A7gmSML6jg7fB:-L3nAG2NE2-0fZU2Udcf:b-rtvr3x) ([raw file](https://github.com/valloric/ycmd/blob/4c06e0fbb95f291414e698e915c7965b69f55bf6/ycmd/completers/cs/cs_completer.py#L819)):*

@micbou Were there any caveats/gotchas in using `SafePopen` on Windows?
I remember there was something whe I made the rustup PR.
</blockquote></details>

I don't think any of those apply in Cygwin, which is the only OS this runs in.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/885#-:-L3nAN3vEePTOu0GPNyK:b-smhocs)*

bstaletic:
>  If there is any interest in this, I can look into it.

Great, but let's not rush. I'd like to see what others have to say about it.

> I have made some changes to increase the coverage number, though I kinda feel I'm maximizing a measurement, rather than code quality.

We shouldn't be trying to hide unlikely and uncovered branches with `# pragma: no cover`. Ideally, code should be covered, but in case that is hard or even impossible for some reason, we should still be aware that some branch isn't covered.

If you need something to return a specific value or throw a specific exception, python provides `@patch` decorator for testing that.

Here's an example of overridden `return` value: https://github.com/Valloric/ycmd/blob/master/ycmd/tests/rust/get_completions_test.py#L58

---

Reviewed 3 of 3 files at r11.

---

*[ycmd/completers/cs/cs_completer.py, line 805 at r10](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3lfH9qzSt00fgJZ0qI:-L3nGMV9pWf970YJg5pa:b9bvssu) ([raw file](https://github.com/valloric/ycmd/blob/4c06e0fbb95f291414e698e915c7965b69f55bf6/ycmd/completers/cs/cs_completer.py#L805)):*
<details><summary><i>Previously, mispencer wrote…</i></summary><blockquote>

It's the direction of conversion (window-to-cygwin or cygwin-to-window).
</blockquote></details>

It's still not clear which direction is which. So this would benefit a lot from a comment explaining what `direction = True`, as opposed to `direction = False` means.

---

*[ycmd/completers/cs/cs_completer.py, line 819 at r10](https://reviewable.io:443/reviews/valloric/ycmd/885#-L3lfW-A7gmSML6jg7fB:-L3nGhWoXpK2nN5YtITW:bclzxbb) ([raw file](https://github.com/valloric/ycmd/blob/4c06e0fbb95f291414e698e915c7965b69f55bf6/ycmd/completers/cs/cs_completer.py#L819)):*
<details><summary><i>Previously, mispencer wrote…</i></summary><blockquote>

I don't think any of those apply in Cygwin, which is the only OS this runs in.
</blockquote></details>

That might as well be true, I know next to nothing about Cygwin.

For reference: https://github.com/Valloric/ycmd/pull/785#issuecomment-312070555
Specifically, the part where @micbou mentions `CREATE_NO_WINDOWS` flag.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/885#-:-L3nG85gSAEIMFlxKEog:bqwhb5f)*

puremourning:
Re: Cygwin

You won't be surprised to hear that I have no particular interest in supporting Cygwin, and personally see it as more things I have to worry about for future changes. But that's a super selfish viewpoint and should not be taken seriously :)

What I would say IMO is that we should only support Cygwin (properly) if the maintenance burden is low, and that we have good evidence that we have strong demand for it, and a _ahem_ willing volunteer to own the integration and maintenance. We currently support it on a best-efforts basis, but if that's not enough then it is worth a discussion.

---

Reviewed 2 of 15 files at r1.
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[appveyor.yml, line 8 at r5](https://reviewable.io:443/reviews/valloric/ycmd/885#-L2HFPcwPm0NvwItNw2M:-L2HFPcxk43-ETi5fY0b:bdbxmlw) ([raw file](https://github.com/valloric/ycmd/blob/df1eb8ea00f705ca44ab3af1f14b09721733c64d/appveyor.yml#L8)):*
> ```YAML
>   # We dropped MSVC 11 support because it lacks features from C++11 and C99.
>   # We dropped MSVC 13 support because it lacks .NET 4.6 on that image
>   #- APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2013
> ```

we can probably remove these lines. i have no issue with dropping older msvc as it is trivial to acquire new versions and you can aiui have multiple versions installed.

---

*[build.py, line 85 at r5](https://reviewable.io:443/reviews/valloric/ycmd/885#-L2HFeKjhYfWGMaTnVP6:-L2HFeKjhYfWGMaTnVP7:b-p8xdiq) ([raw file](https://github.com/valloric/ycmd/blob/df1eb8ea00f705ca44ab3af1f14b09721733c64d/build.py#L85)):*
> ```Python
> 
> def OnCygwin():
>   return platform.system().startswith("CYGWIN")
> ```

is this adding support for cygwin ? My recollection is that we don't (officially) support it. if so is there value in doing that in a different PR ?

---

*[build.py, line 164 at r5](https://reviewable.io:443/reviews/valloric/ycmd/885#-L2HIh3TSJaiPoRNlpfI:-L2HIh3TSJaiPoRNlpfJ:b-5npnz3) ([raw file](https://github.com/valloric/ycmd/blob/df1eb8ea00f705ca44ab3af1f14b09721733c64d/build.py#L164)):*
> ```Python
> 
> 
> def CheckOutput( args, **kwargs ):
> ```

i'm not clear why this is necessary.

---

*[build.py, line 283 at r5](https://reviewable.io:443/reviews/valloric/ycmd/885#-L2HFr6h6w85AlnQLFOq:-L2HFr6h6w85AlnQLFOr:b-t6h77l) ([raw file](https://github.com/valloric/ycmd/blob/df1eb8ea00f705ca44ab3af1f14b09721733c64d/build.py#L283)):*
> ```Python
>   parser.add_argument( '--js-completer', action = 'store_true',
>                        help = 'Enable JavaScript semantic completion engine.' ),
>   parser.add_argument( '--new-cs-completer', action = 'store_true',
> ```

if we are really going to support two completers in parallel i wonder if it would be better to use `--cs-completer=roslyn` or something? New is only new as long as the old one exists. Usually best to avoid distinctions of "new" and "old" because new becomes old very quickly

---

*[build.py, line 551 at r5](https://reviewable.io:443/reviews/valloric/ycmd/885#-L2HJO7CaP9z6v4p_f3r:-L2HJO7CaP9z6v4p_f3s:b-bo8s82) ([raw file](https://github.com/valloric/ycmd/blob/df1eb8ea00f705ca44ab3af1f14b09721733c64d/build.py#L551)):*
> ```Python
>             seven_zip_path = _winreg.QueryValueEx( SZ, 'Path' )[ 0 ]
> 
>       extract_command = [ p.join( seven_zip_path, '7z.exe' ), 'x', file_path ]
> ```

don't we already use 7zip in the make? I seem to recall that we don't have to look in the registry for this?

---

*[build.py, line 555 at r5](https://reviewable.io:443/reviews/valloric/ycmd/885#-L2HJXOF5SpSyVmIIBjU:-L2HJXOF5SpSyVmIIBjV:bgiczc9) ([raw file](https://github.com/valloric/ycmd/blob/df1eb8ea00f705ca44ab3af1f14b09721733c64d/build.py#L555)):*
> ```Python
>       extract_command = [ 'tar', 'xfv', file_path ]
> 
>     subprocess.check_call( extract_command )
> ```

what's extract command on linux/macOS?

---

*[ycmd/completers/cs/cs_completer.py, line 59 at r5](https://reviewable.io:443/reviews/valloric/ycmd/885#-L2HKp5B8E-UnbBSMExX:-L2HKp5B8E-UnbBSMExY:b-j4616i) ([raw file](https://github.com/valloric/ycmd/blob/df1eb8ea00f705ca44ab3af1f14b09721733c64d/ycmd/completers/cs/cs_completer.py#L59)):*
> ```Python
> ROSLYN_OMNISHARP_BINARY = 'OmniSharp'
> MONO_MODE = 1 # TODO
> if utils.OnWindows() or utils.OnCygwin() or MONO_MODE:
> ```

Any reason not to use `PathToFirstExistingExecutable` which sort out this .exe stuff for us?

---

*[ycmd/completers/cs/cs_completer.py, line 207 at r5](https://reviewable.io:443/reviews/valloric/ycmd/885#-L2HLAoCAY6fJ1kRZsSh:-L2HLAoCAY6fJ1kRZsSi:b-f42l5u) ([raw file](https://github.com/valloric/ycmd/blob/df1eb8ea00f705ca44ab3af1f14b09721733c64d/ycmd/completers/cs/cs_completer.py#L207)):*
> ```Python
>       'SetOmnisharpPath'                 : ( lambda self, request_data, args:
>          self._SetOmnisharpPath( request_data, args[ 0 ] ) ),
>       'UseLegacyOmnisharp'                 : ( lambda self, request_data, args:
> ```

is there really a need to support both engines and toggle between them? this sounds painfully coplicated and worse, how many users are really going to understand this and be able to actually do it? 

---

*[ci/travis/travis_install.linux.sh, line 29 at r5](https://reviewable.io:443/reviews/valloric/ycmd/885#-L2HJpiAgR4jvsoOLapq:-L2HJpiAgR4jvsoOLapr:b8s9e3x) ([raw file](https://github.com/valloric/ycmd/blob/df1eb8ea00f705ca44ab3af1f14b09721733c64d/ci/travis/travis_install.linux.sh#L29)):*
> ```Shell
> 
> # Libuv is required for Omnisharp-Roslyn and isn't in accessible repos
> curl -sSL https://github.com/libuv/libuv/archive/v1.18.0.tar.gz | tar zxfv - -C /tmp && cd /tmp/libuv-1.18.0/
> ```

80 characters

---

*[ci/travis/travis_install.linux.sh, line 36 at r5](https://reviewable.io:443/reviews/valloric/ycmd/885#-L2HJtSrRibZIOy_-DEN:-L2HJtSrRibZIOy_-DEO:b-mc9hm4) ([raw file](https://github.com/valloric/ycmd/blob/df1eb8ea00f705ca44ab3af1f14b09721733c64d/ci/travis/travis_install.linux.sh#L36)):*
> ```Shell
> export LD_LIBRARY_PATH="$HOME/libuvinstall/lib"
> export LIBRARY_PATH="$HOME/libuvinstall/lib"
> cd $OLDPWD
> ```

what sets `OLDPWD` ? is that a shell builtin?

i suspect we really want `pushd` and `popd` here, as this seems breakable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/885#-:-L4I8cA7r9IFJid85mQ6:b-lzbics)*

bstaletic:
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[ci/travis/travis_install.linux.sh, line 36 at r5](https://reviewable.io:443/reviews/valloric/ycmd/885#-L2HJtSrRibZIOy_-DEN:-L4J63o30sGTPRAvzlYs:b-bn5k9m) ([raw file](https://github.com/valloric/ycmd/blob/df1eb8ea00f705ca44ab3af1f14b09721733c64d/ci/travis/travis_install.linux.sh#L36)):*

what sets `OLDPWD` ? is that a shell builtin?

i suspect we really want `pushd` and `popd` here, as this seems breakable.
</blockquote></details>

For the record `OLDPWD` is a shell builtin and `cd $OLDPWD` is equivalent to `cd -`, or rather `cd -` is equivalend to `cd $OLDPWD`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/885)*

Valloric:
Closing PR as it's not moving forward; can reopen in the future if desired.

ohjames:
> I am not sure if we want to take features away from the users. Is there an effort to make FixIts work with Roslyn?

I doubt there are very many C# developers who would be content to use a version of C# that is several years old just to be able to use fixits.

PR #891, opened 2017-12-22
Fix --system-libclang on Gentoo amd64, if llvm is compiled with ABI_X86="32 64"

DarthGandalf:
$ ls -ld /usr/lib64/llvm/4/l*
drwxr-xr-x 3 root root  20K Dec 16 02:37 lib32
drwxr-xr-x 3 root root  20K Dec 16 02:37 lib64
drwxr-xr-x 2 root root 4.0K Dec 16 02:37 libexec

This results in GLOB finding the wrong library.

$ ./install.py --clang-completer --system-libclang
...
Using external libclang: /usr/lib64/llvm/4/lib32/libclang.so.4.0
...
/usr/lib64/llvm/4/lib32/libclang.so.4.0: error adding symbols: File in wrong format

FTR, I had to use --system-libclang because of libtinfo.so.5 issue (https://github.com/Valloric/YouCompleteMe/issues/778)

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/891)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -283,12 +283,16 @@ if ( EXTERNAL_LIBCLANG_PATH OR USE_SYSTEM_LIBCLANG )
     else ()
       set( ENV_LIB_PATHS "" )
     endif()
+    find_program( LLVM_CONFIG_EXECUTABLE NAMES llvm-config )
+    if ( LLVM_CONFIG_EXECUTABLE )
+      execute_process( COMMAND ${LLVM_CONFIG_EXECUTABLE} --libdir
+                       OUTPUT_VARIABLE LLVM_CONFIG_PATH
+                       OUTPUT_STRIP_TRAILING_WHITESPACE )
+    endif()
     # On Debian-based systems, llvm installs into /usr/lib/llvm-x.y.
     file( GLOB SYS_LLVM_PATHS "/usr/lib/llvm*/lib" )
     # On FreeBSD , llvm install into /usr/local/llvm-xy
     file ( GLOB FREEBSD_LLVM_PATHS "/usr/local/llvm*/lib")
-    # On Gentoo, llvm installs into /usr/lib/llvm/x/
-    file ( GLOB GENTOO_LLVM_PATHS "/usr/lib*/llvm/*/lib*" )
     # Need TEMP because find_library does not work with an option variable
     # On Debian-based systems only a symlink to libclang.so.1 is created
     find_library( TEMP
@@ -297,11 +301,11 @@ if ( EXTERNAL_LIBCLANG_PATH OR USE_SYSTEM_LIBCLANG )
                   libclang.so.1
                   PATHS
                   ${ENV_LIB_PATHS}
+                  ${LLVM_CONFIG_PATH}
                   /usr/lib
                   /usr/lib/llvm
                   ${SYS_LLVM_PATHS}
                   ${FREEBSD_LLVM_PATHS}
-                  ${GENTOO_LLVM_PATHS}
                   /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib
                   /Library/Developer/CommandLineTools/usr/lib )
     set( EXTERNAL_LIBCLANG_PATH ${TEMP} )


bstaletic:
Of course I forgot about multilib llvm when I posted the first Gentoo specific oull request.

I've confirmed that this works on a non-multilb llvm, so only a few questions.

- Is`llvm-config` a Gentoo specific tool?
  - If `llvm-config` exists on other distros, can others confirm this doesn't break anything else?
- Seems to me like `llvm-config` makes the hand-written glob redundant. Am I correct?

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891#-:-L10DTsaqGDICUJoIfmP:bend0p6)*

DarthGandalf:
`llvm-config` is not specific to Gentoo. But e.g. on Ubuntu it's named `llvm-config-3.8`. I'm not fan of hardcoding all possible versions here, as some people [do](https://github.com/Sarcasm/irony-mode/blob/f6c8810f6b78e2b3d66fb4ffb7513f4f8ff053e3/server/cmake/modules/FindLibClang.cmake).
And I don't see it on macos, though I didn't install clang from homebrew there, maybe that one has it. (I didn't try to install YCM on macos yet at all)

So... some of the other paths are probably not necessary anymore, but I don't know which.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891#-:-L11Z_tAWUQjEACqmIVj:b-wspbgn)*

bstaletic:
I believe debian based distributions have something like `update_alternatives` which would make a symlink from `llvm-config` to `llvm-config-<version>`.

I see this change as one that could clean up out CMakeLists.txt a bit, so it's a welcome change. I just want to make sure we get it right, so I want to hear from other maintainers too.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891#-:-L11e6E8pNAFUtRrrcc7:bcos5o9)*

puremourning:
Thanks for sending a PR!

Re: macOS:

AFAIK`llvm-config` is not shipped with Apple clang, which is the clang that should be used with `--sytem-libclang` on macOS, as this is the only way for modules to work (which is required for iOS).

I wonder if this change will make it inadvertently start using a homebred-installed version of libclang when it didn't before? TBH I'm not sure, and testing it out would be a serious pain.

IN principle, I think using llvm-config is a good-thing(TM), though we have to be a little careful not to break people who previously didn't realise they were relying on us _not_ doing that (however nefariously).

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891#-:-L12O12Yrhaofc2-8w8W:bndpfhf)*

bstaletic:
I think we should test this carefull on all the platforms we have access to.

I can test Mint, Windows and maybe RHEL.
Which leave *BSD land.

---

---

*[cpp/ycm/CMakeLists.txt, line 311 at r1](https://reviewable.io:443/reviews/valloric/ycmd/891#-L12a2rsOufMHah7Aoip:-L12a2rsOufMHah7Aoiq:b-691040) ([raw file](https://github.com/valloric/ycmd/blob/a4e8b0e6d84d72bf9cce8698c043894a8efc9916/cpp/ycm/CMakeLists.txt#L311)):*
> ```CMake
>                   ${SYS_LLVM_PATHS}
>                   ${FREEBSD_LLVM_PATHS}
>                   ${GENTOO_LLVM_PATHS}
> ```

I believe `${GENTOO_LLVM_PATHS}` can be removed if we use `llvm-config`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891#-:-L12aHQUZlQyFhu6zlzD:bjrlm4n)*

bstaletic:
I've just tried llvm-config on Mint. It successfully found the default libclang.so.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891#-:-L1By6B9nesbEZ4mTWfR:b-jfa03z)*

DarthGandalf:

---

*[cpp/ycm/CMakeLists.txt, line 311 at r1](https://reviewable.io:443/reviews/valloric/ycmd/891#-L12a2rsOufMHah7Aoip:-L1D9eDq0cE3ycBDtjN5:bss4a2b) ([raw file](https://github.com/valloric/ycmd/blob/a4e8b0e6d84d72bf9cce8698c043894a8efc9916/cpp/ycm/CMakeLists.txt#L311)):*

I believe `${GENTOO_LLVM_PATHS}` can be removed if we use `llvm-config`.
</blockquote></details>

Removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891)*

Valloric:

---

*[cpp/ycm/CMakeLists.txt, line 295 at r2](https://reviewable.io:443/reviews/valloric/ycmd/891#-L1JkiJLI4Z_WZ8k5jV1:-L1JkiJLI4Z_WZ8k5jV2:b9rvlqh) ([raw file](https://github.com/valloric/ycmd/blob/f912ca740b1e83c33450a72a1eae440721192902/cpp/ycm/CMakeLists.txt#L295)):*
> ```CMake
>     file( GLOB SYS_LLVM_PATHS "/usr/lib/llvm*/lib" )
>     # On FreeBSD , llvm install into /usr/local/llvm-xy
>     file ( GLOB FREEBSD_LLVM_PATHS "/usr/local/llvm*/lib")
> ```

Do we still need the FreeBSD paths if we're now using llvm-config? Does anyone have access to a FreeBSD machine to try it out?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891)*

puremourning:

---

*[cpp/ycm/CMakeLists.txt, line 295 at r2](https://reviewable.io:443/reviews/valloric/ycmd/891#-L1JkiJLI4Z_WZ8k5jV1:-L1KreIhRVJw01mdh7TQ:b-m9qr9x) ([raw file](https://github.com/valloric/ycmd/blob/f912ca740b1e83c33450a72a1eae440721192902/cpp/ycm/CMakeLists.txt#L295)):*

Do we still need the FreeBSD paths if we're now using llvm-config? Does anyone have access to a FreeBSD machine to try it out?
</blockquote></details>

I installed a FreeBSD 11.1 VM. There is no llvm-config by default. The default compiler is clang 4.0.0 which is way too old for YCM now anyway. `--system-libclang` isn't strictly needed on FreeBSD as the precompiled binaries work fine.

I installed llvm 5 `pkg install llvm50` and this put a `llvm-config50` in the path, which this doesn't appear to search for.

 However, the _existing_ path (`/usr/local/llvm*/lib`) does find the llvm50:

```
Using external libclang: /usr/local/llvm50/lib/libclang.so.5.0
```

So, in conclusion: Yes, the BSD paths are still needed, because the required llvm-config on FreeBSD is `llvm-config50`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891)*

bstaletic:
Reviewed 1 of 1 files at r2.

---

*[cpp/ycm/CMakeLists.txt, line 295 at r2](https://reviewable.io:443/reviews/valloric/ycmd/891#-L1JkiJLI4Z_WZ8k5jV1:-L1MEiM5BzzKzKHQl1m7:bc17xf0) ([raw file](https://github.com/valloric/ycmd/blob/f912ca740b1e83c33450a72a1eae440721192902/cpp/ycm/CMakeLists.txt#L295)):*

I installed a FreeBSD 11.1 VM. There is no llvm-config by default. The default compiler is clang 4.0.0 which is way too old for YCM now anyway. `--system-libclang` isn't strictly needed on FreeBSD as the precompiled binaries work fine.

I installed llvm 5 `pkg install llvm50` and this put a `llvm-config50` in the path, which this doesn't appear to search for.

 However, the _existing_ path (`/usr/local/llvm*/lib`) does find the llvm50:

```
Using external libclang: /usr/local/llvm50/lib/libclang.so.5.0
```

So, in conclusion: Yes, the BSD paths are still needed, because the required llvm-config on FreeBSD is `llvm-config50`.

</blockquote></details>

Well, that changes things a bit.

The `llvm-config` on Mint 18 is 3.8, so it found `libclang.so.3.8`. There is `llvm-config-4.0`, but no `llvm-config-5.0`. Mint 18 is based on Ubuntu 16.04.0.
Maybe we could try different `llvm-config<whatever>` instead of globs that match required paths. Not sure how well that would work.

Something like:

```
find_program( LLVM_CONFIG_EXECUTABLE_BSD llvm-config50 )
find_program( LLVM_CONFIG_EXECUTABLE_DEBIAN_BASED llvm-config-5.0 )
...
find_program( LLVM_CONFIG_EXECUTABLE_FALLBACK llvm-config )
```

And if we end up using the last one we print a warning that we have no idea what version we might find and what dragons be there.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891)*

DarthGandalf:
I've got another idea: use `find_package ( Clang )` to make it use `ClangConfig.cmake` which comes with clang itself, instead of calling `llvm-config`. It might be better in some cases, but I'll try and see what happens.

---

---

*[cpp/ycm/CMakeLists.txt, line 295 at r2](https://reviewable.io:443/reviews/valloric/ycmd/891#-L1JkiJLI4Z_WZ8k5jV1:-L1MMHg2e79wisdyFkmE:bx0hzx9) ([raw file](https://github.com/valloric/ycmd/blob/f912ca740b1e83c33450a72a1eae440721192902/cpp/ycm/CMakeLists.txt#L295)):*

Well, that changes things a bit.

The `llvm-config` on Mint 18 is 3.8, so it found `libclang.so.3.8`. There is `llvm-config-4.0`, but no `llvm-config-5.0`. Mint 18 is based on Ubuntu 16.04.0.
Maybe we could try different `llvm-config<whatever>` instead of globs that match required paths. Not sure how well that would work.

Something like:

```
find_program( LLVM_CONFIG_EXECUTABLE_BSD llvm-config50 )
find_program( LLVM_CONFIG_EXECUTABLE_DEBIAN_BASED llvm-config-5.0 )
...
find_program( LLVM_CONFIG_EXECUTABLE_FALLBACK llvm-config )
```

And if we end up using the last one we print a warning that we have no idea what version we might find and what dragons be there.
</blockquote></details>

I wanted to avoid hardcoding versions... but if that's unavoidable, `find_program ( LLVM_CONFIG_EXECUTABLE NAMES llvm-config-5.0 llvm-config50 )` is cleaner than several different `find_program`s. Also someone needs to update the list of versions whenever a new version is released.
I thought `--system-libclang` is where dragons live anyway; is using some particular version found on system more supported than using other versions found on system? Also, why is 4.0 "way too old"? Seems to work fine for me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891#-:-L1MXLzBdqw02xdnm_i3:b-2ct90n)*

bstaletic:

---

*[cpp/ycm/CMakeLists.txt, line 295 at r2](https://reviewable.io:443/reviews/valloric/ycmd/891#-L1JkiJLI4Z_WZ8k5jV1:-L1MY14nyZBSOXJ6ws3p:b-t6h5sz) ([raw file](https://github.com/valloric/ycmd/blob/f912ca740b1e83c33450a72a1eae440721192902/cpp/ycm/CMakeLists.txt#L295)):*
<details><summary><i>Previously, DarthGandalf (Alexey Sokolov) wrote…</i></summary><blockquote>

I wanted to avoid hardcoding versions... but if that's unavoidable, `find_program ( LLVM_CONFIG_EXECUTABLE NAMES llvm-config-5.0 llvm-config50 )` is cleaner than several different `find_program`s. Also someone needs to update the list of versions whenever a new version is released.
I thought `--system-libclang` is where dragons live anyway; is using some particular version found on system more supported than using other versions found on system? Also, why is 4.0 "way too old"? Seems to work fine for me.
</blockquote></details>

Hard coding versions was just for the sake of simplicity, nothing more than that. It's definitely not a good idea.

4.0 is old, because the version ycmd downloads is 5.0.1, so who knows when something might break because of different versions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891)*

puremourning:

---

*[cpp/ycm/CMakeLists.txt, line 295 at r2](https://reviewable.io:443/reviews/valloric/ycmd/891#-L1JkiJLI4Z_WZ8k5jV1:-L1OJcveEE196dQvrFmr:b1r18ir) ([raw file](https://github.com/valloric/ycmd/blob/f912ca740b1e83c33450a72a1eae440721192902/cpp/ycm/CMakeLists.txt#L295)):*

Hard coding versions was just for the sake of simplicity, nothing more than that. It's definitely not a good idea.

4.0 is old, because the version ycmd downloads is 5.0.1, so who knows when something might break because of different versions.
</blockquote></details>

I'm not sure how much effort we should go to in order to support `--system-libclang` which is strongly discouraged, and barely supported. Ultimately users can use the full installation guide and `-DEXTERNAL_LIBCLANG_PATH` and point it wherever they like in those instances.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891)*

bstaletic:

---

*[cpp/ycm/CMakeLists.txt, line 295 at r2](https://reviewable.io:443/reviews/valloric/ycmd/891#-L1JkiJLI4Z_WZ8k5jV1:-L1OKHy1jJ7ZON3irkq_:b-ub3mrs) ([raw file](https://github.com/valloric/ycmd/blob/f912ca740b1e83c33450a72a1eae440721192902/cpp/ycm/CMakeLists.txt#L295)):*

I'm not sure how much effort we should go to in order to support `--system-libclang` which is strongly discouraged, and barely supported. Ultimately users can use the full installation guide and `-DEXTERNAL_LIBCLANG_PATH` and point it wherever they like in those instances.
</blockquote></details>

There's always an option of merging this PR as is and only benefit Gentoo.
I just thought it was a nice oportunitty to tidy up the CMakeLists.txt.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891)*

DarthGandalf:
`ClangConfig.cmake` still requires path to that file, which is different per distro, so doesn't work very well.

---

---

*[cpp/ycm/CMakeLists.txt, line 295 at r2](https://reviewable.io:443/reviews/valloric/ycmd/891#-L1JkiJLI4Z_WZ8k5jV1:-L2GKOLvEqOoeccrGZzj:b3sh7mt) ([raw file](https://github.com/valloric/ycmd/blob/f912ca740b1e83c33450a72a1eae440721192902/cpp/ycm/CMakeLists.txt#L295)):*

There's always an option of merging this PR as is and only benefit Gentoo.
I just thought it was a nice oportunitty to tidy up the CMakeLists.txt.
</blockquote></details>

Let's merge it as is...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891#-:-L2GKZCfXtws8KZ5G9Nl:b-uvixux)*

bstaletic:
Alright, seems like utilising `llvm-config` for every available platform is going to be hard.

This is :lgtm: if you ask me.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891#-:-L2GXPCV0kXaleoKdqIN:bjmo4py)*

puremourning:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891#-:-L2GqKfh5k1a-RumzA5h:bnfp4nl)*

micbou:
@zzbot r+

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891#-:-L2Gr2HK1bOiLQ10O9B-:bi59kwd)*

bstaletic:
@zzbot retry

bstaletic:
Third time's the charm?
@zzbot retry

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/891#-:-L2K6e5-20fCa9lIsrpS:b-qmb0lj)*

PR #897, opened 2018-01-13
[READY] Use our own packages of libclang to prevent CI flakes

puremourning:
releases.llvm.org is under serious load and is causing CI to flake.

The LLVM people say that CI systems shouldn't continually download clang
from release.llvm.org. We normally wouldn't but the caches in our CI are
full and we seem to fail to cache the LLVM full download.

Instead, we just use a package (created by us) containing only libclang,
and download that from bintray (which is free for OSS usage). There is
some limit to the number of downloads etc., so we only enable this in CI, even if it is cached.

In theory, we could also suggest this to users who are struggling with the download, such as from behind certain national firewalls.

I selected bintray as the hosting as it is free for a limited amount of traffic for open source projects. The packages contain the libraries and the LLVM license.

I have prepared and uploaded packages for the CI platforms that we use, and can see in the CI that they are being used and more importantly, the AppVeyor cache is not full anymore, meaning that the cached versions are being used on repeated runs.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/897)

### DIFF
--- a/build.py
+++ b/build.py
@@ -336,6 +336,9 @@ def GetCmakeArgs( parsed_args ):
   use_python2 = 'ON' if PY_MAJOR == 2 else 'OFF'
   cmake_args.append( '-DUSE_PYTHON2=' + use_python2 )
 
+  if OnTravisOrAppVeyor():
+    cmake_args.append( '-DUSE_LIBCLANG_PACKAGE=ON' )
+
   extra_cmake_args = os.environ.get( 'EXTRA_CMAKE_ARGS', '' )
   # We use shlex split to properly parse quoted CMake arguments.
   cmake_args.extend( shlex.split( extra_cmake_args ) )
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -22,6 +22,7 @@ project( ycm_core )
 option( USE_DEV_FLAGS "Use compilation flags meant for YCM developers" OFF )
 option( USE_CLANG_COMPLETER "Use Clang semantic completer for C/C++/ObjC" OFF )
 option( USE_SYSTEM_LIBCLANG "Set to ON to use the system libclang library" OFF )
+option( USE_LIBCLANG_PACKAGE "For ycmd CI only, use packaged libclang, rather than the official LLVM releases" OFF )
 set( PATH_TO_LLVM_ROOT "" CACHE PATH "Path to the root of a LLVM+Clang binary distribution" )
 set( EXTERNAL_LIBCLANG_PATH "" CACHE PATH "Path to the libclang library to use" )
 
@@ -37,17 +38,31 @@ if ( USE_CLANG_COMPLETER AND
     set( CLANG_SHA256
          "c5b105c4960619feb32641ef051fa39ecb913cc0feb6bacebdfa71f8d3cae277" )
     set( CLANG_FILENAME "${CLANG_DIRNAME}.tar.xz" )
+
+    set( LIBCLANG_DIRNAME "libclang-${CLANG_VERSION}-x86_64-apple-darwin" )
+    set( LIBCLANG_FILENAME "${LIBCLANG_DIRNAME}.tar.bz2" )
+    set( LIBCLANG_SHA256
+         "b47661118a4a55e3cfd1a71990ede9e6817d942cd8fabe7185852dfdaf7f01e8" )
   elseif ( WIN32 )
     if( 64_BIT_PLATFORM )
       set( CLANG_DIRNAME "LLVM-${CLANG_VERSION}-win64" )
       set( CLANG_SHA256
            "981543611d719624acb29a2cffd6a479cff36e8ab5ee8a57d8eca4f9c4c6956f" )
+
+      set( LIBCLANG_DIRNAME "libclang-${CLANG_VERSION}-win64" )
+      set( LIBCLANG_SHA256
+           "01f64da1edfa0859f5c63561ba93c0511915f69b9778687dda9ae4b70f68c565" )
     else()
       set( CLANG_DIRNAME "LLVM-${CLANG_VERSION}-win32" )
       set( CLANG_SHA256
            "5de70ab482edb2da7ac20126dc58e23a691498aa644ca23a7b10c32c9ee62157" )
+
+      set( LIBCLANG_DIRNAME "libclang-${CLANG_VERSION}-win32" )
+      set( LIBCLANG_SHA256
+           "b5c7d1c6009599613dd99ec7ebc720bb4be928890e7d46614c8e2adb17c047c7" )
     endif()
     set( CLANG_FILENAME "${CLANG_DIRNAME}.exe" )
+    set( LIBCLANG_FILENAME "${LIBCLANG_DIRNAME}.tar.bz2" )
   elseif ( SYSTEM_IS_FREEBSD )
     if ( 64_BIT_PLATFORM )
       set( CLANG_DIRNAME "clang+llvm-${CLANG_VERSION}-amd64-unknown-freebsd10" )
@@ -75,6 +90,11 @@ if ( USE_CLANG_COMPLETER AND
            "clang+llvm-${CLANG_VERSION}-x86_64-linux-gnu-ubuntu-14.04" )
       set( CLANG_SHA256
            "9e61c6669991e2f0d065348c95917b2c6b697d75098b60ec1c2e9f17093ce012" )
+
+      set( LIBCLANG_DIRNAME
+           "libclang-${CLANG_VERSION}-x86_64-linux-gnu-ubuntu-14.04" )
+      set( LIBCLANG_SHA256
+           "25eeb9183c15ec49eda103ee980569588e8b9c1c690024925a24dedcb9ac94a4" )
     else()
       message( FATAL_ERROR
         "No prebuilt Clang ${CLANG_VERSION} binaries for 32-bit Linux. "
@@ -82,11 +102,26 @@ if ( USE_CLANG_COMPLETER AND
         "See the YCM docs for details on how to use a user-compiled libclang." )
     endif()
     set( CLANG_FILENAME "${CLANG_DIRNAME}.tar.xz" )
+    set( LIBCLANG_FILENAME "${LIBCLANG_DIRNAME}.tar.bz2" )
   endif()
 
-  # Check if the Clang archive is already downloaded and its checksum is correct.
-  # If this is not the case, remove it if needed and download it.
   set( CLANG_DOWNLOAD ON )
+  if ( LIBCLANG_FILENAME AND USE_LIBCLANG_PACKAGE )
+    set( CLANG_FILENAME ${LIBCLANG_FILENAME} )
+    set( CLANG_DIRNAME ${LIBCLANG_DIRNAME} )
+    set( CLANG_SHA256 ${LIBCLANG_SHA256} )
+    set( USING_LIBCLANG_DOWNLOAD ON )
+    set( CLANG_URL
+         "https://bintray.com/puremourning/libclang/download_file?file_path=${CLANG_FILENAME}" )
+    set( CLANG_PACKAGE "libclang" )
+  else()
+    set( CLANG_URL
+         "http://releases.llvm.org/${CLANG_VERSION}/${CLANG_FILENAME}" )
+    set( CLANG_PACKAGE "Clang" )
+  endif()
+
+  # Check if the Clang archive is already downloaded and its checksum is
+  # correct.  If this is not the case, remove it if needed and download it.
   set( CLANG_LOCAL_FILE
        "${CMAKE_SOURCE_DIR}/../clang_archives/${CLANG_FILENAME}" )
 
@@ -101,21 +136,20 @@ if ( USE_CLANG_COMPLETER AND
   endif()
 
   if( CLANG_DOWNLOAD )
-    message( "Downloading Clang ${CLANG_VERSION}" )
+    message( "Downloading ${CLANG_PACKAGE} ${CLANG_VERSION} from ${CLANG_URL}" )
 
-    set( CLANG_URL "http://releases.llvm.org/${CLANG_VERSION}" )
     file(
-      DOWNLOAD "${CLANG_URL}/${CLANG_FILENAME}" "${CLANG_LOCAL_FILE}"
+      DOWNLOAD "${CLANG_URL}" "${CLANG_LOCAL_FILE}"
       SHOW_PROGRESS EXPECTED_HASH SHA256=${CLANG_SHA256}
     )
   else()
-    message( "Using Clang archive: ${CLANG_LOCAL_FILE}" )
+    message( "Using ${CLANG_PACKAGE} archive: ${CLANG_LOCAL_FILE}" )
   endif()
 
   # Copy and extract the Clang archive in the building directory.
   file( COPY "${CLANG_LOCAL_FILE}" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/../" )
   if ( CLANG_FILENAME MATCHES ".+bz2" )
-    execute_process( COMMAND tar -xjf ${CLANG_FILENAME} )
+    execute_process( COMMAND ${CMAKE_COMMAND} -E tar -xjf ${CLANG_FILENAME} )
   elseif( CLANG_FILENAME MATCHES ".+xz" )
     execute_process( COMMAND tar -xJf ${CLANG_FILENAME} )
   elseif( CLANG_FILENAME MATCHES ".+exe" )
@@ -177,7 +211,7 @@ else()
            "available" )
 endif()
 
-if ( PATH_TO_LLVM_ROOT )
+if ( NOT USING_LIBCLANG_DOWNLOAD AND PATH_TO_LLVM_ROOT )
   set( CLANG_INCLUDES_DIR "${PATH_TO_LLVM_ROOT}/include" )
 else()
   set( CLANG_INCLUDES_DIR "${CMAKE_SOURCE_DIR}/llvm/include" )
--- a/make_llvm_package.py
+++ b/make_llvm_package.py
@@ -1,0 +1,330 @@
+#!/usr/bin/env python
+
+# Passing an environment variable containing unicode literals to a subprocess
+# on Windows and Python2 raises a TypeError. Since there is no unicode
+# string in this script, we don't import unicode_literals to avoid the issue.
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+
+import subprocess
+import contextlib
+import os
+import os.path as p
+import platform
+import shutil
+import sys
+import tempfile
+import tarfile
+import hashlib
+
+try:
+  import lzma
+except:
+  from backports import lzma
+
+DIR_OF_THIS_SCRIPT = p.dirname( p.abspath( __file__ ) )
+sys.path.insert( 0, os.path.join( DIR_OF_THIS_SCRIPT, 'ycmd' ) )
+from ycmd import server_utils
+server_utils.SetUpPythonPath()
+
+# Not installing aliases from python-future; it's unreliable and slow.
+from builtins import *  # noqa
+from future.utils import iteritems
+import argparse
+import requests
+from io import BytesIO
+
+
+def OnWindows():
+  return platform.system() == 'Windows'
+
+
+def OnMac():
+  return platform.system() == 'Darwin'
+
+
+LLVM_DOWNLOAD_DATA = {
+  'win32': {
+    'format': 'nsis',
+    'llvm_package': 'LLVM-{llvm_version}-{os_name}.exe',
+    'ycmd_package': 'libclang-{llvm_version}-{os_name}.tar.bz2',
+    'files_to_copy': [
+      os.path.join( 'bin', 'libclang.dll' ),
+      os.path.join( 'lib', 'libclang.lib' ),
+    ]
+  },
+  'win64': {
+    'format': 'nsis',
+    'llvm_package': 'LLVM-{llvm_version}-{os_name}.exe',
+    'ycmd_package': 'libclang-{llvm_version}-{os_name}.tar.bz2',
+    'files_to_copy': [
+      os.path.join( 'bin', 'libclang.dll' ),
+      os.path.join( 'lib', 'libclang.lib' ),
+    ]
+  },
+  'x86_64-apple-darwin': {
+    'format': 'lzma',
+    'llvm_package': 'clang+llvm-{llvm_version}-{os_name}.tar.xz',
+    'ycmd_package': 'libclang-{llvm_version}-{os_name}.tar.bz2',
+    'files_to_copy': [
+      os.path.join( 'lib', 'libclang.dylib' ),
+    ]
+  },
+  'x86_64-linux-gnu-ubuntu-14.04': {
+    'format': 'lzma',
+    'llvm_package': 'clang+llvm-{llvm_version}-{os_name}.tar.xz',
+    'ycmd_package': 'libclang-{llvm_version}-{os_name}.tar.bz2',
+    'files_to_copy': [
+      os.path.join( 'lib', 'libclang.so' ),
+      os.path.join( 'lib', 'libclang.so.5' ),
+      os.path.join( 'lib', 'libclang.so.5.0' ),
+    ]
+  },
+}
+
+
+@contextlib.contextmanager
+def TemporaryDirectory():
+  temp_dir = tempfile.mkdtemp()
+  try:
+    yield temp_dir
+  finally:
+    shutil.rmtree( temp_dir )
+
+
+def DownloadClangLicense( version, destination ):
+  print( 'Downloading license...' )
+  request = requests.get(
+    'https://releases.llvm.org/{version}/LICENSE.TXT'.format( version=version ),
+    stream = True )
+  request.raise_for_status()
+
+  file_name = os.path.join( destination, 'LICENSE.TXT' )
+  with open( file_name, 'wb' ) as f:
+    f.write( request.content )
+
+  request.close()
+
+  return file_name
+
+
+def DownloadClang( url ):
+  print( 'Downloading {0}'.format( url ) )
+
+  request = requests.get( url, stream=True )
+  request.raise_for_status()
+  content = request.content
+  request.close()
+  return content
+
+
+def ExtractClangLZMA( compressed_data, destination ):
+  uncompressed_data = BytesIO( lzma.decompress( compressed_data ) )
+
+  print( 'Extracting...' )
+  with tarfile.TarFile( fileobj=uncompressed_data, mode='r' ) as tar_file:
+    a_member = tar_file.getmembers()[ 0 ]
+    tar_file.extractall( destination )
+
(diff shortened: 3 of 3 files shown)


micbou:
> In theory, we could also suggest this to users who are struggling with the download, such as from behind certain national firewalls.

I really think we should and not only for the reason you gave but also because if downloads from LLVM are failing on CI services, they are going to fail for users too and we'll get reports for that. Other benefits are:
 - faster download (archives are smaller); 
 - less disk usage (archives are stored in `clang_archives`);
 - no 7-Zip requirement on Windows (since we don't have to extract the files from a NSIS installer);
 - less code to maintain in `CMakeLists.txt`;
 - 4 less checksums to update in `CMakeLists.txt` when a new version of Clang is released.

The only issue is the bandwidth limit on Bintray which is 1 TB/month. Since the archives size is ~18 MB, this means a limit of ~55000 downloads/month or ~1800 downloads/day. This should be more than enough; especially since we are caching the downloads.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897#-:-L2kycobh4cwj2NVgZke:b-ioly0n)*

bstaletic:
My concern was also that we might hit our 1TB/month limit, but I doubt we will come close to 1800 downloads per day.

So maybe we really should completely switch to using our hosting instead of llvm.org.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897#-:-L2l4B4cdNqAVaOVjD6Y:b-c62jhu)*

puremourning:
> faster download (archives are smaller);
> less disk usage (archives are stored in clang_archives);
> no 7-Zip requirement on Windows (since we don't have to extract the files from a NSIS installer);
> less code to maintain in CMakeLists.txt;
> 4 less checksums to update in CMakeLists.txt when a new version of Clang is released.

These are all good arguments, but i think it would be wise to start off with just using in CI and monitor how stable that is for a short while. If we're happy with everything we can decide to either roll it out as the default (or only) option, or publish the flag further.

Some arguments _against_ that I thought of:

- users in corporate environments may have specific exceptions for releases.llvm.org in corporate firewalls. changing the download address to "unsanctioned" bintray account could be problematic (i know it would be at my own work)
- changing it would actually net increase the size of the clang_archives folder for current users (it would not attempt to use the currently cached archive, instead downloading a new one), and there is no mechanism to purge the cache at present.
- in the past @valloric has been clear that he doesn't want to be in the business of maintaining binary packages for any OS/vendor combo under the Sun. Making this only work in our CI avoids any question of why we aren't doing it for, say FreeBSD.

> My concern was also that we might hit our 1TB/month limit, but I doubt we will come close to 1800 downloads per day.

I expect that we will be fine on this usage threshold too.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897#-:-L2pUMj-vpBQ__omN4px:blgrgee)*

micbou:
Okay, let's try until the next version of Clang is released. The release is planned for February 21. If this is working well then we'll use our own packages by default.

> users in corporate environments may have specific exceptions for releases.llvm.org in corporate firewalls. changing the download address to "unsanctioned" bintray account could be problematic (i know it would be at my own work)

Even if the Bintray account belongs to one of the maintainers?

> changing it would actually net increase the size of the clang_archives folder for current users (it would not attempt to use the currently cached archive, instead downloading a new one), and there is no mechanism to purge the cache at present.

This only holds until the next release of Clang.

---

Reviewed 3 of 3 files at r1.

---

*[make_llvm_package.py, line 155 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pftI3v0nVd9mKBTCH:-L2pftI3v0nVd9mKBTCI:b-hhr4fa) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L155)):*
> ```Python
>                 source_dir,
>                 bundle_file_name ):
>   print( "Extracting files from {0}".format( source_dir ) )
> ```

We don't seem to extract files here but rather we create an archive containing them.

---

*[make_llvm_package.py, line 189 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pjYY5HJEbYI_Od7_l:-L2pjYY6Z3QMRelzD4Ek:b-a2ep6f) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L189)):*
> ```Python
> def Main():
>   parser = argparse.ArgumentParser()
>   parser.add_argument( '--from-cache', action='store', required=False,
> ```

Flags are optional by default so `required=False` is not needed.

---

*[make_llvm_package.py, line 192 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pjn231kf3cOQDjqfp:-L2pjn231kf3cOQDjqfq:bmcbm7i) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L192)):*
> ```Python
>                        help = 'Use the clang bins from this dir' )
>   parser.add_argument( '--version', action='store', required=True,
>                        help = 'The LLVM Version' )
> ```

Making arguments starting with `--` required is going to be cumbersome. One need to type at least the command
```sh
./make_llvm_package.py --version <version> --output-dir <output_dir> --bt-user <bt_user> --bt-token <bt_token>
```
to run the script. With positional arguments, it would be
```sh
./make_llvm_package.py <version> <output_dir> <bt_user> <bt_token>
```
which is much more convenient.

---

*[make_llvm_package.py, line 194 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2phvOjPf6kFs8N8jy1:-L2phvOkDwkz1ItFivwt:b-3zmjvn) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L194)):*
> ```Python
>                        help = 'The LLVM Version' )
>   parser.add_argument( '--output-dir', action='store', required=True,
>                        help = 'Where to put them' )
> ```

Can't we create the ycmd package in a temporary folder by default and make that argument optional or even remove it?

---

*[make_llvm_package.py, line 199 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pge9SmhsrnYqQL1Ey:-L2pge9SmhsrnYqQL1Ez:bkyh4m5) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L199)):*
> ```Python
>   parser.add_argument( '--bt-token', action='store', required=True,
>                        help = 'Bintray api token' )
>   args = parser.parse_args()
> ```

All that code should be moved to a separate function `ParseArguments`.

---

*[cpp/ycm/CMakeLists.txt, line 25 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pg5-I-C3ifY5RXTiF:-L2pg5-JlU04LMVsykIG:bjs1xjt) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/cpp/ycm/CMakeLists.txt#L25)):*
> ```CMake
> option( USE_CLANG_COMPLETER "Use Clang semantic completer for C/C++/ObjC" OFF )
> option( USE_SYSTEM_LIBCLANG "Set to ON to use the system libclang library" OFF )
> option( USE_LIBCLANG_PACKAGE "For ycmd CI only, use packaged libclang, rather than the official LLVM releases" OFF )
> ```

Line is over 80 characters.

---

*[cpp/ycm/CMakeLists.txt, line 54 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pg6wkph1daHOoW7UX:-L2pg6wkph1daHOoW7UY:b-6xgwe7) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/cpp/ycm/CMakeLists.txt#L54)):*
> ```CMake
>        set( LIBCLANG_DIRNAME "libclang-${CLANG_VERSION}-win64" )
>        set( LIBCLANG_SHA256
>             "2b588f56c157f85179c13a5740a72c70c35e6dcc69d725e2a77f0477a0bae7bf" )
> ```

These two lines and the ones below are not properly aligned.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897#-:-L2plneHvbVlADQrxQAv:bnanvgh)*

Valloric:
Let's try it first in CI and go from there. I'm not too concerned about Bintray being blocked in corp environments since the jcenter maven mirror is hosted there and it's extremely popular (I think Android Studio uses it by default).

:lgtm: with minor comments

---

---

*[make_llvm_package.py, line 192 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pjn231kf3cOQDjqfp:-L2qPC-riFd_XvlkVMpf:b-8je66w) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L192)):*

Making arguments starting with `--` required is going to be cumbersome. One need to type at least the command
```sh
./make_llvm_package.py --version <version> --output-dir <output_dir> --bt-user <bt_user> --bt-token <bt_token>
```
to run the script. With positional arguments, it would be
```sh
./make_llvm_package.py <version> <output_dir> <bt_user> <bt_token>
```
which is much more convenient.
</blockquote></details>

I'd prefer always using named params (`--name`) instead of positional arguments. Positional arguments are far less readable; you have to know which param is for what ahead of time.

---

*[make_llvm_package.py, line 202 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2qPUKdDcU7lxJSm04r:-L2qPUKdDcU7lxJSm04s:b-cid0md) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L202)):*
> ```Python
> 
>   os.makedirs( args.output_dir )
>   for os_name, download_data in iteritems( LLVM_DOWNLOAD_DATA ):
> ```

Oh boy that's a lot of code. Could some of this be split up into helper functions?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897#-:-L2qOFTsb1SKyjvsJFrM:bujlvih)*

micbou:

---

*[make_llvm_package.py, line 192 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pjn231kf3cOQDjqfp:-L2qSJx02DIzz3rtBaHj:b-35voy6) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L192)):*

I'd prefer always using named params (`--name`) instead of positional arguments. Positional arguments are far less readable; you have to know which param is for what ahead of time.
</blockquote></details>

Yes, you need to know the order of the parameters but for named parameters, you need to know their names. In both cases, you are going to run `./make_llvm_package.py --help` to find out. Then you'll have to type much more characters for named parameters.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897)*

Valloric:

---

*[make_llvm_package.py, line 192 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pjn231kf3cOQDjqfp:-L2q_NWzKiqyqYCYW7Gm:b-4ujukg) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L192)):*

Yes, you need to know the order of the parameters but for named parameters, you need to know their names. In both cases, you are going to run `./make_llvm_package.py --help` to find out. Then you'll have to type much more characters for named parameters.
</blockquote></details>

The difference is that when you're looking at a full command invocation in a script, you can tell what's going on if the command uses named params. With positional args, it's far more confusing.

I don't really have a strong preference though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897)*

bstaletic:

---

*[make_llvm_package.py, line 192 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pjn231kf3cOQDjqfp:-L2qTcORoBXNp2ebm6_0:b-tu8jj0) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L192)):*

The difference is that when you're looking at a full command invocation in a script, you can tell what's going on if the command uses named params. With positional args, it's far more confusing.

I don't really have a strong preference though.
</blockquote></details>

I actually like named parameters more as well.
Another positive side of them is that it is easier to remember a few names for the next time than it is to remember positional arguments.

But just as @valloric I don't have a strong preference either.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897)*

Valloric:

---

*[make_llvm_package.py, line 192 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pjn231kf3cOQDjqfp:-L2qaD1l96g0jOXBJZz3:bj5532g) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L192)):*

The difference is that when you're looking at a full command invocation in a script, you can tell what's going on if the command uses named params. With positional args, it's far more confusing.

I don't really have a strong preference though.
</blockquote></details>

As an example, compare:

`./launch-rocket foo bar 15 42`

and 

`./launch-rocket --username foo --password bar --latitude 15 --longitude 42`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897)*

micbou:

---

*[make_llvm_package.py, line 192 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pjn231kf3cOQDjqfp:-L2qcIni4HYdHjI6dygE:b-pt915n) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L192)):*

As an example, compare:

`./launch-rocket foo bar 15 42`

and 

`./launch-rocket --username foo --password bar --latitude 15 --longitude 42`
</blockquote></details>

I can see your point. Ideally, we should reduce the number of required parameters to a minimum. For instance, if `--bt-user` and `--bt-token` are not given, Bintray's user and token could be read from the environment variables `YCM_BINTRAY_USER` and `YCM_BINTRAY_TOKEN`. For the `--version` parameter, its default value could be set to the latest version of Clang like it's done for the version of Boost in the `update_boost.py` script. The goal is that updating the package should be as simple as running the command
```
./make_llvm_package.py
```
without any parameters. Speaking of which, shouldn't we simply name the script `package_llvm.py`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897)*

puremourning:
Review status: 1 of 3 files reviewed at latest revision, 8 unresolved discussions.

---

*[make_llvm_package.py, line 155 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pftI3v0nVd9mKBTCH:-L2qcqVeMQZP1Z6i1fEG:b-e9sxif) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L155)):*

We don't seem to extract files here but rather we create an archive containing them.
</blockquote></details>

This code "extracts" the 'libclang' from the passed in directory and creates an archive from them. But fine, made up another word.

---

*[make_llvm_package.py, line 189 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pjYY5HJEbYI_Od7_l:-L2qd-xjhLw2HuA9rBvN:biwksun) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L189)):*

Flags are optional by default so `required=False` is not needed.
</blockquote></details>

True, but i prefer to be explicit.

---

*[make_llvm_package.py, line 192 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pjn231kf3cOQDjqfp:-L2qd7Lmw9bdQby31LDm:bvvrf06) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L192)):*

I can see your point. Ideally, we should reduce the number of required parameters to a minimum. For instance, if `--bt-user` and `--bt-token` are not given, Bintray's user and token could be read from the environment variables `YCM_BINTRAY_USER` and `YCM_BINTRAY_TOKEN`. For the `--version` parameter, its default value could be set to the latest version of Clang like it's done for the version of Boost in the `update_boost.py` script. The goal is that updating the package should be as simple as running the command
```
./make_llvm_package.py
```
without any parameters. Speaking of which, shouldn't we simply name the script `package_llvm.py`?
</blockquote></details>

My thoughts exactly; unless it is implicitly obvious in some way (like `cat file`), i prefer to always use named arguments, as otherwise you have to somehow remember the order. There's no obvious order in this case, nor is it obvious what the arguments would be. I do have a strong preference for this :)

Whatever, it's only going to be run a few times a year. Let's make a decision, because the cost of bikeshedding is mainly retesting everything, so I would like to do it all in one go.

---

*[make_llvm_package.py, line 194 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2phvOjPf6kFs8N8jy1:-L2qdTyovCc7bzW-uBbW:b7vm0jh) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L194)):*

Can't we create the ycmd package in a temporary folder by default and make that argument optional or even remove it?
</blockquote></details>

We could. The reason this exists is:

- testing that the generated output is correct
- generating the checksum for the cmake script
- paranoia

---

*[make_llvm_package.py, line 199 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pge9SmhsrnYqQL1Ey:-L2qdfxa7U7C8xrWYOqN:b32wtd) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L199)):*

All that code should be moved to a separate function `ParseArguments`.
</blockquote></details>

ok

---

*[make_llvm_package.py, line 202 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2qPUKdDcU7lxJSm04r:-L2qi0rrMocm49bFIyz3:b-7rl0jd) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L202)):*

Oh boy that's a lot of code. Could some of this be split up into helper functions?
</blockquote></details>

done.

---

*[cpp/ycm/CMakeLists.txt, line 25 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pg5-I-C3ifY5RXTiF:-L2qiBWI0e-MARGRK7Kj:bx3fi2v) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/cpp/ycm/CMakeLists.txt#L25)):*

Line is over 80 characters.
</blockquote></details>

So is the line below it. Given that i was unable to find [a non-horrible line continuation](https://stackoverflow.com/questions/7637539/how-to-split-strings-across-multiple-lines-in-cmake) in cmake, i decided to keep this. Let me know if there is a better way to do this, other than something like CONCAT .... ... ?

---

*[cpp/ycm/CMakeLists.txt, line 54 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pg6wkph1daHOoW7UX:-L2qikLiQuTit2NKr7tT:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/cpp/ycm/CMakeLists.txt#L54)):*

These two lines and the ones below are not properly aligned.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897)*

puremourning:
Review status: 1 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[make_llvm_package.py, line 192 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pjn231kf3cOQDjqfp:-L2qk8C2L3n_VED31VQQ:bd7irh9) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L192)):*

My thoughts exactly; unless it is implicitly obvious in some way (like `cat file`), i prefer to always use named arguments, as otherwise you have to somehow remember the order. There's no obvious order in this case, nor is it obvious what the arguments would be. I do have a strong preference for this :)

Whatever, it's only going to be run a few times a year. Let's make a decision, because the cost of bikeshedding is mainly retesting everything, so I would like to do it all in one go.
</blockquote></details>

OK i implemented:

- env vars for user/pass as this is pretty annoying to type in. save it in the .bashrc once and done.
- defaults for everything except version

new command line is:

```
./make_llvm_bundle.py 5.0.1
```

which seems a reasonable compromise

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897)*

micbou:
Reviewed 2 of 2 files at r2, 2 of 2 files at r3.

---

*[make_llvm_package.py, line 192 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pjn231kf3cOQDjqfp:-L2r19jR98M4LwERcIZj:b-83sjoh) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/make_llvm_package.py#L192)):*

OK i implemented:

- env vars for user/pass as this is pretty annoying to type in. save it in the .bashrc once and done.
- defaults for everything except version

new command line is:

```
./make_llvm_bundle.py 5.0.1
```

which seems a reasonable compromise

</blockquote></details>

Nice!

---

*[make_llvm_package.py, line 200 at r3](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2r1J3t03erTcamduQ9:-L2r1J3u_OdtyzfY_4Nb:b-1rgbjg) ([raw file](https://github.com/valloric/ycmd/blob/e0c5196bd11cca366940e7f6a49c3beb0c3550c7/make_llvm_package.py#L200)):*
> ```Python
> def ParseArguments():
>   parser = argparse.ArgumentParser()
>   parser.add_argument( 'version', action='store', help = 'The LLVM Version' )
> ```

No capital letter to `version`?

---

*[make_llvm_package.py, line 316 at r3](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2r1q3qE6Y20V5rTt3T:-L2r1q3qE6Y20V5rTt3U:b-w4teir) ([raw file](https://github.com/valloric/ycmd/blob/e0c5196bd11cca366940e7f6a49c3beb0c3550c7/make_llvm_package.py#L316)):*
> ```Python
>     print( "Checksum for {bundle_file_name}: {sha256}".format(
>       bundle_file_name = bundle_file_name,
>       sha256 = sha256 ) )
> ```

Excellent. I was going to suggest that.

---

*[cpp/ycm/CMakeLists.txt, line 25 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pg5-I-C3ifY5RXTiF:-L2qkX1_1X3G9Uc9jsWl:b-g3fc2r) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/cpp/ycm/CMakeLists.txt#L25)):*

So is the line below it. Given that i was unable to find [a non-horrible line continuation](https://stackoverflow.com/questions/7637539/how-to-split-strings-across-multiple-lines-in-cmake) in cmake, i decided to keep this. Let me know if there is a better way to do this, other than something like CONCAT .... ... ?
</blockquote></details>

Right. I forgot that CMake is not a proper language.

---

*[cpp/ycm/CMakeLists.txt, line 54 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pg6wkph1daHOoW7UX:-L2qkqf7joO8Msz_lkHe:bwap8wb) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/cpp/ycm/CMakeLists.txt#L54)):*

Done.
</blockquote></details>

You updated the two lines below but not the ones above.

---

*[cpp/ycm/CMakeLists.txt, line 97 at r3](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2r3Dx4sI0QZFst4qQb:-L2r3Dx5UyrbguyjIKd6:b-5ewnau) ([raw file](https://github.com/valloric/ycmd/blob/e0c5196bd11cca366940e7f6a49c3beb0c3550c7/cpp/ycm/CMakeLists.txt#L97)):*
> ```CMake
>             "libclang-${CLANG_VERSION}-x86_64-linux-gnu-ubuntu-14.04" )
>        set( LIBCLANG_SHA256
>             "20894c75f578bab6533a93b4ae9370f734093cdb074d944c339067e9f2adaaea" )
> ```

These two lines are misaligned too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897)*

puremourning:

---

*[make_llvm_package.py, line 200 at r3](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2r1J3t03erTcamduQ9:-L2r7OuhSXRuGa3xOtCs:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/e0c5196bd11cca366940e7f6a49c3beb0c3550c7/make_llvm_package.py#L200)):*

No capital letter to `version`?
</blockquote></details>

Done.

---

*[cpp/ycm/CMakeLists.txt, line 54 at r1](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2pg6wkph1daHOoW7UX:-L2r5dc6QOZOkdoImgSI:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/173d45835a7db69e8eb9e6359116b9b9b91a8684/cpp/ycm/CMakeLists.txt#L54)):*

You updated the two lines below but not the ones above.
</blockquote></details>

Done.

---

*[cpp/ycm/CMakeLists.txt, line 97 at r3](https://reviewable.io:443/reviews/valloric/ycmd/897#-L2r3Dx4sI0QZFst4qQb:-L2r5eY8xS_DmVkUDXcL:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/e0c5196bd11cca366940e7f6a49c3beb0c3550c7/cpp/ycm/CMakeLists.txt#L97)):*

These two lines are misaligned too.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897)*

micbou:
:lgtm:

@zzbot r+

---

Reviewed 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897#-:-L2rC9wdeEy9eQfJOTwZ:b-a2cmdb)*

micbou:
Travis is alive!

@zzbot retry

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/897#-:-L32uEs8bxQlSWb7YWcL:bg2hp0b)*

PR #919, opened 2018-02-08
[READY] Cache include paths

bstaletic:
Since @davits said he won't be able to finish #873, I've updated his PR according to the review comments.

All credits go to @davits for actually doing all the work here.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919)

### DIFF
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -39,8 +39,7 @@ from ycmd.completers.cpp.flags import ( Flags, PrepareFlagsForClang,
                                         NoCompilationDatabase,
                                         UserIncludePaths )
 from ycmd.completers.cpp.ephemeral_values_set import EphemeralValuesSet
-from ycmd.completers.general.filename_completer import (
-    GenerateCandidatesForPaths )
+from ycmd.completers.cpp.include_cache import IncludeCache, IncludeList
 from ycmd.responses import NoExtraConfDetected, UnknownExtraConf
 
 CLANG_FILETYPES = set( [ 'c', 'cpp', 'objc', 'objcpp' ] )
@@ -62,6 +61,7 @@ class ClangCompleter( Completer ):
       'max_diagnostics_to_display' ]
     self._completer = ycm_core.ClangCompleter()
     self._flags = Flags()
+    self._include_cache = IncludeCache()
     self._diagnostic_store = None
     self._files_being_compiled = EphemeralValuesSet()
     self._logger = logging.getLogger( __name__ )
@@ -120,20 +120,12 @@ class ClangCompleter( Completer ):
     if quoted_include:
       include_paths.extend( quoted_include_paths )
 
-    paths = []
+    includes = IncludeList()
     for include_path in include_paths:
       unicode_path = ToUnicode( os.path.join( include_path, path_dir ) )
-      try:
-        # We need to pass a unicode string to get unicode strings out of
-        # listdir.
-        relative_paths = os.listdir( unicode_path )
-      except Exception:
-        self._logger.exception( 'Error while listing %s folder.', unicode_path )
-        relative_paths = []
+      includes.AddIncludes( self._include_cache.GetIncludes( unicode_path ) )
 
-      paths.extend( os.path.join( include_path, path_dir, relative_path ) for
-                    relative_path in relative_paths  )
-    return paths
+    return includes.GetIncludes()
 
 
   def ComputeCandidatesInner( self, request_data ):
@@ -141,9 +133,9 @@ class ClangCompleter( Completer ):
     if not flags:
       raise RuntimeError( NO_COMPILE_FLAGS_MESSAGE )
 
-    paths = self.GetIncludePaths( request_data )
-    if paths is not None:
-      return GenerateCandidatesForPaths( paths )
+    includes = self.GetIncludePaths( request_data )
+    if includes is not None:
+      return includes
 
     if self._completer.UpdatingTranslationUnit(
         ToCppStringCompatible( filename ) ):
@@ -326,8 +318,10 @@ class ClangCompleter( Completer ):
 
     return response_builder( message )
 
+
   def _ClearCompilationFlagCache( self ):
     self._flags.Clear()
+
 
   def _FixIt( self, request_data ):
     flags, filename = self._FlagsForRequest( request_data )
@@ -351,6 +345,7 @@ class ClangCompleter( Completer ):
     # in a nice way
 
     return responses.BuildFixItResponse( fixits )
+
 
   def OnFileReadyToParse( self, request_data ):
     flags, filename = self._FlagsForRequest( request_data )
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -683,36 +683,38 @@ def _GetCompilationInfoForFile( database, file_name, file_extension ):
   return None
 
 
-def UserIncludePaths( flags, filename ):
-  quoted_include_paths = [ os.path.dirname( filename ) ]
+def UserIncludePaths( user_flags, filename ):
+  """
+  Returns a tuple ( quoted_include_paths, include_paths )
+
+  quoted_include_paths is a list of include paths that are only suitable for
+  quoted include statement.
+  include_paths is a list of include paths that can be used for angle bracketed
+  and quoted include statement.
+  """
+  quoted_include_paths = [ ToUnicode( os.path.dirname( filename ) ) ]
   include_paths = []
-  if flags:
-    quote_flag = '-iquote'
-    path_flags = ( [ '-isystem', '-I', '/I' ]
-                   if _ShouldAllowWinStyleFlags( flags )
-                   else [ '-isystem', '-I' ] )
+
+  if user_flags:
+    include_flags = { '-iquote':  quoted_include_paths,
+                      '-I':       include_paths,
+                      '-isystem': include_paths }
+    if _ShouldAllowWinStyleFlags( user_flags ):
+      include_flags[ '/I' ] = include_paths
 
     try:
-      it = iter( flags )
-      for flag in it:
-        flag_len = len( flag )
-        if flag.startswith( quote_flag ):
-          quote_flag_len = len( quote_flag )
-          # Add next flag to the include paths if current flag equals to
-          # '-iquote', or add remaining string otherwise.
-          quoted_include_path = ( next( it ) if flag_len == quote_flag_len else
-                                  flag[ quote_flag_len: ] )
-          if quoted_include_path:
-            quoted_include_paths.append( ToUnicode( quoted_include_path ) )
-        else:
-          for path_flag in path_flags:
-            if flag.startswith( path_flag ):
-              path_flag_len = len( path_flag )
-              include_path = ( next( it ) if flag_len == path_flag_len else
-                               flag[ path_flag_len: ] )
-              if include_path:
-                include_paths.append( ToUnicode( include_path ) )
-              break
+      it = iter( user_flags )
+      for user_flag in it:
+        user_flag_len = len( user_flag )
+        for flag in include_flags:
+          if user_flag.startswith( flag ):
+            flag_len = len( flag )
+            include_path = ( next( it ) if user_flag_len == flag_len else
+                             user_flag[ flag_len: ] )
+            if include_path:
+              container = include_flags[ flag ]
+              container.append( ToUnicode( include_path ) )
+            break
     except StopIteration:
       pass
 
--- a/ycmd/completers/cpp/include_cache.py
+++ b/ycmd/completers/cpp/include_cache.py
@@ -1,0 +1,142 @@
+# Copyright (C) 2017 Davit Samvelyan davitsamvelyan@gmail.com
+#                    Synopsys.
+#               2018 ycmd contributors
+#
+# This file is part of ycmd.
+#
+# ycmd is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ycmd is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+# Not installing aliases from python-future; it's unreliable and slow.
+from builtins import *  # noqa
+
+from future.utils import iteritems
+
+import os
+import threading
+
+from collections import defaultdict, namedtuple
+from ycmd import responses
+from ycmd.completers.general.filename_completer import ( GetPathType,
+                                                         GetPathTypeName )
+
+import logging
+_logger = logging.getLogger( __name__ )
+
+""" Represents single include completion candidate.
+name is the name/string of the completion candidate,
+entry_type is an integer indicating whether the candidate is a
+'File', 'Dir' or both (See EXTRA_INFO_MAP in filename_completer). """
+IncludeEntry = namedtuple( 'IncludeEntry', [ 'name', 'entry_type' ] )
+
+
+class IncludeList( object ):
+  """
+  Helper class for combining include completion candidates from
+  several include paths.
+  self._includes is a dictionary whose keys are
+  IncludeEntry `name`s and values are IncludeEntry `entry_type`s.
+  """
+
+  def __init__( self ):
+    self._includes = defaultdict( int )
+
+
+  def AddIncludes( self, includes ):
+    for include in includes:
+      self._includes[ include.name ] |= include.entry_type
+
+
+  def GetIncludes( self ):
+    includes = []
+    for name, include_type in iteritems( self._includes ):
+      includes.append( responses.BuildCompletionData(
+        name, GetPathTypeName( include_type ) ) )
+    return includes
+
+
+class IncludeCache( object ):
+  """
+  Holds a dictionary representing the include path cache.
+  Dictionary keys are the include path directories.
+  Dictionary values are tuples whose first object
+  represents `mtime` of the dictionary key and the other
+  object is an IncludeList.
+  """
+
+  def __init__( self ):
+    self._cache = {}
+    self._cache_lock = threading.Lock()
+
+
+  def GetIncludes( self, path ):
+    includes = self._GetCached( path )
+
+    if includes is None:
+      includes = self._ListIncludes( path )
+      self._AddToCache( path, includes )
+
+    return includes
+
+
+  def _AddToCache( self, path, includes, mtime = None ):
+    if not mtime:
+      mtime = _GetModificationTime( path )
+    # mtime of 0 is "a magic value" to represent inaccessible directory mtime.
+    if mtime:
+      with self._cache_lock:
+        self._cache[ path ] = { 'mtime': mtime, 'includes': includes }
+
+
+  def _GetCached( self, path ):
+    includes = None
+    with self._cache_lock:
+      cache_entry = self._cache.get( path )
+    if cache_entry:
+      mtime = _GetModificationTime( path )
+      if mtime > cache_entry[ 'mtime' ]:
+        includes = self._ListIncludes( path )
+        self._AddToCache( path, includes, mtime )
+      else:
+        includes = cache_entry[ 'includes' ]
+
+    return includes
+
+
+  def _ListIncludes( self, path ):
+    try:
+      names = os.listdir( path )
+    except OSError:
+      _logger.exception( 'Can not list entries for include path %s.', path )
+      return []
+
+    includes = []
+    for name in names:
+      inc_path = os.path.join( path, name )
+      entry_type = GetPathType( inc_path )
+      includes.append( IncludeEntry( name, entry_type ) )
+
+    return includes
+
+
+def _GetModificationTime( path ):
+  try:
+    return os.path.getmtime( path )
+  except OSError:
+    _logger.exception( 'Can not get modification time for include path %s.',
+                       path )
+    return 0
--- a/ycmd/completers/general/filename_completer.py
+++ b/ycmd/completers/general/filename_completer.py
@@ -25,14 +25,18 @@ from builtins import *  # noqa
 
 import logging
 import os
-from collections import defaultdict
 
 from ycmd.completers.completer import Completer
 from ycmd.utils import ( ExpandVariablesInPath, GetCurrentDirectory, OnWindows,
                          re, ToUnicode )
 from ycmd import responses
 
-EXTRA_INFO_MAP = { 1 : '[File]', 2 : '[Dir]', 3 : '[File&Dir]' }
+FILE = 1
+DIR = 2
+# This mapping is also used for the #include completion.
+# We have option 3, because some entries could simultaneously be
+# both a file and a directory.
+EXTRA_INFO_MAP = { FILE : '[File]', DIR : '[Dir]', 3 : '[File&Dir]' }
 
 _logger = logging.getLogger( __name__ )
 
@@ -104,8 +108,8 @@ class FilenameCompleter( Completer ):
     # working directory of ycmd
     working_dir = request_data.get( 'working_dir' )
 
-    return GenerateCandidatesForPaths(
-      _GetAbsolutePaths(
+    return GeneratePathCompletionData(
+      _GetPathCompletionCandidates(
         path_dir,
         self.user_options[ 'filepath_completion_use_working_dir' ],
         filepath,
@@ -138,40 +142,43 @@ def _GetAbsolutePathForCompletions( path_dir,
                          path_dir )
 
 
-def _GetAbsolutePaths( path_dir, use_working_dir, filepath, working_dir ):
+def _GetPathCompletionCandidates( path_dir, use_working_dir,
+                                  filepath, working_dir ):
+
   absolute_path_dir = _GetAbsolutePathForCompletions( path_dir,
                                                       use_working_dir,
                                                       filepath,
                                                       working_dir )
-
+  entries = []
+  unicode_path = ToUnicode( absolute_path_dir )
   try:
     # We need to pass a unicode string to get unicode strings out of
     # listdir.
-    relative_paths = os.listdir( ToUnicode( absolute_path_dir ) )
+    relative_paths = os.listdir( unicode_path )
   except Exception:
     _logger.exception( 'Error while listing %s folder.', absolute_path_dir )
     relative_paths = []
 
-  return ( os.path.join( absolute_path_dir, relative_path )
-           for relative_path in relative_paths )
+  for rel_path in relative_paths:
+    absolute_path = os.path.join( unicode_path, rel_path )
+    entries.append( ( rel_path, GetPathType( absolute_path ) ) )
+
+  return entries
 
 
-def GenerateCandidatesForPaths( absolute_paths ):
-  extra_info = defaultdict( int )
-  basenames = []
-  for absolute_path in absolute_paths:
-    basename = os.path.basename( absolute_path )
-    if extra_info[ basename ] == 0:
-      basenames.append( basename )
-    is_dir = os.path.isdir( absolute_path )
-    extra_info[ basename ] |= ( 2 if is_dir else 1 )
+def GetPathType( path ):
+  return DIR if os.path.isdir( path ) else FILE
 
+
+def GetPathTypeName( path_type ):
+  return EXTRA_INFO_MAP[ path_type ]
+
+
+def GeneratePathCompletionData( entries ):
   completion_dicts = []
-  # Keep original ordering
-  for basename in basenames:
+  for entry in entries:
     completion_dicts.append(
-      responses.BuildCompletionData(
-        basename,
-        EXTRA_INFO_MAP[ extra_info[ basename ] ] ) )
+      responses.BuildCompletionData( entry[ 0 ],
+        GetPathTypeName( entry[ 1 ] ) ) )
 
   return completion_dicts
--- a/ycmd/tests/clang/include_cache_test.py
+++ b/ycmd/tests/clang/include_cache_test.py
@@ -1,0 +1,160 @@
+# Copyright (C) 2017 Davit Samvelyan davitsamvelyan@gmail.com
+#                    Synopsys.
+#
+# This file is part of ycmd.
+#
+# ycmd is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ycmd is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+# Not installing aliases from python-future; it's unreliable and slow.
+from builtins import *  # noqa
+
+import os
+from time import sleep
+
+from nose.tools import eq_
+from hamcrest import ( assert_that,
+                       contains,
+                       contains_inanyorder,
+                       equal_to,
+                       has_entries,
+                       has_entry,
+                       has_properties,
+                       not_ )
+
+from ycmd.completers.cpp.include_cache import IncludeCache
+from ycmd.tests.clang import PathToTestFile
+from ycmd.tests.test_utils import TemporaryTestDir
+
+
+def IncludeCache_NotCached_DirInaccessible_test():
+  include_cache = IncludeCache()
+  eq_( include_cache._cache, {} )
+  includes = include_cache.GetIncludes( PathToTestFile( 'unknown_dir' ) )
+  eq_( includes, [] )
+  eq_( include_cache._cache, {} )
+
+
+def IncludeCache_NotCached_DirAccessible_test():
+  include_cache = IncludeCache()
+  eq_( include_cache._cache, {} )
+  includes = include_cache.GetIncludes( PathToTestFile( 'cache_test' ) )
+  mtime = os.path.getmtime( PathToTestFile( 'cache_test' ) )
+  assert_that( includes, contains( has_properties( {
+                                     'name': 'foo.h',
+                                     'entry_type': 1
+                                   } ) ) )
+  assert_that( include_cache._cache,
+               has_entry( PathToTestFile( 'cache_test' ),
+                          has_entries( { 'mtime': mtime,
+                            'includes': contains( has_properties( {
+                                                    'name': 'foo.h',
+                                                    'entry_type': 1
+                                                  } ) ) } ) ) )
+
+
+def IncludeCache_Cached_NoNewMtime_test():
+  include_cache = IncludeCache()
+  eq_( include_cache._cache, {} )
+  old_includes = include_cache.GetIncludes( PathToTestFile( 'cache_test' ) )
+  old_mtime = os.path.getmtime( PathToTestFile( 'cache_test' ) )
+
+  assert_that( old_includes, contains( has_properties( {
+                                         'name': 'foo.h',
+                                         'entry_type': 1
+                                       } ) ) )
+  assert_that( include_cache._cache,
+               has_entry( PathToTestFile( 'cache_test' ),
+                          has_entries( { 'mtime': old_mtime,
+                            'includes': contains( has_properties( {
+                                                    'name': 'foo.h',
+                                                    'entry_type': 1
+                                                  } ) ) } ) ) )
+
+  new_includes = include_cache.GetIncludes( PathToTestFile( 'cache_test' ) )
+  new_mtime = os.path.getmtime( PathToTestFile( 'cache_test' ) )
+
+  eq_( new_mtime, old_mtime )
+  assert_that( new_includes, contains( has_properties( {
+                                         'name': 'foo.h',
+                                         'entry_type': 1
+                                       } ) ) )
+  assert_that( include_cache._cache,
+               has_entry( PathToTestFile( 'cache_test' ),
+                          has_entries( { 'mtime': new_mtime,
+                            'includes': contains( has_properties( {
+                                                    'name': 'foo.h',
+                                                    'entry_type': 1
+                                                  } ) ) } ) ) )
+
+
+def IncludeCache_Cached_NewMtime_test():
+  with TemporaryTestDir() as tmp_dir:
+    include_cache = IncludeCache()
+    eq_( include_cache._cache, {} )
+    foo_path = os.path.join( tmp_dir, 'foo' )
+    with open( foo_path, 'w' ) as foo_file:
+      foo_file.write( 'foo' )
+
+    old_includes = include_cache.GetIncludes( tmp_dir )
+    old_mtime = os.path.getmtime( tmp_dir )
+    assert_that( old_includes, contains( has_properties( {
+                                           'name': 'foo',
+                                           'entry_type': 1
+                                         } ) ) )
+    assert_that( include_cache._cache,
+                 has_entry( tmp_dir,
+                   has_entries( {
+                     'mtime': old_mtime,
+                     'includes': contains( has_properties( {
+                       'name': 'foo',
+                       'entry_type': 1
+                     } ) )
+                   } ) ) )
+
+    sleep( 2 )
+
+    bar_path = os.path.join( tmp_dir, 'bar' )
+    with open( bar_path, 'w' ) as bar_file:
+      bar_file.write( 'bar' )
+
+    new_includes = include_cache.GetIncludes( tmp_dir )
+    new_mtime = os.path.getmtime( tmp_dir )
+    assert_that( old_mtime, not_( equal_to( new_mtime ) ) )
+    assert_that( new_includes, contains_inanyorder(
+                                 has_properties( {
+                                   'name': 'foo',
+                                   'entry_type': 1
+                                 } ),
+                                 has_properties( {
+                                   'name': 'bar',
+                                   'entry_type': 1
+                                 } )
+                               ) )
+    assert_that( include_cache._cache,
+        has_entry( tmp_dir, has_entries( {
+                              'mtime': new_mtime,
+                              'includes': contains_inanyorder(
+                                has_properties( {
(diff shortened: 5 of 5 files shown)


bstaletic:
Review status: 0 of 5 files reviewed at latest revision, all discussions resolved.

---

*[ycmd/completers/cpp/flags.py, line 678 at r2](https://beta.reviewable.io/reviews/valloric/ycmd/919#-L4ppOb973FGP8_Vvp0s:-L4ppObA4vwwpLpHF8_m:b-tscgbp) ([raw file](https://github.com/valloric/ycmd/blob/a89ab5c08133f28c9989fdf41bde88b0050b4173/ycmd/completers/cpp/flags.py#L678)):*
> ```Python
>                       '-isystem': ( include_paths, True ) }
>     if _ShouldAllowWinStyleFlags( user_flags ):
>       include_flags[ '/I' ] = ( include_paths, False )
> ```

@puremourning mentioned somehow making the use of the include flags at the top of the file.
Considering the flags at the top of the file are somewhat different, what would be the best approach?

---

*[ycmd/completers/cpp/flags.py, line 682 at r2](https://beta.reviewable.io/reviews/valloric/ycmd/919#-L4pqUH48fgecj_20K6r:-L4pqUH48fgecj_20K6s:b-k7xufj) ([raw file](https://github.com/valloric/ycmd/blob/a89ab5c08133f28c9989fdf41bde88b0050b4173/ycmd/completers/cpp/flags.py#L682)):*
> ```Python
>     try:
>       it = iter( user_flags )
>       for user_flag in it:
> ```

@puremourning This shouldn't use `iterkeys` because it iterates over a list, not over dictionary keys.

---

*[ycmd/completers/cpp/include_cache.py, line 115 at r2](https://beta.reviewable.io/reviews/valloric/ycmd/919#-L4pqn4x3IioxoBHg9T4:-L4pqn4x3IioxoBHg9T5:b-e0a4rg) ([raw file](https://github.com/valloric/ycmd/blob/a89ab5c08133f28c9989fdf41bde88b0050b4173/ycmd/completers/cpp/include_cache.py#L115)):*
> ```Python
>         del self._cache[ path ]
>       else:
>         includes = cache_entry[ 1 ]
> ```

@puremourning We could probably update the whole directory entry instead of just deleting it, but...

We still need to do most of the work we are already doing, yet the code would get less readable.
In case you believe this is worthwhile, I'll gladly do it. After all, I'm not the one having troubles with metadata performance.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/919)*

bstaletic:
Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 678 at r2](https://beta.reviewable.io/reviews/valloric/ycmd/919#-L4ppOb973FGP8_Vvp0s:-L4prtcD6NWu8vyx1lGA:bg53ywc) ([raw file](https://github.com/valloric/ycmd/blob/a89ab5c08133f28c9989fdf41bde88b0050b4173/ycmd/completers/cpp/flags.py#L678)):*

@puremourning mentioned somehow making the use of the include flags at the top of the file.
Considering the flags at the top of the file are somewhat different, what would be the best approach?
</blockquote></details>

The reason we're not going "full-hog" and caching everything right now is that the lines above make it easy to test what works the best.

@puremourning Once you say what works best, we could even drop tupples and maybe unconditionally use cache.

---

*[ycmd/completers/cpp/include_cache.py, line 115 at r2](https://beta.reviewable.io/reviews/valloric/ycmd/919#-L4pqn4x3IioxoBHg9T4:-L4ptGqA-4xhlAr_8xOw:bnajdfe) ([raw file](https://github.com/valloric/ycmd/blob/a89ab5c08133f28c9989fdf41bde88b0050b4173/ycmd/completers/cpp/include_cache.py#L115)):*

@puremourning We could probably update the whole directory entry instead of just deleting it, but...

We still need to do most of the work we are already doing, yet the code would get less readable.
In case you believe this is worthwhile, I'll gladly do it. After all, I'm not the one having troubles with metadata performance.
</blockquote></details>

Actually, if we change `_AddToCache(  self, path, includes )` to `_AddToCache( self, path, includes, mtime = None )`, then this can be rewritten like:

```python
  if mtime > cache_entry[ 0 ]:
    _AddToCache( path, self._ListInclude( path ), mtime )
  includes = cache_entry[ 1]
```

This looks like it should work.
Am I missing something?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/919)*

bstaletic:
Coverage can be improved, but before I get to writing more tests, do we want/need `ClearIncludeCache` completer subcommand?

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/919#-:-L4q_8A3965-BVHHU55b:b-l0n4gc)*

davits:
@bstaletic thanks a lot for continuing to work on this :)

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/919#-:-L5OPxgo2G5nOtxfkh4X:b-toymg2)*

bstaletic:
@davits No problem. You have done most of the work though. Also, since I can't test it, if you have the time give it another test.

davits:
Tested, Acceleration confirmed!

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/919#-:-L5P4AxBCN3rUVDw1MY5:b-lqjhgo)*

bstaletic:
Thanks for testing! 👍

Did you test it with only system header caching? How does it behave if you
let it cache all headers?

davits:
In my case there is no much difference, since almost all include paths are given with '-isystem', there are only 2 paths given with '-I'. I think it wouldn't hurt to also cache '-I' flags, that folders rarely change as well, in my case at least.

---

Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/919#-:-L5P6qQHf1vh016n_1IP:b-dg2wxp)*

bstaletic:
Great, your include flags are the opposite of @puremourning's. I'll wait
for his test results and then update the PR appropriately.

Once again, thanks for testing.

bstaletic:
I've pushed the tests.  This is ready, unless someone has more to say.

---

Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 678 at r2](https://reviewable.io:443/reviews/valloric/ycmd/919#-L4ppOb973FGP8_Vvp0s:-L6lrz6J9vpUr0NF9VRc:bolleug) ([raw file](https://github.com/valloric/ycmd/blob/60378e022ff6a316c4eb89ce13ebc59ace0e4440/ycmd/completers/cpp/flags.py#L678)):*

The reason we're not going "full-hog" and caching everything right now is that the lines above make it easy to test what works the best.

@puremourning Once you say what works best, we could even drop tupples and maybe unconditionally use cache.
</blockquote></details>

Okay, we are definitely caching everything.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/919#-:-L6lrjnmcnr66W4bF6-f:bfi813)*

bstaletic:
I squashed the commits.
Since two people have said this works and nobody complained about anything I'll mark this as ready.

---

Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/919#-:-L86syUNpgGP5aIThqgs:bfdg9u)*

bstaletic:
Conflicts resolved.

micbou:
Reviewed 1 of 3 files at r2, 1 of 2 files at r4, 4 of 6 files at r5, 1 of 1 files at r6, 3 of 3 files at r7, 3 of 3 files at r8.

---

*[ycmd/completers/cpp/flags.py, line 669 at r6](https://beta.reviewable.io/reviews/valloric/ycmd/919#-L7-NVJY1RmjwKO3eyw_:-L7-NVJY1RmjwKO3eywa:b9pod3n) ([raw file](https://github.com/valloric/ycmd/blob/096c02a39e4822e16afb467b7f859385ef4ecfbf/ycmd/completers/cpp/flags.py#L669)):*
> ```Python
> def UserIncludePaths( user_flags, filename ):
>   """
>   Returns a tupple ( quoted_include_paths, include_paths )
> ```

`tuple`

---

*[ycmd/completers/cpp/include_cache.py, line 51 at r8](https://beta.reviewable.io/reviews/valloric/ycmd/919#-LAN9ooL-IGhb-iIXYqS:-LAN9ooL-IGhb-iIXYqT:b8zrcnp) ([raw file](https://github.com/valloric/ycmd/blob/5ec9a2a04cba66ac8b4efd5136afb90f5d2ebc88/ycmd/completers/cpp/include_cache.py#L51)):*
> ```Python
>   Helper class for combining include completion candidates from
>   several include paths.
>   slef._includes is a dictionary whose keys are
> ```

`self`.

---

*[ycmd/completers/cpp/include_cache.py, line 76 at r8](https://beta.reviewable.io/reviews/valloric/ycmd/919#-LAN9tlR9mI6PwNh3I8e:-LAN9tlR9mI6PwNh3I8f:bgzj2re) ([raw file](https://github.com/valloric/ycmd/blob/5ec9a2a04cba66ac8b4efd5136afb90f5d2ebc88/ycmd/completers/cpp/include_cache.py#L76)):*
> ```Python
>   Holds a dictionary representing the include path cache.
>   Dictionary keys are the include path directories.
>   Dictionary values are tupples whose first object
> ```

`tuples`

---

*[ycmd/completers/cpp/include_cache.py, line 82 at r8](https://beta.reviewable.io/reviews/valloric/ycmd/919#-LANYyho1pqG9LSk3i3_:-LANYyho1pqG9LSk3i3a:b-bhdztt) ([raw file](https://github.com/valloric/ycmd/blob/5ec9a2a04cba66ac8b4efd5136afb90f5d2ebc88/ycmd/completers/cpp/include_cache.py#L82)):*
> ```Python
> 
>   def __init__( self ):
>     self._cache = {}
> ```

I think a lock is needed for this variable as it can be simultaneously updated by multiple threads in the Clang completer.

---

*[ycmd/completers/cpp/include_cache.py, line 107 at r8](https://beta.reviewable.io/reviews/valloric/ycmd/919#-LANU5iLB5pyR-SpeFbG:-LANU5iLB5pyR-SpeFbH:b-5mnpvx) ([raw file](https://github.com/valloric/ycmd/blob/5ec9a2a04cba66ac8b4efd5136afb90f5d2ebc88/ycmd/completers/cpp/include_cache.py#L107)):*
> ```Python
>     if cache_entry:
>       mtime = _GetModificationTime( path )
>       if mtime > cache_entry[ 'mtime' ]:
> ```

This raises a `TypeError` exception if `mtime` is `None`.

---

*[ycmd/completers/general/filename_completer.py, line 39 at r7](https://beta.reviewable.io/reviews/valloric/ycmd/919#-L8N_3yE5qVm7FPSg9Q6:-L8N_3yE5qVm7FPSg9Q7:byqqe8c) ([raw file](https://github.com/valloric/ycmd/blob/fbd064b60df3f64b4fc5f543a6f73a08ef98f782/ycmd/completers/general/filename_completer.py#L39)):*
> ```Python
> # This mapping is also used for the #include completion.
> # We have option N3, because when using Qt with specific include paths
> # configuration both file & dir entries can exist.
> ```

This comment is misleading. It makes the reader thinks it's specific to Qt. It's not. We should simply say that an entry can be simultaneously a file and a directory when completing include statements.

---

*[ycmd/tests/clang/include_cache_test.py, line 129 at r8](https://beta.reviewable.io/reviews/valloric/ycmd/919#-LAN6UcT-S6czCJ3vMdH:-LAN6UcT-S6czCJ3vMdI:b-mbdikl) ([raw file](https://github.com/valloric/ycmd/blob/5ec9a2a04cba66ac8b4efd5136afb90f5d2ebc88/ycmd/tests/clang/include_cache_test.py#L129)):*
> ```Python
>                    } ) ) )
> 
>     sleep( 10 )
> ```

`sleep( 2 )` (or even `sleep( 1.001 )`) should be enough.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/919)*

bstaletic:
Review status: 2 of 6 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

*[ycmd/completers/cpp/flags.py, line 669 at r6](https://beta.reviewable.io/reviews/valloric/ycmd/919#-L7-NVJY1RmjwKO3eyw_:-LANsYL83fe8-5OsYjLb:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/096c02a39e4822e16afb467b7f859385ef4ecfbf/ycmd/completers/cpp/flags.py#L669)):*

`tuple`
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/include_cache.py, line 51 at r8](https://beta.reviewable.io/reviews/valloric/ycmd/919#-LAN9ooL-IGhb-iIXYqS:-LANriuE3oLOrw3qbMlw:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/5ec9a2a04cba66ac8b4efd5136afb90f5d2ebc88/ycmd/completers/cpp/include_cache.py#L51)):*

`self`.
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/include_cache.py, line 76 at r8](https://beta.reviewable.io/reviews/valloric/ycmd/919#-LAN9tlR9mI6PwNh3I8e:-LANrjHK9VsOpKIyI4xj:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/5ec9a2a04cba66ac8b4efd5136afb90f5d2ebc88/ycmd/completers/cpp/include_cache.py#L76)):*

`tuples`
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/include_cache.py, line 82 at r8](https://beta.reviewable.io/reviews/valloric/ycmd/919#-LANYyho1pqG9LSk3i3_:-LAO0qff546GUGIkUMQL:b-egr7g2) ([raw file](https://github.com/valloric/ycmd/blob/5ec9a2a04cba66ac8b4efd5136afb90f5d2ebc88/ycmd/completers/cpp/include_cache.py#L82)):*

I think a lock is needed for this variable as it can be simultaneously updated by multiple threads in the Clang completer.
</blockquote></details>

Done, I believe. Feel free to tell me if I messed something up.

---

*[ycmd/completers/cpp/include_cache.py, line 107 at r8](https://beta.reviewable.io/reviews/valloric/ycmd/919#-LANU5iLB5pyR-SpeFbG:-LANx4hr77h-OxIk1IKr:b-py7txx) ([raw file](https://github.com/valloric/ycmd/blob/5ec9a2a04cba66ac8b4efd5136afb90f5d2ebc88/ycmd/completers/cpp/include_cache.py#L107)):*

This raises a `TypeError` exception if `mtime` is `None`.
</blockquote></details>

I changed the `_GetModificationTime()` to return 0 if `getmtime()` fails. That should allow us to avoid another try/except.

---

*[ycmd/completers/general/filename_completer.py, line 39 at r7](https://beta.reviewable.io/reviews/valloric/ycmd/919#-L8N_3yE5qVm7FPSg9Q6:-LANsGcvCzd7d7Y39Pha:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/fbd064b60df3f64b4fc5f543a6f73a08ef98f782/ycmd/completers/general/filename_completer.py#L39)):*

This comment is misleading. It makes the reader thinks it's specific to Qt. It's not. We should simply say that an entry can be simultaneously a file and a directory when completing include statements.
</blockquote></details>

Done.

---

*[ycmd/tests/clang/include_cache_test.py, line 129 at r8](https://beta.reviewable.io/reviews/valloric/ycmd/919#-LAN6UcT-S6czCJ3vMdH:-LANsR8h-Nifl9c27b3w:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/5ec9a2a04cba66ac8b4efd5136afb90f5d2ebc88/ycmd/tests/clang/include_cache_test.py#L129)):*

`sleep( 2 )` (or even `sleep( 1.001 )`) should be enough.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/919)*

micbou:
Reviewed 4 of 4 files at r9.

---

*[ycmd/tests/clang/include_cache_test.py, line 44 at r10](https://beta.reviewable.io/reviews/valloric/ycmd/919#-LAPEpnJAxrR2NyrO918:-LAPEpnJAxrR2NyrO919:btuvqkh) ([raw file](https://github.com/valloric/ycmd/blob/be1e559ba7e8ad5ff043321c33187477d2ab8be8/ycmd/tests/clang/include_cache_test.py#L44)):*
> ```Python
> 
> 
> def GetCache_NotCached_DirInaccessible_test():
> ```

Why are the name of the tests starting with `GetCache`? Wouldn't `IncludeCache` be a better choice?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/919)*

bstaletic:
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/tests/clang/include_cache_test.py, line 44 at r10](https://beta.reviewable.io/reviews/valloric/ycmd/919#-LAPEpnJAxrR2NyrO918:-LARspI46f0SQfQK4Stn:bg1ktfs) ([raw file](https://github.com/valloric/ycmd/blob/be1e559ba7e8ad5ff043321c33187477d2ab8be8/ycmd/tests/clang/include_cache_test.py#L44)):*

Why are the name of the tests starting with `GetCache`? Wouldn't `IncludeCache` be a better choice?
</blockquote></details>

You're right. Fixed.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/919)*

micbou:
:lgtm:

---

Reviewed 1 of 1 files at r11.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/919#-:-LAU3fNW0vYe_glbVNhD:bnfp4nl)*

puremourning:
Reviewed 2 of 6 files at r5.

---

*[ycmd/completers/cpp/flags.py, line 673 at r5](https://reviewable.io/reviews/valloric/ycmd/919#-L6lvfGMk6g9F3--_6le:-L6lvfGMk6g9F3--_6lf:b-tvdsqo) ([raw file](https://github.com/valloric/ycmd/blob/3c33b627a175a86cf1d519c1d4d0ec3684e28a27/ycmd/completers/cpp/flags.py#L673)):*
> ```Python
>   quoted_include_paths is a list of include paths that are only suitable for
>   quoted include statement.
>   include_paths is a list of include paths that can be used for angle bracked
> ```

bracket or perhaps bracketed

---

*[ycmd/completers/cpp/include_cache.py, line 115 at r2](https://reviewable.io/reviews/valloric/ycmd/919#-L4pqn4x3IioxoBHg9T4:-LB2JShk00zH5jSyXA2g:b-p1qhun) ([raw file](https://github.com/valloric/ycmd/blob/60378e022ff6a316c4eb89ce13ebc59ace0e4440/ycmd/completers/cpp/include_cache.py#L115)):*

Actually, if we change `_AddToCache(  self, path, includes )` to `_AddToCache( self, path, includes, mtime = None )`, then this can be rewritten like:

```python
  if mtime > cache_entry[ 0 ]:
    _AddToCache( path, self._ListInclude( path ), mtime )
  includes = cache_entry[ 1]
```

This looks like it should work.
Am I missing something?
</blockquote></details>

seems ok. We're not reading the time multiple times, but I think the other path means we don't set includes anymore

---

*[ycmd/completers/cpp/include_cache.py, line 36 at r5](https://reviewable.io/reviews/valloric/ycmd/919#-L6lz6NALSOFg6g-Enud:-L6lz6NALSOFg6g-Enue:by32v31) ([raw file](https://github.com/valloric/ycmd/blob/3c33b627a175a86cf1d519c1d4d0ec3684e28a27/ycmd/completers/cpp/include_cache.py#L36)):*
> ```Python
> 
> 
> class IncludeEntry( object ):
> ```

maybe use a named tuple ? saves on some code. meh. no need to change.

---

*[ycmd/completers/cpp/include_cache.py, line 1 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2KNrwBXLSIQL3LhhE:-LB2KNrwBXLSIQL3LhhF:b-w1z8a9) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L1)):*
> ```Python
> # Copyright (C) 2017 Davit Samvelyan davitsamvelyan@gmail.com
> ```

I suppose as you have made changes to this, you can claim some copyright ownership as well?

---

*[ycmd/completers/cpp/include_cache.py, line 98 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2KmbL2JOdu5VYmw0J:-LB2KmbL2JOdu5VYmw0K:b-ln5gfu) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L98)):*
> ```Python
> 
>   def _AddToCache( self, path, includes, mtime = None ):
>     if not mtime:
> ```

not ? or "is not None" ? Do we want "falsy" to trigger here?

---

*[ycmd/completers/cpp/include_cache.py, line 108 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2IssH3uxv8LFJj0Mm:-LB2IssH3uxv8LFJj0Mn:btkm2lm) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L108)):*
> ```Python
>     includes = None
>     with self._cache_lock:
>       cache_entry = self._cache.get( path )
> ```

do we mutate cache_entry? if so we provably need to hold the lock until we have made any modifications. Otherwise there is a race condition where you acquire a pointer to the entry, then someone else does the same, then you update the list, then someone else updates the list.

Probably not the end of the world if it happens (we just update the cache 2x rather than blocking). Hmm :thinking: 

---

*[ycmd/completers/cpp/include_cache.py, line 112 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2LD8RA9KgcxRRN8uZ:-LB2LD8RA9KgcxRRN8u_:b-ro7xpd) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L112)):*
> ```Python
>       mtime = _GetModificationTime( path )
>       if mtime > cache_entry[ 'mtime' ]:
>         self._AddToCache( path, self._ListIncludes( path ), mtime )
> ```

includes doesn't get set in this branch

---

*[ycmd/completers/cpp/include_cache.py, line 123 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2J_RBCwkdRR9zPEl_:-LB2J_RBCwkdRR9zPEla:b-8hog5a) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L123)):*
> ```Python
>       names = os.listdir( path )
>     except OSError:
>       return []
> ```

any reason not to _logger.exception this ? if not, let's comment that "the directory no longer exists or is no longer accessible, it must be empty" or something/

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919)*

puremourning:
Reviewed 1 of 6 files at r5, 1 of 3 files at r8, 3 of 4 files at r9, 1 of 1 files at r11.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919)*

bstaletic:
Review status: 4 of 6 files reviewed at latest revision, 8 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 673 at r5](https://reviewable.io/reviews/valloric/ycmd/919#-L6lvfGMk6g9F3--_6le:-LB2RhiR5_QaTr6lhOWv:b2shjzj) ([raw file](https://github.com/valloric/ycmd/blob/3c33b627a175a86cf1d519c1d4d0ec3684e28a27/ycmd/completers/cpp/flags.py#L673)):*

bracket or perhaps bracketed
</blockquote></details>

Done.
Opted for bracketed.

---

*[ycmd/completers/cpp/include_cache.py, line 115 at r2](https://reviewable.io/reviews/valloric/ycmd/919#-L4pqn4x3IioxoBHg9T4:-LB2YZJUAX0PrhJvm4RG:b2jfedg) ([raw file](https://github.com/valloric/ycmd/blob/60378e022ff6a316c4eb89ce13ebc59ace0e4440/ycmd/completers/cpp/include_cache.py#L115)):*

seems ok. We're not reading the time multiple times, but I think the other path means we don't set includes anymore
</blockquote></details>

Yeah, that's what I ended up doing.

---

*[ycmd/completers/cpp/include_cache.py, line 36 at r5](https://reviewable.io/reviews/valloric/ycmd/919#-L6lz6NALSOFg6g-Enud:-LB2ULiE5iZNVSOBjxLf:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/3c33b627a175a86cf1d519c1d4d0ec3684e28a27/ycmd/completers/cpp/include_cache.py#L36)):*

maybe use a named tuple ? saves on some code. meh. no need to change.
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/include_cache.py, line 1 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2KNrwBXLSIQL3LhhE:-LB2RxbiErNSzC78MO3H:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L1)):*

I suppose as you have made changes to this, you can claim some copyright ownership as well?
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/include_cache.py, line 98 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2KmbL2JOdu5VYmw0J:-LB2O1A9-CtZangjEy5f:bqkazqs) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L98)):*

not ? or "is not None" ? Do we want "falsy" to trigger here?
</blockquote></details>

Upon reading this again I was sure you were right.
Once I switched to `is not None` the cache completely broke and I have no idea why.

In this method `mtime` can have three values: `None`, `0`, or something valid.

- `None` should mean the `mtime` wasn't passed to the method.
- `0` should mean we couldn't read the `mtime`.
- Valid value should be stored.

---

*[ycmd/completers/cpp/include_cache.py, line 108 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2IssH3uxv8LFJj0Mm:-LB2VL671D4WkOYcDp-D:b-dz63aj) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L108)):*

do we mutate cache_entry? if so we provably need to hold the lock until we have made any modifications. Otherwise there is a race condition where you acquire a pointer to the entry, then someone else does the same, then you update the list, then someone else updates the list.

Probably not the end of the world if it happens (we just update the cache 2x rather than blocking). Hmm :thinking: 
</blockquote></details>

We potentially mutate the cache in case the new `mtime` is newer than the stored one.
I tend to think that one should avoid locks as much as possible, but feel free to prove me wrong and tell me that we should block.

---

*[ycmd/completers/cpp/include_cache.py, line 112 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2LD8RA9KgcxRRN8uZ:-LB2XTPuBJj12ils-Urk:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L112)):*

includes doesn't get set in this branch
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/include_cache.py, line 123 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2J_RBCwkdRR9zPEl_:-LB2X3eB8P0ucWyUj2WR:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L123)):*

any reason not to _logger.exception this ? if not, let's comment that "the directory no longer exists or is no longer accessible, it must be empty" or something/
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919)*

puremourning:
:lgtm: with a few comments

---

Review status: 4 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/cpp/include_cache.py, line 1 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2KNrwBXLSIQL3LhhE:-LB2hXzk-RkPdjV0KXzL:b-a2p1v3) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L1)):*

Done.
</blockquote></details>

I think Synopsis is the company @davits works for, so it should be kept attached.

---

*[ycmd/completers/cpp/include_cache.py, line 108 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2IssH3uxv8LFJj0Mm:-LB2iBxK3rSeY5Uevjbc:b-32ro93) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L108)):*

We potentially mutate the cache in case the new `mtime` is newer than the stored one.
I tend to think that one should avoid locks as much as possible, but feel free to prove me wrong and tell me that we should block.
</blockquote></details>

I'm sure it's fine. Worst case is we stash the dir 2x in a very unlikely race condition

---

*[ycmd/completers/cpp/include_cache.py, line 112 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2LD8RA9KgcxRRN8uZ:-LB2iIqEAuwT0kYbunhY:b-bj9u6p) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L112)):*

Done.
</blockquote></details>

Is there a way to add a test for this? I guess it is a bit of a gap.

---

*[ycmd/completers/cpp/include_cache.py, line 44 at r12](https://reviewable.io/reviews/valloric/ycmd/919#-LB2hsPn5ovvqG9oiNgI:-LB2hsPn5ovvqG9oiNgJ:bquaxtb) ([raw file](https://github.com/valloric/ycmd/blob/7f75f47d2ab02ffe93ba8235eb1b605d9789b095/ycmd/completers/cpp/include_cache.py#L44)):*
> ```Python
> entry_type is an integer indicating whether the candidate is a
> 'File', 'Dir' or both (See EXTRA_INFO_MAP in filename_completer). """
> IncludeEntry = namedtuple( 'IncludeEntry', 'name entry_type' )
> ```

I feel like field names should be a list. This is totally legit, but I find the way named tuple allows whitespace or comma separated strings unusual! Anyway, no action required.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919#-:-LB2iYXGCpuHcd-hw7Fs:b-afkoyi)*

puremourning:
Reviewed 2 of 2 files at r12.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919)*

bstaletic:

---

*[ycmd/completers/cpp/include_cache.py, line 1 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2KNrwBXLSIQL3LhhE:-LB2lzUq1G7p_csvJgPS:b-870ocv) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L1)):*

I think Synopsis is the company @davits works for, so it should be kept attached.
</blockquote></details>

I had no idea. Fixed.

---

*[ycmd/completers/cpp/include_cache.py, line 112 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2LD8RA9KgcxRRN8uZ:-LB2ms1hBiH-P4-qY1gI:b-ld6htn) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L112)):*

Is there a way to add a test for this? I guess it is a bit of a gap.
</blockquote></details>

There is a `IncludeCache_Cached_NewMtime_test()` which makes sure the correct thing is placed in the cache in this branch.
I'm not sure how to write a test that makes sure `includes` is always set.

---

*[ycmd/completers/cpp/include_cache.py, line 44 at r12](https://reviewable.io/reviews/valloric/ycmd/919#-LB2hsPn5ovvqG9oiNgI:-LB2mFXx1Uv2AJBhO2pw:bd68gr3) ([raw file](https://github.com/valloric/ycmd/blob/7f75f47d2ab02ffe93ba8235eb1b605d9789b095/ycmd/completers/cpp/include_cache.py#L44)):*

I feel like field names should be a list. This is totally legit, but I find the way named tuple allows whitespace or comma separated strings unusual! Anyway, no action required.
</blockquote></details>

Done.

This is what happens when you implement something base on the first blog post you find. I didn't know until now that python had named tuples.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919)*

puremourning:
:lgtm:

---

Review status: 5 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/cpp/include_cache.py, line 112 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2LD8RA9KgcxRRN8uZ:-LB2oxWu0FTaxNonS6CS:brggt3d) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L112)):*

There is a `IncludeCache_Cached_NewMtime_test()` which makes sure the correct thing is placed in the cache in this branch.
I'm not sure how to write a test that makes sure `includes` is always set.
</blockquote></details>

Check the results ?

---

*[ycmd/completers/cpp/include_cache.py, line 44 at r12](https://reviewable.io/reviews/valloric/ycmd/919#-LB2hsPn5ovvqG9oiNgI:-LB2orK29t5C3jqnnuPX:bb7hf0f) ([raw file](https://github.com/valloric/ycmd/blob/7f75f47d2ab02ffe93ba8235eb1b605d9789b095/ycmd/completers/cpp/include_cache.py#L44)):*

Done.

This is what happens when you implement something base on the first blog post you find. I didn't know until now that python had named tuples.
</blockquote></details>

hehe

---

*[ycmd/tests/clang/include_cache_test.py, line 135 at r13](https://reviewable.io/reviews/valloric/ycmd/919#-LB2pURnAOCiZKmbQD1x:-LB2pURnAOCiZKmbQD1y:b-ev1de2) ([raw file](https://github.com/valloric/ycmd/blob/4dc0fd2a9c3cb6acdee03302d0e0b2a3c3a2697f/ycmd/tests/clang/include_cache_test.py#L135)):*
> ```Python
>       bar_file.write( 'bar' )
> 
>     new_includes = include_cache.GetIncludes( tmp_dir )
> ```

Rereading the code, I think this was succeeding because `_GetCached` was returning None, then `GetIncludes` was re-cacheing the directory.

We could in theory prove it with mocking, but it's probably not worth it.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919#-:-LB2pzyH0aHHFHV2TcwS:bnfp4nl)*

puremourning:
Reviewed 1 of 1 files at r13.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919)*

bstaletic:

---

*[ycmd/completers/cpp/include_cache.py, line 112 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2LD8RA9KgcxRRN8uZ:-LB2sCb324IBzPjTosKW:bvq49jd) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L112)):*

Check the results ?
</blockquote></details>

The results of `GetIncludes` is checked. (Note that there is no leading underscore here.)
Are you thinking about basically unit testing `_GetIncludes`?

---

*[ycmd/tests/clang/include_cache_test.py, line 135 at r13](https://reviewable.io/reviews/valloric/ycmd/919#-LB2pURnAOCiZKmbQD1x:-LB2rjPE4iBlIlc64Rt6:b-4lk10l) ([raw file](https://github.com/valloric/ycmd/blob/4dc0fd2a9c3cb6acdee03302d0e0b2a3c3a2697f/ycmd/tests/clang/include_cache_test.py#L135)):*

Rereading the code, I think this was succeeding because `_GetCached` was returning None, then `GetIncludes` was re-cacheing the directory.

We could in theory prove it with mocking, but it's probably not worth it.
</blockquote></details>

Yup, you're completely right, though I'm not sure what to mock.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919)*

puremourning:

---

*[ycmd/completers/cpp/include_cache.py, line 112 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2LD8RA9KgcxRRN8uZ:-LB2xG7VCnIskdxCNwFd:b-cph7ts) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L112)):*

The results of `GetIncludes` is checked. (Note that there is no leading underscore here.)
Are you thinking about basically unit testing `_GetIncludes`?
</blockquote></details>

Sorry I meant to delete that comment, after adding the other one about the test.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919)*

micbou:
Reviewed 2 of 2 files at r12, 1 of 1 files at r13.

---

*[ycmd/completers/cpp/include_cache.py, line 123 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2J_RBCwkdRR9zPEl_:-LB5Vdq34egWxAl4dbnd:b59yxpg) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L123)):*

Done.
</blockquote></details>

`_logger.exception` should be used so that we know where the error comes from in the logs.

---

*[ycmd/completers/cpp/include_cache.py, line 142 at r13](https://reviewable.io/reviews/valloric/ycmd/919#-LB5VNtiCTc3sDqRUNYb:-LB5VNtiCTc3sDqRUNYc:b1ak36s) ([raw file](https://github.com/valloric/ycmd/blob/4dc0fd2a9c3cb6acdee03302d0e0b2a3c3a2697f/ycmd/completers/cpp/include_cache.py#L142)):*
> ```Python
>   except OSError:
>     _logger.debug( 'The directory {} no longer exists or is inaccessible.'
>                    .format( path ) )
> ```

Same comment.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919)*

bstaletic:
Review status: 5 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[ycmd/completers/cpp/include_cache.py, line 123 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2J_RBCwkdRR9zPEl_:-LB5WpDfD5UQHFwBtVOF:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L123)):*

`_logger.exception` should be used so that we know where the error comes from in the logs.
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/include_cache.py, line 142 at r13](https://reviewable.io/reviews/valloric/ycmd/919#-LB5VNtiCTc3sDqRUNYb:-LB5WpUv4x0CPIamqBM7:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/4dc0fd2a9c3cb6acdee03302d0e0b2a3c3a2697f/ycmd/completers/cpp/include_cache.py#L142)):*

Same comment.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919)*

micbou:
Reviewed 1 of 1 files at r14.

---

*[ycmd/completers/cpp/include_cache.py, line 123 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2J_RBCwkdRR9zPEl_:-LB5kfXBEw271UNCRoiw:b-d14hr8) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L123)):*

Done.
</blockquote></details>

Sorry, I missed this but we shouldn't format the string when logging. Instead, we should do:
```python
_logger.exception( 'The directory %s no longer exists or is inaccessible.', path )
```
Also, I think the message should give more context e.g. `Cannot list entries for include path %s`.

---

*[ycmd/completers/cpp/include_cache.py, line 142 at r13](https://reviewable.io/reviews/valloric/ycmd/919#-LB5VNtiCTc3sDqRUNYb:-LB5mzL9DhxKTnSqHaL0:b-zc752h) ([raw file](https://github.com/valloric/ycmd/blob/4dc0fd2a9c3cb6acdee03302d0e0b2a3c3a2697f/ycmd/completers/cpp/include_cache.py#L142)):*

Done.
</blockquote></details>

The message here could be `Cannot get modification time for include path %s`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919)*

bstaletic:
Review status: 5 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[ycmd/completers/cpp/include_cache.py, line 123 at r11](https://reviewable.io/reviews/valloric/ycmd/919#-LB2J_RBCwkdRR9zPEl_:-LB5qpi93nReRxCgLDHO:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/ef954e7090031bf28329693f1607162756201848/ycmd/completers/cpp/include_cache.py#L123)):*

Sorry, I missed this but we shouldn't format the string when logging. Instead, we should do:
```python
_logger.exception( 'The directory %s no longer exists or is inaccessible.', path )
```
Also, I think the message should give more context e.g. `Cannot list entries for include path %s`.

</blockquote></details>

Done.

---

*[ycmd/completers/cpp/include_cache.py, line 142 at r13](https://reviewable.io/reviews/valloric/ycmd/919#-LB5VNtiCTc3sDqRUNYb:-LB5qnO-7OwHVfM__TKF:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/4dc0fd2a9c3cb6acdee03302d0e0b2a3c3a2697f/ycmd/completers/cpp/include_cache.py#L142)):*

The message here could be `Cannot get modification time for include path %s`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919)*

micbou:
Nice work.

@zzbot r+

---

Reviewed 1 of 1 files at r15.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/919#-:-LB7WFM33gc7HNEiXUlD:b-732mz1)*

PR #920, opened 2018-02-09
[READY] Add formatting support to language server completer

micbou:
Add support for the [`formatting`](https://microsoft.github.io/language-server-protocol/specification#textDocument_formatting) and [`rangeFormatting`](https://microsoft.github.io/language-server-protocol/specification#textDocument_rangeFormatting) requests to the language server completer through the new command `Format`. Update the API by adding two optional fields to the `run_completer_command` endpoint:
 - `range`: the range to which the formatting is applied. If given, the `rangeFormatting` LSP endpoint is used; otherwise, the `formatting` one is called. This makes more sense than having a separate command for each LSP endpoint;
 - `options`: a set of editor-related options for the current buffer. Available options are `tab_size` and `insert_spaces` which correspond to the options in [the `FormattingOptions` structure](https://microsoft.github.io/language-server-protocol/specification#textDocument_formatting). This is generic enough so that we can add more options later.

The feature is added to the Java completer (since this is the only completer that supports it). Here's a demo:

java-format

Some changes in YCM are needed to support the formatting options and range formatting but formatting the whole buffer should work out of the box.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/920)

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -10,7 +10,7 @@
     <h1>ycmd</h1>
     <p class="sw-info">
         
-        Version: <span class="sw-info-version">0.1.0</span>,     <a href="https://github.com/Valloric/ycmd" class="sw-external-doc">ycmd GitHub page</a>
+        Version: <span class="sw-info-version">0.2.0</span>,     <a href="https://github.com/Valloric/ycmd" class="sw-external-doc">ycmd GitHub page</a>
     
     </p>
     <p><p>ycmd is a code-completion &amp; code-comprehension server.</p>
@@ -1747,7 +1747,7 @@
                             <div  class="panel panel-definition">
                                 <div class="panel-body">
                                             <section class="json-schema-example">
-                                                <pre><code class="lang-json">{<br>    &quot;<span class="hljs-attribute">column_num</span>&quot;: <span class="hljs-value"><span class="hljs-number">20</span></span>,<br>    &quot;<span class="hljs-attribute">command_arguments</span>&quot;: <span class="hljs-value">[<br>        <span class="hljs-string">&quot;RefactorRename&quot;</span>,<br>        <span class="hljs-string">&quot;Testing&quot;</span><br>    ]</span>,<br>    &quot;<span class="hljs-attribute">completer_target</span>&quot;: <span class="hljs-value"><span class="hljs-string">&quot;filetype_default&quot;</span></span>,<br>    &quot;<span class="hljs-attribute">file_data</span>&quot;: <span class="hljs-value">{<br>        &quot;<span class="hljs-attribute">/home/test/dev/test.js</span>&quot;: <span class="hljs-value">{<br>            &quot;<span class="hljs-attribute">contents</span>&quot;: <span class="hljs-value"><span class="hljs-string">&quot;&lt;file contents&gt;&quot;</span></span>,<br>            &quot;<span class="hljs-attribute">filetypes</span>&quot;: <span class="hljs-value">[<br>                <span class="hljs-string">&quot;javascript&quot;</span><br>            ]<br>        </span>}<br>    </span>}</span>,<br>    &quot;<span class="hljs-attribute">filepath</span>&quot;: <span class="hljs-value"><span class="hljs-string">&quot;/home/test/dev/test.js&quot;</span></span>,<br>    &quot;<span class="hljs-attribute">line_num</span>&quot;: <span class="hljs-value"><span class="hljs-number">10</span><br></span>}
+                                                <pre><code class="lang-json">{<br>    &quot;<span class="hljs-attribute">column_num</span>&quot;: <span class="hljs-value"><span class="hljs-number">20</span></span>,<br>    &quot;<span class="hljs-attribute">command_arguments</span>&quot;: <span class="hljs-value">[<br>        <span class="hljs-string">&quot;RefactorRename&quot;</span>,<br>        <span class="hljs-string">&quot;Testing&quot;</span><br>    ]</span>,<br>    &quot;<span class="hljs-attribute">completer_target</span>&quot;: <span class="hljs-value"><span class="hljs-string">&quot;filetype_default&quot;</span></span>,<br>    &quot;<span class="hljs-attribute">file_data</span>&quot;: <span class="hljs-value">{<br>        &quot;<span class="hljs-attribute">/home/test/dev/test.js</span>&quot;: <span class="hljs-value">{<br>            &quot;<span class="hljs-attribute">contents</span>&quot;: <span class="hljs-value"><span class="hljs-string">&quot;&lt;file contents&gt;&quot;</span></span>,<br>            &quot;<span class="hljs-attribute">filetypes</span>&quot;: <span class="hljs-value">[<br>                <span class="hljs-string">&quot;javascript&quot;</span><br>            ]<br>        </span>}<br>    </span>}</span>,<br>    &quot;<span class="hljs-attribute">filepath</span>&quot;: <span class="hljs-value"><span class="hljs-string">&quot;/home/test/dev/test.js&quot;</span></span>,<br>    &quot;<span class="hljs-attribute">line_num</span>&quot;: <span class="hljs-value"><span class="hljs-number">10</span></span>,<br>    &quot;<span class="hljs-attribute">options</span>&quot;: <span class="hljs-value">{<br>        &quot;<span class="hljs-attribute">insert_spaces</span>&quot;: <span class="hljs-value"><span class="hljs-literal">true</span></span>,<br>        &quot;<span class="hljs-attribute">tab_size</span>&quot;: <span class="hljs-value"><span class="hljs-number">4</span><br>    </span>}<br></span>}
                                         </code></pre>
                                         
                                             </section>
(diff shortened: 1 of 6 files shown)


micbou:
Demo when the `insert_spaces` option is `False` (i.e. `noexpandtab` in Vim):



---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/920#-:-L4sAvXGplh8yF9CxeT8:bbz1wil)*

bstaletic:
Reviewed 6 of 6 files at r1.

---

*[docs/openapi.yml, line 107 at r1](https://reviewable.io:443/reviews/valloric/ycmd/920#-L4szc-voR2fgH8SdeKs:-L4szc-voR2fgH8SdeKt:b-5psyip) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/docs/openapi.yml#L107)):*
> ```YAML
>     completion.
> 
>   version: 0.2.0
> ```

Unrelated to this PR, but should we start making git tags for each API version?
That way we can clearly state for how long would a deprecated feature be supported.

---

*[ycmd/completers/language_server/language_server_protocol.py, line 355 at r1](https://reviewable.io:443/reviews/valloric/ycmd/920#-L4t-wCuSCcRJmKRDL8U:-L4t-wCuSCcRJmKRDL8V:bs8q838) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/ycmd/completers/language_server/language_server_protocol.py#L355)):*
> ```Python
>   options = {
>     'tabSize': tab_size,
>     'insertSpaces': insert_spaces,
> ```

Specific servers will have specific extra configurations for `FormattingOptions`.
I think we should allow completers (java completer) to specify more and somehow allow the user to tell us (so we can tell the server) how the user wants his document formatted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/920)*

puremourning:
This looks really good. I like how simple it turned out to be in practice. Main question comes around the default values for the indent options. I feel it would be less likely to start a war if the defaults were actually in `default_settings.json` and could therefore be controlled by clients.

---

---

*[docs/openapi.yml, line 107 at r1](https://reviewable.io:443/reviews/valloric/ycmd/920#-L4szc-voR2fgH8SdeKs:-L5-0Q08OTMqF0JHpMN0:b-j8qpys) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/docs/openapi.yml#L107)):*

Unrelated to this PR, but should we start making git tags for each API version?
That way we can clearly state for how long would a deprecated feature be supported.
</blockquote></details>

Val has been against official versioning in the past if i remember correctly. I think it sort of makes sense but technically we're still in the sort of "we can change anything" state, and we'll probably remain that way indefinitely.

---

*[ycmd/completers/language_server/language_server_completer.py, line 1403 at r1](https://reviewable.io:443/reviews/valloric/ycmd/920#-L5-4AQHpKL4GLFhZa14:-L5-4AQHpKL4GLFhZa15:bk7sdzb) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/ycmd/completers/language_server/language_server_completer.py#L1403)):*
> ```Python
>                                                   request_data[ 'filepath' ],
>                                                   text_edit[ 'range' ] ) )
>                for text_edit in response[ 'result' ] ]
> ```

according to the spec the result can be `null`. In that scenario, do we actually get a `result` entry? Would this throw a `KeyError` ? 

---

*[ycmd/completers/language_server/language_server_protocol.py, line 325 at r1](https://reviewable.io:443/reviews/valloric/ycmd/920#-L5-5IjAXn_0L-zUtJmg:-L5-5IjAXn_0L-zUtJmh:b-a9zlua) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/ycmd/completers/language_server/language_server_protocol.py#L325)):*
> ```Python
>   return {
>     'line': line_num - 1,
>     'character': CodepointsToUTF16CodeUnits( line_value, column_codepoint ) - 1
> ```

oh good catch. how did i miss that :/

i think i only searched for `character` in `language_server_completer` :(

---

*[ycmd/completers/language_server/language_server_protocol.py, line 335 at r1](https://reviewable.io:443/reviews/valloric/ycmd/920#-L5-5i_Vcwur3efgj5Fj:-L5-5i_Vcwur3efgj5Fk:b-duw94n) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/ycmd/completers/language_server/language_server_protocol.py#L335)):*
> ```Python
>       'uri': FilePathToUri( request_data[ 'filepath' ] ),
>     },
>     'options': FormattingOptions( options.get( 'tab_size', 4 ),
> ```

what informed the use of these particular defaults ( 4 and with Tabs ) ?

I guess the issue is that the options are _mandatory_ in the request and the protocol doesn't specify a default. 

I think we should allow the client to specify the defaults, not the server. Or at least put _these_ defaults in `default_settings.json` so that they can be set in client configuration.

defaults for formatting evoke almost zealous responses, so i wouldn't want to be responsible for deciding for users.

---

*[ycmd/completers/language_server/language_server_protocol.py, line 352 at r1](https://reviewable.io:443/reviews/valloric/ycmd/920#-L5-6lUnBe1yF7C0rzgA:-L5-6lUnBe1yF7C0rzgB:b9l0hqz) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/ycmd/completers/language_server/language_server_protocol.py#L352)):*
> ```Python
> 
> 
> def FormattingOptions( tab_size, insert_spaces, **kwargs ):
> ```

The `kwargs` argument is never used in the 2 places this method is called.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/920#-:-L5-7mla8z_YwMjzUXmO:b-krww4s)*

puremourning:
Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/920)*

bstaletic:

---

*[docs/openapi.yml, line 107 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/920#-L4szc-voR2fgH8SdeKs:-L5-8cr27TL5K7CrUzsS:b2nhcgo) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/docs/openapi.yml#L107)):*

Val has been against official versioning in the past if i remember correctly. I think it sort of makes sense but technically we're still in the sort of "we can change anything" state, and we'll probably remain that way indefinitely.
</blockquote></details>

I'm aware of that, but I was thinking "we're versioning the API, that's not far from versioning the whole repo".

As for "we ccan change anything", that comes with bumping the major number.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/920)*

micbou:
Reviewed 5 of 6 files at r1, 6 of 6 files at r2, 1 of 2 files at r3.
Review status: 5 of 6 files reviewed at latest revision, 6 unresolved discussions.

---

*[ycmd/completers/language_server/language_server_completer.py, line 1403 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/920#-L5-4AQHpKL4GLFhZa14:-L53uy6O-xI-57vyXzvB:b-itfjoj) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/ycmd/completers/language_server/language_server_completer.py#L1403)):*

according to the spec the result can be `null`. In that scenario, do we actually get a `result` entry? Would this throw a `KeyError` ? 
</blockquote></details>

No, it will throw a `TypeError` exception. We handle this with the idiom `None or []`.

---

*[ycmd/completers/language_server/language_server_protocol.py, line 335 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/920#-L5-5i_Vcwur3efgj5Fj:-L53vawV5EfJWdmEBjD3:b-wd1ov4) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/ycmd/completers/language_server/language_server_protocol.py#L335)):*

what informed the use of these particular defaults ( 4 and with Tabs ) ?

I guess the issue is that the options are _mandatory_ in the request and the protocol doesn't specify a default. 

I think we should allow the client to specify the defaults, not the server. Or at least put _these_ defaults in `default_settings.json` so that they can be set in client configuration.

defaults for formatting evoke almost zealous responses, so i wouldn't want to be responsible for deciding for users.
</blockquote></details>

Yes, it's clearly not a good idea to specify defaults for these options. I don't really like putting them in `default_settings.json`because these options are not supposed to be global. I can't see a better solution than making these options mandatory in the `/run_completer_command` request. Updated the PR accordingly.

---

*[ycmd/completers/language_server/language_server_protocol.py, line 352 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/920#-L5-6lUnBe1yF7C0rzgA:-L53y2v_3wjX23i5miib:bss4a2b) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/ycmd/completers/language_server/language_server_protocol.py#L352)):*

The `kwargs` argument is never used in the 2 places this method is called.
</blockquote></details>

Removed.

---

*[ycmd/completers/language_server/language_server_protocol.py, line 355 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/920#-L4t-wCuSCcRJmKRDL8U:-L53y4zi2eVWyjkylzUS:b-kb3hhj) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/ycmd/completers/language_server/language_server_protocol.py#L355)):*

Specific servers will have specific extra configurations for `FormattingOptions`.
I think we should allow completers (java completer) to specify more and somehow allow the user to tell us (so we can tell the server) how the user wants his document formatted.
</blockquote></details>

I think it's much simpler to only allow the client (the editor) to customize these options then let the user configures its editor to use a specific formatting (like it's done in PR https://github.com/Valloric/YouCompleteMe/pull/2900 by getting the value of `shiftwidth` and `expandtab`).

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/920)*

puremourning:
:lgtm:

With one leftover debug statement. Thanks!

---

Reviewed 3 of 6 files at r2, 2 of 2 files at r3.

---

*[ycmd/completers/language_server/language_server_completer.py, line 1396 at r4](https://reviewable.io:443/reviews/valloric/ycmd/920#-L54HGklE5Uq_abxCZkl:-L54HGklE5Uq_abxCZkm:b-kbznfx) ([raw file](https://github.com/valloric/ycmd/blob/a19db3b6748ad11ffece783570339cbca941c5af/ycmd/completers/language_server/language_server_completer.py#L1396)):*
> ```Python
>       message = lsp.Formatting( request_id, request_data )
> 
>     print( message )
> ```

leftover debug

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/920#-:-L54HQ0kdT075BE0TI9i:b-bwup2y)*

micbou:
Reviewed 1 of 2 files at r3, 1 of 1 files at r5.

---

*[ycmd/completers/language_server/language_server_completer.py, line 1396 at r4](https://reviewable.io:443/reviews/valloric/ycmd/920#-L54HGklE5Uq_abxCZkl:-L54HpKYhebZ0AV6Xidl:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/a19db3b6748ad11ffece783570339cbca941c5af/ycmd/completers/language_server/language_server_completer.py#L1396)):*

leftover debug
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/920)*

bstaletic:
:lgtm: 
I did leave a comment, but once again it's not a blocker.

---

Reviewed 3 of 6 files at r2, 2 of 2 files at r3, 1 of 1 files at r5.

---

*[ycmd/completers/language_server/language_server_protocol.py, line 355 at r1](https://reviewable.io:443/reviews/valloric/ycmd/920#-L4t-wCuSCcRJmKRDL8U:-L54HYIeo6A75PiT3Q7I:bj7k1a1) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/ycmd/completers/language_server/language_server_protocol.py#L355)):*

I think it's much simpler to only allow the client (the editor) to customize these options then let the user configures its editor to use a specific formatting (like it's done in PR https://github.com/Valloric/YouCompleteMe/pull/2900 by getting the value of `shiftwidth` and `expandtab`).
</blockquote></details>

I like the simplicity of using editor configuration, but it's kind of limiting.
When we say formating, we mean thte braces style too for example. Editors don't have configurations as detailed as formatting usually goes.
That said, Vim does have `cindent` which does a lot of the work. So, even though I'd like this to be more configurable, I won't make it a blocker.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/920#-:-L54Io5j1qGzYKNSgUY4:b-lrcs4f)*

puremourning:

---

*[ycmd/completers/language_server/language_server_protocol.py, line 355 at r1](https://reviewable.io:443/reviews/valloric/ycmd/920#-L4t-wCuSCcRJmKRDL8U:-L54JIDlU7Ldf8qxTzhg:bjj2bk2) ([raw file](https://github.com/valloric/ycmd/blob/848bbe9ce53eed975eee1d36d6b41ba493a8f13b/ycmd/completers/language_server/language_server_protocol.py#L355)):*

I like the simplicity of using editor configuration, but it's kind of limiting.
When we say formating, we mean thte braces style too for example. Editors don't have configurations as detailed as formatting usually goes.
That said, Vim does have `cindent` which does a lot of the work. So, even though I'd like this to be more configurable, I won't make it a blocker.
</blockquote></details>

right now we're exposing all of the supported options as mandatory. we can always revisit if we find servers that document additional options. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/920)*

PR #921, opened 2018-02-10
[READY] Use C++11 enum classes instead of plain enums

bstaletic:
Enumeration classes are supposed to be safer (read: namespaced), so possible name clashes between different `enum`s are avoided. They also prevent the following:

```c++
#include <iostream>
enum class ec1 { A }; // ok
enum class ec2 { A }; // ok
enum e { A }; // ok
int main(void) {
  // if ( ec1::A == ec2::A ) std::cout<<"Error - different enums.\n";
  // if ( A == ec2::A ) std::cout<<"Error - different enums.\n";
  // if ( ec1::A == 0 ) std::cout<<"Error - different types.\n;
  if ( A == 0 ) std::cout<<"Okay, but considered unsafe.\n";
  return 0;
}
```

Should I bump the `CORE_VERSION`?

NOTE:
I also urge everyone to check what our benchmarks say, because travis reported considerable performance hit compared to current master.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/921)

### DIFF
--- a/cpp/ycm/ClangCompleter/ClangHelpers.cpp
+++ b/cpp/ycm/ClangCompleter/ClangHelpers.cpp
@@ -36,15 +36,15 @@ DiagnosticKind DiagnosticSeverityToType( CXDiagnosticSeverity severity ) {
   switch ( severity ) {
     case CXDiagnostic_Ignored:
     case CXDiagnostic_Note:
-      return INFORMATION;
+      return DiagnosticKind::INFORMATION;
 
     case CXDiagnostic_Warning:
-      return WARNING;
+      return DiagnosticKind::WARNING;
 
     case CXDiagnostic_Error:
     case CXDiagnostic_Fatal:
     default:
-      return ERROR;
+      return DiagnosticKind::ERROR;
   }
 }
 
@@ -265,7 +265,7 @@ Diagnostic BuildDiagnostic( DiagnosticWrap diagnostic_wrap,
 
   // If this is an "ignored" diagnostic, there's no point in continuing since we
   // won't display those to the user
-  if ( diagnostic.kind_ == INFORMATION ) {
+  if ( diagnostic.kind_ == DiagnosticKind::INFORMATION ) {
     return diagnostic;
   }
 
--- a/cpp/ycm/ClangCompleter/CompletionData.cpp
+++ b/cpp/ycm/ClangCompleter/CompletionData.cpp
@@ -27,27 +27,27 @@ namespace {
 CompletionKind CursorKindToCompletionKind( CXCursorKind kind ) {
   switch ( kind ) {
     case CXCursor_StructDecl:
-      return STRUCT;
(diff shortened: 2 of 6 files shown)


puremourning:
:lgtm:

Performance regression makes no sense. Using compiler explorer you can prove to yourself that enum class and enum generate the same assembly. 

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/921#-:-L4zDps6aDRF1or0UyUD:b-npsxpj)*

bstaletic:
Should we also define enum sizes i.e `enum class : std::uint8_t enum_class_foo { FOO };`?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/921#-:-L4zI-OR7JP8pwkyWJNi:b-y0gi8h)*

puremourning:
Do we store a lot of them? Do we gain anything in alignment/padding ? Otherwise I honestly doubt it is worth the extra complexity. The default is almost always right. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/921#-:-L4zOJ976w1o5aMJDjRD:b-4rxemg)*

bstaletic:
One `DiagnosticKind` per diagnostic, one `CompletionKind` per completion candidate.
Since this is C++11, they are `int`s. Whatever `int` may be on the specific system. As for alignment, I'm not sure how to tell.

And what do you mean by "extra complexity"? It's just an extra word in declaration.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/921#-:-L4zOgTG2IhtmKBvQ37o:b67zuxf)*

puremourning:
To check alignment/padding you have to look at the struct declaration. Things are aligned at word boundaries typically,  though packing is compiler and arch specific. For example shrinking 4 bytes to 1 byte might just add 3 bytes of padding, depending on the exact order of declarations. 

Put it this way, I’d like to see compiler explorer comparisons to determine what the difference is.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/921#-:-L4zPqnFrCE5_lsx_fXD:b3qx5au)*

micbou:
To be thorough, I compared the benchmarks with and without these changes and didn't observe a difference.

I don't see the point in specifying the enum size. The potential gain in memory usage (due to alignment) is going to be negligible compared to the whole memory usage which is mostly the storage of candidates.

> Should I bump the CORE_VERSION?

No, this kind of changes doesn't warrant a new version.

:lgtm:

@zzbot r+

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/921#-:-L4zbU5ptOZ8G7RgnkHt:b-oddis5)*

PR #922, opened 2018-02-10
[READY] Support for unity-style builds by specifying the translation unit

puremourning:
This allows for more elaborate project structures where the translation unit is not the current file name. We already have some concession to this concept when searching a compilation database for _flags_, but that does not allow for us to tell clang to use a specific file as the TU to parse.

This change allows the extra conf file to specify the translation unit to parse when editing a particular file. This essentially exposes the full flexibility available when executing the compiler from the command line. 

An example scenario:

* unit.cpp

```c++
#include <iostream>

struct Test { int a; int b; };

#include "file.cpp"

int main() {
   Test t;
   do_thing( t );
}
```

* file.cpp

```c++
void do_thing( Test& t ) {
   std::cout << "Test: " << t.a << '\n';
}
```

When editing `file.cpp` i actually want the translation unit that clang sees to be `unit.cpp` otherwise the file is not parseable. Clang understands this by splitting the concept of translation unit file name, and position filename. 

This change replicates that separation in ycmd, allowing such scenarios.

Closes https://github.com/Valloric/YouCompleteMe/issues/2796
Closes https://github.com/Valloric/YouCompleteMe/issues/2795

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/922)


bstaletic:
I thought the changes for this would be more intrusive. I'm positively surprised.

Also, not much wrong for a WIP.

---

Reviewed 11 of 11 files at r1.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 465 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zsyE1lvzU8VbU994n:-L4zsyE2ZH-zPsl8zZjm:bunxxfm) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L465)):*
> ```cpp
>     for ( const Diagnostic& diagnostic : latest_diagnostics_ ) {
>       // Find all diagnostics for the supplied line which have FixIts attached
>       if ( diagnostic.location_.filename_ != filename )
> ```

Correct me if I'm wrong.
This is here in case we have diagnostics in the same translation unit, but in a different file. Correct?
If so, why wasn't it needed before?

---

*[ycmd/completers/cpp/clang_completer.py, line 382 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zyWD9OCrV-8l1Emy4:-L4zyWD9OCrV-8l1Emy5:b-uza8cy) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/ycmd/completers/cpp/clang_completer.py#L382)):*
> ```Python
>     # use the correct one ? But it could be shared ...
>     self._completer.DeleteCachesForFile(
>         ToCppStringCompatible( request_data[ 'filepath' ] ) )
> ```

How could it be wrong?

---

*[ycmd/completers/cpp/flags.py, line 113 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zuTNZi7GnyRXLYohP:-L4zuTNZi7GnyRXLYohQ:b-83a3cr) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/ycmd/completers/cpp/flags.py#L113)):*
> ```Python
> 
> 
>   def FlagsForFile( self,                         # noqa - cyclomatic compexity
> ```

Two ways to solve the complexity issue.

- The `if` block that you added, turn it into a ternary expression. This trics flake8.
- You could also move everythin from `flags = _ExtractFlagsList( results )` to (and includeing) `sanitize_flags = PrepareFlagsForClang( flags,` to a separate function.

And since this now returns `( sanitized_flags, filename )` it could use a docstring.

---

*[ycmd/completers/cpp/flags.py, line 135 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zwPOP3adYbzgONBHR:-L4zwPOQ0DKuRdUGsKt6:b-rewo0w) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/ycmd/completers/cpp/flags.py#L135)):*
> ```Python
>         self.no_extra_conf_file_warning_posted = True
>         raise NoExtraConfDetected
>       return ( [], filename )
> ```

I don't think you need parens for this return statement. Same for all the others.

---

*[ycmd/tests/clang/flags_test.py, line 58 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zwqIJCk_WPiuKLxH4:-L4zwqIJCk_WPiuKLxH5:b-ph439r) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/ycmd/tests/clang/flags_test.py#L58)):*
> ```Python
> 
>   with MockExtraConfModule( FlagsForFile ):
>     ( flags_list, filename ) = flags_object.FlagsForFile( '/foo', False )
> ```

Again, I don't think you need the parens.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/922#-:-L4zyjIBl5h0Fe98BPoR:b1yoafw)*

puremourning:

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 465 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zsyE1lvzU8VbU994n:-L54CX9C8UVUtIoNc5YB:bhy9we3) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L465)):*

Correct me if I'm wrong.
This is here in case we have diagnostics in the same translation unit, but in a different file. Correct?
If so, why wasn't it needed before?
</blockquote></details>

i'm not 100% sure it _wasn't_ a problem before, but certainly now that we have separated file location and translation unit, there are even fewer guarantees

---

*[ycmd/completers/cpp/clang_completer.py, line 382 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zyWD9OCrV-8l1Emy4:-L54CoUM6Q2B-i6mtDpv:bate9un) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/ycmd/completers/cpp/clang_completer.py#L382)):*

How could it be wrong?
</blockquote></details>

i think the comment means that perhaps we should delete caches for the TU ? 

TBH it needs more thought. We should only clear up a TU if the _last_ file which uses it is closed. This will require tracking the relationship between open files and TUs more closely. Right now, i suspect this BufferClose event just doesn't work properly when using alternate TU, unless you happen to close the _actual_ TU filename last.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/922)*

bstaletic:

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 465 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zsyE1lvzU8VbU994n:-L54FQ4MnolM2Ua3ICST:bjlykj8) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L465)):*

i'm not 100% sure it _wasn't_ a problem before, but certainly now that we have separated file location and translation unit, there are even fewer guarantees
</blockquote></details>

Alright, it's fine then.

---

*[ycmd/completers/cpp/clang_completer.py, line 382 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zyWD9OCrV-8l1Emy4:-L54E_Yv7BbWc50zaPyH:b-8z4q5s) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/ycmd/completers/cpp/clang_completer.py#L382)):*

i think the comment means that perhaps we should delete caches for the TU ? 

TBH it needs more thought. We should only clear up a TU if the _last_ file which uses it is closed. This will require tracking the relationship between open files and TUs more closely. Right now, i suspect this BufferClose event just doesn't work properly when using alternate TU, unless you happen to close the _actual_ TU filename last.
</blockquote></details>

In case of separated filename and TU, when unloading, there are two possiblities:

- Current file actually matches TU.
  - Other files may depend on the now unloaded TU. Worst case, we reload the TU.
- Current file doesn't match TU. No problem that I can see.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/922)*

Valloric:
I'm not against this change, though I am curious about use-cases where this is useful. The example in the PR description seems a bit contrived to me. (I'll add that this is mostly for the sake of my curiosity.)

At a quick glance, the changes seem fine to me. 

---

Review status: 7 of 11 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/922#-:-L5a-1eXP9AY4lC-QzIq:b-3gakx3)*

puremourning:
"Unity-style" builds speed up the compilation process by reducing IO (rather than parsing 100x translation units, each including `iostream`, you only read `iostream` once).

They also lead to better code because the compiler has full visibility of the code, so can perform better inlining decisions. This allows a sort of quasi-LTO where LTO isn't an option.

There are drawbacks around incremental builds being arguably slower (you have to reparse the whole mega-unit when changing one file). However, when I applied this to a large (>100 .cc file) lib at work, the compilation time for the lib went from 4 minutes to 5 seconds, so it didn't matter if you rebuilt the whole mega-tu.

This style isn't _that_ uncommon. For example, one place I use this patch is the Vim source which does roughly what's in the example in a few places, typically around conditional compilation. Another example is the [handmade hero](https://handmadehero.org) source. It seems more common in C where there's less convention around file naming (e.g. name after class), but essentially where the `#include` mechanism exists, people will use it).

I'm just about to push some tests, then it should be `READY`

---

Review status: 7 of 11 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cpp/flags.py, line 113 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zuTNZi7GnyRXLYohP:-L5ZKf01TubqWDVvehJr:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/ycmd/completers/cpp/flags.py#L113)):*

Two ways to solve the complexity issue.

- The `if` block that you added, turn it into a ternary expression. This trics flake8.
- You could also move everythin from `flags = _ExtractFlagsList( results )` to (and includeing) `sanitize_flags = PrepareFlagsForClang( flags,` to a separate function.

And since this now returns `( sanitized_flags, filename )` it could use a docstring.
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/flags.py, line 135 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zwPOP3adYbzgONBHR:-L5ZLwSB2bAP5OfdaOWx:b-wu9xfp) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/ycmd/completers/cpp/flags.py#L135)):*

I don't think you need parens for this return statement. Same for all the others.
</blockquote></details>

I have removed them, though as always i maintain that including them makes it obvious that i wish to return a tuple, rather than have some sort of bug.

it's not uncommon (in my experience with python) to have a line like

```
return x,
```

Which ends up causing some surprising error like "too many values to unpack". the dangling comma makes it a tuple. shrug. i prefer that if you want a tuple you use `()`. But i realise i am in the minority here :D

---

*[ycmd/tests/clang/flags_test.py, line 58 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zwqIJCk_WPiuKLxH4:-L5ZNXsiejCRDW2xh-Z0:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/ycmd/tests/clang/flags_test.py#L58)):*

Again, I don't think you need the parens.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/922#-:-L5a1Pb9seM8Xz5Qcbv_:bj1lrqs)*

puremourning:
Review status: 7 of 11 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/cpp/clang_completer.py, line 382 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zyWD9OCrV-8l1Emy4:-L5a3357HCh9Be9i87dC:bj49f04) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/ycmd/completers/cpp/clang_completer.py#L382)):*

In case of separated filename and TU, when unloading, there are two possiblities:

- Current file actually matches TU.
  - Other files may depend on the now unloaded TU. Worst case, we reload the TU.
- Current file doesn't match TU. No problem that I can see.
</blockquote></details>

I might leave this as a FIXME

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/922)*

puremourning:
Added set of tests covering all the places we use `_FlagsForRequest`. The test data is totally convoluted, but demonstrates the behaviour well enough. PR Is READY

bstaletic:
Reviewed 4 of 4 files at r2, 10 of 10 files at r3.

---

*[ycmd/completers/cpp/flags.py, line 135 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zwPOP3adYbzgONBHR:-L5a6OU5ISmCfhuXiplW:b-147om4) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/ycmd/completers/cpp/flags.py#L135)):*

I have removed them, though as always i maintain that including them makes it obvious that i wish to return a tuple, rather than have some sort of bug.

it's not uncommon (in my experience with python) to have a line like

```
return x,
```

Which ends up causing some surprising error like "too many values to unpack". the dangling comma makes it a tuple. shrug. i prefer that if you want a tuple you use `()`. But i realise i am in the minority here :D
</blockquote></details>

I was thinking you'd say that. So I reconsidered posting this many times.

(I just want my name on every file and that's why I do these "fixes"/style PRs.)

---

*[ycmd/tests/clang/diagnostics_test.py, line 402 at r3](https://reviewable.io:443/reviews/valloric/ycmd/922#-L5dAM1dPZANv77Xlaq9:-L5dAM1dPZANv77XlaqA:bbcypv0) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/tests/clang/diagnostics_test.py#L402)):*
> ```Python
>             'line_num': 4,
>             'column_num': 3,
>             'filepath': PathToTestFile( 'unity.h' )
> ```

Windows (or appveyor) decided to have fun and return "C:\\path\\to\\testdata**/**unity.h".

---

*[ycmd/tests/clang/subcommands_test.py, line 897 at r3](https://reviewable.io:443/reviews/valloric/ycmd/922#-L5dBt0yF-CfTRprKyk5:-L5dBt0yF-CfTRprKyk6:buu8xb8) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/tests/clang/subcommands_test.py#L897)):*
> ```Python
>   results = app.post_json( '/run_completer_command', event_data ).json
> 
>   pprint( results )
> ```

For some reason this printed `{'fixits': []}` on appveyor.

---

*[ycmd/tests/clang/testdata/unity.cc, line 3 at r3](https://reviewable.io:443/reviews/valloric/ycmd/922#-L5dCUkOuc1acdSZne_n:-L5dCUkOuc1acdSZne_o:b1iuwae) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/tests/clang/testdata/unity.cc#L3)):*
> ```cpp
> struct Unity
> {
>   int this_is_an_it;
> ```

Was this supposed to be an i**n**t?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/922)*

puremourning:

---

*[ycmd/completers/cpp/flags.py, line 135 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zwPOP3adYbzgONBHR:-L5dNxKAcYjWcBGKfYAM:b3hxste) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/ycmd/completers/cpp/flags.py#L135)):*

I was thinking you'd say that. So I reconsidered posting this many times.

(I just want my name on every file and that's why I do these "fixes"/style PRs.)
</blockquote></details>

hahahahaha

---

*[ycmd/tests/clang/diagnostics_test.py, line 402 at r3](https://reviewable.io:443/reviews/valloric/ycmd/922#-L5dAM1dPZANv77Xlaq9:-L5dOe0gapOTMdQM1qcx:bcb2feo) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/tests/clang/diagnostics_test.py#L402)):*

Windows (or appveyor) decided to have fun and return "C:\\path\\to\\testdata**/**unity.h".
</blockquote></details>

Perhaps this is why we never bothered checking it before. Sigh. Have i mentioned recently how much i love Windows :)

---

*[ycmd/tests/clang/testdata/unity.cc, line 3 at r3](https://reviewable.io:443/reviews/valloric/ycmd/922#-L5dCUkOuc1acdSZne_n:-L5dO4U-9T0MnUGA1olQ:bp30fie) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/tests/clang/testdata/unity.cc#L3)):*

Was this supposed to be an i**n**t?
</blockquote></details>

i noticed the typo, but decided to keep it. i was actually going to rename the variable to `ceci_nest_pas_un_it` if i was going to change it at all.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/922)*

bstaletic:

---

*[ycmd/tests/clang/diagnostics_test.py, line 402 at r3](https://beta.reviewable.io/reviews/valloric/ycmd/922#-L5dAM1dPZANv77Xlaq9:-L5dQDniBoKIqakdEbfQ:b-eipd0m) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/tests/clang/diagnostics_test.py#L402)):*

Perhaps this is why we never bothered checking it before. Sigh. Have i mentioned recently how much i love Windows :)
</blockquote></details>

I'm guessing you could just remove the offending line.

---

*[ycmd/tests/clang/testdata/unity.cc, line 3 at r3](https://beta.reviewable.io/reviews/valloric/ycmd/922#-L5dCUkOuc1acdSZne_n:-L5dQOE23lT9uJ5lceMB:b389zmb) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/tests/clang/testdata/unity.cc#L3)):*

i noticed the typo, but decided to keep it. i was actually going to rename the variable to `ceci_nest_pas_un_it` if i was going to change it at all.
</blockquote></details>

In lights of recent vim easter eggs discussion. I like that. But I don't think this typo is at all important and can be left as is.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/922)*

puremourning:

---

*[ycmd/tests/clang/diagnostics_test.py, line 402 at r3](https://beta.reviewable.io/reviews/valloric/ycmd/922#-L5dAM1dPZANv77Xlaq9:-L5dQqlr5ySTSXzX_bg-:b-zgfhpz) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/tests/clang/diagnostics_test.py#L402)):*

I'm guessing you could just remove the offending line.
</blockquote></details>

The line in question is kind of the fundamental thing that's being tested: that the diags are produced for the files included by `unity.cc` not `unitya.cc`

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/922)*

bstaletic:

---

*[ycmd/tests/clang/diagnostics_test.py, line 402 at r3](https://beta.reviewable.io/reviews/valloric/ycmd/922#-L5dAM1dPZANv77Xlaq9:-L5dRzv6CiAy_JhnLRfO:b-xj7g7p) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/tests/clang/diagnostics_test.py#L402)):*

The line in question is kind of the fundamental thing that's being tested: that the diags are produced for the files included by `unity.cc` not `unitya.cc`
</blockquote></details>

Would it work to use `PathToTestFile( '' ) + /unity.h`? Terribly hackish, but... \*shrug\*

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/922)*

bstaletic:

---

*[ycmd/tests/clang/diagnostics_test.py, line 402 at r3](https://reviewable.io:443/reviews/valloric/ycmd/922#-L5dAM1dPZANv77Xlaq9:-L5hR554DwVkYk_4T0lg:b6y94vg) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/tests/clang/diagnostics_test.py#L402)):*

Would it work to use `PathToTestFile( '' ) + /unity.h`? Terribly hackish, but... \*shrug\*
</blockquote></details>

`'filepath': os.path.normpath( PathToTestFile( 'unity.h' )` should solve this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/922)*

puremourning:

---

*[ycmd/tests/clang/diagnostics_test.py, line 402 at r3](https://reviewable.io:443/reviews/valloric/ycmd/922#-L5dAM1dPZANv77Xlaq9:-L5hRni8SI1arVNc36f-:bdum3jq) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/tests/clang/diagnostics_test.py#L402)):*

`'filepath': os.path.normpath( PathToTestFile( 'unity.h' )` should solve this.
</blockquote></details>

There’s another issue as well. I have a plan but it requires some work and probablybwibdows updates. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/922)*

puremourning:
Marking NOT READY due to the problems on Windows that need sorting out.

puremourning:
Currently only NOT READY because i need to write a test for a diagnostic+fixit with same line/column but different filename. otherwise should work, particularly with the java upgrade PR merged as that makes the java tests less flaky.

---

Review status: 13 of 20 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/922#-:-L6l-K2_9waAc4DSLZs_:b-vw1c8s)*

puremourning:
Review status: 13 of 20 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/tests/clang/subcommands_test.py, line 897 at r3](https://beta.reviewable.io/reviews/valloric/ycmd/922#-L5dBt0yF-CfTRprKyk5:-L6l1Vuw5AsZK4O9-wwb:b-y5r51v) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/tests/clang/subcommands_test.py#L897)):*

For some reason this printed `{'fixits': []}` on appveyor.
</blockquote></details>

There was a reason for this which was just a bug. I fixed it.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/922)*

puremourning:
I added the test (it failed without the filename check). so this is ready.

puremourning:
Review status: 12 of 20 files reviewed at latest revision, all discussions resolved.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 465 at r1](https://reviewable.io:443/reviews/valloric/ycmd/922#-L4zsyE1lvzU8VbU994n:-L6l72gpXuGxC0aEmOjl:bicduu2) ([raw file](https://github.com/valloric/ycmd/blob/b416c086b6812f5794dc180e3af9c740f4b84c94/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L465)):*

Alright, it's fine then.
</blockquote></details>

It always was a problem. I have a test now that failed without this line (and likely would have failed in reality even without a unity setup), and passes now.

I had to use `boost::filesystem` to compare the paths because windows.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/922)*

bstaletic:
Provided the tests are passing, this is :lgtm:.

---

Reviewed 7 of 7 files at r4, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/922#-:-L6l7BRq5xx-Vj9ffNYW:b-o8ruh1)*

micbou:
Reviewed 5 of 11 files at r1, 3 of 4 files at r2, 10 of 10 files at r3, 6 of 7 files at r4, 1 of 1 files at r5.

---

*[README.md, line 232 at r3](https://beta.reviewable.io/reviews/valloric/ycmd/922#-L5cLiJI-wd68yNwI_R3:-L5cLiJJfpoRl9z0vIUc:b-toyy0n) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/README.md#L232)):*
> ```Markdown
>     directory.
> 
>   - `override_filename`: (optional) a string indicating the name of the fie to
> ```

`file`

---

*[cpp/ycm/ClangCompleter/TranslationUnit.h, line 62 at r3](https://beta.reviewable.io/reviews/valloric/ycmd/922#-L5cNVkoK6BhEIPhV4Nb:-L5cNVkoK6BhEIPhV4Nc:b-8dhiz8) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/cpp/ycm/ClangCompleter/TranslationUnit.h#L62)):*
> ```objc
> 
>   YCM_EXPORT std::vector< CompletionData > CandidatesForLocation(
>     const std::string& filename,
> ```

We seem to put `&` next to the variable in other places so it should be `const std::string &filename` if we want to be consistent. This applies to all the changes in that file and some of them in `TranslationUnit.cpp`.

---

*[ycmd/completers/cpp/clang_completer.py, line 380 at r3](https://beta.reviewable.io/reviews/valloric/ycmd/922#-L5cNrEaWRkB1kXl0YRZ:-L5cNrEbvtyeoBgJGQny:b-qldq6s) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/completers/cpp/clang_completer.py#L380)):*
> ```Python
>   def OnBufferUnload( self, request_data ):
>     # TODO/FIXME: The filepath here is (possibly) wrong. Call FlagsForFile and
>     # use the correct one ? But it could be shared ...
> ```

No spaces before `?` and `...`.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/922)*

puremourning:

---

*[README.md, line 232 at r3](https://reviewable.io:443/reviews/valloric/ycmd/922#-L5cLiJI-wd68yNwI_R3:-L74u5FzTiL1PXEdBN4n:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/README.md#L232)):*

`file`
</blockquote></details>

Done.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.h, line 62 at r3](https://reviewable.io:443/reviews/valloric/ycmd/922#-L5cNVkoK6BhEIPhV4Nb:-L74u6IzGAjccfJSnbxE:b-ybg9ow) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/cpp/ycm/ClangCompleter/TranslationUnit.h#L62)):*

We seem to put `&` next to the variable in other places so it should be `const std::string &filename` if we want to be consistent. This applies to all the changes in that file and some of them in `TranslationUnit.cpp`.
</blockquote></details>

You're right. This is one of those habit things. Once I switched about 3 years ago to attaching the `*` or `&` to the type, it made so much sense I find it hard not to anymore.

i think i got them all.

---

*[ycmd/completers/cpp/clang_completer.py, line 380 at r3](https://reviewable.io:443/reviews/valloric/ycmd/922#-L5cNrEaWRkB1kXl0YRZ:-L74vABIiRZgTnRK9CTs:b48yqrn) ([raw file](https://github.com/valloric/ycmd/blob/e193ed0b44d8216926db362d2eb2658c763061e0/ycmd/completers/cpp/clang_completer.py#L380)):*

No spaces before `?` and `...`.
</blockquote></details>

Yup. Also the comment is useless. I have rewritten it to state the actual problem and what might be done (in future patch) to solve it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/922)*

micbou:
:lgtm:

@zzbot r+

---

Reviewed 1 of 11 files at r1, 4 of 4 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/922#-:-L7512kD7s92xc_MjKL-:b-a2cmdb)*

puremourning:
@zzbot retry

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/922#-:-L75Z0HuDcxlmBi0KiTn:b-kc0oa4)*

puremourning:
@zzbot r=micbou

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/922#-:-L75Z6Nu16IAVvdvRSQQ:byfyio0)*

bstaletic:
@puremourning zzbot is doing fine. This PR is just queued behind another one that's going through the testing phase at the moment.

puremourning:
i was hoping it might fire the appveyour build, which flaked.

PR #931, opened 2018-02-18
[READY] Use file contents from server state if available when polling asynchronous diagnostics

micbou:
While I was playing with the Java completer, I noticed that matches from asynchronous diagnostics are sometimes incorrectly displayed:

file-contents-change-diagnostics-issue

This is due to the file contents from the `/receive_messages` request not being up to date anymore when the language server are publishing new diagnostics. In the example above, when inserting the second `public String test;`, the diagnostic offsets are converted from UTF-16 to UTF-8 using the contents:
```java
package com.youcompleteme;
  
public class Test {
  public String test;
}
```
instead of
```java
package com.youcompleteme;
  
public class Test {
  public String test;
  public String test;
}
```

The solution is to store new file contents in the `ServerFileState` object and use that contents (if available) when polling asynchronous diagnostics:

file-contents-change-diagnostics-fix

Added a test that fails without this change.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/931)

### DIFF
--- a/ycmd/completers/completer_utils.py
+++ b/ycmd/completers/completer_utils.py
@@ -24,11 +24,14 @@ from builtins import *  # noqa
 
 # Must not import ycm_core here! Vim imports completer, which imports this file.
 # We don't want ycm_core inside Vim.
+import logging
 import os
 import re
 from collections import defaultdict
 from future.utils import iteritems
 from ycmd.utils import ToCppStringCompatible, ToUnicode, ReadFile
+
+_logger = logging.getLogger( __name__ )
 
 
 class PreparedTriggers( object ):
@@ -222,4 +225,8 @@ def GetFileContents( request_data, filename ):
   if filename in file_data:
     return ToUnicode( file_data[ filename ][ 'contents' ] )
 
-  return ToUnicode( ReadFile( filename ) )
+  try:
+    return ToUnicode( ReadFile( filename ) )
+  except IOError:
+    _logger.exception( 'Error reading file {}'.format( filename ) )
+    return ''
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -896,10 +896,12 @@ class LanguageServerCompleter( Completer ):
     # However, we _also_ return them here to refresh diagnostics after, say
     # changing the active file in the editor, or for clients not supporting the
     # polling mechanism.
-    uri = lsp.FilePathToUri( request_data[ 'filepath' ] )
+    filepath = request_data[ 'filepath' ]
+    uri = lsp.FilePathToUri( filepath )
+    contents = utils.SplitLines( GetFileContents( request_data, filepath ) )
     with self._server_info_mutex:
       if uri in self._latest_diagnostics:
-        return [ _BuildDiagnostic( request_data, uri, diag )
+        return [ _BuildDiagnostic( contents, uri, diag )
                  for diag in self._latest_diagnostics[ uri ] ]
 
 
@@ -1010,7 +1012,30 @@ class LanguageServerCompleter( Completer ):
     if notification[ 'method' ] == 'window/showMessage':
       return responses.BuildDisplayMessageResponse(
         notification[ 'params' ][ 'message' ] )
-    elif notification[ 'method' ] == 'window/logMessage':
+
+    if notification[ 'method' ] == 'textDocument/publishDiagnostics':
+      params = notification[ 'params' ]
+      uri = params[ 'uri' ]
+
+      try:
+        filepath = lsp.UriToFilePath( uri )
+      except lsp.InvalidUriException:
+        _logger.exception( 'Ignoring diagnostics for unrecognized URI' )
+        return None
+
+      with self._server_info_mutex:
+        if filepath in self._server_file_state:
+          contents = self._server_file_state[ filepath ].contents
+        else:
+          contents = GetFileContents( request_data, filepath )
+      contents = utils.SplitLines( contents )
+      return {
+        'diagnostics': [ _BuildDiagnostic( contents, uri, x )
+                         for x in params[ 'diagnostics' ] ],
+        'filepath': filepath
+      }
+
+    if notification[ 'method' ] == 'window/logMessage':
       log_level = [
         None, # 1-based enum from LSP
         logging.ERROR,
@@ -1022,20 +1047,6 @@ class LanguageServerCompleter( Completer ):
       params = notification[ 'params' ]
       _logger.log( log_level[ int( params[ 'type' ] ) ],
                    SERVER_LOG_PREFIX + params[ 'message' ] )
-    elif notification[ 'method' ] == 'textDocument/publishDiagnostics':
-      params = notification[ 'params' ]
-      uri = params[ 'uri' ]
-      try:
-        filepath = lsp.UriToFilePath( uri )
-        response = {
-          'diagnostics': [ _BuildDiagnostic( request_data, uri, x )
-                           for x in params[ 'diagnostics' ] ],
-          'filepath': filepath
-        }
-        return response
-      except lsp.InvalidUriException:
-        _logger.exception( 'Ignoring diagnostics for unrecognized URI' )
-        pass
 
     return None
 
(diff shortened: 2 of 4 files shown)


puremourning:
Thanks for fixing this!

a couple of ideas below.

---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/language_server/language_server_completer.py, line 1026 at r1](https://reviewable.io:443/reviews/valloric/ycmd/931#-L5dxh5hPBxhedEOsXjT:-L5dxh5hPBxhedEOsXjU:b-hz610j) ([raw file](https://github.com/valloric/ycmd/blob/441e0658358d7dfc15d2d436308fd54bc1de8abb/ycmd/completers/language_server/language_server_completer.py#L1026)):*
> ```Python
>         return None
> 
>       if filepath in self._server_file_state:
> ```

`_server_file_state` is a private member of the base class, so we should try not to use it directly.

Perhaps we should have a `self.GetServerFileContents( filename )` provided by `LanguageServerCompleter` instead.

On that thought, I suppose we actually have this issue potentially any time that we get a response from the server. The LSP itself is supposedly versioned, so you're supposed to only deal with messages that match the file version they sent.

Perhaps what we should _really_ do is actually implement that, e.g. add `LanguageServerCompleter.GetFileContentsForResponse` which checks the server file state, and compares the version number, raising an exception if it doesn't match. If that is right (needs some thought) we can do it under a different PR though.

---

*[ycmd/completers/language_server/language_server_protocol.py, line 109 at r1](https://reviewable.io:443/reviews/valloric/ycmd/931#-L5dyb1ZtJkCSQGJOp_S:-L5dyb1ZtJkCSQGJOp_T:bu17skd) ([raw file](https://github.com/valloric/ycmd/blob/441e0658358d7dfc15d2d436308fd54bc1de8abb/ycmd/completers/language_server/language_server_protocol.py#L109)):*
> ```Python
>     self.state = ServerFileState.CLOSED
>     self.checksum = None
>     self.contents = ''
> ```

When I wrote this class I considered if we needed this, always thinking that we did, but it would be nice (for resources and performance) if we didn't need it. Turns out we do, which is a touch sad, but it's better to have working software than any other type of software :)

---

*[ycmd/tests/java/diagnostics_test.py, line 367 at r1](https://reviewable.io:443/reviews/valloric/ycmd/931#-L5dz4qKVxN1K-Uo-oWH:-L5dz4qKVxN1K-Uo-oWI:bbilij7) ([raw file](https://github.com/valloric/ycmd/blob/441e0658358d7dfc15d2d436308fd54bc1de8abb/ycmd/tests/java/diagnostics_test.py#L367)):*
> ```Python
> 
>   def PollForMessagesInAnotherThread( filepath, contents ):
>     for message in PollForMessages( app,
> ```

You might want to consider setting the poll timeout to something enormous (like 5 minutes) to ensure that this really tests the scenario exactly. If this call times out and re-sends (imaging a long delay between staring the thread and doing the next step), it will clobber the "new" contents with the 'old' contents. A really long timeout would make it more robust. There's a value you can patch like `@patch.object( completer, 'MESSAGE_POLL_TIMEOUT', 60*5 )` or something.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/931#-:-L5e-87qtv95_7anmqgV:btwnfnm)*

puremourning:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/language_server/language_server_completer.py, line 1026 at r1](https://reviewable.io:443/reviews/valloric/ycmd/931#-L5dxh5hPBxhedEOsXjT:-L5e00P4-JUxqosdMKLh:b-gp91tb) ([raw file](https://github.com/valloric/ycmd/blob/441e0658358d7dfc15d2d436308fd54bc1de8abb/ycmd/completers/language_server/language_server_completer.py#L1026)):*

`_server_file_state` is a private member of the base class, so we should try not to use it directly.

Perhaps we should have a `self.GetServerFileContents( filename )` provided by `LanguageServerCompleter` instead.

On that thought, I suppose we actually have this issue potentially any time that we get a response from the server. The LSP itself is supposedly versioned, so you're supposed to only deal with messages that match the file version they sent.

Perhaps what we should _really_ do is actually implement that, e.g. add `LanguageServerCompleter.GetFileContentsForResponse` which checks the server file state, and compares the version number, raising an exception if it doesn't match. If that is right (needs some thought) we can do it under a different PR though.

</blockquote></details>

what a load of crap. i thought this was java_completer.py. Of course you can use `_server_file_state` here. Sorry.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/931)*

puremourning:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/language_server/language_server_completer.py, line 1026 at r1](https://reviewable.io:443/reviews/valloric/ycmd/931#-L5dxh5hPBxhedEOsXjT:-L5e0OPwXzBa5OsbcYcB:b-hg3n6w) ([raw file](https://github.com/valloric/ycmd/blob/441e0658358d7dfc15d2d436308fd54bc1de8abb/ycmd/completers/language_server/language_server_completer.py#L1026)):*
> _server_file_state

on the other hand, `_server_file_state` requires protection using `with self._server_info_mutex`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/931)*

micbou:
Reviewed 3 of 4 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---

*[ycmd/completers/language_server/language_server_completer.py, line 1026 at r1](https://reviewable.io:443/reviews/valloric/ycmd/931#-L5dxh5hPBxhedEOsXjT:-L5e19gj9YR-pLFmxauS:bwcx7uf) ([raw file](https://github.com/valloric/ycmd/blob/441e0658358d7dfc15d2d436308fd54bc1de8abb/ycmd/completers/language_server/language_server_completer.py#L1026)):*

> _server_file_state

on the other hand, `_server_file_state` requires protection using `with self._server_info_mutex`
</blockquote></details>

Right. I forgot the lock.

---

*[ycmd/completers/language_server/language_server_protocol.py, line 109 at r1](https://reviewable.io:443/reviews/valloric/ycmd/931#-L5dyb1ZtJkCSQGJOp_S:-L5e2GTWvAtMZIdPaj_d:bqc1xr) ([raw file](https://github.com/valloric/ycmd/blob/441e0658358d7dfc15d2d436308fd54bc1de8abb/ycmd/completers/language_server/language_server_protocol.py#L109)):*

When I wrote this class I considered if we needed this, always thinking that we did, but it would be nice (for resources and performance) if we didn't need it. Turns out we do, which is a touch sad, but it's better to have working software than any other type of software :)
</blockquote></details>

This is mostly going to increase memory usage. The impact on performance should be really low.

---

*[ycmd/tests/java/diagnostics_test.py, line 367 at r1](https://reviewable.io:443/reviews/valloric/ycmd/931#-L5dz4qKVxN1K-Uo-oWH:-L5e24GMRGwxlWZHqKr2:blcf5b5) ([raw file](https://github.com/valloric/ycmd/blob/441e0658358d7dfc15d2d436308fd54bc1de8abb/ycmd/tests/java/diagnostics_test.py#L367)):*
> If this call times out and re-sends (imaging a long delay between staring the thread and doing the next step), it will clobber the "new" contents with the 'old' contents. 

We are only polling with the old contents so I am not sure how that's possible.

---

*[ycmd/tests/java/diagnostics_test.py, line 391 at r1](https://reviewable.io:443/reviews/valloric/ycmd/931#-L5e5gX-DyH5nutfr2V_:-L5e5gX-DyH5nutfr2Va:b-w27vnm) ([raw file](https://github.com/valloric/ycmd/blob/441e0658358d7dfc15d2d436308fd54bc1de8abb/ycmd/tests/java/diagnostics_test.py#L391)):*
> ```Python
>   app.post_json( '/event_notification', event_data ).json
> 
>   expiration = time.time() + 5
> ```

Had to increase the timeout. I hope it's enough. We don't want the timeout to be too long in case the test is really failing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/931)*

puremourning:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/931#-:-L5e9LSC0j4BHfRqA7bW:bnfp4nl)*

puremourning:
Reviewed 2 of 4 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/931)*

micbou:
Commits squashed. Should be ready now.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/931#-:-L5edEp4EcRWMpBC1mpD:byll5pa)*

bstaletic:
:lgtm:
r=puremourning

---

Reviewed 2 of 4 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/931#-:-L5gGzqXRtTvdeQapNgS:b2a1uz3)*

micbou:
@zzbot r=puremourning

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/931#-:-L5hVpFhGO64IERN26j_:b-7f0b5i)*

PR #933, opened 2018-02-19
[READY] Do not check if server is running immediately after starting it

micbou:
Given that `Popen` is returning immediately, the process will always be running at the `_ServerIsRunning` call even if it's going to crash. A simple way to confirm this is to comment the `-jar` argument in the command used to start jdt.ls: `_ServerIsRunning` still returns `True` in that scenario.

### DIFF
--- a/ycmd/completers/java/java_completer.py
+++ b/ycmd/completers/java/java_completer.py
@@ -361,12 +361,6 @@ class JavaCompleter( language_server_completer.LanguageServerCompleter ):
                                                stdout = PIPE,
                                                stderr = stderr )
 
-      if not self._ServerIsRunning():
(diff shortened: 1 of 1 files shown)


puremourning:
:lgtm: makes sense

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/933#-:-L5kKC0VL8PfWueKjMNc:b88jd50)*

bstaletic:
:lgtm: Definitely does.
@zzbot r=puremourning

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/933#-:-L5kKOoMUVEBkSqlx-is:bmkmj52)*

PR #935, opened 2018-02-20
simplify LetterNode, improve cache behaviour

dgel:
EDIT: I just saw the Unicode PR, which makes this one superfluous. Sorry for taking your time 

I made some changes to LetterNode almost 2 years ago that reduced memory usage and improved performance. There were a few niggles I had about it however: every node had a vector, but only 1 actually used it and we were using indexes in the main algorithm and storing indexes in the data structure at the same time. This PR is intended to fix that. 

The main change is that most of the data is no longer stored: the bool storing case of letters was unused, the indices were already implicit in the indexes used in Candidate and so could be removed as well. In the end, all that is stored is one array of offsets per letter, all in one vector which should improve cache locality.

I've used a similar setup for benchmarking as described [here](https://gist.github.com/dgel/0ee4709015288ead82c14e2c079710bd) for the previous PR. some results (using 1000 samples):

shorthands:

cs: case-sensitive
cis: case-insensitive
lc: lowercase
mc: mixed case

## Speed

 variant | construction | match cis, lc | match cs, lc | match cis, mc | match cs, mc 
---------|--------------|---------------|--------------|---------------|--------------
 original | 11.1 us | 416 ns | 358 ns | 409 ns | 431 ns 
 modified | 7.9 us | 201 ns | 337 ns | 220 ns | 324 ns

## memory usage

variant | 100000 nodes of 4 chars | 10000 nodes of 78 chars
-------|-------------------------|------------------------
original | 220MB | 422MB
modified | 200MB | 383MB

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/935)

### DIFF
--- a/cpp/ycm/Candidate.cpp
+++ b/cpp/ycm/Candidate.cpp
@@ -65,18 +65,17 @@ Candidate::Candidate( const std::string &text )
     word_boundary_chars_( GetWordBoundaryChars( text ) ),
     text_is_lowercase_( IsLowercase( text ) ),
     letters_present_( LetterBitsetFromString( text ) ),
-    root_node_( new LetterNode( text ) ) {
+    root_node_( text ) {
 }
 
 
 Result Candidate::QueryMatchResult( const std::string &query,
                                     bool case_sensitive ) const {
-  LetterNode *node = root_node_.get();
+  size_t node_index = 0;
   int index_sum = 0;
 
   for ( char letter : query ) {
-    const NearestLetterNodeIndices *nearest =
-      node->NearestLetterNodesForLetter( letter );
+    const NearestLetterNodeIndices *nearest = root_node_.NearestLetterNodesForLetter( node_index, letter );
 
     if ( !nearest ) {
       return Result();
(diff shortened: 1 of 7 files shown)


puremourning:
Wow! Thanks for making a PR!


> EDIT: I just saw the Unicode PR, which makes this one superfluous. 

I haven't looked at the detail of this, but I wonder if there are aspects of this PR we can learn from/use in the Unicode PR, or could do as a follow up? I see you've added some comments on that PR, which are super valuable, so thanks for that too!

dgel:
You're welcome, I've been happily using YouCompleteMe for several years now and wanted to contribute a small thing back =) I don't think any of this is directly applicable to the Unicode PR, since it just does a direct search over the candidates. I think that is the right way to go, so not much room to improve there.

Valloric:
Thanks for your amazing PR! :)

Like @puremourning, the unicode PR is making this a bit complicated... had this PR come in before that one, it would have likely been merged.

>  I don't think any of this is directly applicable to the Unicode PR, since it just does a direct search over the candidates. I think that is the right way to go, so not much room to improve there.

It's unfortunate that we can't get some knowledge from this PR over to that one... Should we close this PR then if there's not much that can be merged given the unicode PR?

dgel:
Hi, yeah feel free to close this one, thanks for looking it over =)

PR #936, opened 2018-02-21
[READY] Add clang-cl support

bstaletic:
According to https://github.com/Valloric/ycmd/pull/789#issuecomment-367151834, #789 didn't include support for `clang-cl` without `--driver-mode=cl` specified.

I did not add `clang-cl` to the C++ regex. Since there is no `clang-cl++` I think we should not touch language flags and let clang decide if it's C or C++.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/936)


bstaletic:
Appveyor flaked, but codecov says +2 lines total, 100% covered.

puremourning:
:lgtm: Nice one.

Would be nice to confirm that this solves the issue for the related report, though it seems that there are still issues with the preamble generation on windows.

---

Reviewed 12 of 12 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/936#-:-L5z3e1JjulsNzrXyIo0:b-uqw949)*

bstaletic:
Well, I have included completion tests which all pass. I'm not sure what else can we do.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/936#-:-L5z4eoOZkCfD6NERlhW:b-mvv5tx)*

micbou:
Thanks for working on this.

---

---

*[ycmd/completers/cpp/flags.py, line 248 at r1](https://reviewable.io:443/reviews/valloric/ycmd/936#-L5zPawj2rMv00A43msS:-L5zPawj2rMv00A43msT:b-p02vez) ([raw file](https://github.com/valloric/ycmd/blob/11ddf75d0763f84a5f348c4161b5d221e7d02b1c/ycmd/completers/cpp/flags.py#L248)):*
> ```Python
>   if OnWindows():
>     # clang-cl implies --driver-mode=cl
>     enable_windows_style_flags = 'clang-cl' in flags[ 0 ]
> ```

This should be done after the loop since we may ignore it if `--driver-mode` is used. We also need to support the `.exe` extension, the `cl` compiler, and the case must be ignored. I would suggest the following regex:
```python
CL_COMPILER_REGEX = re.compile( r'(?:cl|clang-cl)(.exe)?$', re.IGNORECASE )
```

---

*[ycmd/completers/cpp/flags.py, line 250 at r1](https://reviewable.io:443/reviews/valloric/ycmd/936#-L5zPOBJ6tC1KGPm71QK:-L5zPOBJ6tC1KGPm71QL:b-6ps888) ([raw file](https://github.com/valloric/ycmd/blob/11ddf75d0763f84a5f348c4161b5d221e7d02b1c/ycmd/completers/cpp/flags.py#L250)):*
> ```Python
>     enable_windows_style_flags = 'clang-cl' in flags[ 0 ]
>     # Iterate in reverse because we only care
>     # about the last occurance of --driver-mode flag.
> ```

occurrence.

---

*[ycmd/completers/cpp/flags.py, line 256 at r1](https://reviewable.io:443/reviews/valloric/ycmd/936#-L5z67v5fEjRAHoNFC1q:-L5z67v6RTbiHRb06ZBz:b4j85ub) ([raw file](https://github.com/valloric/ycmd/blob/11ddf75d0763f84a5f348c4161b5d221e7d02b1c/ycmd/completers/cpp/flags.py#L256)):*
> ```Python
>       if flag.startswith( '--driver-mode' ):
>         enable_windows_style_flags = flag == '--driver-mode=cl'
>         break
> ```

We can directly `return flag == '--driver-mode=cl'` here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/936#-:-L5zRtiZMxNpc1QurAmk:b-4v1jp5)*

bstaletic:
Review status: 8 of 9 files reviewed at latest revision, 3 unresolved discussions.

---

*[ycmd/completers/cpp/flags.py, line 248 at r1](https://reviewable.io:443/reviews/valloric/ycmd/936#-L5zPawj2rMv00A43msS:-L5zY0R8BuKlx6aHiCcB:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/11ddf75d0763f84a5f348c4161b5d221e7d02b1c/ycmd/completers/cpp/flags.py#L248)):*

This should be done after the loop since we may ignore it if `--driver-mode` is used. We also need to support the `.exe` extension, the `cl` compiler, and the case must be ignored. I would suggest the following regex:
```python
CL_COMPILER_REGEX = re.compile( r'(?:cl|clang-cl)(.exe)?$', re.IGNORECASE )
```
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/flags.py, line 250 at r1](https://reviewable.io:443/reviews/valloric/ycmd/936#-L5zPOBJ6tC1KGPm71QK:-L5zY1G3yZU7TdMUifHO:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/11ddf75d0763f84a5f348c4161b5d221e7d02b1c/ycmd/completers/cpp/flags.py#L250)):*

occurrence.
</blockquote></details>

Done.

---

*[ycmd/completers/cpp/flags.py, line 256 at r1](https://reviewable.io:443/reviews/valloric/ycmd/936#-L5z67v5fEjRAHoNFC1q:-L5zY47Ac9nW6mlIRX7C:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/11ddf75d0763f84a5f348c4161b5d221e7d02b1c/ycmd/completers/cpp/flags.py#L256)):*

We can directly `return flag == '--driver-mode=cl'` here.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/936)*

micbou:
:lgtm:

@zzbot r+

---

Reviewed 11 of 12 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/936#-:-L5zgXd-evhECVHAq5dm:b-a2cmdb)*

PR #938, opened 2018-02-24
[READY] Skip stray Windows paths in flags

### LINKED ISSUE #937: Clang error parsing headers?
Hello, All. I have an issue I've been able to distill to a simple test case. I've been able to test on two platforms, Mac and Windows. Mac doesn't have an issue, but Windows does.

Of course, the setup is everything, so here goes.

First, things you'll need:

- Windows environment (of course)
- YouCompleteMe (all set up, etc.)

Next, here's the code:

[cmake_test.zip](https://github.com/Valloric/ycmd/files/1753390/cmake_test.zip)

Extract it to C:\cmake_test. Then open Vim. Using Vim, open C:\cmake_test\main.cpp. Then open C:\cmake_test\head.h. You'll notice I made a silly coding error: I forget the "c" on the definition of the class Classy:

`head.h, line 6: lass Classy {`

That's wrong. On Mac, if you were to generate the compile_commands.json via CMake, ycmd detects my error:

`unknown type name 'lass'; did you mean 'class'? (Fixit)`

On Windows, I get a red error:

`ClangParseError: An AST deserialization error occurred while parsing the translation unit.`

If you like, I can give instructions on how I generate the compile_commands.json. That should be unnecessary, though. I'm not sure if this is a ycmd issue or a Clang issue.

Of course, if I fix my error and compile, the program works (and of course appears to do nothing) on both platforms.

Any ideas? Is this related to #892?

micbou:
When using a compilation database, the flags generally end with `-c /path/to/file` where `/path/to/file` is the compiled file. Libclang chokes on this if we don't remove the path. We have some logic to handle that case where, in addition to several conditions, we detect that the flag is a path if it contains the `/` separator. However, this doesn't take care of the case where the flag is a Windows path with `\` as separators.

Fixes #937.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/938)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -429,8 +429,7 @@ def _RemoveUnusedFlags( flags, filename, enable_windows_style_flags ):
     # "foo.cpp" when we are compiling "foo.h" because the comp db doesn't have
     # flags for headers. The returned flags include "foo.cpp" and we need to
     # remove that.
-    if _SkipStrayFilenameFlag( flag,
-                               current_flag,
+    if _SkipStrayFilenameFlag( current_flag,
                                previous_flag,
                                enable_windows_style_flags ):
       continue
@@ -440,8 +439,7 @@ def _RemoveUnusedFlags( flags, filename, enable_windows_style_flags ):
   return new_flags
 
 
-def _SkipStrayFilenameFlag( flag,
-                            current_flag,
+def _SkipStrayFilenameFlag( current_flag,
                             previous_flag,
                             enable_windows_style_flags ):
   current_flag_starts_with_slash = current_flag.startswith( '/' )
@@ -452,7 +450,11 @@ def _SkipStrayFilenameFlag( flag,
 
   previous_flag_is_include = ( previous_flag in INCLUDE_FLAGS or
                                ( enable_windows_style_flags and
-                                 flag in INCLUDE_FLAGS_WIN_STYLE ) )
+                                 previous_flag in INCLUDE_FLAGS_WIN_STYLE ) )
+
+  current_flag_may_be_path = ( '/' in current_flag or
+                               ( enable_windows_style_flags and
+                                 '\\' in current_flag ) )
 
   return ( not ( current_flag_starts_with_dash or
                  ( enable_windows_style_flags and
@@ -460,7 +462,7 @@ def _SkipStrayFilenameFlag( flag,
            ( not ( previous_flag_starts_with_dash or
                    ( enable_windows_style_flags and
                      previous_flag_starts_with_slash ) ) or
-             ( not previous_flag_is_include and '/' in flag ) ) )
+             ( not previous_flag_is_include and current_flag_may_be_path ) ) )
 
 
 # Return the path to the macOS toolchain root directory to use for system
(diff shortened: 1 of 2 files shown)


bstaletic:
Something is off here. Appveyor says that this PR breaks completion in CL driver mode.

---

Reviewed 2 of 2 files at r1.

---

*[ycmd/tests/clang/flags_test.py, line 458 at r1](https://reviewable.io:443/reviews/valloric/ycmd/938#-L65dvMUM5BosvlLh1z0:-L65dvMUM5BosvlLh1z1:bm0kfkz) ([raw file](https://github.com/valloric/ycmd/blob/f165bfac0b2ff711fbb7b83e993aa74b4aeb440f/ycmd/tests/clang/flags_test.py#L458)):*
> ```Python
> 
>   # cl path with Windows separators
>   expected = [ 'path\\to\\cl.EXE', '-foo', '-xc++', '-bar', 'include_dir' ]
> ```

`cl.exe` is Microsoft's native/default compiler. Right? If so, don't we need to add `--driver-mode=cl` to the flags?

I know this isn't strictly related to this PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/938#-:-L65dSBLGuRuOgqBu0_u:b-efci86)*

micbou:
`previous_flag_is_include` was done on the current flag instead of the previous when using Windows flags. Added a test that fails without the fix.

---

Reviewed 2 of 2 files at r2.

---

*[ycmd/tests/clang/flags_test.py, line 458 at r1](https://reviewable.io:443/reviews/valloric/ycmd/938#-L65dvMUM5BosvlLh1z0:-L66AJZdb_FSL1F1nKst:b-f8c3lc) ([raw file](https://github.com/valloric/ycmd/blob/f165bfac0b2ff711fbb7b83e993aa74b4aeb440f/ycmd/tests/clang/flags_test.py#L458)):*

`cl.exe` is Microsoft's native/default compiler. Right? If so, don't we need to add `--driver-mode=cl` to the flags?

I know this isn't strictly related to this PR.
</blockquote></details>

libclang is smart enough to deduce `--driver-mode=cl` in that case. I tested it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/938#-:-L66Aa54jzKXmfEyyMM8:b-9s84jb)*

bstaletic:
I wonder how did we not notice that before.
Nice work. I hope CL mode won't cause us any more issues.
Provided the tests are passing, this is :lgtm:.

---

Reviewed 2 of 2 files at r2.

---

*[ycmd/tests/clang/flags_test.py, line 458 at r1](https://reviewable.io:443/reviews/valloric/ycmd/938#-L65dvMUM5BosvlLh1z0:-L66BcITifptvtXzhmQA:b-ahle5g) ([raw file](https://github.com/valloric/ycmd/blob/f165bfac0b2ff711fbb7b83e993aa74b4aeb440f/ycmd/tests/clang/flags_test.py#L458)):*

libclang is smart enough to deduce `--driver-mode=cl` in that case. I tested it.
</blockquote></details>

Good to know.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/938#-:-L66Bz09WeJvAJq-FsvQ:b-6x6bod)*

puremourning:
:lgtm: @zzbot r=bstaletic

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/938#-:-L67LzNtotqiOaJpCy-9:b-yi2rvi)*

PR #940, opened 2018-02-26
[READY] Lower minimum Python version requirements on Travis and CircleCI

micbou:
As discussed on Gitter some time ago, we should try to test for the oldest version of Python that we are supposed to support on Linux and macOS (e.g. RHEL 7 has Python 2.7.5 installed by default so if we test for Python 2.7.14, we are not sure it's going to work on that platform).

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/940)


puremourning:
Codecov result looks spurious.

@zzbot r=bstaletic

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/940#-:-L6IpoMx2AUZZfjoliT3:bgm6cf0)*

PR #951, opened 2018-03-10
[READY] Expand environment variables in extra conf options

micbou:
This PR adds support for environment variables in the `extra_conf_globlist` and `global_ycm_extra_conf` options.

Fixes https://github.com/Valloric/YouCompleteMe/issues/2944.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/951)

### DIFF
--- a/ycmd/extra_conf_store.py
+++ b/ycmd/extra_conf_store.py
@@ -33,7 +33,8 @@ import logging
 from threading import Lock
 from ycmd import user_options_store
 from ycmd.responses import UnknownExtraConf, YCM_EXTRA_CONF_FILENAME
-from ycmd.utils import LoadPythonSource, PathsToAllParentFolders
+from ycmd.utils import ( ExpandVariablesInPath, LoadPythonSource,
+                         PathsToAllParentFolders )
 from fnmatch import fnmatch
 
 
@@ -191,15 +192,15 @@ def Load( module_file, force = False ):
 
 
 def _MatchesGlobPattern( filename, glob ):
-  """Returns true if a filename matches a given pattern. A '~' in glob will be
-  expanded to the home directory and checking will be performed using absolute
-  paths with symlinks resolved (except on Windows). See the documentation of
-  fnmatch for the supported patterns."""
+  """Returns true if a filename matches a given pattern. Environment variables
+  and a '~' in glob will be expanded and checking will be performed using
+  absolute paths with symlinks resolved (except on Windows). See the
+  documentation of fnmatch for the supported patterns."""
 
   # NOTE: os.path.realpath does not resolve symlinks on Windows.
   # See https://bugs.python.org/issue9949
   realpath = os.path.realpath( filename )
-  return fnmatch( realpath, os.path.realpath( os.path.expanduser( glob ) ) )
+  return fnmatch( realpath, os.path.realpath( ExpandVariablesInPath( glob ) ) )
 
 
 def _ExtraConfModuleSourceFilesForFile( filename ):
@@ -233,5 +234,5 @@ def _RandomName():
 
 
 def _GlobalYcmExtraConfFileLocation():
-  return os.path.expanduser(
+  return ExpandVariablesInPath(
     user_options_store.Value( 'global_ycm_extra_conf' ) )
--- a/ycmd/completers/general/filename_completer.py
+++ b/ycmd/completers/general/filename_completer.py
@@ -29,7 +29,8 @@ import re
 from collections import defaultdict
 
 from ycmd.completers.completer import Completer
-from ycmd.utils import GetCurrentDirectory, OnWindows, ToUnicode
+from ycmd.utils import ( ExpandVariablesInPath, GetCurrentDirectory, OnWindows,
+                         ToUnicode )
 from ycmd import responses
 
 EXTRA_INFO_MAP = { 1 : '[File]', 2 : '[Dir]', 3 : '[File&Dir]' }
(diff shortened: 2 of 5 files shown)


ycmd/extra_conf_store.py
```diff
@@ -233,5 +234,5 @@ def _RandomName():
 
 
 def _GlobalYcmExtraConfFileLocation():
-  return os.path.expanduser(
-    user_options_store.Value( 'global_ycm_extra_conf' ) )
+  return os.path.expandvars( os.path.expanduser(
```
puremourning: Maybe we should use vim's `expand()` function instead of Python's ?

I wonder if that would be more obvious to users? Just a thought.

micbou:

---

*[ycmd/extra_conf_store.py, line 237 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/951#-L7FtONLyDaFLpqHjcHp-r1-237:-L7FuJKO5JKJonhR9H2e:buq93vk) ([raw file](https://github.com/valloric/ycmd/blob/1189260f3219354e0fc652a8e68f7472bc2f4e02/ycmd/extra_conf_store.py#L237)):*

Maybe we should use vim's `expand()` function instead of Python's ?

I wonder if that would be more obvious to users? Just a thought.
</blockquote></details>

We could but this is still useful for other clients.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/951)*

bstaletic:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---

*[ycmd/extra_conf_store.py, line 237 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/951#-L7FtONLyDaFLpqHjcHp-r1-237:-L7FveE37EHve2KLosYk:bff5wk7) ([raw file](https://github.com/valloric/ycmd/blob/1189260f3219354e0fc652a8e68f7472bc2f4e02/ycmd/extra_conf_store.py#L237)):*

We could but this is still useful for other clients.
</blockquote></details>

I think we should stick with the python's way because ycmd should be client agnostic.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/951#-:-L7FvqCh9z2JfSx_B-gK:bnfp4nl)*

puremourning:
:lgtm: with one optional comment. trigger the bot if you don't think its worth it 

---

---

*[ycmd/extra_conf_store.py, line 203 at r1](https://reviewable.io:443/reviews/valloric/ycmd/951#-L7FwR190to31z0qGYDd:-L7FwR190to31z0qGYDe:b-dqcx8x) ([raw file](https://github.com/valloric/ycmd/blob/1189260f3219354e0fc652a8e68f7472bc2f4e02/ycmd/extra_conf_store.py#L203)):*
> ```Python
>   realpath = os.path.realpath( filename )
>   return fnmatch( realpath, os.path.realpath(
>     os.path.expanduser( os.path.expandvars( glob ) ) ) )
> ```

Should we maybe add a `utils.ExpandShellVars( ... )` rather than remembering to do this in a few places ?

---

*[ycmd/extra_conf_store.py, line 237 at r1](https://reviewable.io:443/reviews/valloric/ycmd/951#-L7FtONLyDaFLpqHjcHp-r1-237:-L7FwAPn3fHruuxUV1Nd:b-llmayh) ([raw file](https://github.com/valloric/ycmd/blob/1189260f3219354e0fc652a8e68f7472bc2f4e02/ycmd/extra_conf_store.py#L237)):*

I think we should stick with the python's way because ycmd should be client agnostic.
</blockquote></details>

yeh, duh. Sorry. I read the "fixes" and assumed this code was YouCompleteMe not ycmd.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/951#-:-L7FwirhlCTvOK-Qpn_X:b-yaym8h)*

puremourning:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/951)*

micbou:
Reviewed 2 of 2 files at r1, 4 of 4 files at r2, 1 of 1 files at r3.

---

*[ycmd/extra_conf_store.py, line 203 at r1](https://reviewable.io:443/reviews/valloric/ycmd/951#-L7FwR190to31z0qGYDd:-L7FyVCLE9gGld8d03ok:b-h70d0k) ([raw file](https://github.com/valloric/ycmd/blob/1189260f3219354e0fc652a8e68f7472bc2f4e02/ycmd/extra_conf_store.py#L203)):*

Should we maybe add a `utils.ExpandShellVars( ... )` rather than remembering to do this in a few places ?
</blockquote></details>

Good idea. I named it `ExpandVariablesInPath` though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/951)*

puremourning:
:lgtm:

---

Reviewed 3 of 4 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/951#-:-L7G0kh1Cdvphv5K7sxp:bnfp4nl)*

bstaletic:
:lgtm: 
@zzbot r=puremourning

---

Reviewed 3 of 4 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/951#-:-L7G3HGnQLLv-b9D0I-O:bgshp04)*

micbou:
@zzbot r+

---

Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/951#-:-L7HBZWY5cf-ODtXMK0X:bi59kwd)*

PR #954, opened 2018-03-12
[READY] Reset prefix when setting the start column

micbou:
Like the query, we need to reset the prefix when setting the start column in the request.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/954)

### DIFF
--- a/ycmd/request_wrap.py
+++ b/ycmd/request_wrap.py
@@ -154,9 +154,10 @@ class RequestWrap( object ):
       self[ 'line_value' ],
       column_num )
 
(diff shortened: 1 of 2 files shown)


puremourning:
:lgtm: good catch

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/954#-:-L7N-huBhoCuNu_4dsR7:bbvzhhj)*

bstaletic:
:lgtm:
@zzbot r=puremourning

PR #955, opened 2018-03-12
[READY] Disable Ninja auto-detection

micbou:
The build system [Ninja](https://ninja-build.org/) is supposed to be faster for incremental builds but this project is too small to notice a difference (a few seconds at best). Also, this is the second time a user has an issue with it because the `ninja` executable on the PATH is not the build system but another software. See https://github.com/Valloric/YouCompleteMe/issues/2941#issuecomment-372259295 (the first time was reported on Gitter). For these reasons, I am proposing to completely remove support for this build system.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/955)

### DIFF
--- a/.circleci/install_dependencies.sh
+++ b/.circleci/install_dependencies.sh
@@ -12,19 +12,18 @@ set -e
 brew update || brew update
 
 # List of homebrew formulae to install in the order they appear.
-# We require CMake, Node, Go, Mono and Ninja for our build and tests, and all
+# We require CMake, Node, Go, and Mono for our build and tests, and all
 # the others are dependencies of pyenv.
 REQUIREMENTS="cmake
               node.js
               go
               mono
-              ninja
               readline
               autoconf
               pkg-config
               openssl"
 
-# Install CMake, Node, Go, Mono, Ninja, pyenv and dependencies.
+# Install CMake, Node, Go, Mono, pyenv and dependencies.
 for pkg in $REQUIREMENTS; do
   # Install package, or upgrade it if it is already installed.
   brew install $pkg || brew outdated $pkg || brew upgrade $pkg
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,6 @@ addons:
      - libc++-dev
      # Required to build the OmniSharp server.
      - mono-devel
-     - ninja-build
      # Everything below is a Python build dep (though it depends on Python
      # version). We need them because pyenv builds Python.
      - libssl-dev
--- a/build.py
+++ b/build.py
@@ -296,12 +296,12 @@ def CustomPythonCmakeArgs( args ):
 
 
 def GetGenerator( args ):
+  if args.ninja:
+    return 'Ninja'
   if OnWindows():
     return 'Visual Studio {version}{arch}'.format(
         version = args.msvc,
         arch = ' Win64' if platform.architecture()[ 0 ] == '64bit' else '' )
-  if PathToFirstExistingExecutable( ['ninja'] ):
-    return 'Ninja'
   return 'Unix Makefiles'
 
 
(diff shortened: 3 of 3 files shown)


bstaletic:
I'm up for this. :lgtm: 

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/955#-:-L7PrQTGza_TalLLh-3x:b-xf9sxu)*

puremourning:
Ninja automatically picks the right level of parallelism, which is why I always use it when the build system/generator supports it. It's also quieter.

anecdotally, I also think Ninja is faster for incremental builds (e.g. after using `--build-dir`). Perhaps we can keep it, but just ditch the auto-Ninja-detection (instead add `--with-generator=Ninja` or something ?).

I also think that 2 users having issues isn't such a large problem.

So I guess I would vote against this as it stands.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/955#-:-L7PplR3E3XFHLLOn722:blwsofb)*

micbou:
> Ninja automatically picks the right level of parallelism, which is why I always use it when the build system/generator supports it.

We already do that in the `build.py` script by specifying the `-j` option.

> It's also quieter.

True but we have the `--quiet` option which is even more quiet.

> anecdotally, I also think Ninja is faster for incremental builds (e.g. after using --build-dir).

Like I said, the project is too small to make a difference.

> Perhaps we can keep it, but just ditch the auto-Ninja-detection (instead add --with-generator=Ninja or something ?).

Won't this annoy you to have to pass the option each time you run the script?

> I also think that 2 users having issues isn't such a large problem.

Sure but we also have better things to do that explain to users that it's not working because they don't have the right `ninja` in their path. If there was a real non-cosmetic advantage over Make for this project, we could at least give a good reason why we look for Ninja but no, we can't even do that.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/955#-:-L7PzMcCs2GoBbj3Krsw:b20676w)*

puremourning:
OK fair enough :)

> We already do that in the build.py script by specifying the -j option.

I guess we do. I was sort of thinking about the incremental case, but that's actually a rare case.

> True but we have the --quiet option which is even more quiet.

I should search out whoever it was that added that!

> Like I said, the project is too small to make a difference.

I don't find that at work. I find that on a slow filesystem the difference is more pronounced, but tbh this again isn't a common use case.

> Won't this annoy you to have to pass the option each time you run the script?

Meh, I already have fairly sophisticated `build.py` invocations. I could probably remember to add one more, but sure - 'if there's an option to not have an option, take the no-option option' :)

> Sure but we also have better things to do

You had me at "better things to do" ! :D

Before we merge, it's worth checking with @valloric who (IIRC) added support for Ninja in the first place.

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/955#-:-L7QMO-a-GSdxY5fys6b:bvvedb2)*

Valloric:
I'd hate to see ninja support for this to go away; we should just remove ninja auto-detection. I'd totally pass the flag every time. :)

Note that we're likely to see issue requests about adding ninja auto-detection if we remove it. :D

micbou:
I removed Ninja auto-detection and added the `--ninja` option to build with Ninja.

---

Reviewed 3 of 3 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/955#-:-L8e67pUBPM8Qkn23v_Y:blefzxl)*

Valloric:
> I removed Ninja auto-detection and added the --ninja option to build with Ninja.

:dancer: Woohoo! :D

:lgtm: Thanks for the PR!

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/955#-:-L8erFtpAofOewtG3heJ:b-l2ug2s)*

bstaletic:
@zzbot r=bstaletic

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/955#-:-L8ewR3L0BmraiZjeoij:b1kuxdk)*

PR #957, opened 2018-03-13
[READY] Send the full query to the language server when forced

puremourning:
jdt.ls does not return anything useful for top-level completions,
presumably because with no query to work from it has no way to filter
the almost infinite set of possible suggestions.

However it does provide useful completion at the top level (including
automatic imports) when it is given the query string to work with. The
main drawback of this is that the filtering that gets applied is done by
whatever algorithm jdt.ls applies, not what ycmd applies. This can lead
to confusion and breaking the user's mental model of how ycmd completion
works.

Automatic imports are particularly useful for top-level completions,
which are (in general) forced by the user.

So as a compromise, we use the column_codepoint (i.e. including the
'query') when the user forces semantic completion and use the
start_codepoint (i.e. first char after trigger/beginning of identifier),
when 'triggered'.

This allows for powerful features and a simple mental model: if you
force completion, you might get more relevant results, but the filtering
is necessarily different.

As this won't necessarily be universal, it is provided as an API hook.

If we find that multiple language servers behave this way, we can promote the override to the base implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/957)

### DIFF
--- a/ycmd/completers/java/java_completer.py
+++ b/ycmd/completers/java/java_completer.py
@@ -426,6 +426,24 @@ class JavaCompleter( language_server_completer.LanguageServerCompleter ):
       self._CleanUp()
 
 
+  def GetCodepointForCompletionRequest( self, request_data ):
+    """Returns the 1-based codepoint offset on the current line at which to make
+    the completion request"""
+    # When the user forces semantic completion, we pass the actual cursor
+    # position to jdt.ls.
+
+    # At the top level (i.e. without a semantic trigger), there are always way
+    # too many possible candidates for jdt.ls to return anything useful. This is
+    # because we don't send the currently typed characters to jdt.ls. The
+    # general idea is that we apply our own post-filter and sort. However, in
+    # practice we never get a full set of possibilities at the top-level. So, as
+    # a compromise, we allow the user to force us to send the "query" to the
+    # semantic engine, and thus get good completion results at the top level,
+    # even if this means the "filtering and sorting" is not 100% ycmd flavor.
+    return ( request_data[ 'column_codepoint' ]
+             if utils.ForceSemanticCompletion( request_data )
+             else request_data[ 'start_codepoint' ] )
+
 
   def HandleNotificationInPollThread( self, notification ):
     if notification[ 'method' ] == 'language/status':
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -597,6 +597,9 @@ class LanguageServerCompleter( Completer ):
   Completions
 
   - The implementation should not require any code to support completions
+  - (optional) Override GetCodepointForCompletionRequest if you wish to change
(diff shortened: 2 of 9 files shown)


puremourning:
If you want a before/after test, it's as simple as say typing `Array` at the top level and hitting ctrl-space. Before, you'd be lucky to get any meaningful results (because the completion request is "before the A") but now, you should see all the various Array classes, like `ArrayList` etc.

micbou:
I don't think we need to introduce a new key. We can re-use `force_semantic` by adding it to the `_computed_key` dictionary like this:
```python
'force_semantic': ( self._GetForceSemantic, self._SetForceSemantic )
```
where `_GetForceSemantic` is defined as:
```python
def _GetForceSemantic( self ):
  return self._request.get( 'force_semantic', False )
```
and `_SetForceSemantic` is the same as `_SetForcedSemanticCompletion`.
                     

---

Review status: 0 of 8 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/957#-:-L7X-7zDkYdGyCAErrVA:b-7obwbm)*

micbou:
At this point, I would completely remove the `ForceSemanticCompletion` function. If we want to be sure that `request_data[ 'force_semantic' ]` is a boolean, we can return `bool( ... )` in `_GetForceSemantic`. Aside from that, :lgtm:.

---

Reviewed 5 of 8 files at r1, 4 of 5 files at r2.
Review status: 8 of 9 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/957#-:-L7o8gK-8f802Td0zhp6:bs3d3hr)*

puremourning:
I think we can/should remove ForceSemanticCompletion in another non-functional PR. Reason is it is actually used in quite a few places, and i think we can also now refactor ShouldUseFileTypeCompleter so it no longer returns the tuple (which is a mess). In the interest of keeping this PR minimal, do you agree?

---

Review status: 8 of 9 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/957#-:-L7offQvnqk988_dQQ98:bv9njlj)*

bstaletic:
I don't see a reason not to. :lgtm: 

---

Reviewed 8 of 8 files at r1, 3 of 5 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/957#-:-L7ohbHzxPpKUOcQkBp2:b1esv2u)*

micbou:
Okay, let's do that in another PR.

@zzbot r+

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/957#-:-L7oinTuzIZLTAT8Cehn:br8qjjo)*

PR #961, opened 2018-03-18
[READY] Fix ycm_core_tests incremental builds on Windows

micbou:
When building the `ycm_core_tests` target for the first time on Windows, we override the `BUILD_SHARED_LIBS` variable to dynamically link the `gmock` library. The Boost libraries are still statically built because they are compiled before the option is set. However, if we rebuild the `ycm_core_tests` target, `BUILD_SHARED_LIBS` is now set and CMake tries to build dynamic libraries of Boost which fails because the Python dynamic library cannot be found (which is expected since there is no `link_directories` directive to build Boost). To fix that, we need to force building Boost statically.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/961)

### DIFF
--- a/cpp/BoostParts/CMakeLists.txt
+++ b/cpp/BoostParts/CMakeLists.txt
@@ -36,7 +36,7 @@ include_directories(
   ${PYTHON_INCLUDE_DIRS}
   )
 
-add_library( BoostParts ${SOURCES} )
+add_library( BoostParts STATIC ${SOURCES} )
 
 #############################################################################
 


puremourning:
Not that I have ever done this, nor would I expect anyone to, but this change does mean that setting `BUILD_SHARED_LIBS=ON` now actually works. (I mean it works in that it builds, though it doesn't actually build any shared libs other than ycm_core). Point is, I tested build_shared_libs=ON with master and it fails to compile. With this change it compiles.

:lgtm: 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/961#-:-L7ufztrd-lZlbLaxEnI:b-a8vfe)*

bstaletic:
:lgtm:
@zzbot r=puremourning

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/961#-:-L7ySNNNoMvZPFUO1cbv:bsrfr92)*

micbou:
@zzbot retry

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/valloric/ycmd/961#-:-L8-AD3_GCndeMcH8Lu8:b-kc0oa4)*

PR #966, opened 2018-03-21
Show the right message in the clang completer when _LocationForGoTo is called and the file is being parsed

Emigre:
In a C++ file, if a user tries to use one of the GoTo commands while the file is being parsed, no location is returned and the user receives a `Can't jump to declaration` `RuntimeError` message or similar. If the user waits a couple of seconds for the parsing to end, the command will work. 

The error message is misleading, I think that the user should be shown a `PARSING_FILE_MESSAGE` as in [line 151](https://github.com/Emigre/ycmd/blob/4de05405155803e8d0bf80262ff630815f3ca9c6/ycmd/completers/cpp/clang_completer.py#L151) of the modified file. With this change, the user sees a `ValueError: Still parsing file, no completions yet.` message and knows that he has to wait.

**How to reproduce**
- Create a C++ project.
- Enter a file and quickly try to use`YcmCompleter GoToDeclaration`

**Expected result**
- If the file is being parsed at that moment the user is shown a `Still parsing file, no completions yet` information message. 
- Then if he waits and tries again when it has been parsed, the command works as expected.

**Actual result**
- If the file is being parsed at that moment the user is shown a `Can't jump to declaration` error message. 
- Then if he waits and tries again when it has been parsed, the command works as expected.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/966)

### DIFF
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -211,6 +211,10 @@ class ClangCompleter( Completer ):
     if not flags:
       raise ValueError( NO_COMPILE_FLAGS_MESSAGE )
 
+    if self._completer.UpdatingTranslationUnit(
+        ToCppStringCompatible( filename ) ):
+      raise ValueError( PARSING_FILE_MESSAGE )
+
     files = self.GetUnsavedFilesVector( request_data )
     line = request_data[ 'line_num' ]
     column = request_data[ 'column_num' ]


Emigre:
I noticed what I think is a bug, I'm not familiar with this project so please correct me if I'm wrong. Thanks.

Emigre:
When moving from an already parsed file to another file in another tab and then going back to the previous file, I notice that the file is parsed again, with the corresponding delay.

Is this normal? Wouldn't it make sense to keep the old parse information, for the sake of speed?

puremourning:
Thanks for sending a PR!

We would need to include test cases that cover this however...

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/966#-:-L8O_-ki3aDinU17gJgA:b-mxca38)*

Emigre:
@puremourning you are right. I've taken a look around, I'm not familiar with the project and I don't really know where to add the test or how to setup it all. Would you mind adding the test yourself? It's a three-lines change and I'd say it would be faster if you could. I imagine that you are familiar with the project.

Emigre:
I've found what I think is another error, this one in the Python completer. I've written a fix for it, but it causes the tests to break. I can run the tests but I'm not sure how to change them for the fix - I don't really know how they work.

I haven't commited or pushed that fix. It is, again, related to errors not being shown to the user properly. I could add it to this PR, or to a new one, but I'm not able to change the tests on my own, so I'm not entirely sure of what to do.

micbou:
> I've found what I think is another error, this one in the Python completer. I've written a fix for it, but it causes the tests to break. I can run the tests but I'm not sure how to change them for the fix - I don't really know how they work.

Could you raise an issue with a description of the problem so that we can take a look and potentially fix it?

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/966#-:-L99zx4o6_Wv9OU02u9B:bckzbkh)*

Emigre:
sure @micbou, I will do it later. Thanks

Emigre:
thanks for #1088 @micbou and apologies, I haven't had the time to look into this. Thanks for the fix

PR #973, opened 2018-03-27
[READY] Add TSServer version to debug info

micbou:
This is more complicated than expected for a number of reasons:
 - we need to send the `status` request to get the version;
 - we can't just send the request when getting the debug info because we want to know the version even if TSServer crashed so we send the request immediately after the server started and store the result in a variable;
 - we don't want the request to block;
 - we need a lock for the variable because it can be accessed by multiple threads.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/973)

### DIFF
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -176,9 +176,7 @@ class TypeScriptCompleter( Completer ):
     self._tsserver_is_running = threading.Event()
 
     # Start a thread to read response from TSServer.
-    self._thread = threading.Thread( target = self._ReaderLoop, args = () )
-    self._thread.daemon = True
-    self._thread.start()
+    utils.StartThread( self._ReaderLoop )
 
     # Used to map sequence id's to their corresponding DeferredResponse
     # objects. The reader loop uses this to hand out responses.
@@ -200,13 +198,8 @@ class TypeScriptCompleter( Completer ):
 
 
   def _SetServerVersion( self ):
-    def SetServerVersionInThread():
-      with self._tsserver_version_lock:
-        self._tsserver_version = self._SendRequest( 'status' )[ 'version' ]
-
-    self._thread = threading.Thread( target = SetServerVersionInThread )
-    self._thread.daemon = True
-    self._thread.start()
+    with self._tsserver_version_lock:
+      self._tsserver_version = self._SendRequest( 'status' )[ 'version' ]
 
 
   def _StartServer( self ):
@@ -236,7 +229,7 @@ class TypeScriptCompleter( Completer ):
 
       self._tsserver_is_running.set()
 
-      self._SetServerVersion()
+      utils.StartThread( self._SetServerVersion )
 
 
   def _ReaderLoop( self ):
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -153,11 +153,12 @@ class TypeScriptCompleter( Completer ):
 
     self._logfile = None
 
+    self._tsserver_lock = threading.RLock()
     self._tsserver_binary_path = FindTsserverBinary()
-
     self._tsserver_handle = None
-    self._tsserver_version_lock = threading.Lock()
     self._tsserver_version = None
+    # Used to read response only if TSServer is running.
+    self._tsserver_is_running = threading.Event()
 
     # Used to prevent threads from concurrently writing to
     # the tsserver process' stdin
@@ -169,11 +170,6 @@ class TypeScriptCompleter( Completer ):
 
     # Used to prevent threads from concurrently accessing the sequence counter
     self._sequenceid_lock = threading.Lock()
-
-    self._server_lock = threading.RLock()
-
-    # Used to read response only if TSServer is running.
-    self._tsserver_is_running = threading.Event()
 
     # Start a thread to read response from TSServer.
     utils.StartThread( self._ReaderLoop )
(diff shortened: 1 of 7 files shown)


Valloric:
:lgtm: Thanks!

---

---

*[ycmd/completers/typescript/typescript_completer.py, line 235 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/973#-L8dKRv96KqcRt75Nd--:-L8dKRv96KqcRt75Nd-0:b7acm28) ([raw file](https://github.com/valloric/ycmd/blob/962bef41bc915044bee54cfb3336e94f20a54e10/ycmd/completers/typescript/typescript_completer.py#L235)):*
>    self._thread = threading.Thread( target = SetServerVersionInThread )
>     self._thread.daemon = True
>     self._thread.start()

I think this is a fairly common pattern in ycmd; create a thread, make it daemon, start it. Maybe create a helper for this? Just a suggestion.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/973#-:-L8dKpbT0CKfVzobH6OV:b-g2a5fb)*

puremourning:

---

*[ycmd/completers/typescript/typescript_completer.py, line 231 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/973#-L8dR8FX8C0rjZInQ_OO:-L8dR8FX8C0rjZInQ_OP:b-51vz47) ([raw file](https://github.com/valloric/ycmd/blob/962bef41bc915044bee54cfb3336e94f20a54e10/ycmd/completers/typescript/typescript_completer.py#L231)):*
> ```Python
>     def SetServerVersionInThread():
>       with self._tsserver_version_lock:
>         self._tsserver_version = self._SendRequest( 'status' )[ 'version' ]
> ```

Do we really want to/need to hold this lock the entire time? If this request hangs for whatever reason, then the DebugInfo request will hang waiting for it. Perhaps we should only acquire the lock after the request returns. While the request is outstanding we could report "Checking..." or something in the debug info, or just not include it if it is currently set to None.

Not a huge deal.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/973)*

micbou:
Reviewed 4 of 6 files at r2.
Review status: 4 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/completers/typescript/typescript_completer.py, line 231 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/973#-L8dR8FX8C0rjZInQ_OO:-L8dZ--13IyIXbgcdJl6:baeq68i) ([raw file](https://github.com/valloric/ycmd/blob/962bef41bc915044bee54cfb3336e94f20a54e10/ycmd/completers/typescript/typescript_completer.py#L231)):*
We now hold the lock only when setting the variable and we reuse `_server_lock` (renamed to `_tsserver_lock`).

>  While the request is outstanding we could report "Checking..." or something in the debug info, or just not include it if it is currently set to None.

TSServer's startup is fast enough that I don't think it's worth the trouble. Also, this is rather annoying to test.

---

*[ycmd/completers/typescript/typescript_completer.py, line 235 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/973#-L8dKRv96KqcRt75Nd--:-L8dYx7gCP3IdiN34aZW:byx4o9i) ([raw file](https://github.com/valloric/ycmd/blob/962bef41bc915044bee54cfb3336e94f20a54e10/ycmd/completers/typescript/typescript_completer.py#L235)):*

>    self._thread = threading.Thread( target = SetServerVersionInThread )
>     self._thread.daemon = True
>     self._thread.start()

I think this is a fairly common pattern in ycmd; create a thread, make it daemon, start it. Maybe create a helper for this? Just a suggestion.
</blockquote></details>

Agreed. Added the `StartThread` helper.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/973)*

puremourning:
Review status: 4 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[ycmd/completers/typescript/typescript_completer.py, line 231 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/973#-L8dR8FX8C0rjZInQ_OO:-L8diBIm8GQfYoUPEhDH:b-4h1dkz) ([raw file](https://github.com/valloric/ycmd/blob/962bef41bc915044bee54cfb3336e94f20a54e10/ycmd/completers/typescript/typescript_completer.py#L231)):*

We now hold the lock only when setting the variable and we reuse `_server_lock` (renamed to `_tsserver_lock`).

>  While the request is outstanding we could report "Checking..." or something in the debug info, or just not include it if it is currently set to None.

TSServer's startup is fast enough that I don't think it's worth the trouble. Also, this is rather annoying to test.
</blockquote></details>

OK cool.  I think that this means if the debug info request is processed prior to the version being set that we end up setting the version to 'null' in the debug info response. I think that's ok as long as it doesn't cause an exception.

The 'simple' way to test it would be to @patch.object( <this>, '_tsserver_version', None ) and run the debug info test.

It's probably OK even if it does raise an exception tbh.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/973)*

puremourning:
Reviewed 1 of 2 files at r1, 6 of 6 files at r2.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/973)*

bstaletic:
:lgtm:
@zzbot r=valloric

---

Reviewed 1 of 2 files at r1, 5 of 6 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/973#-:-L9J9ZVY5yLI6zaZqf6z:b-5mwjnx)*

PR #974, opened 2018-03-28
[READY] Replace re module with regex

micbou:
The `\w` pattern from the `re` module does not match Unicode characters with combining marks (e.g. `e` + acute accent). This causes identifiers that contains such characters to not be properly extracted from the buffer. The solution is to replace the `re` module with [regex](https://pypi.python.org/pypi/regex/). This module has a compiled component so we can't simply vendor it. Instead, we download it in the `build.py` script as a pip package and install it to `third_party/regex` in the `py2` folder for Python 2 and `py3` for Python 3. We use two folders because YCM is going to import this module and if it's running on a different version of Python than ycmd, this will fail. We need to fall back to the `re` module in that case.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/974)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,6 @@ docs/package-lock.json
 
 # jdt.ls
 third_party/eclipse.jdt.ls
+
+# regex module
+third_party/regex/py*
--- a/build.py
+++ b/build.py
@@ -85,6 +85,8 @@ JDTLS_SHA256 = (
   '4fe3ca50d2b7011f7323863bdf77a16979e5d3a2a534d69e1ef32742cc443061'
 )
 
+REGEX_MODULE_VERSION = '2018.02.21'
+
 
 def OnMac():
   return platform.system() == 'Darwin'
@@ -155,10 +157,8 @@ def NumCores():
 
 
 def CheckCall( args, **kwargs ):
-  quiet = kwargs.get( 'quiet', False )
-  kwargs.pop( 'quiet', None )
-  status_message = kwargs.get( 'status_message', None )
-  kwargs.pop( 'status_message', None )
+  quiet = kwargs.pop( 'quiet', False )
+  status_message = kwargs.pop( 'status_message', None )
 
   if quiet:
     _CheckCallQuiet( args, status_message, **kwargs )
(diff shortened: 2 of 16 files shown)


bstaletic:
:lgtm:

Once again, amazing work on unicode support, @micbou.

---

Reviewed 16 of 16 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/974#-:-L8i6MkZ1XN9g-yfPAH_:b-gu34ge)*

micbou:
I made it so that the script does not exit if the installation of the `regex` module fails. @puremourning Would that work for you or do we need to add an option (e.g. `--no-regex`) to disable installation of that module?

---

Reviewed 14 of 16 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/974#-:-L9QfOVb6xxVQq22L0z-:b-xma8ar)*

micbou:
> Out of curiosity do we know if there is a perf difference between re and regex ?

I found [this benchmark](https://pypi.python.org/pypi/re2/#performance) comparing `re`, `regex`, and `re2` (wrapper of Google's RE2) and it looks like `regex` performs better than `re`.

---

Review status: 12 of 16 files reviewed at latest revision, 1 unresolved discussion.

---

*[build.py, line 533 at r3](https://beta.reviewable.io/reviews/valloric/ycmd/974#-L9vhwzE5Rxr5KUW57YH:-L9wHyBY0Vbpb_2PDVvf:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/5674e8c25eaf4bbad947899c30e594b968819d41/build.py#L533)):*

Should we stick the version in a constant at the top to make it marginally easier to upgrade/change?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/974#-:-L9wE_TG0oAT-use32JQ:b-c4uhtz)*

micbou:
Conflicts fixed.

---

Reviewed 2 of 4 files at r4, 4 of 4 files at r5.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/974#-:-LA11bvDCYUlDu4rV6XI:b-wlrssi)*

bstaletic:
@zzbot r+

---

Reviewed 1 of 1 files at r2, 1 of 1 files at r3, 4 of 4 files at r4, 4 of 4 files at r5.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/974#-:-LA1MEWh1WUfBSaRk2yp:bi59kwd)*

PR #975, opened 2018-03-28
[READY] Add smart diacritic matching

micbou:
This PR adds smart [diacritic](http://www.unicode.org/glossary/#diacritic) matching on top of the smart case matching, e.g.:
  - `e` matches `e`, `é`, `E`, `É`;
  - `E` matches `E`, `É` but not `e`, `é`;
  - `é` matches `é`, `É` but not `e`, `E`;
  - `É` matches `É` but not `e`, `é`, `E`.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/975)


bstaletic:
With this PR, the `Character` struct also contains the `base_` and `is_base` properties. Doesn't that require the `UnicodeTable.inc` to be updated?

---

Reviewed 9 of 9 files at r1.

---

*[cpp/ycm/Candidate.cpp, line 121 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/975#-L8i1BTM4J0v6In3fAd7:-L8i0JRA5mQ41lmNI4uZ:b-v9twkn) ([raw file](https://github.com/valloric/ycmd/blob/6c721052a94f67e85af1ad24a1b008782f47f953/cpp/ycm/Candidate.cpp#L121)):*
> ```cpp
>          ( !query_character->IsUppercase() &&
>            query_character->EqualsIgnoreCase( *candidate_character ) ) ||
>          *query_character == *candidate_character ) {
> ```

The condition looks really ugly right now. Can we extract it to a separate function without introducing overhead? 

---

*[cpp/ycm/Character.cpp, line 105 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/975#-L8i5ADQ9I6tVqjw2QZe:-L8i5ADQ9I6tVqjw2QZf:b-p5eisc) ([raw file](https://github.com/valloric/ycmd/blob/6c721052a94f67e85af1ad24a1b008782f47f953/cpp/ycm/Character.cpp#L105)):*
> ```cpp
>         break;
>       default:
>         base_.append( code_point->FoldedCase() );
> ```

Is `base_` left uninitialised if `is_base_` is `false`?
If so, doesn't that interfere with the changes in `ComputeBytesPresent()`?

---

*[cpp/ycm/Word.cpp, line 287 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/975#-L8i3cnW0031UZLEsdwT:-L8i3cnW0031UZLEsdwU:bjpp044) ([raw file](https://github.com/valloric/ycmd/blob/6c721052a94f67e85af1ad24a1b008782f47f953/cpp/ycm/Word.cpp#L287)):*
> ```cpp
> void Word::ComputeBytesPresent() {
>   for ( const auto &character : characters_ ) {
>     for ( uint8_t byte : character->Base() ) {
> ```

Why are we only calculating based on the `Base()` part, instead of the `FoldedCase()`?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/975#-:-L8i2m_e3EPQXcIcPjna:b-89qycf)*

micbou:
Refactored the matching condition, added tests, and bumped the core version.

> With this PR, the Character struct also contains the base_ and is_base properties. Doesn't that require the UnicodeTable.inc to be updated?

No because it's the `CodePoint` struct that's using the Unicode table, not `Character`.

---

Reviewed 7 of 9 files at r1, 5 of 5 files at r2.

---

*[cpp/ycm/Candidate.cpp, line 121 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/975#-L8i1BTM4J0v6In3fAd7:-L99o4HT3aY1IcQGRf6t:b-j9at90) ([raw file](https://github.com/valloric/ycmd/blob/6c721052a94f67e85af1ad24a1b008782f47f953/cpp/ycm/Candidate.cpp#L121)):*

The condition looks really ugly right now. Can we extract it to a separate function without introducing overhead? 
</blockquote></details>

Moved the condition to the `Character` class. Strangely enough, this seems to slightly improve performance.

---

*[cpp/ycm/Character.cpp, line 105 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/975#-L8i5ADQ9I6tVqjw2QZe:-L99oSoi5FLLCeWIRew9:baprhj2) ([raw file](https://github.com/valloric/ycmd/blob/6c721052a94f67e85af1ad24a1b008782f47f953/cpp/ycm/Character.cpp#L105)):*

Is `base_` left uninitialised if `is_base_` is `false`?
If so, doesn't that interfere with the changes in `ComputeBytesPresent()`?
</blockquote></details>

`base_` is a string so it's always initialized. It will be empty if all code points are, for instance, combining marks, which should never happen in practice (such code points don't form a valid character). But even if it does, this shouldn't be a problem. `ComputeBytesPresent` will simply ignore that character.

---

*[cpp/ycm/Word.cpp, line 287 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/975#-L8i3cnW0031UZLEsdwT:-L8mfegP76k1dJ0ep36g:b-ye49g8) ([raw file](https://github.com/valloric/ycmd/blob/6c721052a94f67e85af1ad24a1b008782f47f953/cpp/ycm/Word.cpp#L287)):*

Why are we only calculating based on the `Base()` part, instead of the `FoldedCase()`?
</blockquote></details>

`ComputeBytesPresent` purpose is to do a preliminary filtering before the actual matching by comparing the bytes of two words. If we don't use the base version of the characters, `e` (`\x65`) will not match `é` (`\xc3\xa9`).

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/975#-:-L8i6bMzEwSlO14jaVGN:b-a1txn6)*

bstaletic:
:lgtm:
Great job!

---

Reviewed 5 of 5 files at r2.

---

*[cpp/ycm/Candidate.cpp, line 121 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/975#-L8i1BTM4J0v6In3fAd7:-L9A3AgWBE_FdTBu0hLZ:bl7ei3e) ([raw file](https://github.com/valloric/ycmd/blob/6c721052a94f67e85af1ad24a1b008782f47f953/cpp/ycm/Candidate.cpp#L121)):*

Moved the condition to the `Character` class. Strangely enough, this seems to slightly improve performance.
</blockquote></details>

The performance improvement makes no sense to me, but sound great.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/975#-:-L9A3zht-iQF8VXNVPlH:bvcm1o0)*

Valloric:
Awesome, thank you! :lgtm: 

@zzbot r+

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/975#-:-L9GjC5V7xRDUNHzpsrV:b-2209gs)*

micbou:
@zzbot retry

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/975#-:-L9HSzgQ5h0p3y1NpEub:b-kc0oa4)*

PR #976, opened 2018-03-28
[READY] Add FixIt command to TypeScript completer

micbou:
This is implemented similarly to the Clang completer i.e. the FixIts are stored in the diagnostics. Demo:

typescript-fixit

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/976)

### DIFF
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -1,5 +1,4 @@
-# Copyright (C) 2015-2016 Google Inc.
-#               2016-2017 ycmd contributors
+# Copyright (C) 2015-2018 ycmd contributors
 #
 # This file is part of ycmd.
 #
@@ -30,6 +29,7 @@ import re
 import subprocess
 import itertools
 import threading
+from collections import defaultdict
 from functools import partial
 
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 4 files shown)


bstaletic:
:lgtm: 

---

Reviewed 4 of 4 files at r1.

---

*[ycmd/responses.py, line 236 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/976#-L8iADIG71EhPkJl8sq_:-L8iADIH0UnjFCDkBcvG:bocyvmz) ([raw file](https://github.com/valloric/ycmd/blob/9a8f1eb0c83fe4646ad20166f458def5523aa54e/ycmd/responses.py#L236)):*
> ```Python
>     'text': diagnostic.text_,
>     'kind': kind,
>     'fixit_available': len( diagnostic.fixits_ ) > 0,
> ```

Unrelated to this PR, but should we make a suggestion for LSP to support something like this?\\\\\

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/976#-:-L8iBjOzAZQ4s48uc20B:bg7igkv)*

puremourning:
:lgtm: 

The one thought I had is that we optimistically generate the fixits as ycmd fixits. One option would be to store the raw TSServer data and only convert when applied. This would then defer that work to the synchronous FixIt command rather than the OnFileReadyToParse event. It's probably not a noticeable difference, and I don't write typescript in anger so I don't know how likely it is that it would be.

@zzbot r+

---

---

*[ycmd/responses.py, line 236 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/976#-L8iADIG71EhPkJl8sq_:-L96_-qmDGp6auvm6gO5:bcemmex) ([raw file](https://github.com/valloric/ycmd/blob/9a8f1eb0c83fe4646ad20166f458def5523aa54e/ycmd/responses.py#L236)):*

Unrelated to this PR, but should we make a suggestion for LSP to support something like this?\\\\\
</blockquote></details>

you can do the LSP thing, by sending a codeaction request, but we would need to send it for every diagnostic would would be a performance issue, and tbh might not do exactly what we want.

---

*[ycmd/responses.py, line 155 at r2](https://beta.reviewable.io/reviews/valloric/ycmd/976#-L96ZmnpAmBZ9cIh8b3W:-L96ZmnpAmBZ9cIh8b3X:bpwf28x) ([raw file](https://github.com/valloric/ycmd/blob/033b5ac083523555161b80f640f3cf00083eadca/ycmd/responses.py#L155)):*
> ```
>   def __init__ ( self, ranges,
>                        location,
> ```
 
seems strange to break at second comma and align with ranges, rather than break at comma and align with first argument. 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/976#-:-L96aPfU0hkjKdjYR1PM:b-4qess7)*

bstaletic:
Reviewed 1 of 1 files at r2.

---

*[ycmd/responses.py, line 236 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/976#-L8iADIG71EhPkJl8sq_:-L96fJvF1fzrOxv8_fiU:b-xci413) ([raw file](https://github.com/valloric/ycmd/blob/9a8f1eb0c83fe4646ad20166f458def5523aa54e/ycmd/responses.py#L236)):*

you can do the LSP thing, by sending a codeaction request, but we would need to send it for every diagnostic would would be a performance issue, and tbh might not do exactly what we want.
</blockquote></details>

Still not as nice as if it would be if part of the protocol.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/976)*

micbou:
@zzbot r+

---

Reviewed 3 of 4 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4.

---

*[ycmd/responses.py, line 155 at r2](https://beta.reviewable.io/reviews/valloric/ycmd/976#-L96ZmnpAmBZ9cIh8b3W:-L96gP1u8qNh7kBCaRGH:bt2vr1b) ([raw file](https://github.com/valloric/ycmd/blob/033b5ac083523555161b80f640f3cf00083eadca/ycmd/responses.py#L155)):*

> ```
>   def __init__ ( self, ranges,
>                        location,
> ```
 
seems strange to break at second comma and align with ranges, rather than break at comma and align with first argument. 
</blockquote></details>

Fixed.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/976#-:-L96gS2oEZOa2ympu5Mj:bi59kwd)*

PR #979, opened 2018-03-29
Fix building with system boost on Fedora

eklitzke:
Without this ycm_core.so builds but doesn't have an init function exported. Which is really confusing -- if you try to run in py3 you get an error message about ycm_core.so being built for py2, and vice versa if you try to run in py2.

I would much rather the code to an autotools-style linking test to find the right boost lib, but I don't know enough about CMake to contribute that.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/979)


eklitzke:
Submitted this too soon, the build is still broken (although less so with this). Will re-open after I figure out why it isn't exporting an init symbol.

bstaletic:
Is the fedora's boost version 1.65?
That version is buggy and won't work.

PR #984, opened 2018-04-03
[READY] Always return detailed completions in TypeScript completer

micbou:
This limit of 100 detailed completions is often a source of confusion to users. See issues https://github.com/Valloric/YouCompleteMe/issues/1878, https://github.com/Valloric/YouCompleteMe/issues/2040, and https://github.com/Valloric/YouCompleteMe/issues/2955. With completions being async in YCM, this is less a problem to always return detailed completions even if there are a lot of suggestions. Removing this limit is also important to support automatic imports (we need detailed completions for that).

Closes https://github.com/Valloric/YouCompleteMe/issues/2955.

### DIFF
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -382,11 +382,6 @@ class TypeScriptCompleter( Completer ):
       'offset': request_data[ 'start_codepoint' ]
     } )
 
-    # A less detailed version of the completion data is returned
-    # if there are too many entries. This improves responsiveness.
-    if len( entries ) > MAX_DETAILED_COMPLETIONS:
-      return [ _ConvertCompletionData(e) for e in entries ]
-
     detailed_entries = self._SendRequest( 'completionEntryDetails', {
       'file':       request_data[ 'filepath' ],
       'line':       request_data[ 'line_num' ],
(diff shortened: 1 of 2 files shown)


puremourning:
Hmm. This is the opposite of what I did for jdt.ls, which I found to be unusably slow with very many completion suggestions (in my case there was a 'class' with thousands of methods). I actually [implemented this exact restriction](https://github.com/Valloric/ycmd/blob/master/ycmd/completers/language_server/language_server_completer.py#L842) in the language_server_completer.

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/984#-:-L9Cj4cyB3uOlEq1xaRo:bvhyh16)*

micbou:
The difference is that only one additional request is needed to get the detailed completions for TypeScript while it's one request for each completion for the language server completer.

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/984#-:-L9Cjo3mEDUXWziWL5Xc:b-ampfit)*

bstaletic:
:lgtm:
@zzbot r=puremourning

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/984#-:-L9DuKSL5Q7DsDRfE-gB:bsrfr92)*

PR #986, opened 2018-04-04
Allow ycmd to pass jdt.ls options on initialization

hdiniz:
There are many options supported by jdt.ls that need to be sent by a
client over Initialize and DidConfigurationChange notifications.
This change allows a specific setting to be passed based on an
user_option. Other settings can be integrated due to other use cases.

This proposal can be used to further extend eclipse.jdt.ls options. But for my current use case, I need a way to disable the Gradle import done by jdt.ls (buildship) because I manage the eclipse files manually (.classpath, etc)... This is due to limitations in buildship to make proper import of Gradle-based Android projects. Which is a use case that I see appear on other open issues for jdt.ls language clients.

Most significant perhaps, is the move of DidChangeConfiguration notification from inside the Initialize response handler.. And leaving it up to language clients to implement as needed. This was necessary because jdt.ls resets its settings if DidChangeConfiguration is received with an empty object.

Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/986)

### DIFF
--- a/ycmd/completers/java/java_completer.py
+++ b/ycmd/completers/java/java_completer.py
@@ -86,6 +86,7 @@ WORKSPACE_ROOT_PATH = os.path.abspath( os.path.join(
 #    ycmd instance
 #  - An option is available to re-use workspaces
 CLEAN_WORKSPACE_OPTION = 'java_jdtls_use_clean_workspace'
+ENABLE_GRADLE_IMPORT  = 'java_jdlts_enable_gradle_import'
 
 
 def ShouldEnableJavaCompleter():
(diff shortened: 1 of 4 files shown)


bstaletic:
The builds are failing because of the following:

```
/Users/distiller/project/ycmd/completers/java/java_completer.py:91:1: E302 expected 2 blank lines, found 1
/Users/distiller/project/ycmd/completers/language_server/language_server_completer.py:1266:81: E501 line too long (81 > 80 characters)
```

hdiniz:
@bstaletic I fixed those styling problems. There is still a test failure, but I don't see how that is relevant. Maybe a flaky test case?

hdiniz:
Dear @bstaletic, these tests are passing on my setup, could you help me check this?

Best regards

bstaletic:
The tests were flakes. If you take a look right now, all of them are passing (except the codecov nonsense which you can disregard).

Thanks a lot for working on this.
I've taken a look at the implementation and it is using the "global" options for server (jdt.ls) configuration.
It would be more flexible if it was done through `.ycm_extra_conf.py` to allow per-project configurations, which is also the reason I have not given this PR a proper review - I'd like to hear from others, especially @puremourning, who wrote most of the jdt.ls completer.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/986#-:-L9jd4H74bhfQdfJx7oF:b-xe4cr6)*

bstaletic:
Personally, I agree with @micbou that this should be done through the extra conf, because it allows the easiest per-project configuration.

Valloric:
Closing PR as it's not moving forward; can reopen in the future if desired.

PR #1002, opened 2018-04-15
[READY] Improve message when regex module installation fails

micbou:
When pip is not available, the regex module cannot be installed but the script still continues since this module is not required. However, the absence of message or the `FAILED` message when `--quiet` is given may let the user thinks that the whole installation failed. We improve that by printing a warning if the regex module cannot be installed. When installation is successful:
```sh
$ ./build.py
...
Collecting regex==2018.02.21
Installing collected packages: regex
Successfully installed regex
$ ./build.py --quiet
...
Installing regex module...OK
```
When pip is not available:
```sh
$ ./build.py
...
/usr/bin/python: No module named pip
WARNING: pip is required to install the regex module.
Unicode will not be fully supported without this module.
$ ./build.py --quiet
...
Installing regex module...SKIP
WARNING: pip is required to install the regex module.
Unicode will not be fully supported without this module.
```
When another error occurs (a connection error in that case; note how the exception from pip is helpful):
```sh
$ ./build.py
...
Collecting regex==2018.02.21
Exception:
...
TypeError: unsupported operand type(s) for -=: 'Retry' and 'int'
WARNING: cannot install the regex module. Unicode will not be fully supported.
$ ./build.py --quiet
...
Installing regex module...SKIP
WARNING: cannot install the regex module. Unicode will not be fully supported.
```

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1002)

### DIFF
--- a/build.py
+++ b/build.py
@@ -168,9 +168,9 @@ def CheckCall( args, **kwargs ):
 
 def _CheckCallQuiet( args, status_message, **kwargs ):
   if not status_message:
-    status_message = 'Running {0}'.format( args[ 0 ] )
+    status_message = 'Running {}'.format( args[ 0 ] )
 
-  # __future_ not appear to support flush= on print_function
+  # __future__ not appear to support flush= on print_function
   sys.stdout.write( status_message + '...' )
   sys.stdout.flush()
 
@@ -531,27 +531,47 @@ def BuildYcmdLib( args ):
 
 
 def InstallRegexModule( args ):
+  # We don't exit the script if the regex module cannot be installed; ycmd is
+  # still usable without this module.
+  if args.quiet:
+    sys.stdout.write( 'Installing regex module...' )
+    sys.stdout.flush()
+
   regex_dir = p.join( DIR_OF_THIRD_PARTY, 'regex', 'py{}'.format( PY_MAJOR ) )
   pip_command = [ sys.executable, '-m', 'pip', 'install', '--upgrade',
                   'regex=={}'.format( REGEX_MODULE_VERSION ), '-t', regex_dir ]
+
   # We need to add the --system option on Debian-like distributions. See
   # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=830892
   try:
-    pip_help_output = subprocess.check_output( pip_command + [ '--help' ],
-                                               stderr = subprocess.STDOUT )
-    if '--system' in pip_help_output.decode( 'utf8' ):
-      pip_command.append( '--system' )
-  except subprocess.CalledProcessError:
-    pass
+    output = subprocess.check_output(
+      pip_command + [ '--help' ], stderr = subprocess.STDOUT ).decode( 'utf8' )
+  except subprocess.CalledProcessError as error:
+    output = error.output.decode( 'utf8' )
 
-  # Do not exit if installing the regex module fails; ycmd is still usable
-  # without this module.
+  # Return early if pip is not available.
+  if 'No module named pip' in output:
+    message = 'SKIP\n' if args.quiet else output
+    message += ( 'WARNING: pip is required to install the regex module.\n'
+                 'Unicode will not be fully supported without this module.' )
+    print( message )
+    return
+
+  if '--system' in output:
+    pip_command.append( '--system' )
+
   try:
-    CheckCall( pip_command,
-               quiet = args.quiet,
-               status_message = 'Installing regex module' )
-  except SystemExit:
-    pass
+    if args.quiet:
+      subprocess.check_call( pip_command, stdout = subprocess.PIPE,
+                                          stderr = subprocess.PIPE )
+      print( 'OK' )
+    else:
+      subprocess.check_call( pip_command )
+  except subprocess.CalledProcessError:
+    if args.quiet:
+      print( 'SKIP' )
+    print( 'WARNING: cannot install the regex module. '
+           'Unicode will not be fully supported.' )
 
 
 def EnableCsCompleter( args ):


puremourning:
Reviewed 1 of 1 files at r1.

---

*[build.py, line 555 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/1002#-LAFSNb8-C1EeMIWatI6:-LAFSNb8-C1EeMIWatI7:b2pfr16) ([raw file](https://github.com/valloric/ycmd/blob/0598793c7cac373b6a1419ddc65ec07f8c5a3b55/build.py#L555)):*
> ```Python
>   # Do not exit if installing the regex module fails; ycmd is still usable
>   # without this module.
>   try:
> ```

Is there any point running this if `pip_available` is `False` ?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/1002)*

micbou:
Reviewed 1 of 1 files at r1.

---

*[build.py, line 555 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/1002#-LAFSNb8-C1EeMIWatI6:-LAFSqod2aEtil9BhJYV:b-egnoyu) ([raw file](https://github.com/valloric/ycmd/blob/0598793c7cac373b6a1419ddc65ec07f8c5a3b55/build.py#L555)):*

Is there any point running this if `pip_available` is `False` ?
</blockquote></details>

We could return early when `pip_available` is `False` but we would have to add one more `args.quiet` condition and it doesn't hurt to run again the command even if we know it's going to fail. If you feel that's too confusing, I can change it.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/1002)*

micbou:
Removed the `--disable-pip-version-check` option since it doesn't exist in old versions of pip.

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/1002#-:-LAKBeAk74MdpzOzIccL:bqvyspk)*

puremourning:
On an unrelated note, the effect of this pip command at work is kinda annoying. Due to the vagaries of the great firewall of office, the script just hangs when trying to run the eventual pip command (should pip be found). 
I have locally just changed the method to return immediately, but thought I'd mention it in passing. (no action required).
If that becomes untenable, I might push a PR for a --offline switch or something that doesn't attempt to contact them internets.

:lgtm: with consideration of the one comment . thanks for this.

---

---

*[build.py, line 555 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/1002#-LAFSNb8-C1EeMIWatI6:-LAKQZ2JAzUTHKy6Dn8a:b-d0gc7x) ([raw file](https://github.com/valloric/ycmd/blob/8e15f0d13375ec5b320904ecf110cf5b6a69e757/build.py#L555)):*

We could return early when `pip_available` is `False` but we would have to add one more `args.quiet` condition and it doesn't hurt to run again the command even if we know it's going to fail. If you feel that's too confusing, I can change it.
</blockquote></details>

I just the logic here is quite fiddly (not helped tbh because some muppet added a --quiet switch which means things like this are more work than necessary!), so if it can be simplified in terms of data flow, then I think it's worth it, if it isn't too much retesting etc. I realise there is no "regression test" for this script, so changes can be painful. I'll leave it up to you :)

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/1002#-:-LAKQt048AjM4i2t4bwn:byg48ai)*

puremourning:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/1002)*

micbou:
> the script just hangs when trying to run the eventual pip command (should pip be found).

Does it hang indefinitely?

> If that becomes untenable, I might push a PR for a --offline switch or something that doesn't attempt to contact them internets.

There is also the possibility to add a switch to disable the installation of the regex module.

---

Reviewed 1 of 1 files at r3, 1 of 1 files at r4.

---

*[build.py, line 555 at r1](https://beta.reviewable.io/reviews/valloric/ycmd/1002#-LAFSNb8-C1EeMIWatI6:-LAKXTev5BlmvxuVkhmD:beb8odv) ([raw file](https://github.com/valloric/ycmd/blob/8e15f0d13375ec5b320904ecf110cf5b6a69e757/build.py#L555)):*

I just the logic here is quite fiddly (not helped tbh because some muppet added a --quiet switch which means things like this are more work than necessary!), so if it can be simplified in terms of data flow, then I think it's worth it, if it isn't too much retesting etc. I realise there is no "regression test" for this script, so changes can be painful. I'll leave it up to you :)
</blockquote></details>

I've rewritten the code so that we now directly return if pip is not available. While this results in more lines, the logic should be a little bit easier to follow.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/1002#-:-LAKXN7d7I9lTKnMjRf6:b-hi583u)*

puremourning:
It hangs long enough that I ctrl-c it. It's not indefinite; it's a foible of the fact that there's no route to the internet, so the request just times out eventually. The timeout is quite long :)

There is the option for a --without-regex or something, but again, it's not worth it just for me. If we get a lot of reports, let's consider it.

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/1002#-:-LAKfP-87Yd9ualqNsoF:bi2xaik)*

bstaletic:
This is :lgtm: if you ask me.
I'm not triggering zzbot because I'm not sure if @puremourning would like to add something more to the discussion.

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/1002#-:-LAMzWN75aDa3BfVNbmg:bsvel6n)*

micbou:
PR is now ready.

> There is the option for a --without-regex or something, but again, it's not worth it just for me. If we get a lot of reports, let's consider it.

Agreed.

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/1002#-:-LAN0QeNBXj65piCCezF:bjq6hnm)*

puremourning:
:lgtm:

@zzbot r+

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/1002#-:-LAhb33x3vnSMHpJ4bvG:b-a2cmdb)*

micbou:
@zzbot retry

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/valloric/ycmd/1002#-:-LAk-6glCJsJ0H6RK48f:b-kc0oa4)*

PR #1010, opened 2018-04-23
[READY] Cache flags by file and client data

micbou:
We currently cache the compilation flags by file path but we should also cache them according to the `client_data` parameter given to the `FlagsForFile` function since this parameter may be used to change the list of flags. In other words, the `flags_for_file` variable should be indexed by the `( filename, client_data )` tuple. However, `client_data` is a dictionary and dictionaries are not hashable so they can't be used as part of dictionaries' keys. We need to introduce a new class `HashableDict` which is a dictionary wrapper implementing the `__hash__` method, and convert `client_data` to that class.

In addition, this PR simplifies the use of the `client_data` parameter in the `FlagsForFile` function by always passing the parameter as a dictionary. For instance, instead of having to write:
```python
def FlagsForFile( filename, **kwargs ):
  client_data = kwargs.get( 'client_data' )
  if client_data:
    return { 'flags': client_data.get( 'flags', [] ) }
  return { 'flags': [] }
```
one can now write:
```python
def FlagsForFile( filename, **kwargs ):
  return { 'flags': kwargs[ 'client_data' ].get( 'flags', [] ) }
```

The API docs are updated to mention the optional `extra_conf_data` field in a request.

Here's a demo with the Vim client (but this could work for any client supporting the `extra_conf_data` feature) that shows the benefit of these changes :

ycmd-cache-client-data

In that demo, the compilation flags are given by setting the `g:ycm_compiler_flags` option. Without these changes, the user would have to run the `ClearCompilationFlagCache` command or restart the server after modifying the `g:ycm_compiler_flags` option for the flags to be updated.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010)

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -285,99 +285,7 @@
                                             <div class="col-md-6 sw-request-model">
                             <div  class="panel panel-definition">
                                 <div class="panel-body">
-                                        
-                                                <section class="json-schema-properties">
-                                                    <dl>
-                                                            <dt data-property-name="line_num">
-                                                                <span class="json-property-name">line_num:</span>
-                                                                
-                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/LineNumber">LineNumber</a>
-                                                                </span>
-                                                                <span class="json-property-range" title="Value limits"></span>
-                                                                
-                                                                    <span class="json-property-required"></span>
-                                                            </dt>
-                                                            <dd>
-                                                                
-                                                                <div class="json-inner-schema">
-                                                                    
-                                                                </div>
-                                                            </dd>
-                                                            <dt data-property-name="column_num">
-                                                                <span class="json-property-name">column_num:</span>
-                                                                
-                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/ColumnNumber">ColumnNumber</a>
-                                                                </span>
-                                                                <span class="json-property-range" title="Value limits"></span>
-                                                                
-                                                                    <span class="json-property-required"></span>
-                                                            </dt>
-                                                            <dd>
-                                                                
-                                                                <div class="json-inner-schema">
-                                                                    
-                                                                </div>
-                                                            </dd>
-                                                            <dt data-property-name="filepath">
-                                                                <span class="json-property-name">filepath:</span>
-                                                                
-                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/FilePath">FilePath</a>
-                                                                </span>
-                                                                <span class="json-property-range" title="Value limits"></span>
-                                                                
-                                                                    <span class="json-property-required"></span>
-                                                            </dt>
-                                                            <dd>
-                                                                
-                                                                <div class="json-inner-schema">
-                                                                    
-                                                                </div>
-                                                            </dd>
-                                                            <dt data-property-name="file_data">
-                                                                <span class="json-property-name">file_data:</span>
-                                                                
-                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/FileDataMap">FileDataMap</a>
-                                                                </span>
-                                                                <span class="json-property-range" title="Value limits"></span>
-                                                                
-                                                                    <span class="json-property-required"></span>
-                                                            </dt>
-                                                            <dd>
-                                                                
-                                                                <div class="json-inner-schema">
-                                                                    
-                                                                </div>
-                                                            </dd>
-                                                            <dt data-property-name="completer_target">
-                                                                <span class="json-property-name">completer_target:</span>
-                                                                
-                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/CompleterTarget">CompleterTarget</a>
-                                                                </span>
-                                                                <span class="json-property-range" title="Value limits"></span>
-                                                                
-                                                            </dt>
-                                                            <dd>
-                                                                
-                                                                <div class="json-inner-schema">
-                                                                    
-                                                                </div>
-                                                            </dd>
-                                                            <dt data-property-name="working_dir">
-                                                                <span class="json-property-name">working_dir:</span>
-                                                                
-                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/WorkingDirectory">WorkingDirectory</a>
-                                                                </span>
-                                                                <span class="json-property-range" title="Value limits"></span>
-                                                                
-                                                            </dt>
-                                                            <dd>
-                                                                
-                                                                <div class="json-inner-schema">
-                                                                    
-                                                                </div>
-                                                            </dd>
-                                                    </dl>
-                                                </section>
+                                            <a class="json-schema-ref" href="#/definitions/SimpleRequest">SimpleRequest</a>
                                 </div>
                             </div></div>
                                         </div>
@@ -1016,6 +924,20 @@
                                                                 <span class="json-property-name">working_dir:</span>
                                                                 
                                                                 <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/WorkingDirectory">WorkingDirectory</a>
+                                                                </span>
+                                                                <span class="json-property-range" title="Value limits"></span>
+                                                                
+                                                            </dt>
+                                                            <dd>
+                                                                
+                                                                <div class="json-inner-schema">
+                                                                    
+                                                                </div>
+                                                            </dd>
+                                                            <dt data-property-name="extra_conf_data">
+                                                                <span class="json-property-name">extra_conf_data:</span>
+                                                                
+                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/ExtraConfData">ExtraConfData</a>
                                                                 </span>
                                                                 <span class="json-property-range" title="Value limits"></span>
                                                                 
@@ -1832,6 +1754,20 @@
                                                                 <span class="json-property-name">working_dir:</span>
                                                                 
                                                                 <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/WorkingDirectory">WorkingDirectory</a>
+                                                                </span>
+                                                                <span class="json-property-range" title="Value limits"></span>
+                                                                
+                                                            </dt>
+                                                            <dd>
+                                                                
+                                                                <div class="json-inner-schema">
+                                                                    
+                                                                </div>
+                                                            </dd>
+                                                            <dt data-property-name="extra_conf_data">
+                                                                <span class="json-property-name">extra_conf_data:</span>
+                                                                
+                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/ExtraConfData">ExtraConfData</a>
                                                                 </span>
                                                                 <span class="json-property-range" title="Value limits"></span>
                                                                 
@@ -2883,6 +2819,27 @@
                                     </section>
                     </div>
                 </div>        
+                <div id="definition-ExtraConfData" class="panel panel-definition">
+                        <div class="panel-heading">
+                                <h3 class="panel-title"><a name="/definitions/ExtraConfData"></a>ExtraConfData:
+                                    <span class="json-property-type">
+                <span class="json-property-type">object</span>
+                <span class="json-property-range" title="Value limits"></span>
+                
+                </span>
+                                </h3>
+                        </div>
+                    <div class="panel-body">
+                                <section class="json-schema-description">
+                                    <p>A dictionary passed to the <code>FlagsForFile</code> function of the
+                            <code>.ycm_extra_conf.py</code> file as the <code>client_data</code> parameter. This optional
+                            field should be used to give users a way to customize the
+                            <code>.ycm_extra_conf.py</code> file with their own set of options.</p>
+                            
+                                </section>
+                            
+                    </div>
+                </div>        
                 <div id="definition-FileData" class="panel panel-definition">
                         <div class="panel-heading">
                                 <h3 class="panel-title"><a name="/definitions/FileData"></a>FileData:
@@ -3675,6 +3632,20 @@
                                                     <span class="json-property-name">working_dir:</span>
                                                     
                                                     <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/WorkingDirectory">WorkingDirectory</a>
+                                                    </span>
+                                                    <span class="json-property-range" title="Value limits"></span>
+                                                    
+                                                </dt>
+                                                <dd>
+                                                    
+                                                    <div class="json-inner-schema">
+                                                        
+                                                    </div>
+                                                </dd>
+                                                <dt data-property-name="extra_conf_data">
+                                                    <span class="json-property-name">extra_conf_data:</span>
+                                                    
+                                                    <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/ExtraConfData">ExtraConfData</a>
                                                     </span>
                                                     <span class="json-property-range" title="Value limits"></span>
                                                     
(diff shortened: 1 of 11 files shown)


bstaletic:
> In addition, this PR simplifies the use of the client_data

This is my python ignorance speaking, but how (or rather why) do the two extra confs in your example differ?

---

Reviewed 9 of 11 files at r1, 1 of 1 files at r2, 4 of 4 files at r3.

---

*[ycmd/completers/cpp/clang_completer.py, line 458 at r2](https://reviewable.io/reviews/valloric/ycmd/1010#-LAzQ_HK7SkX-L02PQeu:-LAzQ_HK7SkX-L02PQev:b-gr0etn) ([raw file](https://github.com/valloric/ycmd/blob/4d711ed9b0bd554598a7cf3ff71f9e0c2ca8247d/ycmd/completers/cpp/clang_completer.py#L458)):*
> ```Python
>                filename )
> 
>     client_data = request_data[ 'extra_conf_data' ]
> ```

Doesn't this potentially raise `TypeError`?

---

*[ycmd/completers/cpp/flags.py, line 131 at r3](https://reviewable.io/reviews/valloric/ycmd/1010#-LBjy4pC0p3qqIsiOgBi:-LBjy4pC0p3qqIsiOgBj:b-303gfu) ([raw file](https://github.com/valloric/ycmd/blob/1d8419bdc70bfddaa1b8e7c9a3370d0530c2a0f0/ycmd/completers/cpp/flags.py#L131)):*
> ```Python
>     # 1-python-statement synchronisation for "free" (via the GIL)
>     try:
>       return self.flags_for_file[ filename, client_data ]
> ```

Am I reading this right? Is the ket for the `flags_for_file` dictionary a `( filename, client_data )` tuple?

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010#-:-LBjzdWh18taV9TcjwXO:b-7pkd04)*

micbou:
With the proposed changes, the `client_data` parameter is always passed to the `FlagsForFile` function as a dictionary so there is no need to check that the parameter exists and that it's not `None`. That's why there is no need to do
```python
client_data = kwargs.get( 'client_data' )
if client_data:
  ...
```
in the second extra conf.

---

Reviewed 6 of 11 files at r1, 1 of 1 files at r2, 4 of 4 files at r3.

---

*[ycmd/completers/cpp/clang_completer.py, line 458 at r2](https://reviewable.io/reviews/valloric/ycmd/1010#-LAzQ_HK7SkX-L02PQeu:-LBk0olhAZ0W8Divsh3L:bccb31) ([raw file](https://github.com/valloric/ycmd/blob/4d711ed9b0bd554598a7cf3ff71f9e0c2ca8247d/ycmd/completers/cpp/clang_completer.py#L458)):*

Doesn't this potentially raise `TypeError`?
</blockquote></details>

No because `extra_conf_data` (which is in fact `client_data`) is defined in the same way as `force_semantic` and other fields of the `RequestWrap` object.

---

*[ycmd/completers/cpp/flags.py, line 131 at r3](https://reviewable.io/reviews/valloric/ycmd/1010#-LBjy4pC0p3qqIsiOgBi:-LBk0yex7z9aErCZOzOp:brpkxiu) ([raw file](https://github.com/valloric/ycmd/blob/1d8419bdc70bfddaa1b8e7c9a3370d0530c2a0f0/ycmd/completers/cpp/flags.py#L131)):*

Am I reading this right? Is the ket for the `flags_for_file` dictionary a `( filename, client_data )` tuple?
</blockquote></details>

Yes, `flags_for_file` is indexed by the `( filename, client_data )` tuple. This makes sense because the `FlagsForFile` function depends on these two parameters: `filename` and `client_data`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010#-:-LBk-pel3ww0s6qgFXwR:by3u1nu)*

bstaletic:
Thanks for clearing it up for me. :lgtm: 
Codecov has been rather annoying lately.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010#-:-LBkA1ZQAr-cZ0QkVG-9:b-k7u6hs)*

puremourning:
:lgtm: with a few small nits

---

Review status: 10 of 11 files reviewed at latest revision, all discussions resolved.

---

*[docs/openapi.yml, line 152 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJ6n-_Ab_DGG16kFpG:-LCJ6n-_Ab_DGG16kFpH:bf2e88k) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/docs/openapi.yml#L152)):*
Is this a valid definition? 

According to the JSON schema (if I read it correctly), `type`:

> 6.1.1.  type

>   The value of this keyword MUST be either a string or an array.  If it
   is an array, elements of the array MUST be strings and MUST be
   unique.

>   String values MUST be one of the six primitive types ("null",
   "boolean", "object", "array", "number", or "string"), or "integer"
   which matches any number with a zero fractional part.

 >  An instance validates if and only if the instance is in any of the
   sets listed for this keyword.

i.e. I think we should be using `object` here.

---

*[ycmd/request_wrap.py, line 97 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJ8wliCA7jt3cL2NTP:-LCJ8wliCA7jt3cL2NTQ:b8dk20l) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/ycmd/request_wrap.py#L97)):*
> ```Python
>       'lines': ( self._CurrentLines, None ),
> 
>       'extra_conf_data': ( self._GetExtraConfData, None )
> ```

we could add a trailing comma here

---

*[ycmd/utils.py, line 503 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJAX7-9tbsbUZXfcw7:-LCJAX7-9tbsbUZXfcw8:b-bucgj2) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/ycmd/utils.py#L503)):*
> ```Python
> 
> 
> class HashableDict( collections.Mapping ):
> ```

curious why we can't/don't subclass `dict` here ? I suspect there is some engineering reason why but it wasn't immediately obvious to me :)

---

*[ycmd/completers/cpp/flags.py, line 95 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJB2f800_MF3qnSoZj:-LCJB2f800_MF3qnSoZk:b3uxn5o) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/ycmd/completers/cpp/flags.py#L95)):*
> ```Python
>   def __init__( self ):
>     # It's caches all the way down...
>     self.flags_for_file = {}
> ```

I think we should document here that the key of this mapping is a tuple of filename and client data ?

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010#-:-LCJBg08EUpLyS16lmXD:b-5wa6l7)*

puremourning:
Reviewed 5 of 11 files at r1, 1 of 1 files at r2, 4 of 4 files at r3, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010)*

bstaletic:
Reviewed 1 of 1 files at r4.

---

*[ycmd/request_wrap.py, line 97 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJ8wliCA7jt3cL2NTP:-LCJBw8F4lWxZVNc5WQt:b-de4ltj) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/ycmd/request_wrap.py#L97)):*

we could add a trailing comma here
</blockquote></details>

I personally dislike trailing commas, but whatever.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010)*

puremourning:

---

*[ycmd/request_wrap.py, line 97 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJ8wliCA7jt3cL2NTP:-LCJDuYS1G_F4ueaUxM9:b-ubvidu) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/ycmd/request_wrap.py#L97)):*

I personally dislike trailing commas, but whatever.
</blockquote></details>

I find it useful for:

- reordering
- adding new elements
- etc.

i.e. each entry has the same form, rather than the last one having a special form

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010)*

bstaletic:

---

*[ycmd/request_wrap.py, line 97 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJ8wliCA7jt3cL2NTP:-LCJGeis306rpZxjP_yW:b-7rnvb7) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/ycmd/request_wrap.py#L97)):*

I find it useful for:

- reordering
- adding new elements
- etc.

i.e. each entry has the same form, rather than the last one having a special form
</blockquote></details>

Yes, I know the up sides. Yet it looks ugly to me and like "there's something missing here".
But won't mind it if @micbou decides to add the trailing comma.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010)*

micbou:
Reviewed 1 of 1 files at r4, 3 of 3 files at r5, 2 of 2 files at r6.

---

*[docs/openapi.yml, line 152 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJ6n-_Ab_DGG16kFpG:-LCJKc_21WV8lPc1V0PR:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/docs/openapi.yml#L152)):*

Is this a valid definition? 

According to the JSON schema (if I read it correctly), `type`:

> 6.1.1.  type

>   The value of this keyword MUST be either a string or an array.  If it
   is an array, elements of the array MUST be strings and MUST be
   unique.

>   String values MUST be one of the six primitive types ("null",
   "boolean", "object", "array", "number", or "string"), or "integer"
   which matches any number with a zero fractional part.

 >  An instance validates if and only if the instance is in any of the
   sets listed for this keyword.

i.e. I think we should be using `object` here.

</blockquote></details>

Done.

---

*[ycmd/request_wrap.py, line 97 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJ8wliCA7jt3cL2NTP:-LCJP3w1AfddWsgZuxCL:b-kskff0) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/ycmd/request_wrap.py#L97)):*

Yes, I know the up sides. Yet it looks ugly to me and like "there's something missing here".
But won't mind it if @micbou decides to add the trailing comma.
</blockquote></details>

I added it then changed my mind. I don't like it either.

---

*[ycmd/utils.py, line 503 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJAX7-9tbsbUZXfcw7:-LCJIk1d3K49g9rKZMUc:b-26mx8u) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/ycmd/utils.py#L503)):*

curious why we can't/don't subclass `dict` here ? I suspect there is some engineering reason why but it wasn't immediately obvious to me :)
</blockquote></details>

That's a good question. I didn't really think about it since I based the implementation on [frozendict](https://github.com/slezica/python-frozendict/blob/b27053e4d11f5891319fd29eda561c130ba3112a/frozendict/__init__.py#L3). To be honest, I am not sure how `__repr__` could be implemented if `HashableDict` were a subclass of `dict`.

---

*[ycmd/completers/cpp/flags.py, line 95 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJB2f800_MF3qnSoZj:-LCJPOF9EtIY8X9xV29H:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/ycmd/completers/cpp/flags.py#L95)):*

I think we should document here that the key of this mapping is a tuple of filename and client data ?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010)*

puremourning:

---

*[ycmd/request_wrap.py, line 97 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJ8wliCA7jt3cL2NTP:-LCK0hUiE4yBiHG36iey:b2qvtd5) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/ycmd/request_wrap.py#L97)):*

I added it then changed my mind. I don't like it either.
</blockquote></details>

It seems even PEP8 isn't prescriptive, though in most if its examples the trailing comma is always added: https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas

---

*[ycmd/utils.py, line 503 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJAX7-9tbsbUZXfcw7:-LCK-Z4O8bM7HevdMWDd:b25d1qw) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/ycmd/utils.py#L503)):*

That's a good question. I didn't really think about it since I based the implementation on [frozendict](https://github.com/slezica/python-frozendict/blob/b27053e4d11f5891319fd29eda561c130ba3112a/frozendict/__init__.py#L3). To be honest, I am not sure how `__repr__` could be implemented if `HashableDict` were a subclass of `dict`.
</blockquote></details>

Could we get the `__repr__` of `dict`? Like 

```python

def __repr__( self ):
  return '<HashableDict {0}>'.format( super( HashableDict, self ).__repr__() )
```

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010)*

puremourning:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010#-:-LCK0rUtC00oAo2w3KLR:bnfp4nl)*

puremourning:
@homu r+

---

---

*[ycmd/request_wrap.py, line 97 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJ8wliCA7jt3cL2NTP:-LCKaZcg0NZL1X7c8vtF:bz35y44) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/ycmd/request_wrap.py#L97)):*

We could consider adding [flake8-commas](https://github.com/PyCQA/flake8-commas) to `test_requirements.txt`. It's going to be annoying to update the whole codebase though. There are a lot of places where we don't put a trailing comma (mostly in the tests).
</blockquote></details>

not in this PR :)

---

*[ycmd/utils.py, line 503 at r4](https://reviewable.io/reviews/valloric/ycmd/1010#-LCJAX7-9tbsbUZXfcw7:-LCKaJbQ2gk1_B6or5jz:bdcv6l8) ([raw file](https://github.com/valloric/ycmd/blob/5728eec5eee7b125b30fff8d3bdec6d21bc61bf5/ycmd/utils.py#L503)):*

Yes it's working. However, I know why we don't inherit from `dict`: we want an immutable dictionary like `frozendict`.
</blockquote></details>

Good point.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010#-:-LCKakNb736AoZwwz9Jd:b-iwbgq9)*

puremourning:
again! @zzbot r+

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1010#-:-LCKarL_-lEb1-AbpANK:b2n96mw)*

PR #1015, opened 2018-04-26
[READY] Force MSVC to treat source files as UTF-8 encoded

micbou:
Compilation fails with a lot of these errors:
```
error C3688: invalid literal suffix '唷'; literal operator or literal operator template 'operator ""唷' not found
```
when the system local is Chinese on Windows. This is fixed by forcing MSVC to use UTF-8.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1015)


micbou:
@zzbot retry

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1015#-:-LB3_E8G0xb7_lr0Sogv:b-kc0oa4)*

PR #1020, opened 2018-04-30
[READY] Use sysconfig instead of distutils.sysconfig

micbou:
[The `distutils.sysconfig` module](https://docs.python.org/3/distutils/apiref.html?highlight=distutils%20sysconfig#module-distutils.sysconfig) has been moved to a separate package on Debian-based distributions for Python 3. Use [`sysconfig`](https://docs.python.org/3/library/sysconfig.html) instead as suggested in [this Debian bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=893924).

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1020)

### DIFF
--- a/build.py
+++ b/build.py
@@ -7,21 +7,21 @@ from __future__ import print_function
 from __future__ import division
 from __future__ import absolute_import
 
-from distutils import sysconfig
 from shutil import rmtree
 from tempfile import mkdtemp
 import errno
+import hashlib
 import multiprocessing
 import os
 import os.path as p
 import platform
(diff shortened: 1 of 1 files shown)


bstaletic:
Considering that `sysconfig` stand alone module works on python 2.7+ and that the tests pass on all operating systems, this is :lgtm:.

Valloric:
:lgtm:

Thanks for the PR!

@zzbot r+

PR #1022, opened 2018-05-05
[READY] Check for broken Anaconda build flags

bstaletic:
As discussed in gitter, we figured out what makes ycmd broken when using Anaconda python.

Here's the result of this PR:
```
bstaletic@gentoo ycmd  (git)-[anaconda_check]-% ./build.py --clang-completer
ERROR: Anaconda's libraries are too old to support libclang.
       If you want to build ycm_core.so with Anaconda, specify --system-libclang.
       Alternatively, build ycm_core.so without Anaconda.
```

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1022)

### DIFF
--- a/build.py
+++ b/build.py
@@ -387,6 +387,15 @@ def ParseArguments():
        not args.all_completers ):
     sys.exit( 'ERROR: you can\'t pass --system-libclang without also passing '
               '--clang-completer or --all as well.' )
+
+  if ( 'Anaconda' in sys.version and
+     args.clang_completer and
+     not args.system_libclang ):
+    sys.exit( 'ERROR: Anaconda\'s libraries are too'
+              ' old to support libclang.\n'
+              '       If you want to build ycm_core.so with Anaconda, '
+              'specify --system-libclang.\n'
+              '       Alternatively, build ycm_core.so without Anaconda.')
   return args
 
 


puremourning:

---

*[build.py, line 397 at r1](https://reviewable.io/reviews/valloric/ycmd/1022#-LBlZB8z5UgXJUQ-5ULE:-LBlZB8z5UgXJUQ-5ULF:b-wi9fl8) ([raw file](https://github.com/valloric/ycmd/blob/7419c5d58b193457a2e6dd2c65bddd0b3be1d175/build.py#L397)):*
> ```Python
>               ' old to support libclang.\n'
>               '       If you want to build ycm_core.so with Anaconda, '
>               'specify --system-libclang.\n'
> ```

What makes us think system libclang would even work?

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1022)*

puremourning:

---

*[build.py, line 398 at r1](https://reviewable.io/reviews/valloric/ycmd/1022#-LBlZJi--x2vkRTik-eu:-LBlZJi--x2vkRTik-ev:b-xuhtpp) ([raw file](https://github.com/valloric/ycmd/blob/7419c5d58b193457a2e6dd2c65bddd0b3be1d175/build.py#L398)):*
> ```Python
>               '       If you want to build ycm_core.so with Anaconda, '
>               'specify --system-libclang.\n'
>               '       Alternatively, build ycm_core.so without Anaconda.')
> ```

Users don’t know what ycm_core.so is. Hey know only about install.py and/or build.py

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1022)*

bstaletic:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[build.py, line 397 at r1](https://reviewable.io/reviews/valloric/ycmd/1022#-LBlZB8z5UgXJUQ-5ULE:-LBlaktM1dzK2no2ebw-:b-2r1v8r) ([raw file](https://github.com/valloric/ycmd/blob/7419c5d58b193457a2e6dd2c65bddd0b3be1d175/build.py#L397)):*

What makes us think system libclang would even work?
</blockquote></details>

It does work on my system, so it's the best advice we can give our users.

---

*[build.py, line 398 at r1](https://reviewable.io/reviews/valloric/ycmd/1022#-LBlZJi--x2vkRTik-eu:-LBlb2Wg48c2od1SUa1E:b2rsa4g) ([raw file](https://github.com/valloric/ycmd/blob/7419c5d58b193457a2e6dd2c65bddd0b3be1d175/build.py#L398)):*

Users don’t know what ycm_core.so is. Hey know only about install.py and/or build.py
</blockquote></details>

Is `build YCM` or `build YouCompleteMe` better? Or do you have an even better idea?

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1022)*

puremourning:
I would say we just just flat out refuse to build with anaconda python. It just causes too many issues and isn’t supported even if it works. We also tend to avoid recommending system libclang.

bstaletic:
I'd argue that, since system libclang is already clearly unsupported, it should be clear that anaconda is not supported either.
On the other hand, many are using anaconda, so I feel like we should do our best to make it work. Considering that it's actually a matter of linking against an old glibc, anaconda with system's libclaang seems like a pretty safe advice.

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1022#-:-LBllcmC6VwVwkA3vhjp:bkebb4m)*

bstaletic:
Considering that the latest anaconda works, even with the bundled libclang, provided the user has a libtinfo.so.5, I'll close the PR.

Instead, let's just make an FAQ saying that anaconda isn't officially supported because it sometimes breaks.

PR #1024, opened 2018-05-08
Support java source from jar package

brookhong:
eclipse.jdt.ls returns uri for class from jar package in scheme like 'jdt://contents/.*\.class.*', please see https://github.com/eclipse/eclipse.jdt.ls/blob/ae0081c28744dcdda117e97bf14b1bec9ca7f911/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/managers/ContentProviderManager.java#L40

For example,

    jdt://contents/spring-boot-autoconfigure-1.5.8.RELEASE.jar/org.springframework.boot.autoconfigure/SpringBootApplication.class?\u003dsurfingkeys/%5C/Users%5C/brook%5C/.m2%5C/repository%5C/org%5C/springframework%5C/boot%5C/spring-boot-autoconfigure%5C/1.5.8.RELEASE%5C/spring-boot-autoconfigure-1.5.8.RELEASE.jar%3Corg.springframework.boot.autoconfigure(SpringBootApplication.class

was returned for class SpringBootApplication.

This change is to parse java source path from source jar package, so that source code from jar package could be viewed.

Test case:

    ./run_tests.py ycmd/tests/utils_test.py:ParseSourcefileFromJdtUri_test --skip-build

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1024)

### DIFF
--- a/ycmd/completers/completer.py
+++ b/ycmd/completers/completer.py
@@ -252,6 +252,10 @@ class Completer( with_metaclass( abc.ABCMeta, object ) ):
       # We don't want expose this subcommand because it is not really needed
       # for the user but it is useful in tests for tearing down the server
       subcommands.remove( 'StopServer' )
+      # Don't expose ClassFileContents, user need not run the command manually
+      # client of ycmd will call this command whenever a uri like
+      # jdt://contents/*/.class is opened
+      subcommands.remove( 'ClassFileContents' )
     except ValueError:
       pass
     return subcommands
--- a/ycmd/completers/java/java_completer.py
+++ b/ycmd/completers/java/java_completer.py
@@ -254,6 +254,9 @@ class JavaCompleter( language_server_completer.LanguageServerCompleter ):
       'OrganizeImports': (
         lambda self, request_data, args: self.OrganizeImports( request_data )
       ),
+      'ClassFileContents': (
+        lambda self, request_data, args: self.ClassFileContents( request_data )
+      ),
     }
 
 
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -1030,11 +1030,7 @@ class LanguageServerCompleter( Completer ):
       params = notification[ 'params' ]
       uri = params[ 'uri' ]
 
-      try:
-        filepath = lsp.UriToFilePath( uri )
-      except lsp.InvalidUriException:
-        _logger.exception( 'Ignoring diagnostics for unrecognized URI' )
-        return None
+      filepath = lsp.UriToFilePath( uri )
 
       with self._server_info_mutex:
         if filepath in self._server_file_state:
@@ -1327,6 +1323,19 @@ class LanguageServerCompleter( Completer ):
         raise RuntimeError( 'Cannot jump to location' )
     else:
       raise RuntimeError( 'Cannot jump to location' )
+
+
+  def ClassFileContents( self, request_data ):
+    """Issues the classFileContents request and returns the result as
+    response."""
+    if not self.ServerIsReady():
+      raise RuntimeError( 'Server is initializing. Please wait.' )
+
+    request_id = self.GetConnection().NextRequestId()
+    return self.GetConnection().GetResponse(
+      request_id,
+      lsp.ClassFileContents( request_id, request_data ),
+      REQUEST_TIMEOUT_COMMAND )
 
 
   def GoToReferences( self, request_data ):
@@ -1738,12 +1747,12 @@ def _LocationListToGoTo( request_data, response ):
 def _PositionToLocationAndDescription( request_data, position ):
   """Convert a LSP position to a ycmd location."""
   try:
-    filename = lsp.UriToFilePath( position[ 'uri' ] )
-    file_contents = GetFileLines( request_data, filename )
-  except lsp.InvalidUriException:
-    _logger.debug( "Invalid URI, file contents not available in GoTo" )
-    filename = ''
-    file_contents = []
+    if responses.IsJdtContentUri( position[ 'uri' ] ):
+      filename = position[ 'uri' ]
+      file_contents = []
+    else:
+      filename = lsp.UriToFilePath( position[ 'uri' ] )
+      file_contents = GetFileLines( request_data, filename )
   except IOError:
     # It's possible to receive positions for files which no longer exist (due to
     # race condition). UriToFilePath doesn't throw IOError, so we can assume
@@ -1793,12 +1802,7 @@ def _BuildRange( contents, filename, r ):
 
 def _BuildDiagnostic( contents, uri, diag ):
   """Return a ycmd diagnostic from a LSP diagnostic."""
-  try:
-    filename = lsp.UriToFilePath( uri )
-  except lsp.InvalidUriException:
-    _logger.debug( 'Invalid URI received for diagnostic' )
-    filename = ''
-
+  filename = lsp.UriToFilePath( uri )
   r = _BuildRange( contents, filename, diag[ 'range' ] )
 
   return responses.BuildDiagnosticData( responses.Diagnostic(
@@ -1811,12 +1815,7 @@ def _BuildDiagnostic( contents, uri, diag ):
 
 def TextEditToChunks( request_data, uri, text_edit ):
   """Returns a list of FixItChunks from a LSP textEdit."""
-  try:
-    filepath = lsp.UriToFilePath( uri )
-  except lsp.InvalidUriException:
-    _logger.debug( 'Invalid filepath received in TextEdit' )
-    filepath = ''
-
+  filepath = lsp.UriToFilePath( uri )
   contents = GetFileLines( request_data, filepath )
   return [
     responses.FixItChunk( change[ 'newText' ],
--- a/ycmd/completers/language_server/language_server_protocol.py
+++ b/ycmd/completers/language_server/language_server_protocol.py
@@ -33,6 +33,8 @@ from ycmd.utils import ( ByteOffsetToCodepointOffset,
                          url2pathname,
                          urljoin )
 
+from ycmd.responses import IsJdtContentUri
+
 
 INSERT_TEXT_FORMAT = [
   None, # 1-based
@@ -69,12 +71,6 @@ SEVERITY = [
   'Information',
   'Hint',
 ]
-
-
-class InvalidUriException( Exception ):
-  """Raised when trying to convert a server URI to a file path but the scheme
-  was not supported. Only the file: scheme is supported"""
-  pass
 
 
 class ServerFileStateStore( dict ):
@@ -193,6 +189,7 @@ def Initialize( request_id, project_directory ):
     'rootPath': project_directory,
     'rootUri': FilePathToUri( project_directory ),
     'initializationOptions': {
+      'classFileContentsSupport': True
       # We don't currently support any server-specific options.
     },
     'capabilities': {
@@ -277,6 +274,12 @@ def Definition( request_id, request_data ):
   return BuildRequest( request_id,
                        'textDocument/definition',
                        BuildTextDocumentPositionParams( request_data ) )
+
+
+def ClassFileContents( request_id, request_data ):
+  return BuildRequest( request_id, 'java/classFileContents', {
+    'uri': request_data[ 'filepath' ]
+  } )
 
 
 def CodeAction( request_id, request_data, best_match_range, diagnostics ):
@@ -391,12 +394,14 @@ def ExecuteCommand( request_id, command, arguments ):
 
 
 def FilePathToUri( file_name ):
+  if IsJdtContentUri( file_name ):
+    return file_name
   return urljoin( 'file:', pathname2url( file_name ) )
 
 
 def UriToFilePath( uri ):
-  if uri[ : 5 ] != "file:":
-    raise InvalidUriException( uri )
+  if IsJdtContentUri( uri ):
+    return uri
 
   return os.path.abspath( url2pathname( uri[ 5 : ] ) )
 
--- a/ycmd/responses.py
+++ b/ycmd/responses.py
@@ -133,13 +133,21 @@ def BuildCompletionResponse( completion_datas,
   }
 
 
+def IsJdtContentUri( uri ):
+  return uri[ : 6 ] == "jdt://"
+
+
 # location.column_number_ is a byte offset
 def BuildLocationData( location ):
+  filename = ''
+  if location.filename_:
+    filename = location.filename_
+    if not IsJdtContentUri( filename ):
+      filename = os.path.normpath( filename )
   return {
     'line_num': location.line_number_,
     'column_num': location.column_number_,
-    'filepath': ( os.path.normpath( location.filename_ )
-                  if location.filename_ else '' ),
+    'filepath': filename,
   }
 
 
@@ -209,7 +217,10 @@ class Location( object ):
     self.line_number_ = line
     self.column_number_ = column
     if filename:
-      self.filename_ = os.path.realpath( filename )
+      if IsJdtContentUri( filename ):
+        self.filename_ = filename
+      else:
+        self.filename_ = os.path.realpath( filename )
     else:
       # When the filename passed (e.g. by a server) can't be recognized or
       # parsed, we send an empty filename. This at least allows the client to
--- a/ycmd/tests/java/diagnostics_test.py
+++ b/ycmd/tests/java/diagnostics_test.py
@@ -50,7 +50,6 @@ from ycmd.completers import completer
 
 from pprint import pformat
 from mock import patch
-from ycmd.completers.language_server import language_server_protocol as lsp
 from ycmd import handlers
 
 
@@ -456,46 +455,6 @@ public class Test {
 
 
 @IsolatedYcmd
-@patch(
-  'ycmd.completers.language_server.language_server_protocol.UriToFilePath',
-  side_effect = lsp.InvalidUriException )
-def FileReadyToParse_Diagnostics_InvalidURI_test( app, uri_to_filepath, *args ):
-  StartJavaCompleterServerInDirectory( app,
-                                       PathToTestFile( DEFAULT_PROJECT_DIR ) )
-
-  filepath = TestFactory
-  contents = ReadFile( filepath )
-
-  # It can take a while for the diagnostics to be ready
-  expiration = time.time() + 10
-  while True:
-    try:
-      results = _WaitForDiagnosticsToBeReady( app, filepath, contents )
-      print( 'Completer response: {0}'.format(
-        json.dumps( results, indent=2 ) ) )
-
-      uri_to_filepath.assert_called()
-
-      assert_that( results, has_item(
-        has_entries( {
-          'kind': 'WARNING',
-          'text': 'The value of the field TestFactory.Bar.testString is not '
-                  'used',
-          'location': LocationMatcher( '', 15, 19 ),
-          'location_extent': RangeMatcher( '', ( 15, 19 ), ( 15, 29 ) ),
-          'ranges': contains( RangeMatcher( '', ( 15, 19 ), ( 15, 29 ) ) ),
-          'fixit_available': False
-        } ),
-      ) )
-
-      return
-    except AssertionError:
-      if time.time() > expiration:
-        raise
-      time.sleep( 0.25 )
-
-
-@IsolatedYcmd
 def FileReadyToParse_ServerNotReady_test( app ):
   filepath = TestFactory
   contents = ReadFile( filepath )
@@ -680,38 +639,6 @@ def OnBufferUnload_ServerNotRunning_test( app ):
                                filetype = 'java' )
     result = app.post_json( '/event_notification', event_data ).json
     assert_that( result, equal_to( {} ) )
-
-
-@IsolatedYcmd
-def PollForMessages_InvalidUri_test( app, *args ):
-  StartJavaCompleterServerInDirectory(
-    app,
-    PathToTestFile( 'simple_eclipse_project' ) )
-
-  filepath = TestFactory
-  contents = ReadFile( filepath )
-
-  with patch(
-    'ycmd.completers.language_server.language_server_protocol.UriToFilePath',
-    side_effect = lsp.InvalidUriException ):
-
-    for tries in range( 0, 5 ):
-      response = app.post_json( '/receive_messages',
-                                BuildRequest(
-                                  filetype = 'java',
-                                  filepath = filepath,
-                                  contents = contents ) ).json
-      if response is True:
-        break
-      elif response is False:
-        raise AssertionError( 'Message poll was aborted unexpectedly' )
-      elif 'diagnostics' in response:
-        raise AssertionError( 'Did not expect diagnostics when file paths '
-                              'are invalid' )
-
-      time.sleep( 0.5 )
-
-  assert_that( response, equal_to( True ) )
 
 
 @IsolatedYcmd
--- a/ycmd/tests/java/subcommands_test.py
+++ b/ycmd/tests/java/subcommands_test.py
@@ -35,6 +35,7 @@ from pprint import pformat
 import requests
 
 from ycmd.utils import ReadFile
+from ycmd.responses import IsJdtContentUri
 from ycmd.completers.java.java_completer import NO_DOCUMENTATION_MESSAGE
 from ycmd.tests.java import ( DEFAULT_PROJECT_DIR,
                               PathToTestFile,
@@ -45,7 +46,6 @@ from ycmd.tests.test_utils import ( BuildRequest,
                                     LocationMatcher,
                                     WithRetry )
 from mock import patch
-from ycmd.completers.language_server import language_server_protocol as lsp
 from ycmd import handlers
 from ycmd.completers.language_server.language_server_completer import (
   ResponseTimeoutException,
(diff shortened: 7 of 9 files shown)


brookhong:
ycm

bstaletic:
Thanks for the PR. It would be great if we could get this to work.

We are aware that jdt.ls can return a path to a `*.class` file. Since it is basically a binary file we decided to just drop responses returning those URIs. So, if we can get this to actually work, more changes are required.

I can not try this right now, but does this require any changes on the client side?

Appveyor says the [`ParseSourcefileFromJdtUri_test`](https://ci.appveyor.com/project/Valloric/ycmd/build/3648) failed on Windows.

---

Reviewed 3 of 5 files at r1.
Review status: 3 of 5 files reviewed at latest revision, all discussions resolved.

---

*[ycmd/utils.py, line 76 at r1](https://reviewable.io/reviews/valloric/ycmd/1024#-LBzI63A0-CR-_oV0D78:-LBzI63A0-CR-_oV0D79:b94mf9z) ([raw file](https://github.com/valloric/ycmd/blob/f372543bb6f8828f334a5567778de94473b8a536/ycmd/utils.py#L76)):*
> ```Python
>     import zipfile
>     paths = filepath[8:].split("::")
>     archive = zipfile.ZipFile(paths[0], 'r')
> ```

Use spaces inside `( parens )` and `[ brackets ]`.

---

*[ycmd/completers/language_server/language_server_protocol.py, line 403 at r1](https://reviewable.io/reviews/valloric/ycmd/1024#-LBzDyt4EQvAWb7cbox2:-LBzDyt6EvpBx6Y61D3G:b6xsrsj) ([raw file](https://github.com/valloric/ycmd/blob/f372543bb6f8828f334a5567778de94473b8a536/ycmd/completers/language_server/language_server_protocol.py#L403)):*
> ```Python
>     return filepath
>   if uri [ : 5 ] != "file:":
>     raise InvalidUriException( uri )
> ```

Like I mentioned we used to "drop" `jdt://*.class` URIs. This was done here, by raising an `InvalidUriException`.
If we can get `jdt://` to work, we don't need the exception any more, so it should be dropped (along with the tests for the same).

---

*[ycmd/tests/utils_test.py, line 589 at r1](https://reviewable.io/reviews/valloric/ycmd/1024#-LBzFXQO5bSJjI5gxhT0:-LBzFXQO5bSJjI5gxhT1:b-w2yrfx) ([raw file](https://github.com/valloric/ycmd/blob/f372543bb6f8828f334a5567778de94473b8a536/ycmd/tests/utils_test.py#L589)):*
> ```Python
> def ParseSourcefileFromJdtUri_test():
>   lcd = utils.GetCurrentDirectory()
>   jarPath = lcd + "/ycmd/tests/testdata/java/spring-boot-autoconfigure-1.5.8" \
> ```

The ycmd style is to use parens instead of back slashes for multiline statements.

Instead of `lcd + "/ycmd/tests..."` you should be using `PathToTestFile()`, like this:
`jarPath = PathToTestFile( 'java', 'spring-boot-autoconfigure-1.5.8.RELEASE.jar...')`
Note that directories should not be passed all together, instead path joining should be left up to Python. That way Windows gets `\\` separators and *nix gets `/` separators.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1024#-:-LBzChpd2p0FGAbBXurF:b20f32o)*

brookhong:
Thanks for review.

The PR is just trying to find *.sources.jar synced by maven(gradle?), then locate the source file, so there is no client change.

    $ tree ~/.m2/repository/org/springframework/boot/spring-boot-autoconfigure/1.5.8.RELEASE/
    /Users/brook/.m2/repository/org/springframework/boot/spring-boot-autoconfigure/1.5.8.RELEASE/
    ├── _remote.repositories
    ├── m2e-lastUpdated.properties
    ├── spring-boot-autoconfigure-1.5.8.RELEASE-sources.jar
    ├── spring-boot-autoconfigure-1.5.8.RELEASE-sources.jar.sha1
    ├── spring-boot-autoconfigure-1.5.8.RELEASE.jar
    ├── spring-boot-autoconfigure-1.5.8.RELEASE.jar.sha1
    ├── spring-boot-autoconfigure-1.5.8.RELEASE.pom
    └── spring-boot-autoconfigure-1.5.8.RELEASE.pom.sha1

In case of there is no *.sources.jar fetched, we still need `InvalidUriException`. The others are fixed.

Also tested and verified in windows.

bstaletic:
@brookhong 
We should have more tests for this. `UriToFilePath` is tested in `ycmd/tests/language_server/language_server_protocol_test.py`. Even more importantly, we need tests that confirm that `GoTo` commands work with `jdt://contents/` URIs.

Also, please use [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1024) for PR reviews.

@tsmaeder Could you provide a link to the documentation of it? That could greatly simplify this PR.

---

Reviewed 4 of 4 files at r2.

---

*[ycmd/utils.py, line 76 at r1](https://reviewable.io/reviews/valloric/ycmd/1024#-LBzI63A0-CR-_oV0D78:-LC-1ubj2Sak5qA_Yba4:btoj4qj) ([raw file](https://github.com/valloric/ycmd/blob/f372543bb6f8828f334a5567778de94473b8a536/ycmd/utils.py#L76)):*

Use spaces inside `( parens )` and `[ brackets ]`.
</blockquote></details>

You missed adding spaces in this block.

---

*[ycmd/utils.py, line 517 at r2](https://reviewable.io/reviews/valloric/ycmd/1024#-LC-359SApaCuRUZYrHY:-LC-359SApaCuRUZYrHZ:bpocpvp) ([raw file](https://github.com/valloric/ycmd/blob/e02f32d6262b8a902785fd1e3cdab2bd22bd030a/ycmd/utils.py#L517)):*
> ```Python
>   pstart = query.index( '/' )
>   jarEnd = query.index( '<', pstart )
>   fileStart = query.index( '(', jarEnd )
> ```

Ycmd uses snake_case for variables.

---

*[ycmd/utils.py, line 520 at r2](https://reviewable.io/reviews/valloric/ycmd/1024#-LC-1TkdB5o9fRfdhJBZ:-LC-1TkdB5o9fRfdhJB_:b-ojuo75) ([raw file](https://github.com/valloric/ycmd/blob/e02f32d6262b8a902785fd1e3cdab2bd22bd030a/ycmd/utils.py#L520)):*
> ```Python
>   if pstart > 0 and jarEnd > pstart and fileStart > jarEnd:
>     filePath = query[ pstart + 1:jarEnd ]
>     if os.name != 'nt':
> ```

Instead of `os.name != 'nt'` use `not OnWindows()` which is defined in this same file.

---

*[ycmd/utils.py, line 522 at r2](https://reviewable.io/reviews/valloric/ycmd/1024#-LC-3h4z8U8SuO8W5deo:-LC-3h4z8U8SuO8W5dep:bfszda0) ([raw file](https://github.com/valloric/ycmd/blob/e02f32d6262b8a902785fd1e3cdab2bd22bd030a/ycmd/utils.py#L522)):*
> ```Python
>     if os.name != 'nt':
>       filePath = filePath.replace( '\\', '' )
>     filePath = filePath.replace( '.RELEASE.jar', '.RELEASE-sources.jar' )
> ```

This line looks odd. Why are we appending `-sources` here? I'm assuming the `-source` archive holds the `*.java` files, while the "default" one holds `*.class` files.
Sorry if this is just my inexperience with java projects speaking.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1024#-:-LC-4TbWEzZgj01ngTXv:b-1x8d7n)*

brookhong:
@tsmaeder Do you mean this SourceContentProvider -- https://github.com/eclipse/eclipse.jdt.ls/blob/2127f12fb3700dd783b75d973e4f2ecc134b9ddf/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/SourceContentProvider.java#L25 ?

ycmd integrates with jdt.ls through LSP, and from https://github.com/eclipse/eclipse.jdt.ls/wiki/Language-Server-Protocol-support , there is no interface for fetching source.

bstaletic:
@brookhong Take a look at [this](https://github.com/eclipse/eclipse.jdt.ls/wiki/Language-Server-Protocol-Extensions). Unfortunately, the parameters are not specified.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1024#-:-LC-QwXF4fRTuHAp6gy3:bgn5loz)*

fbricon:
Agree with @tsmaeder, parsing URIs is definitely not a good idea. Format might change, it would not support decompiled sources ...
While I reckon our documentation sucks, the only parameter required for `java/classFileContents` is the [TextDocumentIdentifier](https://github.com/eclipse/eclipse.jdt.ls/blob/0bba050b76ea867564bd6221e78782bc69d09cfa/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/lsp/JavaProtocolExtensions.java#L31), calling is probably way simpler.

Incidentally, in the next jdt.ls release, jdt:// uris won't be returned unless clients declare their support for it(WIP https://github.com/eclipse/eclipse.jdt.ls/issues/649)

brookhong:
Thanks @fbricon for comments, so https://github.com/eclipse/eclipse.jdt.ls/issues/657 will also got fixed next release?

fbricon:
@brookhong no, eclipse/eclipse.jdt.ls#657 is an orthogonal issue. Someone's encoding the URI, but shouldn't

brookhong:
Thanks @tsmaeder and @fbricon for referring to `java/classFileContents`, which works great.

I have modified my PR to use `java/classFileContents`, and made relevant changes on client, PR is https://github.com/Valloric/YouCompleteMe/pull/3018

@bstaletic, please help to review the changes.

brookhong:
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[ycmd/completers/language_server/language_server_protocol.py, line 403 at r1](https://reviewable.io/reviews/valloric/ycmd/1024#-LBzDyt4EQvAWb7cbox2:-LC9vDPt5iOosDnrh7Rj:bwpsnfb) ([raw file](https://github.com/valloric/ycmd/blob/f372543bb6f8828f334a5567778de94473b8a536/ycmd/completers/language_server/language_server_protocol.py#L403)):*

Like I mentioned we used to "drop" `jdt://*.class` URIs. This was done here, by raising an `InvalidUriException`.
If we can get `jdt://` to work, we don't need the exception any more, so it should be dropped (along with the tests for the same).
</blockquote></details>

I did not remove it, I'm not sure if there is other case that need `InvalidUriException`?

---

*[ycmd/utils.py, line 522 at r2](https://reviewable.io/reviews/valloric/ycmd/1024#-LC-3h4z8U8SuO8W5deo:-LC9vduVAbeExf6pnPtk:b-qmmtph) ([raw file](https://github.com/valloric/ycmd/blob/e02f32d6262b8a902785fd1e3cdab2bd22bd030a/ycmd/utils.py#L522)):*

This line looks odd. Why are we appending `-sources` here? I'm assuming the `-source` archive holds the `*.java` files, while the "default" one holds `*.class` files.
Sorry if this is just my inexperience with java projects speaking.
</blockquote></details>

With extension -- java/classFileContents from jdt.ls, we don't need to fetch source from jar package now.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1024)*

brookhong:
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[ycmd/utils.py, line 76 at r1](https://reviewable.io/reviews/valloric/ycmd/1024#-LBzI63A0-CR-_oV0D78:-LC9wAZi3-PCO_1oLClE:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/f372543bb6f8828f334a5567778de94473b8a536/ycmd/utils.py#L76)):*

You missed adding spaces in this block.
</blockquote></details>

Done.

---

*[ycmd/utils.py, line 517 at r2](https://reviewable.io/reviews/valloric/ycmd/1024#-LC-359SApaCuRUZYrHY:-LC9wDXf5CFskigbuNVx:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/e02f32d6262b8a902785fd1e3cdab2bd22bd030a/ycmd/utils.py#L517)):*

Ycmd uses snake_case for variables.
</blockquote></details>

Done.

---

*[ycmd/utils.py, line 520 at r2](https://reviewable.io/reviews/valloric/ycmd/1024#-LC-1TkdB5o9fRfdhJBZ:-LC9wCH6Dkrz0iCanYV_:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/e02f32d6262b8a902785fd1e3cdab2bd22bd030a/ycmd/utils.py#L520)):*

Instead of `os.name != 'nt'` use `not OnWindows()` which is defined in this same file.
</blockquote></details>

Done.

---

*[ycmd/utils.py, line 522 at r2](https://reviewable.io/reviews/valloric/ycmd/1024#-LC-3h4z8U8SuO8W5deo:-LC9wELg5R2WmJ0nudgo:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/e02f32d6262b8a902785fd1e3cdab2bd22bd030a/ycmd/utils.py#L522)):*
<details><summary><i>Previously, brookhong wrote…</i></summary><blockquote>

With extension -- java/classFileContents from jdt.ls, we don't need to fetch source from jar package now.
</blockquote></details>

Done.

---

*[ycmd/tests/utils_test.py, line 589 at r1](https://reviewable.io/reviews/valloric/ycmd/1024#-LBzFXQO5bSJjI5gxhT0:-LC9wFXt1W_utZ4rTQT5:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/f372543bb6f8828f334a5567778de94473b8a536/ycmd/tests/utils_test.py#L589)):*

The ycmd style is to use parens instead of back slashes for multiline statements.

Instead of `lcd + "/ycmd/tests..."` you should be using `PathToTestFile()`, like this:
`jarPath = PathToTestFile( 'java', 'spring-boot-autoconfigure-1.5.8.RELEASE.jar...')`
Note that directories should not be passed all together, instead path joining should be left up to Python. That way Windows gets `\\` separators and *nix gets `/` separators.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1024)*

fbricon:
@brookhong you need to add extendedClientCapabilities.classFileContentsSupport=true to jdt.ls' initializationOptions, see https://github.com/eclipse/eclipse.jdt.ls/issues/649#issuecomment-387775466, else it will not work with the next jdt.ls release

puremourning:
@fbricon Thanks for the heads up! However, I think we will probably wait to update the initialisation options when we do take the next version. IIRC we currently don't have a nice way internally to handle that and adding something now might be YAGNI. Our regression tests will catch when it breaks :) and we will have to make a decision on it then.

I think we can at least add a comment/FIXME/TODO.

fbricon:
@puremourning that's up to you. Next jdt.ls release planned for Tuesday, May 15th, 2018.

brookhong:
@puremourning  @bstaletic Is it possible to get this merged?

brookhong:
Paste gitter conversation here for clarification.

    brookhong @brookhong 22:59
    @bstaletic any thoughts on PR Valloric/ycmd#1024

    Boris Staletic @bstaletic 23:02
    Sorry about the delay.
    Exposing the functionality through a ClassFileContents subcommand seems dirty.
    It's basically exposing an implementation detail to the users and forcing them to think if they need to touch that subcommand or not.
    It would be much better if we could do some magic and have all of the current subcommands (GoTo for example) work even on jdt:// URIs.

    brookhong @brookhong May 25 23:09
    I have had another implementation that returns class file content within response of GoToDefinition, but the problem of it is that content will disappear if user reloads/leaves a buffer for jdt uri.

    Boris Staletic @bstaletic 23:13
    I had an idea, but it's quite hacky.
    Once we get the contents, we could write it to an actual file and point the GoTo response to that file.
    This then imposes a question of how do we track the "temporary" files, what about duplicate files (two GoTo requests pointing to the same .class file) and where should we store those.
    There's also a question of performance in this case...
    @micbou Any thoughts?

    brookhong @brookhong 23:15
    Yes, I also thought about using a temp file, but as you said, it has problems, and makes things complicated.

    Boris Staletic @bstaletic 23:16
    I'm really not sure what's the best option.

    brookhong @brookhong 23:18
    With current implementation, class file content will be re-fetched by a ClassFileContents call to server whenever user reloads/re-enters a buffer for jdt uri, <C-o>/<C-i> will also be happy with it.

    Boris Staletic @bstaletic 23:19
    Yes, but it exposes the user needs to manually run ClassFileContents and I'm not the only one who's against that.

    brookhong @brookhong 23:23
    No, user need not run the command manually, client of ycmd will take care of that automatically.
    autocmd BufReadCmd jdt://contents/*/.class call s:ReadClassFileFromJDT()
    Whenever a uri like jdt://contents/*/.class is opened, a ClassFileContents call will be made to jdt.ls automatically to fetch class file content.
    Please see https://github.com/Valloric/YouCompleteMe/pull/3018/files
    Actually this is similar with logic of reading git commit uri by https://github.com/tpope/vim-fugitive

    Boris Staletic @bstaletic 23:30
    Okay, that does seem better than I initially thought. I'd still like to hear from @micbou and @puremourning, as it still feels kinda dirty to expose that subcommand even if a user doesn't need to trigger it manually, because it will be there when users starts typing :YcmCompleter and then tries to complete the subcommand using wildmenu or <C-d>.

    brookhong @brookhong 23:34
    Yep, that could be a problem, I'll try to see if there is a way to make ClassFileContents internal.

    Boris Staletic @bstaletic 23:34
    Once again, sorry about the delay.

    brookhong @brookhong 23:35
    No problem. @puremourning @micbou please also give your thoughts.

    brookhong @brookhong 00:08
    Updated the PR to make the ClassFileContents command not exposed.

bstaletic:
Reviewed 6 of 8 files at r3.
Review status: 3 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[ycmd/responses.py, line 137 at r3](https://reviewable.io/reviews/valloric/ycmd/1024#-LCA2LeX9P_RkL-bjW_u:-LCA2LeX9P_RkL-bjW_v:b-nn76mb) ([raw file](https://github.com/valloric/ycmd/blob/9640bcf5d50d3fb884c5f953e2616171ba6c15b8/ycmd/responses.py#L137)):*
> ```Python
> 
> def IsJdtContentUri( uri ):
>   return uri[ : 15 ] == "jdt://contents/"
> ```

Is the format guaranteed to stay `jdt://contents/` or would it be more future proof to use only `jdt://`?

---

*[ycmd/completers/language_server/language_server_protocol.py, line 403 at r1](https://reviewable.io/reviews/valloric/ycmd/1024#-LBzDyt4EQvAWb7cbox2:-LC-48ph6Oous4AEC_tc:baockgi) ([raw file](https://github.com/valloric/ycmd/blob/f372543bb6f8828f334a5567778de94473b8a536/ycmd/completers/language_server/language_server_protocol.py#L403)):*
<details><summary><i>Previously, brookhong wrote…</i></summary><blockquote>

I did not remove it, I'm not sure if there is other case that need `InvalidUriException`?
</blockquote></details>

Currently we only have jdt.ls that uses the LSP interface. As far as I know, jdt.ls is passing either `jdt://` or `file://` both should work with this PR and thus the `InvalidUriException` won't ever be raised.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1024)*

puremourning:
First and foremost, thanks so much for your patience and for your interest in contributing to YCM/ycmd!!

I just sort of scanned through this. It seems the design is to expose on our API the custom URI that jdt.ls uses, and for the client to then be responsible to sort of just-know that it should call some specific completer subcommand in order to resolve that.
I think that approach is not general enough for my taste, although it likely does solve this one particular use case.
If other computers are brought in then we now have a proliferation of special cases that we rely on client implementations to have to know about. We must at the _very least_ document this in our API and somehow explain that to the client authors. This is not a great situation IMO.

Overall i think the implementation also is somewhat breaking all the paradigms of design (such that they are) that we have internally so I'm not super keen on it. I feel to do it right we need at least 2 new general mechanisms, but the overal complexity of that seems high for what I really think of as a pretty niche feature. How often does it come up that you need jdt.ls to decompile the code to view ? In my experience it's pretty rare, but happy to take an opinion on how valuable this feature is. My current opinion is that it isn't sufficiently valuable or general to justify the complexity of a general solution or the somewhat hackiness of the special-casing in this PR as it stands.

---

Review status: 1 of 9 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[ycmd/responses.py, line 136 at r5](https://reviewable.io/reviews/valloric/ycmd/1024#-LDS5sxY2nWqDlEimII-:-LDS5sxY2nWqDlEimII0:b-15xosw) ([raw file](https://github.com/valloric/ycmd/blob/131df787fba0ef16256aba8b5fab4d586ac587ff/ycmd/responses.py#L136)):*
> ```Python
> 
> 
> def IsJdtContentUri( uri ):
> ```

This is breaking encapsulation. If we're going to go down the route of having custom uri handlers, then we should allow the completer abstraction to hide it. This "core" area of the code shouldn't even know that JDT exists. Indeed, in a future ycmd, it might not.

---

*[ycmd/responses.py, line 145 at r5](https://reviewable.io/reviews/valloric/ycmd/1024#-LDS6Gu-9uVOtv-0pVDV:-LDS6Gu08VPpWRME1Kfb:b-xvjf8l) ([raw file](https://github.com/valloric/ycmd/blob/131df787fba0ef16256aba8b5fab4d586ac587ff/ycmd/responses.py#L145)):*
> ```Python
>   if location.filename_:
>     filename = location.filename_
>     if not IsJdtContentUri( filename ):
> ```

this is somewhat a logical fallacy. i think what this code should be saying is "if this is a real filesystem path, normalise it", not "if this is anything other than one very-specific-case-that-we-have-implemented, then normalise it"

---

*[ycmd/responses.py, line 220 at r5](https://reviewable.io/reviews/valloric/ycmd/1024#-LDS6ZD7DDYVj3a6yogg:-LDS6ZD7DDYVj3a6yogh:b-2guby3) ([raw file](https://github.com/valloric/ycmd/blob/131df787fba0ef16256aba8b5fab4d586ac587ff/ycmd/responses.py#L220)):*
> ```Python
>     self.column_number_ = column
>     if filename:
>       if IsJdtContentUri( filename ):
> ```

same here

---

*[ycmd/completers/completer.py, line 258 at r5](https://reviewable.io/reviews/valloric/ycmd/1024#-LDS6b1c3ZgsdKyayU4-:-LDS6b1dEocGBeZa-yYp:bh48tqg) ([raw file](https://github.com/valloric/ycmd/blob/131df787fba0ef16256aba8b5fab4d586ac587ff/ycmd/completers/completer.py#L258)):*
> ```Python
>       # client of ycmd will call this command whenever a uri like
>       # jdt://contents/*/.class is opened
>       subcommands.remove( 'ClassFileContents' )
> ```

again, this is breaking encapsulation. If the java completer doesn't want to expose this subcommand, then don't include it in the returned mapping.

---

*[ycmd/completers/language_server/language_server_completer.py, line 1337 at r5](https://reviewable.io/reviews/valloric/ycmd/1024#-LDS6oeE8Q6KHK1ztj28:-LDS6oeE8Q6KHK1ztj29:b-wl129f) ([raw file](https://github.com/valloric/ycmd/blob/131df787fba0ef16256aba8b5fab4d586ac587ff/ycmd/completers/language_server/language_server_completer.py#L1337)):*
> ```Python
>     return self.GetConnection().GetResponse(
>       request_id,
>       lsp.ClassFileContents( request_id, request_data ),
> ```

I think this is a custom JDT message, so should be implemented in the jdt-specidic JavaCompleter class, not in the language-agnostic LAnguageServerCompleter class. Again, the encapsulation is breaking down by implementing this at this level: it implies that this is general, when it is not.

Bear in mind that there is a high likelihood of ycmd exploiting this general LSP completer for other things in the future. Indeed, there are such ongoing projects.

---

*[ycmd/completers/language_server/language_server_completer.py, line 1750 at r5](https://reviewable.io/reviews/valloric/ycmd/1024#-LDS7FtGCyeRGIix64d0:-LDS7FtGCyeRGIix64d1:b-gs3y4d) ([raw file](https://github.com/valloric/ycmd/blob/131df787fba0ef16256aba8b5fab4d586ac587ff/ycmd/completers/language_server/language_server_completer.py#L1750)):*
> ```Python
>   """Convert a LSP position to a ycmd location."""
>   try:
>     if responses.IsJdtContentUri( position[ 'uri' ] ):
> ```

again, this should be more like "if this is a real file, read the file" not "if this is a very specific thing which is not a file, don't read the file"

Removing the exception handling means that you're asserting all uris will be files or jdt:// things. This is not true, as not only could jdt.ls start sending other uris, but other lsp servers could reginster/send other schemes. I think we need to keep the protection (in the form of an exception).

Why not have the InvalidUriException hold a property that is the uri scheme and have a scheme for registering support for specific uris? This is ultimately what VScode extensions do and is somewhat general, bundling the custom completer-specific stuff into the completer that knows/cares about the server.

We don't want to over-generalise, but equally we don't want to have special cases littered all over the general code.

---

*[ycmd/completers/language_server/language_server_protocol.py, line 192 at r5](https://reviewable.io/reviews/valloric/ycmd/1024#-LDS84Jj8byKt5GzLhhg:-LDS84Jj8byKt5GzLhhh:b-h7fi2f) ([raw file](https://github.com/valloric/ycmd/blob/131df787fba0ef16256aba8b5fab4d586ac587ff/ycmd/completers/language_server/language_server_protocol.py#L192)):*
> ```Python
>     'rootUri': FilePathToUri( project_directory ),
>     'initializationOptions': {
>       'classFileContentsSupport': True
> ```

This is a completer-specific argument which would not be supported, nor necessarily have the same semantics, on other servers. We need a way to provide arbitrary initialisation options in a completer-agnostic way. This requires more engineering that just injecting jdt-speciric things int he general completer code.

---

*[ycmd/completers/language_server/language_server_protocol.py, line 280 at r5](https://reviewable.io/reviews/valloric/ycmd/1024#-LDS8LT-B2GZO67pnsgo:-LDS8LT-B2GZO67pnsgp:b5ajgsn) ([raw file](https://github.com/valloric/ycmd/blob/131df787fba0ef16256aba8b5fab4d586ac587ff/ycmd/completers/language_server/language_server_protocol.py#L280)):*
> ```Python
> 
> def ClassFileContents( request_id, request_data ):
>   return BuildRequest( request_id, 'java/classFileContents', {
> ```

i don't think java-specifics should be in the language-agnostic protocol implementation.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1024#-:-LDS8W2HB6DqCYHKVHW7:bc0mjva)*

PR #585, opened 2016-08-28
[READY] Update JediHTTP submodule

micbou:
You rarely (never?) want to jump to an import statement in Python.

Update JediHTTP submodule to include https://github.com/vheon/JediHTTP/pull/22.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/585)

PR #1028, opened 2018-05-13
[READY] Rewrite Python completer

micbou:
[Jedi 0.12.0](https://github.com/davidhalter/jedi/blob/master/CHANGELOG.rst#0120-2018-04-15) introduces two important features that obsolete JediHTTP:
 - ability to complete for a different Python than the one it's running on;
 - cannot crash when importing compiled modules.

Rewrite the Python completer to use directly Jedi. This introduces a breaking change in that the Python interpreter path cannot be set anymore with the `RestartServer` command (such command doesn't make sense without a server). More importantly, the new completer allows users to configure the interpreter path and `sys.path` through the extra conf file by defining the `PythonSettings( **kwargs )`, `PythonSysPath( **kwargs )` functions or the `python_settings` variable. See the changes in `README.md` for an explanation on how they work. Note that it's still possible to set the interpreter path with the `python_binary_path` option but using the extra conf file is preferred.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)


micbou:
Reviewed 36 of 37 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---

*[ycmd/completers/python/python_completer.py, line 141 at r3](https://reviewable.io/reviews/valloric/ycmd/1028#-LCPxw7A-yxtQbpwZ38P:-LCPxw7A-yxtQbpwZ38Q:b-nwpc10) ([raw file](https://github.com/valloric/ycmd/blob/bfa673d81ba6ea607791431b9ae13323339bb1c9/ycmd/completers/python/python_completer.py#L141)):*
> ```Python
>     settings.update( self._SettingsForRequest( request_data ) )
>     settings[ 'interpreter_path' ] = environment.executable
>     settings[ 'sys_path' ].extend( environment.get_sys_path() )
> ```

It's quite subtle but this line may change `request_data[ 'extra_conf_data' ]` without PR https://github.com/Valloric/ycmd/pull/1029. That's why the PR is marked as WIP.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

micbou:

---

*[.gitmodules, line 24 at r3](https://reviewable.io/reviews/valloric/ycmd/1028#-LCPzziN4-1HxTLGtBK0:-LCQBvvF4WqZ7lGrqH8e:b-caishf) ([raw file](https://github.com/valloric/ycmd/blob/bfa673d81ba6ea607791431b9ae13323339bb1c9/.gitmodules#L24)):*

Parso is the new jedi dependency. Is that the reason we have it in our `third_party`?
</blockquote></details>

Yes, Jedi doesn't bundle it.

---

*[ycmd/completers/python/python_completer.py, line 141 at r3](https://reviewable.io/reviews/valloric/ycmd/1028#-LCPxw7A-yxtQbpwZ38P:-LCQDeqDDX0GPalHUCIH:bxcbbrb) ([raw file](https://github.com/valloric/ycmd/blob/bfa673d81ba6ea607791431b9ae13323339bb1c9/ycmd/completers/python/python_completer.py#L141)):*

Perhaps mark this PR as `[NOT READY]`?
</blockquote></details>

Yes, that's better.

---

*[ycmd/completers/python/python_completer.py, line 227 at r3](https://reviewable.io/reviews/valloric/ycmd/1028#-LCPzWRs1hlLQpyx_thJ:-LCQCHIi-2QSU7JyyjRP:bob93mj) ([raw file](https://github.com/valloric/ycmd/blob/bfa673d81ba6ea607791431b9ae13323339bb1c9/ycmd/completers/python/python_completer.py#L227)):*

What's the reason for not passing `follow_imports = True` to `goto_assignments`.
</blockquote></details>

This has already been attempted in PR https://github.com/Valloric/ycmd/pull/585 and it comes with its own set of issues. See @puremourning's example.

---

*[ycmd/tests/__init__.py, line 87 at r2](https://reviewable.io/reviews/valloric/ycmd/1028#-LCP2IbG2yN2OL0M7eAc:-LCQF6rV3MqbnNONbFAL:b-i3aoqe) ([raw file](https://github.com/valloric/ycmd/blob/edf830f0ba6cdeb6c80e6c510860da44cb8e3fa1/ycmd/tests/__init__.py#L87)):*

I understand that now the extra conf can be discovered  from `ycmd/ycmd` directory, but why do we need to ignore it in isolated tests?
</blockquote></details>

Because some of the tests in `tests/extra_conf_store_test.py` are failing otherwise and it's more convenient to ignore the file here than in those tests.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

bstaletic:

---

*[ycmd/completers/python/python_completer.py, line 227 at r3](https://reviewable.io/reviews/valloric/ycmd/1028#-LCPzWRs1hlLQpyx_thJ:-LCQJi_m7hNeDAG240l3:b7xzoft) ([raw file](https://github.com/valloric/ycmd/blob/bfa673d81ba6ea607791431b9ae13323339bb1c9/ycmd/completers/python/python_completer.py#L227)):*

This has already been attempted in PR https://github.com/Valloric/ycmd/pull/585 and it comes with its own set of issues. See @puremourning's example.
</blockquote></details>

Alright, that's fine.

On a side note, if we take @puremourning's example into account and attempt `:YcmCompleter GoToDefinition` on `OBFUSCATED_NAME` inside the `format` method I get the `RuntimeError: Can't jump to builtin module` exception.

It might be a little confusing when first encountered.
And the same happens when I drop the `as OBFUSCATED_NAME` part.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

micbou:
I added support for the `settings` dictionary in the extra conf as discussed. Here's an example of a `.ycm_extra_conf.py` file using that variable:
```python
settings = {
  'python': {
    'interpreter_path': '~/project/virtual_env/bin/python',
    'sys_path': [ '~/project/third_party/module' ]
  }
}
```
The idea behind this dictionary is that we can extend it to other completers. For instance, we could extend it this way for the Clang completer:
```python
settings = {
  'cfamily': {
    'flags': [ '-x', 'c++', '-Wall', ... ]
  }
}
```
This would allow users to simultaneously configure Python and C-family languages in their extra conf:
```python
settings = {
  'cfamily': {
    ...
  },
  'python': {
    ...
  }
}
```  

---

Reviewed 1 of 1 files at r4, 7 of 11 files at r5.
Review status: 32 of 36 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028#-:-LCzCMQY1Cw0L-LBd1WJ:b2ukqfh)*

bstaletic:
Nice work! I'm guessing Codecov won't be complaining once the tests don't flake.
:lgtm: 

---

Reviewed 1 of 1 files at r4, 11 of 11 files at r5.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028#-:-LD0xfhL8Snq3wc6O82y:b-i37ejy)*

micbou:
Renamed the `PythonSettings` function to `Settings`. It now takes a new parameter `filetype` which is the current filetype i.e. `python` when called from the Python completer. The goal is to extend that function to other completers so that it's  possible to configure multiple languages through that function e.g.
```python
def Settings( **kwargs ):
  filetype == kwargs[ 'filetype' ]
  if filetype == 'cpp':
    return { 'flags': [ '-Wall', '-Werror', ... ] }
  if filetype == 'python':
    return { 'interpreter_path': '~/project/my_virtual_env/bin/python' }  
```

---

Reviewed 4 of 13 files at r5, 1 of 1 files at r6, 7 of 10 files at r7, 1 of 1 files at r8, 5 of 5 files at r9.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028#-:-LD2Oh7w-hnzlJ6fObf9:beo0ha)*

bstaletic:
Great work. :lgtm_strong: 

---

Reviewed 1 of 1 files at r6, 8 of 10 files at r7, 1 of 1 files at r8, 5 of 5 files at r9.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028#-:-LD2XtqS2dYwcJBQgQeW:b-7ztoc7)*

puremourning:
:lgtm:

---

Review status: 37 of 38 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[README.md, line 267 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE-tDfb4QJaoEtQEhel:-LE-tDfb4QJaoEtQEhem:bpaa9fw) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/README.md#L267)):*
> ```Markdown
>    example.
> 
> The return value is a dictionary containing optionally the following items:
> ```

I would prefer to put (optional) next to each entry like we did for FlagsForFile, rather than awkwardly squeezing "optionally" in this part of the sentence. 

---

*[README.md, line 284 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE-tllPAJZV8nIaez6u:-LE-tllPAJZV8nIaez6v:b7tu3om) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/README.md#L284)):*
> ```Markdown
>       'interpreter_path': '~/project/virtual_env/bin/python',
>       'sys_path': [ '~/project/third_party/module' ]
>     }
> ```

We should encourage this to return {} in the case the language is not understood. This allows it to be forward compatible. I realise there's a PR to tolerate "no return" from this method, but I still think it is good practice for a function defined to return a dict, to actually always return one.

---

*[README.md, line 320 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE-vWMf7XU-980o7A_F:-LE-vWMf7XU-980o7A_G:b-g5r5zc) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/README.md#L320)):*
> ```Markdown
> 
> This function allows further customization of the Python path `sys.path`. Its
> parameters are the possible items returned by the `PythonSettings` function:
> ```

PythonSettings ? I think it is called Settings now ?

---

*[ycmd/completers/python/python_completer.py, line 58 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE-wKVaE-eIA6kFRkSU:-LE-wKVaE-eIA6kFRkSV:b-yznh3e) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L58)):*
> ```Python
>   def OnFileReadyToParse( self, request_data ):
>     environment = self._EnvironmentForRequest( request_data )
>     self._SysPathForFile( request_data, environment )
> ```

it seems odd that the return from this method is not used.

I assume it is populating some cache or other? As a general approach I'm not a big fan of relying on the side-effects of method calls.

Would rather have a method self._CacheSysPathForFile ... or something that does what it says rather than implicitly doing something side-effecty

---

*[ycmd/completers/python/python_completer.py, line 78 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE02bIS9jgYgYpVBtcK:-LE02bIS9jgYgYpVBtcL:bi7hb2n) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L78)):*
> ```Python
>     # We don't warn the user if no extra conf file is found.
>     if module:
>       if hasattr( module, 'Settings' ):
> ```

Should we check `and callable( getattr( module, 'Settings' ) )` here ?

Now that I think about it, it's quite subtle that `settings` is a dict and `Settings` is a function. I think it would be quite easy to make that mistake and just get an exception here.

---

*[ycmd/completers/python/python_completer.py, line 144 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE04t9wDeR7__-AAd6R:-LE04t9wDeR7__-AAd6S:b-xtec0j) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L144)):*
> ```Python
>     # We don't warn the user if no extra conf file is found.
>     if module:
>       if hasattr( module, 'PythonSysPath' ):
> ```

`and callable...` ?

---

*[ycmd/completers/python/python_completer.py, line 194 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE066Ch-nDUxwTJPY-R:-LE066Ch-nDUxwTJPY-S:b-jqghrg) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L194)):*
> ```Python
> 
>   def ComputeCandidatesInner( self, request_data ):
>     with self._jedi_lock:
> ```

perhaps we should get the completions under the lock, then convert to our format outside of the lock? That's somewhat assuming that the return value thingy is safe to use in that way.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028#-:-LE086743Soy83f8uoG6:bnfp4nl)*

micbou:

---

*[README.md, line 267 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE-tDfb4QJaoEtQEhel:-LE0I6315ZyY7Z-xAYVf:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/README.md#L267)):*

I would prefer to put (optional) next to each entry like we did for FlagsForFile, rather than awkwardly squeezing "optionally" in this part of the sentence. 
</blockquote></details>

Done.

---

*[README.md, line 284 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE-tllPAJZV8nIaez6u:-LE0IE-OCwwi0vY9B8q_:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/README.md#L284)):*

We should encourage this to return {} in the case the language is not understood. This allows it to be forward compatible. I realise there's a PR to tolerate "no return" from this method, but I still think it is good practice for a function defined to return a dict, to actually always return one.
</blockquote></details>

Done.

---

*[README.md, line 320 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE-vWMf7XU-980o7A_F:-LE0IErV3WgIN-7_xSqZ:bvka32d) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/README.md#L320)):*

PythonSettings ? I think it is called Settings now ?
</blockquote></details>

Forgot to rename this one. Fixed.

---

*[ycmd/completers/python/python_completer.py, line 58 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE-wKVaE-eIA6kFRkSU:-LE0NhcN53MK0UowFqWn:bdt5szi) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L58)):*

it seems odd that the return from this method is not used.

I assume it is populating some cache or other? As a general approach I'm not a big fan of relying on the side-effects of method calls.

Would rather have a method self._CacheSysPathForFile ... or something that does what it says rather than implicitly doing something side-effecty
</blockquote></details>

Yes, it's for the cache. The return value will be used when implementing modules preloading so let's just remove these lines for now.

---

*[ycmd/completers/python/python_completer.py, line 78 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE02bIS9jgYgYpVBtcK:-LE0KOBy5wpWydFF7kv2:b-n88un8) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L78)):*

Should we check `and callable( getattr( module, 'Settings' ) )` here ?

Now that I think about it, it's quite subtle that `settings` is a dict and `Settings` is a function. I think it would be quite easy to make that mistake and just get an exception here.
</blockquote></details>

Maybe we should use the exact same name and detect if `Settings`/`settings` is a dictionary or a function since there is no point to define both in a `.ycm_extra_conf.py` file?

---

*[ycmd/completers/python/python_completer.py, line 194 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE066Ch-nDUxwTJPY-R:-LE0Kxr71bitAtwOJRKH:b5lp6y) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L194)):*

perhaps we should get the completions under the lock, then convert to our format outside of the lock? That's somewhat assuming that the return value thingy is safe to use in that way.
</blockquote></details>

Unfortunately, I don't think accessing the properties of the completions and calling the `docstring` method is thread safe.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

puremourning:

---

*[ycmd/completers/python/python_completer.py, line 78 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE02bIS9jgYgYpVBtcK:-LE0YBWA2vAiiqE_JLc0:b-c2wsgh) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L78)):*

Maybe we should use the exact same name and detect if `Settings`/`settings` is a dictionary or a function since there is no point to define both in a `.ycm_extra_conf.py` file?
</blockquote></details>

I'm not really sure we need both. the settings dict is just the same as:

```python
settings = {
 ...
}

def Settings( **kwargs ):
  return settings
```

maybe we should just have the method and put an example in the docs that shows that a simple approach like above works?

---

*[ycmd/completers/python/python_completer.py, line 194 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE066Ch-nDUxwTJPY-R:-LE0YdS9DqvgHpHdwg81:b32wtd) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L194)):*

Unfortunately, I don't think accessing the properties of the completions and calling the `docstring` method is thread safe.
</blockquote></details>

ok

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

bstaletic:

---

*[ycmd/completers/python/python_completer.py, line 78 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE02bIS9jgYgYpVBtcK:-LE0l87F8Z_SYZilQMKj:b4fm4yd) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L78)):*

I'm not really sure we need both. the settings dict is just the same as:

```python
settings = {
 ...
}

def Settings( **kwargs ):
  return settings
```

maybe we should just have the method and put an example in the docs that shows that a simple approach like above works?
</blockquote></details>

@micbou's reason for including both was "a function is more fexible but a variable is more convenient".

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

puremourning:

---

*[ycmd/completers/python/python_completer.py, line 78 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE02bIS9jgYgYpVBtcK:-LE0sUMwDPtaVMnj-pov:b7ziyqn) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L78)):*

@micbou's reason for including both was "a function is more fexible but a variable is more convenient".
</blockquote></details>

I just argue that a variable is only marginally more convent (really, just 1 LOC) for the user, whereas in the code we have to double up our cases, support and tests

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

bstaletic:

---

*[ycmd/completers/python/python_completer.py, line 78 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE02bIS9jgYgYpVBtcK:-LE0tx5LBP0FJfYLOMYc:b-v3zmzy) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L78)):*

I just argue that a variable is only marginally more convent (really, just 1 LOC) for the user, whereas in the code we have to double up our cases, support and tests
</blockquote></details>

Fair enough. I just wanted to point out @micbou's reasoining. I too don't find a variable that much convenient.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

micbou:

---

*[ycmd/completers/python/python_completer.py, line 78 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE02bIS9jgYgYpVBtcK:-LE4QW1hCh1tTTXkvYsy:bk7xkk2) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L78)):*

Fair enough. I just wanted to point out @micbou's reasoining. I too don't find a variable that much convenient.
</blockquote></details>

Okay, I am considering dropping the `settings` dictionary (it can always be reintroduced later under a different name) but I'd like to know your opinion on the `Settings` function. The goal is to extend that function to all completers: `FlagsForFile` would be deprecated and replaced with `Settings` and the LSP completers would use that function to pass settings to their language servers. Are you fine with the idea?

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

bstaletic:

---

*[ycmd/completers/python/python_completer.py, line 78 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE02bIS9jgYgYpVBtcK:-LE4V4gH9i8r5musS_jM:b-krfkj) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L78)):*

Okay, I am considering dropping the `settings` dictionary (it can always be reintroduced later under a different name) but I'd like to know your opinion on the `Settings` function. The goal is to extend that function to all completers: `FlagsForFile` would be deprecated and replaced with `Settings` and the LSP completers would use that function to pass settings to their language servers. Are you fine with the idea?
</blockquote></details>

Yes, `Settings` sounds like a good idea.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

micbou:

---

*[ycmd/completers/python/python_completer.py, line 58 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE-wKVaE-eIA6kFRkSU:-LEEkW_60pbcjh4kvKnr:bpe8vmc) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L58)):*

Yes, it's for the cache. The return value will be used when implementing modules preloading so let's just remove these lines for now.
</blockquote></details>

Turns out it's not only for the cache but also for loading the extra conf file. We have to put these lines back. Should I add a comment?

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

puremourning:

---

*[ycmd/completers/python/python_completer.py, line 58 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE-wKVaE-eIA6kFRkSU:-LEGNRlH0DewmYeAi-vC:bli61cs) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L58)):*

Turns out it's not only for the cache but also for loading the extra conf file. We have to put these lines back. Should I add a comment?
</blockquote></details>

I would still like to have a method that does what it says, rather than relying on a side-effect (even with a comment). Side-effects can change when changing the code later.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

micbou:

---

*[ycmd/completers/python/python_completer.py, line 58 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE-wKVaE-eIA6kFRkSU:-LEGPpTf3nyohMKXAGds:bytbvfo) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L58)):*

I would still like to have a method that does what it says, rather than relying on a side-effect (even with a comment). Side-effects can change when changing the code later.
</blockquote></details>

In that case, I think I am going to include module preloading to this PR as I don't want to spend time on something that I am going to revert in a separate PR. Does that work for you?

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

puremourning:

---

*[ycmd/completers/python/python_completer.py, line 58 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE-wKVaE-eIA6kFRkSU:-LEGWAhV1U_VLEXLqmrv:ba6r71g) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L58)):*

In that case, I think I am going to include module preloading to this PR as I don't want to spend time on something that I am going to revert in a separate PR. Does that work for you?
</blockquote></details>

Oh I see. Sorry I remember you mentioned that before. Based on it being somewhat short-lived, I'm happy with a comment, and module loading in a separate PR. I'm also happy with incorporating the 2 PRs if that also makes sense. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

micbou:

---

*[ycmd/completers/python/python_completer.py, line 58 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE-wKVaE-eIA6kFRkSU:-LEGn0Gv8aswrgFD8ZiJ:b-f8sf1p) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L58)):*

Oh I see. Sorry I remember you mentioned that before. Based on it being somewhat short-lived, I'm happy with a comment, and module loading in a separate PR. I'm also happy with incorporating the 2 PRs if that also makes sense. 
</blockquote></details>

Added a comment.

---

*[ycmd/completers/python/python_completer.py, line 78 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE02bIS9jgYgYpVBtcK:-LEGjhO-8nDfTGSqGStK:bbwyy2o) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L78)):*

Yes, `Settings` sounds like a good idea.
</blockquote></details>

I removed the `settings` dictionary.

---

*[ycmd/completers/python/python_completer.py, line 144 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE04t9wDeR7__-AAd6R:-LEGkMttAcObkDWhewhk:b1oietc) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L144)):*

`and callable...` ?
</blockquote></details>

If `PythonSysPath` is not callable, the error `TypeError: ... object is not callable` is displayed to the user. Isn't that better than failing silently? There is also the solution to raise a more explicit error.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028)*

micbou:
Jedi's issue https://github.com/davidhalter/jedi/issues/1130 is blocking the PR as it breaks the Python completer on Ubuntu 18.04 (unless the `python3-distutils` package is installed).

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028#-:-LEGo-UsE_2w8kWtAA_G:bbxcibz)*

bstaletic:
This `python3-distutils` thing is getting "fun".

---

---

*[ycmd/completers/python/python_completer.py, line 144 at r12](https://reviewable.io/reviews/valloric/ycmd/1028#-LE04t9wDeR7__-AAd6R:-LEGvWr-45oCgU3xyO-U:bp5k3k9) ([raw file](https://github.com/valloric/ycmd/blob/b3541e59ac5fa91689f1e00396d4379a5bb009fe/ycmd/completers/python/python_completer.py#L144)):*

If `PythonSysPath` is not callable, the error `TypeError: ... object is not callable` is displayed to the user. Isn't that better than failing silently? There is also the solution to raise a more explicit error.
</blockquote></details>

I really don't think that we should fail silently here. The "default" exception seems to be explicit enough.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028#-:-LEGvig7ATtMFy4OO_36:b6pfrr)*

puremourning:
:lgtm:

---

Reviewed 4 of 5 files at r14, 4 of 7 files at r16, 1 of 1 files at r17, 1 of 1 files at r18, 1 of 1 files at r19, 1 of 1 files at r21.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028#-:-LEkk9MDDA1XXtn_Aftr:bnfp4nl)*

micbou:
@zzbot r+

---

Review status: :shipit: complete! 2 of 2 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028#-:-LEkmbXk2Q03XyqJBv-g:bi59kwd)*

micbou:
@zzbot retry

---

Review status: :shipit: complete! 2 of 2 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028#-:-LElQJU42Kk09_uhOGFq:b-kc0oa4)*

micbou:
@zzbot retry

---

Review status: :shipit: complete! 2 of 2 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028#-:-LElfKgcBBufXCMi0X6M:b-kc0oa4)*

bstaletic:
@zzbot retry

---

Review status: :shipit: complete! 2 of 2 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1028#-:-LEmv4Nd2VtK5q1NTKGH:b-kc0oa4)*

bstaletic:
Come on!
@zzbot retry

yan12125:
The sixth failure... Is it possible to bypass zzbot check and simply merge this? Looking forward this feature and have used this branch for several weeks.

bstaletic:
@yan12125 We could use the merge button on github, but it could mess up the bot. With so many flakes and zzbot causing double builds, I sometimes really wonder if the bot is worth it. \*sigh\*
@zzbot retry

PR #1029, opened 2018-05-13
[READY] Prevent users from modifying extra conf data

micbou:
Users can change the values of the extra conf data by doing something like
```python
def FlagsForFile( filename, **kwargs ):
  flags = kwargs[ 'client_data' ][ 'flags' ]  
  flags.append( '-Wall' ) # Change the value of request_data[ 'extra_conf_data' ][ 'flags' ]
  return { 'flags': flags }
```
in the `.ycm_extra_conf.py` file. This was already possible prior to PR https://github.com/Valloric/ycmd/pull/1010 but it's now a problem because cache is invalidated if extra conf data has changed. We prevent that by doing a deep copy when getting a value of a `HashableDict`'s key.

Also, implement the `__eq__` and `__ne__` methods in `HashableDict` as recommended by the official documentation on [`__hash__`](https://docs.python.org/2/reference/datamodel.html#object.__hash__)
> If a class does not define a __cmp__() or __eq__() method it should not define a __hash__() operation either

and [`__eq__`](https://docs.python.org/2/reference/datamodel.html#object.__eq__)
> There are no implied relationships among the comparison operators. The truth of x==y does not imply that x!=y is false. Accordingly, when defining __eq__(), one should also define __ne__() so that the operators will behave as expected.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1029)

### DIFF
--- a/ycmd/tests/request_wrap_test.py
+++ b/ycmd/tests/request_wrap_test.py
@@ -24,8 +24,8 @@ from __future__ import absolute_import
 # Not installing aliases from python-future; it's unreliable and slow.
 from builtins import *  # noqa
 
-from hamcrest import ( assert_that, calling, empty, equal_to, has_entry,
-                       has_string, matches_regexp, raises )
+from hamcrest import ( assert_that, calling, contains, empty, equal_to,
+                       has_entry, has_string, matches_regexp, raises )
 from nose.tools import eq_
 
 from ycmd.utils import ToBytes
@@ -382,13 +382,18 @@ def ExtraConfData_test():
   wrap = RequestWrap( PrepareJson() )
   assert_that( wrap[ 'extra_conf_data' ], empty() )
 
-  wrap = RequestWrap( PrepareJson( extra_conf_data = { 'key': 'value' } ) )
+  wrap = RequestWrap( PrepareJson( extra_conf_data = { 'key': [ 'value' ] } ) )
   extra_conf_data = wrap[ 'extra_conf_data' ]
-  assert_that( extra_conf_data, has_entry( 'key', 'value' ) )
+  assert_that( extra_conf_data, has_entry( 'key', contains( 'value' ) ) )
   assert_that(
     extra_conf_data,
-    has_string( matches_regexp( "^<HashableDict {u?'key': u?'value'}>$" ) )
+    has_string( matches_regexp( "^<HashableDict {u?'key': \[u?'value'\]}>$" ) )
   )
+
   # Check that extra_conf_data can be used as a dictionary's key.
   assert_that( { extra_conf_data: 'extra conf data' },
                has_entry( extra_conf_data, 'extra conf data' ) )
+
+  # Check that extra_conf_data's values are immutable.
+  extra_conf_data[ 'key' ].append( 'another_value' )
+  assert_that( extra_conf_data, has_entry( 'key', contains( 'value' ) ) )
--- a/ycmd/tests/utils_test.py
+++ b/ycmd/tests/utils_test.py
@@ -582,3 +582,14 @@ def GetCurrentDirectory_Py2NoCurrentDirectory_test():
 def GetCurrentDirectory_Py3NoCurrentDirectory_test():
   with patch( 'os.getcwd', side_effect = FileNotFoundError ): # noqa
     eq_( utils.GetCurrentDirectory(), tempfile.gettempdir() )
+
+
+def HashableDict_Equality_test():
+  dict1 = { 'key': 'value' }
+  dict2 = { 'key': 'another_value' }
+  ok_(     utils.HashableDict( dict1 ) == utils.HashableDict( dict1 ) )
+  ok_( not utils.HashableDict( dict1 ) != utils.HashableDict( dict1 ) )
+  ok_( not utils.HashableDict( dict1 ) == dict1 )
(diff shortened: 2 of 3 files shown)


bstaletic:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1029#-:-LCPjvT3133KoYB0JJ6C:bnfp4nl)*

puremourning:
seems ok, though it's a little sad that we have to jump through such hoops,

---

---

*[ycmd/utils.py, line 536 at r1](https://reviewable.io/reviews/valloric/ycmd/1029#-LCQ6bCz7yBPjk7t7j1o:-LCQ6bCz7yBPjk7t7j1p:b-r5bram) ([raw file](https://github.com/valloric/ycmd/blob/2080ce6cc965065814dadd2b1ddf6c81acd0788b/ycmd/utils.py#L536)):*
> ```Python
> 
> 
>   def __eq__( self, other ):
> ```

we should have module tests for this that ensure a == b and a != b work for a generic one of these 

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1029#-:-LCQ77Oz88j6-W0vMrjH:b-ybxdqw)*

micbou:
Yes, it's annoying.

---

Reviewed 2 of 2 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---

*[ycmd/utils.py, line 536 at r1](https://reviewable.io/reviews/valloric/ycmd/1029#-LCQ6bCz7yBPjk7t7j1o:-LCQAmD51WBI10V29Qi6:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/2080ce6cc965065814dadd2b1ddf6c81acd0788b/ycmd/utils.py#L536)):*

we should have module tests for this that ensure a == b and a != b work for a generic one of these 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1029#-:-LCQAxlm88wBR1lOO3hc:b-kdscrg)*

puremourning:
:lgtm:

---

---

*[ycmd/utils.py, line 504 at r3](https://reviewable.io/reviews/valloric/ycmd/1029#-LC_Ydkb4Rnm2ZOl4Pt0:-LC_Ydkb4Rnm2ZOl4Pt1:bxzrs3a) ([raw file](https://github.com/valloric/ycmd/blob/3d2bd527c731bee1f3eb40bc7c03416f080b12c2/ycmd/utils.py#L504)):*
> ```Python
> 
> 
> class HashableDict( collections.Mapping ):
> ```

Should we call a knife a knife and name this class UserDataDict ? We could call it ImmutableHashableDict, but that is a bit of a mouthful and we only use it for one thing.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1029#-:-LC_YxQB7lYNcJOeOqIn:bnfp4nl)*

puremourning:
Reviewed 2 of 2 files at r1, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1029)*

micbou:

---

*[ycmd/utils.py, line 504 at r3](https://reviewable.io/reviews/valloric/ycmd/1029#-LC_Ydkb4Rnm2ZOl4Pt0:-LC_Zco7DIyt0dcc0nZk:b-p0phwf) ([raw file](https://github.com/valloric/ycmd/blob/3d2bd527c731bee1f3eb40bc7c03416f080b12c2/ycmd/utils.py#L504)):*
> Should we call a knife a knife and name this class UserDataDict ?

I feel the class is a little too generic to call it `UserDataDict`. Also, we could use it for other things like replacing `frozendict`.

> We could call it ImmutableHashableDict, but that is a bit of a mouthful

A hashable object must be immutable so calling it `ImmutableHashableDict` would be redundant.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1029)*

puremourning:

---

*[ycmd/utils.py, line 504 at r3](https://reviewable.io/reviews/valloric/ycmd/1029#-LC_Ydkb4Rnm2ZOl4Pt0:-LC_bSkZDr8Br31ZWa7Z:b-uhma7n) ([raw file](https://github.com/valloric/ycmd/blob/3d2bd527c731bee1f3eb40bc7c03416f080b12c2/ycmd/utils.py#L504)):*

> Should we call a knife a knife and name this class UserDataDict ?

I feel the class is a little too generic to call it `UserDataDict`. Also, we could use it for other things like replacing `frozendict`.

> We could call it ImmutableHashableDict, but that is a bit of a mouthful

A hashable object must be immutable so calling it `ImmutableHashableDict` would be redundant.
</blockquote></details>

Ah ok then

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1029)*

puremourning:
@zzbot r+

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1029#-:-LC_bYr32PHB6DUCfvVB:bi59kwd)*

micbou:
@zzbot r+

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1029#-:-LC_zTbuEhFRLWTshZmR:bi59kwd)*

PR #1035, opened 2018-05-20
[READY] Prioritize compilation database over global extra conf

micbou:
Add the property `is_global_ycm_extra_conf` to the extra conf module and use this property to load flags in this order of priority:
 - from a `.ycm_extra_conf.py` file in the current directory or any parent directory;
 - from a `compile_commands.json` file in the current directory or any parent directory;
 - from the file set by the `global_ycm_extra_conf` option.

Closes https://github.com/Valloric/YouCompleteMe/issues/2777.

### DIFF
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -35,7 +35,6 @@ from ycmd import responses
 from ycmd.utils import re, ToBytes, ToCppStringCompatible, ToUnicode
 from ycmd.completers.completer import Completer
 from ycmd.completers.cpp.flags import ( Flags, PrepareFlagsForClang,
-                                        NoCompilationDatabase,
                                         UserIncludePaths )
 from ycmd.completers.cpp.ephemeral_values_set import EphemeralValuesSet
 from ycmd.completers.cpp.include_cache import IncludeCache, IncludeList
@@ -416,11 +415,8 @@ class ClangCompleter( Completer ):
       flags = []
       filename = request_data[ 'filepath' ]
 
-    try:
-      database_directory = self._flags.FindCompilationDatabase(
-          os.path.dirname( filename ) ).database_directory
-    except NoCompilationDatabase:
-      database_directory = None
+    database = self._flags.FindCompilationDatabase( filename )
+    database_directory = database.database_directory if database else None
 
     database_item = responses.DebugInfoItem(
       key = 'compilation database path',
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -81,10 +81,6 @@ EMPTY_FLAGS = {
 }
 
 
-class NoCompilationDatabase( Exception ):
-  pass
-
-
 class Flags( object ):
   """Keeps track of the flags necessary to compile a file.
   The flags are loaded from user-created python files (hereafter referred to as
@@ -132,17 +128,7 @@ class Flags( object ):
     except KeyError:
       pass
 
-    module = extra_conf_store.ModuleForSourceFile( filename )
-    try:
-      results = self._GetFlagsFromExtraConfOrDatabase( module,
-                                                       filename,
-                                                       client_data )
-    except NoCompilationDatabase:
-      if not self.no_extra_conf_file_warning_posted:
-        self.no_extra_conf_file_warning_posted = True
-        raise NoExtraConfDetected
-      return [], filename
-
+    results = self._GetFlagsFromExtraConfOrDatabase( filename, client_data )
     if not results or not results.get( 'flags_ready', True ):
       return [], filename
 
@@ -179,11 +165,25 @@ class Flags( object ):
     return sanitized_flags, filename
 
 
-  def _GetFlagsFromExtraConfOrDatabase( self, module, filename, client_data ):
-    if not module:
-      return self._GetFlagsFromCompilationDatabase( filename )
+  def _GetFlagsFromExtraConfOrDatabase( self, filename, client_data ):
+    module = extra_conf_store.ModuleForSourceFile( filename )
+    if module and not module.is_global_ycm_extra_conf:
+      return _CallExtraConfFlagsForFile( module, filename, client_data )
 
-    return _CallExtraConfFlagsForFile( module, filename, client_data )
+    database = self.FindCompilationDatabase( filename )
+    if database:
+      return self._GetFlagsFromCompilationDatabase( database, filename )
+
+    if module:
+      return _CallExtraConfFlagsForFile( module, filename, client_data )
+
+    # No compilation database and no extra conf found. Warn the user if not
+    # already warned.
+    if not self.no_extra_conf_file_warning_posted:
+      self.no_extra_conf_file_warning_posted = True
+      raise NoExtraConfDetected
+
+    return None
 
 
   def Clear( self ):
@@ -192,11 +192,10 @@ class Flags( object ):
     self.file_directory_heuristic_map.clear()
 
 
-  def _GetFlagsFromCompilationDatabase( self, file_name ):
+  def _GetFlagsFromCompilationDatabase( self, database, file_name ):
     file_dir = os.path.dirname( file_name )
-    file_root, file_extension = os.path.splitext( file_name )
+    _, file_extension = os.path.splitext( file_name )
 
-    database = self.FindCompilationDatabase( file_dir )
     compilation_info = _GetCompilationInfoForFile( database,
                                                    file_name,
                                                    file_extension )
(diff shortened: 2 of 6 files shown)


puremourning:
Just to be sure... this is a breaking change. I think it’s ok as long as we’re sure we want to make it. Personally, I don’t think this is a net improvement, but we know at least one person disagrees :)

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1035#-:-LCxMvkX8J4owIHYnC3G:b-m9zv8w)*

micbou:
This is only a breaking change if some users are relying on the current behavior which is highly doubtful.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1035#-:-LCxOyqoA5DbcnNPv9NM:b-loboww)*

bstaletic:
Yes, it is a breaking change technically. I personally think this is an improvement.
I really don't think anyone would create a fallback extra conf then create a project specific database just so he could say "nah... I like the generic stuff better".

---

Reviewed 1 of 6 files at r1.
Review status: 1 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1035#-:-LCxO3bz-_QC5u_nJpkD:bg5npbq)*

bstaletic:
Mostly looks good. :lgtm: 

We *could* make a configuration option that would set the desired prority order, but it would be far more complex than what I would be comfortable maintain. I vote for making this (breaking) change and just deal with (in my opinion unlikely) issue reports.

---

Reviewed 5 of 6 files at r1.

---

*[ycmd/completers/cpp/flags.py, line 177 at r1](https://reviewable.io/reviews/valloric/ycmd/1035#-LCxSSuN7JF2Notu3Vqe:-LCxSSuN7JF2Notu3Vqf:b-s4fdk9) ([raw file](https://github.com/valloric/ycmd/blob/96ade826eb692a0bec8bf27763206d6ab938b198/ycmd/completers/cpp/flags.py#L177)):*
> ```Python
>       return self._GetFlagsFromCompilationDatabase( database, filename )
> 
>     if module:
> ```

Maybe put a comment pointing out that this is a global extra conf.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1035#-:-LCxTAg93-GdESeeemZR:bgh5z6d)*

bstaletic:

---

*[ycmd/completers/cpp/flags.py, line 177 at r1](https://reviewable.io/reviews/valloric/ycmd/1035#-LCxSSuN7JF2Notu3Vqe:-LCx_KAQ0uLR_jVZht-B:b-lpa2gv) ([raw file](https://github.com/valloric/ycmd/blob/96ade826eb692a0bec8bf27763206d6ab938b198/ycmd/completers/cpp/flags.py#L177)):*

Maybe put a comment pointing out that this is a global extra conf.
</blockquote></details>

According to Codecov this is the only branch that has not been covered by the tests.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1035)*

micbou:
I'd like to add that the current behavior is neither documented nor tested. Yes, the changes in that PR are passing our current test suite.

---

Reviewed 6 of 6 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---

*[ycmd/completers/cpp/flags.py, line 177 at r1](https://reviewable.io/reviews/valloric/ycmd/1035#-LCxSSuN7JF2Notu3Vqe:-LCz49Rc3wIPe5xMx8RH:b-p12hoq) ([raw file](https://github.com/valloric/ycmd/blob/96ade826eb692a0bec8bf27763206d6ab938b198/ycmd/completers/cpp/flags.py#L177)):*

According to Codecov this is the only branch that has not been covered by the tests.
</blockquote></details>

Added some comments and a test that covers this branch.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1035#-:-LCz4LVJ4rbOYms-Jrp9:bgb5jfr)*

bstaletic:
That thought occurred to me at one point. Not sure what to think of that.

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1035#-:-LCzAQrA4BLrKxa5rsrl:bdaak39)*

puremourning:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1035#-:-LE-mQToAQi2NcXFE4XM:bnfp4nl)*

micbou:
@zzbot r+

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1035#-:-LE29Q4V5efm-Hb25OoF:bi59kwd)*

bstaletic:
Come on, be a good bot... @zzbot r+

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1035#-:-LE4SS2e6LR2zJLjjuXY:b-88yjbq)*

Fabbi:
you should probably warn users after doing such things.. So I will be the first Issue report:

I have a global config file which searches for local `compile_commands.json` and does some things afterwards (like finding the right compile commands for `foo.hpp` files that don't have a `foo.cpp` entry in `compile_commands.json` or use a `clang_complete` file if there is no `compile_commands.json` and so on..). I want this behaviour on every project without putting a `.ycm_extra_conf.py` in every folder, so now I have to wrap my module for this to work again...
Also Spacemacs uses a similar approach.

bstaletic:
We were aware that this was a breaking change. However there were other users who wanted this kind of change and, frankly, it makes sense to favour thte compilation database tailored for a specific project over a generic global conf.
Apologies for breaking your workflow, but as you can see, there is no way to make everybody happy.

micbou:
> finding the right compile commands for foo.hpp files that don't have a foo.cpp entry in compile_commands.json

What's the heuristic? If it's general enough, we could implement it directly in ycmd.

> use a clang_complete file if there is no compile_commands.json 

There is no issue in that case.

Fabbi:
>frankly, it makes sense to favour thte compilation database tailored for a specific project over a generic global conf.

Yeah, I get that. But knowing that would have made it easier to find out why the setup broke 😃 

> What's the heuristic? If it's general enough, we could implement it directly in ycmd.

The Idea is to find entries that include the header file and take the "nearest" one. `Spacemacs` has an (poorly written and flawed) example [global_conf.py](https://github.com/syl20bnr/spacemacs/blob/master/layers/%2Btools/ycmd/global_conf.py), but it can be used as an entry point. I could also see if I can make mine available; I rewrote their config. 
From their header comment:
```
# This script tries to get the compilation flags for a translation unit using
# the following logic:
#
# 1) If there is a compile_commands.json in a parent directory:
#   a) If the file is a header file:
#     - search for the header file itself in db
#     - search for a sibling source file in the same directory (i.e. a source
#       file with the same name but different extension)
#     - search for a source file that includes our header's path
#     - search for the nearest source file in db
#
#   b) If the file is a source file:
#     - search for the source file itself
#     - search for the nearest source file in db
#
# 2) If no compile_commands.json, search for a .clang_complete:
#   - get flags from .clang_complete
#
# 3) Always try to add extra flags from a .ycm_extra_flags file in a parent
#    directory. (like --sysroot="/path/to/your/toolchain/libc" if you are cross-compiling)
```

bstaletic:
For header files, ycmd does check the database and looks for "a sibling source file". However we don't currently look for a source file that includes the header. I am also not sure what "search for the nearest source file in db" means.

Do you have an example that would look for a source file that includes the requested header?

Fabbi:
I opened a new thread for this conversation

sam-lunt:
Just wanted to chime in that this change was a problem for me as well. My global `ycm_extra_conf.py` script was getting flags for headers that didn't correspond to an implementation file, but now those headers have no flags, not even reasonable defaults.

I think that determining whether a header is included by a nearby file is probably overkill, but it would be nice if the presence of a compilation_database didn't shut down header files so completely. For example, my script was checking if there were any `.cpp` / `.c` / `.cc` files in the same directory. If there was and they all used the same flags, then those flags were used for the header as well.

I'm sure there's better logic that could be used, but I think it's important to make some attempt for headers beyond `s/.hpp$/.cpp/` (etc.).

If there's a better place to discuss this, please let me know. I'd be happy to create a pull request myself, but I figured this merits some discussion first.

bstaletic:
@sam-lunt It doesn't "shut down header files completely". Checking for `foo.cpp`, `foo.cc` etc in the same directory as the opened `foo.h` is already done, but we're repeating ourselves. The discussion is in #1063.

puremourning:
the extra config is preferred over the compilation database if it is an ancestor directory in the filesystem. presumably you are using the "global extra conf" setting? 

If you put the "global" extra conf in your home directory, then it will load the extra conf file rather than the compilation database, so you are in control.

sam-lunt:
@puremourning Thanks, that's a good idea.

Issue #847, opened 2017-09-27
Using TypeScript for filetype javascript


maximus8891:
Method [PathToFiletypeCompleterPluginLoader](https://github.com/Valloric/ycmd/blob/master/ycmd/completers/completer_utils.py#L162) searches for completers for a given filetype by their folder names. This way you can't have several completers handling one filetype.

I discovered this while I was looking for ways to apply the TypeScript completer for javascript files abandoning Tern altogether (it does not support class parameters).

Since each completer has method `SupportedFiletypes`, I see it as a better place to test if a completer can be used for a filetype. I would have to extend `[ 'typescript' ]` to `['typescript', 'javascript']`.  Of course, you wouldn't want to instantiate each completer to do this check. It makes more sense to make this method static.

bstaletic:
> This way you can't have several completers handling one filetype.

The thing is, we don't want to support more then one completer per filetype.  It would be a big maintainance burden.

What we do want, though, is having the best available completer for a certain filetype. So even though we won't be making the suggested change, I think this is a good opportunity to discuss if we should use TSserver for javascript and abandon Tern.

Let me ask the first question. Is there anything TSserver does worse than Tern?

maximus8891:
Tern does not support class properties: https://babeljs.io/docs/plugins/transform-class-properties/. They are in-built in Typescript.

For me TSserver required no config, unlike Tern, which wanted for some reason to have .tern-project file. I found it hard to find docs and examples for Tern.

With my js project setup TSserver found definitions in various files, while Tern could not handle it.

These points might sound a bit random, but I had a better experience with TS.

micbou:
I created [a branch that enables TSServer for JS files](https://github.com/micbou/YouCompleteMe/tree/tsserver-javascript) so that we can easily compare both engines ([TSServer](https://github.com/Microsoft/TypeScript/wiki/Standalone-Server-(tsserver)) and [Tern](http://ternjs.net/)). Feel free to try it and report back.

An obvious advantage for TSServer is that it has built-in diagnostics while Tern doesn't (there is [a lint plugin for Tern](https://github.com/angelozerr/tern-lint) but it doesn't seem to be maintained).

maximus8891:
Thanks @micbou, I'll give it a whirl.
I also managed to make it work by copying the content of typescript folder into the javascript folder.

puremourning:
Are class properties a JavaScript language feature? Have you asked the tern people about supporting it?

[...]

PR #1036, opened 2018-05-21
[READY] Use TypeScript completer for JavaScript

micbou:
Since [Tern is not maintained anymore](https://github.com/ternjs/tern#tern) and TSServer offers a lot more features (automatic import, diagnostics, and the `GoToType`, `FixIt`, `Format`, and `OrganizeImports` commands), we should use the TypeScript completer for JavaScript. However, we don't want to break our Tern users so we fall back to Tern if TypeScript is not installed.

Move the Tern tests to a separate folder and adapt the TypeScript tests to JavaScript.

Closes https://github.com/Valloric/ycmd/issues/847.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036)


bstaletic:
> However, we don't want to break our Tern users so we fall back to Tern if TypeScript is not installed.

This *could* still be a breaking change if there's a user who prefers Tern for javascript and TSServer for typescript.

All around, this looks good to me, but I'll let @puremourning be the judge. One question though. If we are going to do this, considering that we don't usually have two servers for one language, shouldn't we have a plan for eventually dropping Tern?

---

Reviewed 54 of 55 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036#-:-LD0zoXnB7rjWPj-9TJ8:b-h8l512)*

micbou:
> This could still be a breaking change if there's a user who prefers Tern for javascript and TSServer for typescript.

That's true. Maybe we need an option after all. The issue with an option is that it still requires the user to restart the server which is inconvenient.

> If we are going to do this, considering that we don't usually have two servers for one language, shouldn't we have a plan for eventually dropping Tern?

Put the Tern completer in low maintenance mode and drop it when it becomes irrelevant?

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036#-:-LD15xhy4wemS0m6Hjcr:bbzxedk)*

ycmd/tests/javascript/testdata/jsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
```
maximus8891: Is the config filename supposed to be `jsconfig.json` or `tsconfig.json`?

micbou: Both are supported. The only difference is that `allowJs` must be set to `true` in `tsconfig.json` for JavaScript projects:
```json
{
  "compilerOptions": {
    "allowJs": true
  }
}
```

maximus8891: 👍

bstaletic:
Reviewed 6 of 6 files at r2.

---

*[ycmd/tests/shutdown_test.py, line 93 at r2](https://reviewable.io/reviews/valloric/ycmd/1036#-LD50GSV81KtlmwWMxgx:-LD50GSV81KtlmwWMxgy:b-rdwx3h) ([raw file](https://github.com/valloric/ycmd/blob/1e5bb67a4fdae6b426e2a82e4919c68a00ed2965/ycmd/tests/shutdown_test.py#L93)):*
> ```Python
>                   'python',
>                   'typescript' ]
>                   'rust' ]
> ```

These two closing brackets are causing the tests to fail.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036)*

micbou:
Added the `use_tern_completer` option to force the use of Tern for JavaScript even if TypeScript is installed. The option is disabled by default.

---

Reviewed 51 of 55 files at r1, 6 of 6 files at r2, 2 of 2 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036#-:-LD7XWde9ocOKx-P9u1e:b-c7divk)*

puremourning:
I still worry that this change is breaking by default. I have typescript installed and I build ycmd with --js-completer, but I use and have configuration for tern in my projects (i.e. a .tern-project). If I install this version of ycmd, my project stops working. I don't think we want that.

Perhaps we need to be more sympathetic with our compatibility? For example, if we detect a .tern-project (which we already do) and tern is set up, use the Tern completer, otherwise use the typescript completer ?

---

---

*[a discussion](https://reviewable.io/reviews/valloric/ycmd/1036#-LE-iYsf7IrRKxPLz3vj:-LE-iYsf7IrRKxPLz3vk:bddw2y7) (no related file):*
Does tsserver support things like node and jquery plugins/depencies etc. like tern does? If so do we know how you set them up? How do we set up things like dependency paths when using require js and other such? I assume this is all handled in the jsconfig.json ?

---

*[ycmd/completers/javascript/hook.py, line 32 at r5](https://reviewable.io/reviews/valloric/ycmd/1036#-LE-egC4DbI96G-1E67a:-LE-egC4DbI96G-1E67b:b-f1x8hd) ([raw file](https://github.com/valloric/ycmd/blob/92777a58139a38d24dadd643237df182d658a54f/ycmd/completers/javascript/hook.py#L32)):*
> ```Python
> 
> def GetCompleter( user_options ):
>   if ( not user_options[ 'use_tern_completer' ] and
> ```

I would rather avoid having an apology-option if we can avoid it. I think an alternative would be have ShouldEnableTernConpleter start returning False when .tern-project isn't found.

---

*[ycmd/tests/javascript/testdata/imports.js, line 1 at r5](https://reviewable.io/reviews/valloric/ycmd/1036#-LE-hXVw0m6bE67NcHmS:-LE-hXVw0m6bE67NcHmT:b6fjzat) ([raw file](https://github.com/valloric/ycmd/blob/92777a58139a38d24dadd643237df182d658a54f/ycmd/tests/javascript/testdata/imports.js#L1)):*
> ```JavaScript
> import func from "library";
> ```

IIRC this is ECMAscript 5? How does one tell tsserver to use previous versions? Is there some typescript config file that we should document/link to etc.?

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036#-:-LE-dPql8hEBj7GpRnyQ:b4iswnt)*

micbou:
> For example, if we detect a .tern-project (which we already do) and tern is set up, use the Tern completer, otherwise use the typescript completer ?

We can't detect if a `.tern-project` file exists because the request is not available when loading a completer. If we really want to be backward compatible, we first need to check if ycmd was built with `--js-completer` and use the Tern completer if that's the case; otherwise, we fall back to the TypeScript completer if TypeScript is installed. My concern with that approach is that if the user want to use the TypeScript completer for JavaScript but Tern has already been installed through the `--js-completer` option, he needs to manually remove the `third_party/tern_runtime/node_modules` folder. This is something that we'll need to document.

---

---

*[a discussion](https://reviewable.io/reviews/valloric/ycmd/1036#-LE-iYsf7IrRKxPLz3vj:-LE6WTxyBDwsd5Kg3D1S:bk2r864) (no related file):*
> Does tsserver support things like node and jquery plugins/depencies etc. like tern does? If so do we know how you set them up?

Yes, TSServer does it by looking at the dependencies in `package.json`. The dependencies can also be set in the `jsconfig.json` file. See https://code.visualstudio.com/Docs/languages/javascript#_automatic-type-acquisition.

> How do we set up things like dependency paths when using require js and other such?

[AMD is not supported](https://github.com/Microsoft/TypeScript/issues/6942) and will probably never be (since the technology is obsolete). That's IMO the only reason to keep Tern.

---

*[ycmd/completers/javascript/hook.py, line 32 at r5](https://reviewable.io/reviews/valloric/ycmd/1036#-LE-egC4DbI96G-1E67a:-LE6c4xwAruuRqrnEeHs:bnuzhhj) ([raw file](https://github.com/valloric/ycmd/blob/92777a58139a38d24dadd643237df182d658a54f/ycmd/completers/javascript/hook.py#L32)):*

I would rather avoid having an apology-option if we can avoid it. I think an alternative would be have ShouldEnableTernConpleter start returning False when .tern-project isn't found.
</blockquote></details>

I removed the option. We now attempt to load the Tern completer first then the TypeScript completer.

---

*[ycmd/tests/javascript/testdata/imports.js, line 1 at r5](https://reviewable.io/reviews/valloric/ycmd/1036#-LE-hXVw0m6bE67NcHmS:-LE6XoBv395i4VfYOSvj:b-lbbftv) ([raw file](https://github.com/valloric/ycmd/blob/92777a58139a38d24dadd643237df182d658a54f/ycmd/tests/javascript/testdata/imports.js#L1)):*
> IIRC this is ECMAscript 5?

No, it's ES2015 (ES 6). 

> How does one tell tsserver to use previous versions?

By specifying the `target` option in `jsconfig.json` (or `tsconfig.json`) e.g.
```json
{
    "compilerOptions": {
        "target": "es5"
    }
}
```

> Is there some typescript config file that we should document/link to etc.?

We could include these links:
 - https://code.visualstudio.com/Docs/languages/javascript
 - https://code.visualstudio.com/docs/languages/jsconfig
 - http://json.schemastore.org/jsconfig

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036#-:-LE6Rmii812QEwlsIKmd:b-t0nngw)*

puremourning:
Right that makes sense.

---

Reviewed 9 of 13 files at r6, 2 of 2 files at r7.

---

*[a discussion](https://reviewable.io/reviews/valloric/ycmd/1036#-LE-iYsf7IrRKxPLz3vj:-LEdwdj-8aqBwwRDKaX6:bkmg7zg) (no related file):*

> Does tsserver support things like node and jquery plugins/depencies etc. like tern does? If so do we know how you set them up?

Yes, TSServer does it by looking at the dependencies in `package.json`. The dependencies can also be set in the `jsconfig.json` file. See https://code.visualstudio.com/Docs/languages/javascript#_automatic-type-acquisition.

> How do we set up things like dependency paths when using require js and other such?

[AMD is not supported](https://github.com/Microsoft/TypeScript/issues/6942) and will probably never be (since the technology is obsolete). That's IMO the only reason to keep Tern.
</blockquote></details>

So tsserver assumes that you use npm ?

what's AMD ?  

I love how things in javascript land become obsolete. Doesn't matter how many people are using them, someone on the internet declares that it is obsolete and that some equally crappy thing has taken its place. 

Obsolete means no longer used or useful. Not that it is no longer recommended. slash rant.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036#-:-LEdxVbw44JcyS1WJj0O:b-mx9h11)*

micbou:

---

*[a discussion](https://reviewable.io/reviews/valloric/ycmd/1036#-LE-iYsf7IrRKxPLz3vj:-LEe-8hq6FuJnYMiAM-p:bfn2now) (no related file):*
> So tsserver assumes that you use npm ?

Most likely since you need npm to install it but I can't say for sure.

> what's AMD ?

AMD stands for Asynchronous Module Definition and is the API used by requirejs to load modules.

> I love how things in javascript land become obsolete. Doesn't matter how many people are using them, someone on the internet declares that it is obsolete and that some equally crappy thing has taken its place.

> Obsolete means no longer used or useful. Not that it is no longer recommended. slash rant.

I suppose "superseded" would be a better choice of word.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036)*

Valloric:
Normally I'd be very concerned with us breaking people by switching to a different backend. But since Tern is now unmaintained, IMO this concern should be much relaxed. Whoever is using Tern today will eventually have to migrate away from it because it's dead.

Thus, I think even a hard break where we stop supporting Tern completely might be acceptable. We could do this in stages where we deprecate Tern support and rip it out a few months later.

Thoughts?

---

Review status: 0 of 2 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036#-:-LEvQMSID5GksGdILGbj:b-roj7m9)*

puremourning:
As I said before, tern isn’t dead exactly: it still works and I still use it, and tsserver doesn’t support things in it that I need.

 So in a survey of 1, 100% of Bens still need tern support.

 If that means I have to maintain it in my fork, so be it, but I would be surprised if me and my colleagues are the only  ones on that situation. 

---

Review status: 0 of 2 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036#-:-LExB8g0Cj9kCiansG3e:b-pv2q1v)*

maximus8891:
What are those features that tsserver doesn't support, out of curiosity?

puremourning:
requirejs

micbou:
@puremourning The current implications of the PR should satisfy you:
 - nothing changes for previous users of Tern;
 - TSServer will be used for new users unless the `--js-completer` or `--tern-completer` options are passed when running the `install.py` script; in that case, Tern will be used. These options are hidden to encourage people to use TSServer;
 - for users that were previously using Tern and want to switch to TSServer, they need to manually remove the `third_party/tern_runtime/node_modules` folder.

If the PR is merged, I propose to do the following changes in the YCM documentation:
 - update the JavaScript features summary with the TSServer ones;
 - remove all mention of the `--js-completer` option;
 - move the Tern section to its own page on the wiki;
 - add a link to the wiki for people that still want to use Tern;
 - give instructions on how to switch to TSServer if Tern were previously used.

What I would suggest about maintenance is to keep it at a minimum:
 - no changes should be done to the Tern completer;
 - if some of the Tern tests are failing, remove them.

> What are those features that tsserver doesn't support, out of curiosity?

[AMD support](https://github.com/Microsoft/TypeScript/issues/6942) which is used by [requirejs](http://requirejs.org/).

---

Reviewed 5 of 5 files at r8, 2 of 2 files at r9, 1 of 1 files at r10.
Review status: 0 of 2 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036#-:-LExWHs40sqaeZR9tM96:bzdfz1e)*

kevindurb:
Is there any way i can help to get this in? this would be an amazing feature for me.

micbou:
> I'd also add that the wiki page for Tern should mention that support for it is deprecated.

Agreed.

> Is there any way i can help to get this in? this would be an amazing feature for me.

Not much. It's ready to be merged. This just needs approval from @puremourning.

---

Review status: 0 of 2 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036#-:-LGQBF-B8xTZvzUYfIZM:b-x2fzam)*

bstaletic:
Reviewed 2 of 2 files at r3, 1 of 1 files at r5, 3 of 13 files at r6, 1 of 5 files at r8.
Review status: 1 of 2 LGTMs obtained

---

*[a discussion](https://reviewable.io/reviews/valloric/ycmd/1036#-LE-iYsf7IrRKxPLz3vj:-LGRi0Yk0NzGngmnq2xC:b-b1chrl) (no related file):*

OK. :lgtm:

One thing we should consider doing on the wiki is publish some sort of tsserver migration guide. Even if it is just a link to the new documentation and a statement about why we changed it. For many users, I expect it won't be completely obvious.

Finally, does anyone know what "drives" tsserver's diagnostics? My use case is that we use jshint and our build will fail if jshint fails with our specific config (a bit like ycmd with flake8). I rely on syntactic to run jshint for me. Should I migrate to tsserver, this would (probably?) stop working. Typically we attempt to disable syntactic when the language supports diagnostics (presumably we'll need a client change for this anyway to include javascript in the diags list). Thoughts on that?
</blockquote></details>

A wiki page does sound useful.

Quick googling didn't return anything useful for some combinations of "jshint" and "tsserver".

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036)*

micbou:
Review status: 1 of 2 LGTMs obtained

---

*[a discussion](https://reviewable.io/reviews/valloric/ycmd/1036#-LE-iYsf7IrRKxPLz3vj:-LGRgk3BAqVMflTO6Ayu:b-1winhg) (no related file):*
> One thing we should consider doing on the wiki is publish some sort of tsserver migration guide. Even if it is just a link to the new documentation and a statement about why we changed it. For many users, I expect it won't be completely obvious.

Sure.

> Finally, does anyone know what "drives" tsserver's diagnostics?

The TypeScript compiler `tsc`.

> Typically we attempt to disable syntactic when the language supports diagnostics (presumably we'll need a client change for this anyway to include javascript in the diags list). Thoughts on that?

I think we shouldn't do that because it can be useful to get diagnostics from multiple sources.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036)*

fftyjw:
Hope this can be merged as soon as possible, it would be very useful for me.

bstaletic:
This is :lgtm: as far as I can see. Since I don't use javascript and don't have a great idea about tern vs tsserver, I'll let someone else call @zzbot.

---

Reviewed 3 of 13 files at r6, 1 of 2 files at r7, 1 of 5 files at r8, 1 of 1 files at r10.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1036#-:-LE-jUcW6LdUeh9Pmmm6:b-ai6puf)*

Valloric:
@zzbot status

micbou:
@zzbot r+

micbou:
@zzbot retry

bstaletic:
@zzbot r+

PR #1038, opened 2018-05-23
[READY] Fix GetDoc command on symbols declared in system headers

micbou:
The `GetDoc` command fails to return documentation on symbols declared in system headers and, in particular, headers included with the `-isystem` flag. This is due to libclang ignoring comments from these headers.

If a symbol is declared in such header, get the documentation directly from the corresponding translation unit. This works because comments in the main file of a translation unit are not ignored.

Fixes https://github.com/Valloric/YouCompleteMe/issues/3020.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1038)


bstaletic:
I left a comment, but this is still :lgtm:.

For what it's worth, my solution to the same problem was a little more complicated. Nice work.

---

Reviewed 8 of 8 files at r1, 1 of 1 files at r2.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 592 at r1](https://reviewable.io/reviews/valloric/ycmd/1038#-LDDNAus7FEGTI0UYzBZ:-LDDNAus7FEGTI0UYzB_:b-ij9nik) ([raw file](https://github.com/valloric/ycmd/blob/cdc2fba7db9ee6609eb31f8f4a9ee052fcbb2dc8/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L592)):*
> ```cpp
>   CXFile file = clang_getFile( clang_translation_unit_,
>                                location.filename_.c_str() );
>   CXSourceLocation source_location = clang_getLocation(
> ```

`TranslationUnit::GetCursor()` and this function share the same algorithm that "returns" a `CXSourceLocation` from `clang_translation_unit_` and a location. Maybe we should move those lines in `GetSourceLocation()`.

Something like:
```c++
CXSourceLocation GetSourceLocation( std::string filename, int line, int column ) {
  CXFile file = clang_getFile( clang_translation_unit_, filename.c_str() );
  return clang_getLocation( clang_translation_unit_, file, line, column );
}
```

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1038#-:-LDDP1KM7xlO10AwKLUm:b-3fttju)*

micbou:
Reviewed 7 of 8 files at r1, 1 of 1 files at r2, 3 of 3 files at r3.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.cpp, line 592 at r1](https://reviewable.io/reviews/valloric/ycmd/1038#-LDDNAus7FEGTI0UYzBZ:-LDDfJ_Z1JVbFi09y-mK:b-896fix) ([raw file](https://github.com/valloric/ycmd/blob/cdc2fba7db9ee6609eb31f8f4a9ee052fcbb2dc8/cpp/ycm/ClangCompleter/TranslationUnit.cpp#L592)):*

`TranslationUnit::GetCursor()` and this function share the same algorithm that "returns" a `CXSourceLocation` from `clang_translation_unit_` and a location. Maybe we should move those lines in `GetSourceLocation()`.

Something like:
```c++
CXSourceLocation GetSourceLocation( std::string filename, int line, int column ) {
  CXFile file = clang_getFile( clang_translation_unit_, filename.c_str() );
  return clang_getLocation( clang_translation_unit_, file, line, column );
}
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1038)*

micbou:
Bumped the core version.

---

Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1038#-:-LDGMwH6EcWzglhttfyL:b-ks6e1q)*

bstaletic:
Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---

*[cpp/ycm/ClangCompleter/TranslationUnit.h, line 124 at r4](https://reviewable.io/reviews/valloric/ycmd/1038#-LDGTT6_8AM9dkU5LnoD:-LDGTT6_8AM9dkU5LnoE:b-lowcu0) ([raw file](https://github.com/valloric/ycmd/blob/c6c1d894f19dc7a9c9e0c93e2bd7583af432259e/cpp/ycm/ClangCompleter/TranslationUnit.h#L124)):*
> ```objc
>   void UpdateLatestDiagnostics();
> 
>   // These four methods must be called under the clang_access_mutex_ lock.
> ```

Unrelated to this PR, but we should take another look at [the clang thread safety annotations](https://releases.llvm.org/6.0.0/tools/clang/docs/ThreadSafetyAnalysis.html).

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1038)*

puremourning:
:lgtm: with one comment

---

Reviewed 5 of 8 files at r1, 3 of 3 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.

---

*[cpp/ycm/ClangCompleter/ClangCompleter.cpp, line 272 at r5](https://reviewable.io/reviews/valloric/ycmd/1038#-LE-YobPCzwDUIunHHmq:-LE-YobPCzwDUIunHHmr:bhqmhbj) ([raw file](https://github.com/valloric/ycmd/blob/f60f125fed5e33528bad691d5f40d6ed5b74d885/cpp/ycm/ClangCompleter/ClangCompleter.cpp#L272)):*
> ```cpp
>                                                    column,
>                                                    unsaved_files,
>                                                    reparse ) );
> ```

by passing reparse here and later (in GetDocsForLocation) we invoke 2x reparse operations, which could be significant on large TUs. I suppose we can always pass FALSE to GetDocsForLocation because:

- if we created a new TU, it just got parsed
- otherwise, if reparse is true, we did it here, else we don't want to anyway.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1038#-:-LE-_Ss2B-jlsVsr3qbx:b-2ecocm)*

micbou:

---

*[cpp/ycm/ClangCompleter/ClangCompleter.cpp, line 272 at r5](https://reviewable.io/reviews/valloric/ycmd/1038#-LE-YobPCzwDUIunHHmq:-LE-f5Wv1Uue8oQ_pouW:bclafwf) ([raw file](https://github.com/valloric/ycmd/blob/f60f125fed5e33528bad691d5f40d6ed5b74d885/cpp/ycm/ClangCompleter/ClangCompleter.cpp#L272)):*

Done. I added a comment explaining why we never parse a second time.
</blockquote></details>

Now that I think about it, are we sure we are parsing the translation unit when we create it? If that's the case then why are we always passing `reparse` to `GetDeclarationLocation`?

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1038)*

micbou:

---

*[cpp/ycm/ClangCompleter/ClangCompleter.cpp, line 272 at r5](https://reviewable.io/reviews/valloric/ycmd/1038#-LE-YobPCzwDUIunHHmq:-LE-fh5q1yYasOd6wXDV:b-is2ycr) ([raw file](https://github.com/valloric/ycmd/blob/f60f125fed5e33528bad691d5f40d6ed5b74d885/cpp/ycm/ClangCompleter/ClangCompleter.cpp#L272)):*

Now that I think about it, are we sure we are parsing the translation unit when we create it? If that's the case then why are we always passing `reparse` to `GetDeclarationLocation`?
</blockquote></details>

Maybe we need to do this instead:
```cpp
if ( unit->LocationIsInSystemHeader( location ) ) {
  unit = translation_unit_store_.GetOrCreate( location.filename_,
                                                unsaved_files,
                                                flags );
  return unit->GetDocsForLocation( location, unsaved_files, reparse );
}
return unit->GetDocsForLocation( location, unsaved_files, false );
```

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1038)*

micbou:
Reviewed 1 of 1 files at r7, 1 of 1 files at r8.

---

*[cpp/ycm/ClangCompleter/ClangCompleter.cpp, line 272 at r5](https://reviewable.io/reviews/valloric/ycmd/1038#-LE-YobPCzwDUIunHHmq:-LEP-D8C-JrVukfc50gD:b3du0ct) ([raw file](https://github.com/valloric/ycmd/blob/f60f125fed5e33528bad691d5f40d6ed5b74d885/cpp/ycm/ClangCompleter/ClangCompleter.cpp#L272)):*

Maybe we need to do this instead:
```cpp
if ( unit->LocationIsInSystemHeader( location ) ) {
  unit = translation_unit_store_.GetOrCreate( location.filename_,
                                                unsaved_files,
                                                flags );
  return unit->GetDocsForLocation( location, unsaved_files, reparse );
}
return unit->GetDocsForLocation( location, unsaved_files, false );
```
</blockquote></details>

I went for only passing the `reparse` parameter if the location is in a system header and not reparsing otherwise. The reason is that we can imagine a scenario where the docstring of a system header is modified by an external source while editing code.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1038)*

puremourning:
:lgtm:

---

Reviewed 1 of 1 files at r7, 1 of 1 files at r8.

---

*[cpp/ycm/ClangCompleter/ClangCompleter.cpp, line 272 at r5](https://reviewable.io/reviews/valloric/ycmd/1038#-LE-YobPCzwDUIunHHmq:-LEdwKos0GB0MVnEDSyO:b32wtd) ([raw file](https://github.com/valloric/ycmd/blob/f60f125fed5e33528bad691d5f40d6ed5b74d885/cpp/ycm/ClangCompleter/ClangCompleter.cpp#L272)):*

I went for only passing the `reparse` parameter if the location is in a system header and not reparsing otherwise. The reason is that we can imagine a scenario where the docstring of a system header is modified by an external source while editing code.
</blockquote></details>

ok

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1038#-:-LEdwSx79qNyQIUOh_ci:bnfp4nl)*

micbou:
Thanks for the review!

@zzbot r++

---

Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1038#-:-LEe0nxz7SzTbghFReDf:b6hpibu)*

micbou:
@zzbot r+

---

Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1038#-:-LEe0wMO1LbwTQ26mUrV:bi59kwd)*

PR #1040, opened 2018-05-23
[READY] Use equals default on structs

bstaletic:
Clang-tidy fails to detect these for `struct`s. More importantly, if a constructor is declared as `Foo(){};`, even though it is no-op, it needs to be called upon object creation, but if it is declared as `Foo() = default;` then the standard calls it a constructor with a trivial ABI and the compiler is allowed to skip calling a trivial ABI constructor. This optimisation is present in all three compilers that we support.

If we want to have cang-tidy report these consistently, we have to ban `struct{int foo;};` in favour of `class{public: int foo;};`.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1040)

### DIFF
--- a/cpp/ycm/ClangCompleter/CompletionData.h
+++ b/cpp/ycm/ClangCompleter/CompletionData.h
@@ -50,7 +50,7 @@ enum class CompletionKind {
 // The user can also enable a "preview" window that will show extra information
 // about a completion at the top of the buffer.
 struct CompletionData {
-  CompletionData() {}
+  CompletionData() = default;
   CompletionData( const CXCompletionResult &completion_result );
 
   // What should actually be inserted into the buffer. For a function like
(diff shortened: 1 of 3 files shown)


micbou:
:lgtm_strong:.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1040#-:-LDDx1NDDSS5gW-1C34Y:b-uwkovz)*

puremourning:
:lgtm: @zzbot r+

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1040#-:-LDEKCojEl6OoFPAKxX1:by5ago3)*

bstaletic:
@zzbot retry

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1040#-:-LDFZ7k81hVeeBj1oaq2:b-kc0oa4)*

PR #1043, opened 2018-05-24
Define __CUDACC__ for CUDA source files

midchildan:
Sorry for the repeated PRs. I've just noticed that the standard CUDA runtime library headers make extensive use of the `__CUDACC__` macro, which should be defined when clang or nvcc compiles CUDA source files [1]. For some reason, it appears that the clang completer doesn't define `__CUDACC__`, which causes problems when parsing headers such as `cuda_runtime_api.h`. The changes in this PR inject `-D__CUDACC__` when CUDA source files are found.

[1] https://llvm.org/docs/CompileCudaWithLLVM.html

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1043)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -390,7 +390,7 @@ def _AddLanguageFlagWhenAppropriate( flags, enable_windows_style_flags ):
   # compiling CUDA source files with a C++ compiler
   if any( fl.endswith( '.cu' ) or fl.endswith( '.cuh' )
           for fl in reversed( flags ) ):
-    return [ first_flag, '-x', 'cuda' ] + flags[ 1: ]
+    return [ first_flag, '-x', 'cuda', '-D__CUDACC__' ] + flags[ 1: ]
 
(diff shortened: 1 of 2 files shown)


micbou:
That's strange. If I manually add the `-D__CUDACC__` flag for [this code](https://gist.github.com/anonymous/855e277884eb6b388cd2f00d956c2fd4), I get the warning that the `__CUDACC__` macro is redefined. In fact, the macro seems to be already defined in the `clang_includes/include/__clang_cuda_runtime_wrapper.h` file (which is included through the `-resource-dir` flag) so I am not sure about this change.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1043#-:-LDHv5tkC2X7Ljsg2Bgu:btxx6fp)*

midchildan:
I did some further investigation and I couldn't reproduce this problem on a different computer. I'll close this PR for now, since its starting to look like an issue with my local CUDA installation.

PR #1045, opened 2018-05-26
[RFC] Use Clang thread safety analysis annotations

bstaletic:
This PR
- Adds `-Wthread-safety` to CMake and extra conf.
- Wraps STL mutexes with annotated classes in `cpp/ycm/Mutex.h`.
- Places appropriate annotations on member variables and methods.
- Places NO_THREAD_SAFETY_ANALYSIS on methods deemed unsafe.
- Bumps travis clang version to 3.5

### Wrapping STL

I'm not too happy about this because, as it stands now, nothing prevents anyone from using `std::mutex` and thus avoid the thread safety annotations.

Libc++ is supposed to have its `<mutex>` header properly annotated since 2016 according to this [PR](https://reviews.llvm.org/D14731), but I couldn't get it to work when I tried passing `-stdlib=libc++` to clang. [This gist](https://gist.github.com/bstaletic/9084c975107db6c5d1b5e7396514ca11) can be used for testing non-wrapped STL mutexes and needs to be compiled with `-Wthread-safety`. If successful, it should produce exactly one warning with the following content:

```
foo.cpp:24:3: warning: mutex 'mu' is still held at the end of function [-Wthread-safety-analysis]
  }                          // WARNING!  Failed to unlock mu. - Not triggered with UniqueLock or LockGuard
  ^
foo.cpp:22:8: note: mutex acquired here
    mu.lock();
       ^
1 warning generated.
```

### NO_THREAD_SAFETY_ANALYSIS

This appeared in six places throughout our codebase:
- `CandidateRepository::ClearCandidates`
- `CharacterRepository::ClearCharacters`
- `CodePointRepository::ClearCodePoints`
- `IdentifierDatabase::GetCandidateSet`
- `TranslationUnit::IsCurrentlyUpdating`
- `TranslationUnit::Reparse( std::vector< CXUnsavedFile > )`

`NO_THREAD_SAFETY_ANALYSIS` simply silences the thread safety warnings inside a single function. Other options include really adding the locks that clang says are missing or specifying that the caller of these six functions must already hold a mutex before making the call.
The reason I didn't just put locks whenever clang complained is because some of these are in places where performance really matters and we all know that another word for "a mutex" is "a bottleneck".

### Clang 3.5

Clang 3.4 doen't have support for thread annotations, so I had to increase the clang version on travis. We could introduce `-DTHREAD_SAFETY_ANNOTATIONS` which would be guarded by `if( COMPILER_IS_CLANG AND NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.4 )`.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1045)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ addons:
      # 4.8 is the first version of GCC with good enough C++11 support.
      - g++-4.8
      # Install Clang 3.4 and its standard library.
-     - clang
+     - clang-3.5
      - clang-tidy-3.9
      - libc++-dev
      # Required to build the OmniSharp server.
--- a/ci/travis/travis_install.sh
+++ b/ci/travis/travis_install.sh
@@ -18,8 +18,8 @@ unset -f cd popd pushd
 mkdir -p ${HOME}/bin
 
 if [ "${YCM_COMPILER}" == "clang" ]; then
-  ln -s /usr/bin/clang++ ${HOME}/bin/c++
-  ln -s /usr/bin/clang ${HOME}/bin/cc
+  ln -s /usr/bin/clang++-3.5 ${HOME}/bin/c++
+  ln -s /usr/bin/clang-3.5 ${HOME}/bin/cc
   # Tell CMake to compile with libc++ when using Clang.
   export EXTRA_CMAKE_ARGS="${EXTRA_CMAKE_ARGS} -DHAS_LIBCXX11=ON"
 else
--- a/cpp/ycm/.ycm_extra_conf.py
+++ b/cpp/ycm/.ycm_extra_conf.py
@@ -42,6 +42,7 @@ flags = [
 '-Werror',
 '-Wno-long-long',
 '-Wno-variadic-macros',
+'-Wthread-safety',
 '-fexceptions',
 '-DNDEBUG',
 # You 100% do NOT need -DUSE_CLANG_COMPLETER and/or -DYCM_EXPORT in your flags;
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -451,6 +451,12 @@ endif()
 
 #############################################################################
 
+if( COMPILER_IS_CLANG )
+  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wthread-safety" )
+endif()
+
+#############################################################################
+
 if( SYSTEM_IS_SUNOS )
   # SunOS needs this setting for thread support
   set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthreads" )
--- a/cpp/ycm/CandidateRepository.cpp
+++ b/cpp/ycm/CandidateRepository.cpp
@@ -25,6 +25,9 @@
 #  include "ClangCompleter/CompletionData.h"
 #endif // USE_CLANG_COMPLETER
 
+using std::lock_guard;
+using std::mutex;
+
 namespace YouCompleteMe {
 
 namespace {
@@ -37,12 +40,12 @@ const size_t MAX_CANDIDATE_SIZE = 80;
 
 
 CandidateRepository *CandidateRepository::instance_ = nullptr;
-std::mutex CandidateRepository::instance_mutex_;
+Mutex CandidateRepository::instance_mutex_;
 
 CandidateRepository &CandidateRepository::Instance() {
   // This lock is required as magic statics are not thread-safe on MSVC 12.
   // See https://msdn.microsoft.com/en-us/library/hh567368#concurrencytable
-  std::lock_guard< std::mutex > locker( instance_mutex_ );
+  LockWrapper< lock_guard< mutex > > locker( instance_mutex_ );
 
   if ( !instance_ ) {
     static CandidateRepository repo;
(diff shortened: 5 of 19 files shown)


bstaletic:
I turned `LockGuard` and `UniqueLock` wrapper classes into a single template so now we can write
```c++
LockWrapper< std::lock_guard< std::mutex > > lock( some_mutex );
LockWrapper< std::unique_lock< std::mutex > > lock( some_mutex );
```
or, with `using std::foo` statements
```c++
LockWrapper< lock_guard< mutex > lock( some_mutex );
LockWrapper< unique_lock< mutex > > lock( some_mutex );
```

---

Review status: 0 of 19 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1045#-:-LDYrcRu8iWFgIjlPU3S:b-lxtc3s)*

puremourning:
Gave this a light review.

I had 2 overall thoughts:

1. This adds quite a lot of complexity and macros and SFINAE shenanigans which is quite difficult to get right. I would say that if it delivers high value and low ongoing burden then that's ok
2. As you mentioned, we've simply silenced the errors in all places where we're currently wrong. My thoughts on that are that there's little point having the validation if we're just going to silence it. The point being that it's only really useful if it is always right?

---

Review status: 0 of 19 files reviewed at latest revision, all discussions resolved.

---

*[cpp/ycm/Mutex.h, line 69 at r1](https://reviewable.io/reviews/valloric/ycmd/1045#-LE09I227byXBstoHq5E:-LE09I239ME0oWforuL0:btpe638) ([raw file](https://github.com/valloric/ycmd/blob/43f80ebfeaa0021e8fe5771337a3a5b3c39b1c20/cpp/ycm/Mutex.h#L69)):*
> ```objc
> namespace YouCompleteMe {
> 
> // NOTE: Wrappers for std::mutex, std::lock_guard andstd::unique_lock
> ```

space after and

---

*[cpp/ycm/Mutex.h, line 92 at r1](https://reviewable.io/reviews/valloric/ycmd/1045#-LE0A2YH6KCZwQBxcGZN:-LE0A2YH6KCZwQBxcGZO:bms4x78) ([raw file](https://github.com/valloric/ycmd/blob/43f80ebfeaa0021e8fe5771337a3a5b3c39b1c20/cpp/ycm/Mutex.h#L92)):*
> ```objc
>   LockWrapper( Mutex &m ) ACQUIRE( m ) : lt_( m.native_handle() ) {}
> 
>   template < typename T = LockType, typename std::enable_if<
> ```

when declaring the SFINAE default template arguments, I like to name them as what they return like

```c++
template< typename T, typename IsMutex = std::is_same< ... >::type >
```

I prefer that to this `= 0` form because it adds some semantic information for the human reader.

But I tend to avoid this kind of shenanigans as much as possible.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1045#-:-LE0BQ0l0dFUwf7URnht:b4h2gnz)*

puremourning:
Right sure, so I guess the next step is to work that out and see where we are with it. 

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1045#-:-LE0DyCHDT9i0AIPGcPg:bibni8r)*

bstaletic:
As far as "false positive" goes, I don't see anything we do listed in [known limitations](https://clang.llvm.org/docs/ThreadSafetyAnalysis.html#known-limitations).

Another thing I'm not sure is if I've annotated the `LockWrapper` correctly. because `std::unique_lock( std::mutex, std::try_to_lock_t )` doesn't necessarily lock. Which we consistently check with `owns_lock()`. There is a [TRY_ACQUIRE()](https://clang.llvm.org/docs/ThreadSafetyAnalysis.html#try-acquire-bool-try-acquire-shared-bool) attribute, but I didn't manage to get it working.

---

---

*[cpp/ycm/Mutex.h, line 92 at r1](https://reviewable.io/reviews/valloric/ycmd/1045#-LE0A2YH6KCZwQBxcGZN:-LE0FJnjDyryjnHT8y5M:b-pv8ysz) ([raw file](https://github.com/valloric/ycmd/blob/43f80ebfeaa0021e8fe5771337a3a5b3c39b1c20/cpp/ycm/Mutex.h#L92)):*

We can remove the SFINEA "shenanigans" completely, the only downside would be having two extra and invalid methods for the `std::locker_guard` wrapper.
</blockquote></details>

As for `= 0` instead of what looks more clear to a human, here's the reasoning.

This one is a "typed template parameter" unlike your proposed alternative (which is a "non-type template parameter"). That means that `= 0` variant is a part of the signature, while the other one isn't.
This should be nicer one for completion and also allow possible overloads. Granted, for this specific case ability to overload isn't a good argument - there most likely won't ever be a reason to overload a member function with everything the same except the template parameter.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1045#-:-LE0EMJoECBe85wXT1YH:b-b7vdq8)*

bstaletic:
Thinking about the silenced warnings, I'm sure those were not false positives. So either those member functions aren't meant to be thread safe, or we we forgot the mutex.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1045#-:-LE0GsuK7EORoTW31Lyq:b44h7zz)*

bstaletic:
Upon further inspection of the thread safety library coplaints, @micbou and I both agree it's not worth it, so I'm closing this PR.

PR #1051, opened 2018-06-10
[READY] Inform user if maximum number of diagnostics is exceeded

micbou:
See issue https://github.com/Valloric/YouCompleteMe/issues/2392.

If the number of diagnostics exceed the value of `max_diagnostics_to_display`, a diagnostic with the message "Maximum number of diagnostics exceeded." is added at the top of the current file to inform the user.

The language server completer didn't honor the `max_diagnostics_to_display` option. It now does.

Similarly to [the `-ferror-limit` Clang flag](https://clang.llvm.org/docs/UsersManual.html#cmdoption-ferror-limit), a value of `0` for `max_diagnostics_to_display` now means that the number of diagnostics is unlimited.

Closes https://github.com/Valloric/YouCompleteMe/issues/2392.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1051)

### DIFF
--- a/ycmd/completers/completer.py
+++ b/ycmd/completers/completer.py
@@ -169,6 +169,8 @@ class Completer( with_metaclass( abc.ABCMeta, object ) ):
   def __init__( self, user_options ):
     self.user_options = user_options
     self.min_num_chars = user_options[ 'min_num_of_chars_for_completion' ]
+    self.max_diagnostics_to_display = user_options[
+        'max_diagnostics_to_display' ]
     self.prepared_triggers = (
         completer_utils.PreparedTriggers(
             user_trigger_map = user_options[ 'semantic_triggers' ],
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -54,8 +54,6 @@ INCLUDE_REGEX = re.compile( '(\s*#\s*(?:include|import)\s*)(?:"[^"]*|<[^>]*)' )
 class ClangCompleter( Completer ):
   def __init__( self, user_options ):
     super( ClangCompleter, self ).__init__( user_options )
-    self._max_diagnostics_to_display = user_options[
-      'max_diagnostics_to_display' ]
     self._completer = ycm_core.ClangCompleter()
     self._flags = Flags()
     self._include_cache = IncludeCache()
@@ -357,8 +355,9 @@ class ClangCompleter( Completer ):
 
     diagnostics = _FilterDiagnostics( diagnostics )
     self._diagnostic_store = DiagnosticsToDiagStructure( diagnostics )
-    return [ responses.BuildDiagnosticData( x ) for x in
-             diagnostics[ : self._max_diagnostics_to_display ] ]
+    return responses.BuildDiagnosticResponse( diagnostics,
+                                              request_data[ 'filepath' ],
+                                              self.max_diagnostics_to_display )
 
 
   def OnBufferUnload( self, request_data ):
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -59,8 +59,6 @@ class CsharpCompleter( Completer ):
     self._solution_for_file = {}
     self._completer_per_solution = {}
     self._diagnostic_store = None
-    self._max_diagnostics_to_display = user_options[
-      'max_diagnostics_to_display' ]
     self._solution_state_lock = threading.Lock()
 
     if not os.path.isfile( PATH_TO_OMNISHARP_BINARY ):
@@ -203,8 +201,9 @@ class CsharpCompleter( Completer ):
 
     self._diagnostic_store = DiagnosticsToDiagStructure( diagnostics )
 
-    return [ responses.BuildDiagnosticData( x ) for x in
-             diagnostics[ : self._max_diagnostics_to_display ] ]
+    return responses.BuildDiagnosticResponse( diagnostics,
+                                              request_data[ 'filepath' ],
+                                              self.max_diagnostics_to_display )
 
 
   def _QuickFixToDiagnostic( self, request_data, quick_fix ):
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -914,8 +914,11 @@ class LanguageServerCompleter( Completer ):
     contents = GetFileLines( request_data, filepath )
     with self._server_info_mutex:
       if uri in self._latest_diagnostics:
-        return [ _BuildDiagnostic( contents, uri, diag )
-                 for diag in self._latest_diagnostics[ uri ] ]
+        diagnostics = [ _BuildDiagnostic( contents, uri, diag )
+                        for diag in self._latest_diagnostics[ uri ] ]
+        return responses.BuildDiagnosticResponse(
+          diagnostics, filepath, self.max_diagnostics_to_display )
+
 
 
   def PollForMessagesInner( self, request_data, timeout ):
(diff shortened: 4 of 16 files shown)


bstaletic:
:lgtm:

---

Reviewed 15 of 15 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1051#-:-LEdDCEt4oOKj_PmX9i9:bnfp4nl)*

puremourning:

---

*[ycmd/responses.py, line 248 at r1](https://reviewable.io/reviews/valloric/ycmd/1051#-LEdp7jV0HhAt0Mxhx4C:-LEdp7jV0HhAt0Mxhx4D:b-2uy6wu) ([raw file](https://github.com/valloric/ycmd/blob/e88d0d0940913c97545e5ab3bd02ea4d12fffdd5/ycmd/responses.py#L248)):*
> ```Python
>     location = Location( 1, 1, filename )
>     location_extent = Range( location, location )
>     diagnostics.append( Diagnostic(
> ```

I don't think it is a problem, and certainly no need to change it, but there is a slight irony that in the case that we exceed the max number of diags, we send exactly the max number of diags _plus one_ diagnostics.

---

*[ycmd/completers/cpp/clang_completer.py, line 361 at r1](https://reviewable.io/reviews/valloric/ycmd/1051#-LEdpkhN-8GHJx6BJ_Gt:-LEdpkhN-8GHJx6BJ_Gu:b5fi50y) ([raw file](https://github.com/valloric/ycmd/blob/e88d0d0940913c97545e5ab3bd02ea4d12fffdd5/ycmd/completers/cpp/clang_completer.py#L361)):*
> ```Python
>     self._diagnostic_store = DiagnosticsToDiagStructure( diagnostics )
>     return responses.BuildDiagnosticResponse( diagnostics,
>                                               filename,
> ```

hmm in this case, the special extra diagnostic is associated with the file name of the TU. In the case of a huge unity build, this might not be the file the user is editing. I feel like we should be sticking this diagnostic on the request_data[ 'filepath' ] so that it is prominent for the user?

---

*[ycmd/completers/language_server/language_server_completer.py, line 643 at r1](https://reviewable.io/reviews/valloric/ycmd/1051#-LEdqBqy8gkBJ7-rHK_w:-LEdqBqy8gkBJ7-rHK_x:b-nmw3ip) ([raw file](https://github.com/valloric/ycmd/blob/e88d0d0940913c97545e5ab3bd02ea4d12fffdd5/ycmd/completers/language_server/language_server_completer.py#L643)):*
> ```Python
>     super( LanguageServerCompleter, self ).__init__( user_options )
> 
>     self._max_diagnostics_to_display = user_options[
> ```

we might consider hoisting this up into the Completer base class? It's now appearing in a number of places.

---

*[ycmd/tests/language_server/language_server_completer_test.py, line 456 at r1](https://reviewable.io/reviews/valloric/ycmd/1051#-LEdt9hc4zhZp2CCiGTQ:-LEdt9hc4zhZp2CCiGTR:b-kqvy1b) ([raw file](https://github.com/valloric/ycmd/blob/e88d0d0940913c97545e5ab3bd02ea4d12fffdd5/ycmd/tests/language_server/language_server_completer_test.py#L456)):*
> ```Python
> 
> def LanguageServerCompleter_Diagnostics_MaxDiagnosticsNumberExceeded_test():
>   completer = MockCompleter( { 'max_diagnostics_to_display': 1 } )
> ```

what was the reason to use a mock here, rather than requesting the actual diagnostics? I was curious to see how this affected diags across multiple files as reported by jdt.ls. I realise that the diagnostics tests for java are quite fiddly.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1051)*

puremourning:
:lgtm: with a couple observations

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1051#-:-LEdysqT4lIOHzaEwQU0:b-f8xmpv)*

micbou:
Reviewed 15 of 15 files at r1, 5 of 5 files at r2.

---

*[ycmd/responses.py, line 248 at r1](https://reviewable.io/reviews/valloric/ycmd/1051#-LEdp7jV0HhAt0Mxhx4C:-LEduMrbDhY9n2u6JJ59:b-9rime2) ([raw file](https://github.com/valloric/ycmd/blob/e88d0d0940913c97545e5ab3bd02ea4d12fffdd5/ycmd/responses.py#L248)):*

I don't think it is a problem, and certainly no need to change it, but there is a slight irony that in the case that we exceed the max number of diags, we send exactly the max number of diags _plus one_ diagnostics.
</blockquote></details>

Yes, I thought about it and I know that Clang handles this by removing a diagnostic but I think that, while technically correct, it's not intuitive e.g. the limit is 30 and there are 31 errors but only 29 of the 30 expected errors appear?

---

*[ycmd/completers/cpp/clang_completer.py, line 361 at r1](https://reviewable.io/reviews/valloric/ycmd/1051#-LEdpkhN-8GHJx6BJ_Gt:-LEe3klo-4KhjsJ0ejRY:b-b45qjt) ([raw file](https://github.com/valloric/ycmd/blob/e88d0d0940913c97545e5ab3bd02ea4d12fffdd5/ycmd/completers/cpp/clang_completer.py#L361)):*

hmm in this case, the special extra diagnostic is associated with the file name of the TU. In the case of a huge unity build, this might not be the file the user is editing. I feel like we should be sticking this diagnostic on the request_data[ 'filepath' ] so that it is prominent for the user?
</blockquote></details>

I completely forgot about the TU thing. Thanks for catching this. Fixed.

---

*[ycmd/completers/language_server/language_server_completer.py, line 643 at r1](https://reviewable.io/reviews/valloric/ycmd/1051#-LEdqBqy8gkBJ7-rHK_w:-LEe4ZMp5X2AMsy_Jtj8:bctpdcc) ([raw file](https://github.com/valloric/ycmd/blob/e88d0d0940913c97545e5ab3bd02ea4d12fffdd5/ycmd/completers/language_server/language_server_completer.py#L643)):*

we might consider hoisting this up into the Completer base class? It's now appearing in a number of places.
</blockquote></details>

Makes sense. Done.

---

*[ycmd/tests/language_server/language_server_completer_test.py, line 456 at r1](https://reviewable.io/reviews/valloric/ycmd/1051#-LEdt9hc4zhZp2CCiGTQ:-LEe4c9yCK3c7EsEI_v5:b94sxfj) ([raw file](https://github.com/valloric/ycmd/blob/e88d0d0940913c97545e5ab3bd02ea4d12fffdd5/ycmd/tests/language_server/language_server_completer_test.py#L456)):*

what was the reason to use a mock here, rather than requesting the actual diagnostics? I was curious to see how this affected diags across multiple files as reported by jdt.ls. I realise that the diagnostics tests for java are quite fiddly.
</blockquote></details>

I didn't add tests to the Java completer because I don't want to make the CI builds even more flaky.

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1051)*

puremourning:
:lgtm:

---

Review status: 0 of 2 LGTMs obtained (and 2 stale)

---

*[ycmd/completers/cpp/clang_completer.py, line 361 at r1](https://reviewable.io/reviews/valloric/ycmd/1051#-LEdpkhN-8GHJx6BJ_Gt:-LEkj7JJ1lS06aZlmMG7:btxguhx) ([raw file](https://github.com/valloric/ycmd/blob/e88d0d0940913c97545e5ab3bd02ea4d12fffdd5/ycmd/completers/cpp/clang_completer.py#L361)):*

I completely forgot about the TU thing. Thanks for catching this. Fixed.
</blockquote></details>

:D its my pet feature :)

---

*[ycmd/tests/language_server/language_server_completer_test.py, line 456 at r1](https://reviewable.io/reviews/valloric/ycmd/1051#-LEdt9hc4zhZp2CCiGTQ:-LEkjEPr0Zn6rEVyQFwR:b-t8xni1) ([raw file](https://github.com/valloric/ycmd/blob/e88d0d0940913c97545e5ab3bd02ea4d12fffdd5/ycmd/tests/language_server/language_server_completer_test.py#L456)):*

I didn't add tests to the Java completer because I don't want to make the CI builds even more flaky.
</blockquote></details>

Makes sense :/

---


*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1051#-:-LEkjH69ATJWXkyV3COY:bnfp4nl)*

micbou:
@zzbot r+

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1051#-:-LEkm2c3CI17Yc4UZz-G:bi59kwd)*

PR #1052, opened 2018-06-10
[READY] Add the regex module to sys.path in ycmd exclusively

micbou:
Since the `regex` module has a compiled component, adding it to `sys.path` in YCM may lead to a segmentation fault if the version of Python that YCM is running on is different than the one used to compile the module (e.g. YCM running on Python 3.6 while the `regex` module is compiled for Python 3.5). Note that the crash only seems to happen on macOS.

This is most likely going to fix issue https://github.com/Valloric/YouCompleteMe/issues/3048.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1052)

### DIFF
--- a/ycmd/server_utils.py
+++ b/ycmd/server_utils.py
@@ -124,6 +124,16 @@ def CompatibleWithCurrentCore():
 def SetUpPythonPath():
   sys.path.insert( 0, os.path.join( DIR_OF_CURRENT_SCRIPT, '..' ) )
 
+  # We don't add this path in AddNearestThirdPartyFoldersToSysPath because
+  # loading the regex module in YCM may cause a segmentation fault if the module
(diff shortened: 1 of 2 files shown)


micbou:
Fixed the tests in `server_utils_test.py`.

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1052#-:-LEe3V1f4tYtrve8eVhE:b-4kce0j)*

bstaletic:
Nice one. :lgtm: 
@zzbot r=puremourning

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1052#-:-LEeqajv-jAbF0Vd5X-l:b-az3i0x)*

PR #1053, opened 2018-06-12
[WIP] Parallel sort using Intel's PSTL

bstaletic:
After a lot of benchmarking, I've found out that if the maximum number of candidates to sort *and* total number of candidates both exceed 2048 the parallel excution starts giving a performance boost.

This is the benchmark of thte current master and this PR branch.
https://gist.github.com/bstaletic/cc9c0d54cb3f9a238b9f66cfe75808e0

### PSTL dependencies

- Intel's Threading Building Blocks library
- OpenMP 4.0 (MSVC supports only 2.0)
- `make`
  - The TBB repo does contain cmake files like `TBBBuild.cmake` which can be used to build TBB, but all it does is shelling out to make.
  - `make` is already a part of Ubuntu's `build-essentials` package.
- C++11 compiler (MSVC should work, but I have not tested it).

### MSVC support

Their documentation states that MSVC should be able to compile the library and there are even some `windows.cl.inc` file that the main `Makefile` sources when on Windows and using `cl.exe`. However, I have not tried any of this. If someone can prove that using PSTL and TBB won't hurt performance on Windows without OpenMP I'll gladly update the PR.

### What about parallel filtering?

That should also be possible, here's my first attempt of doing such:

```diff
diff --git a/cpp/ycm/PythonSupport.cpp b/cpp/ycm/PythonSupport.cpp
index 2b041a44..d63cf8fc 100644
--- a/cpp/ycm/PythonSupport.cpp
+++ b/cpp/ycm/PythonSupport.cpp
@@ -77,19 +77,24 @@ pylist FilterAndSortCandidates(
     pybind11::gil_scoped_release unlock;
     Word query_object( query );
 
-    for ( size_t i = 0; i < num_candidates; ++i ) {
-      const Candidate *candidate = repository_candidates[ i ];
 
-      if ( candidate->IsEmpty() || !candidate->ContainsBytes( query_object ) ) {
-        continue;
-      }
-
-      Result result = candidate->QueryMatchResult( query_object );
-
-      if ( result.IsSubsequence() ) {
-        result_and_objects.emplace_back( result, i );
-      }
-    }
+    size_t i = 0;
+    std::for_each( std::execution::par,
+                   repository_candidates.begin(),
+                   repository_candidates.begin() + num_candidates,
+                   [ &query_object, &result_and_objects, &i ]
+                   ( const Candidate *candidate ) -> void {
+                     if ( !candidate->IsEmpty() &&
+                          candidate->ContainsBytes( query_object ) ) {
+                       Result result = candidate->QueryMatchResult( query_object );
+
+                       if ( result.IsSubsequence() ) {
+                         result_and_objects.emplace_back( result, i);
+                       }
+                     }
+                     ++i;
+
+                   } );
 
     PartialSort( result_and_objects, max_candidates );
   }
```

The important details to notice here are a switch from a C style for loop to `std::for_each` and the use of `std::execution::par` as the execution policy.

Why is this not a part of this PR? because, as it is now, it causes the following exception:
```
terminate called after throwing an instance of 'pybind11::error_already_set'
  what():  IndexError: list index out of range
```
Frankly, I'm not sure why the exception is thrown.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1053)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -35,3 +35,9 @@
 [submodule "third_party/cregex"]
 	path = third_party/cregex
 	url = https://github.com/micbou/regex.git
+[submodule "third_party/parallelstl"]
+	path = third_party/parallelstl
+	url = https://github.com/intel/parallelstl
+[submodule "third_party/tbb"]
+	path = third_party/tbb
+	url = https://github.com/01org/tbb/
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,8 @@ addons:
      - clang
      - clang-tidy-3.9
      - libc++-dev
+     # Required by Intel TBB
+     - libomp-dev
      # Required to build the OmniSharp server.
      - mono-devel
      # Everything below is a Python build dep (though it depends on Python
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -90,6 +90,12 @@ get_python_inc(),
 # versions.
 if platform.system() != 'Windows':
   flags.append( '-std=c++11' )
+  flags.extend( [ '-isystem', os.path.join( DIR_OF_THIRD_PARTY,
+                                            'tbb',
+                                            'include' ),
+                  '-isystem', os.path.join( DIR_OF_THIRD_PARTY,
+                                            'parallelstl',
+                                            'include' ) ] )
 
 
 # Set this to the absolute path to the folder (NOT the file!) containing the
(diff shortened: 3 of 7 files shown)


bstaletic:
What's going on with CircleCI (or Travis)?


WRT parallel filtering, it's definitely something missing some mutex.

PR #1055, opened 2018-06-15
[READY] Set break_property_ type to uint8_t

bstaletic:
As it stands now (on master), the `CodePoint::break_property_` is 4 bytes large, requiring one byte of padding before it and causing the whole class to require 7 bytes of padding at the end. If we change the type of the `enum class BreakProperty { ... }` to `enum class BreakProperty : uint8_t { ... } `, we save 3 bytes for the enum size, 1 byte of padding before the `break_property_` and 4 bytes of boundary alignment (which still leaves 3 bytes of boundary alignment) - leading to 8 bytes of saved memory per `CodePoint` instance.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1055)

### DIFF
--- a/cpp/ycm/CodePoint.h
+++ b/cpp/ycm/CodePoint.h
@@ -28,7 +28,7 @@ namespace YouCompleteMe {
 // http://www.unicode.org/reports/tr29/#Grapheme_Cluster_Break_Property_Values
 // NOTE: The properties must take the same value as the ones defined in the
 // update_unicode.py script.
-enum class BreakProperty {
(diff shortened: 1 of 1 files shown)


micbou:
Thanks! :lgtm:

---

Reviewed 1 of 1 files at r1.
Review status: 0 of 2 LGTMs obtained (and 1 stale)

---



*Comments from [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1055#-:-LF2_TC5A71rRYGNbFhB:b-ftzx47)*

PR #1071, opened 2018-07-23
In case of error, show which directory doesn't exist

sylvestre:
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1071)


micbou:
@zzbot r+

PR #1072, opened 2018-07-23
Detect python 3.7 installation

wizzard:
Detect python 3.7 installation.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1072)


micbou:
@zzbot retry

PR #1075, opened 2018-08-05
[READY] Remove nodejs workaround

micbou:
Opening a file without setting the encoding may fail on Python 2. See issue https://github.com/Valloric/YouCompleteMe/issues/3101. We could use the `ReadFile` utility function to fix the issue but I think we are better off dropping the `nodejs` workaround altogether (looking for the `nodejs` executable before `node` because `nodejs` is the name of the Node binary on Debian-based distributions). Users should install the `nodejs-legacy` package on such distributions.


bstaletic:
@zzbot r=ppuremourning

PR #1080, opened 2018-08-11
[READY] Replace Boost canonical function with our own implementation

micbou:
The following error occurs when applying a FixIt in a C/C++ buffer that's not yet saved to disk:
```
RuntimeError: boost::filesystem::canonical: No such file or directory
```
This happens because `boost::filesystem::canonical` raises an exception if the file does not exist. We need our own implementation that works on non-existing files.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1080)


bstaletic:
Homu can't be summoned from Reviewable.
@zzbot r+

bstaletic:
Homu is being an idiot...
@zzbot retry

PR #1081, opened 2018-08-11
GetType on C++ member functions should return their type.

thisismiller:
Previously, GetType on a C++ member function would always show the text
"bound member function type".  This is because a cursor on an instance
method is of kind MemberRefExpr, which itself is a reference to a
declaration of a function that has a type.  Undoing that layer of
referencing will return a useful result.

We need to be careful though, as member variables also show up as
MemberRefExpr, and undoing the layer of ref on them will remove
cv-qualification.  Checking explicitly if the cursor is pointing to a
CXXMethod seems to work fine, but honestly the clang API seems a little
weird in this case.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1081)

### DIFF
--- a/cpp/ycm/ClangCompleter/TranslationUnit.cpp
+++ b/cpp/ycm/ClangCompleter/TranslationUnit.cpp
@@ -327,6 +327,13 @@ std::string TranslationUnit::GetTypeAtLocation(
     return "Internal error: cursor not valid";
   }
 
+  if (clang_getCursorKind(cursor) == CXCursor_MemberRefExpr) {
+    CXCursor ref = clang_getCursorReferenced(cursor);
+    if (clang_getCursorKind(ref) == CXCursor_CXXMethod) {
+      cursor = ref;
+    }
+  }
+
   CXType type = clang_getCursorType( cursor );
 
   std::string type_description =
--- a/cpp/ycm/ClangCompleter/TranslationUnit.cpp
+++ b/cpp/ycm/ClangCompleter/TranslationUnit.cpp
@@ -327,9 +327,14 @@ std::string TranslationUnit::GetTypeAtLocation(
     return "Internal error: cursor not valid";
   }
 
-  if (clang_getCursorKind(cursor) == CXCursor_MemberRefExpr) {
-    CXCursor ref = clang_getCursorReferenced(cursor);
-    if (clang_getCursorKind(ref) == CXCursor_CXXMethod) {
+  // Cursors on member functions return a rather unhelpful type text of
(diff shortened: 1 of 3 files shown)


thisismiller:
I don't see a way to strip the attribute from anything that libeling gives me, and I'm honestly a bit wavering on if stripping all attributes would be the most helpful.  Things like `__attribute__((pure))` give meaningful information.

The closest that I can figure out is to call `clang_getFunctionTypeCallingConv()`, and see if it returns `CXCallingConv_X86ThisCall`.  If so, then take the returned type spelling and apply a series of replacements ("thiscall" -> "", "((," -> "((", ",))" -> "))", "__attribute__(())" -> "") aimed at removing `thiscall` from a list of attributes if others exist.  I'd have to somehow limit this to the range on an `__attribute__` so that I don't rewrite a type, such as `thiscall_t` to just `_t`.

I'm personally more inclined to just adjust the tests.  I don't have a win32 box to develop the above on, and it'd be a good amount of extra code complexity to make the types slightly less verbose in a development environment that probably doesn't see a lot of use?

My plan here would be to claim that an expected output that starts and ends with a / is a regex, add a `_regex` beneath `_eq` that performs regex matching, and adjust RunGetSemanticTest to check for /.../ as an expected output string and call `_regex` instead.  The `expected` string for the member function tests would then be `/int \(int\)(?: __attribute__\(\(thiscall\)\)?/`.

Does that sound alright, or would you prefer to see it solved some other way?

thisismiller:
Converting to use hamcrest indeed was a simpler way to go.  I think it should be right, but I'm partly pushing to get the win32 run via appveyor, so we'll see how that goes...

I tried to add an explicit `__attribute__((thiscall))` function to the tests, to try and push the thiscall problem to be something that's testable on all platforms.  It turns out that clang "helpfully" strips this attribute on platforms where it will have no effect, and :YcmDiags actually surfaces that as well.  So Win32-only failure it is...


And tangentially, hamcrest functions in surprising ways when you don't understand how to use it right:

```
>>> assert_that( { 'a': 1 }, has_entry( 'a', equal_to(3) ) )
AssertionError: 
Expected: a dictionary containing ['a': <3>]
     but: was <{'a': 1}>
>>> assert_that( { 'a': 1 }, { 'a': equal_to(3) } )  # silently passes
>>> # ???
```

thisismiller:
Oh, I didn't see your other comments before pushing.  I can convert it back to a regex if you'd prefer.

micbou:
@zzbot r+

PR #1084, opened 2018-08-19
[READY] Improve standard library index searching

micbou:
See https://github.com/Valloric/YouCompleteMe/pull/3113#pullrequestreview-147458804.


bstaletic:
@zzbot r+

PR #1087, opened 2018-08-22
[READY] Ignore extra conf files outside tests directory

micbou:
Some tests fail if a `.ycm_extra_conf.py` file is found outside the ycmd directory (e.g. YCM's `.ycm_extra_conf.py` file). We prevent that by mocking the `PathsToAllParentFolder` function so that it considers the `tests` folder as the root directory.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1087)


puremourning:
@zzbot r+

PR #1088, opened 2018-08-22
[READY] Raise proper exception for commands when file is still being parsed in Clang completer

micbou:
Same as PR https://github.com/Valloric/ycmd/pull/966 but extended to all commands (not only the `GoTo*` ones) and with tests. Error messages are updated since they are not specific to completion.

Closes https://github.com/Valloric/ycmd/pull/966.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1088)


bstaletic:
@zzbot r+

micbou:
@zzbot retry

PR #1092, opened 2018-08-29
Update submodule from nsf/gocode to mdempsky/gocode

shafreeck:
nsf/gocode is not maintained anymore and it does not work
above go1.10.4

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1092)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -35,3 +35,6 @@
 [submodule "third_party/gocode"]
 	path = third_party/gocode
 	url = https://github.com/mdempsky/gocode.git
+[submodule "third_party/github.com/mdempsky/gocode"]
+	path = third_party/github.com/mdempsky/gocode
+	url = https://github.com/mdempsky/gocode
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,9 +13,6 @@
 [submodule "third_party/requests"]
 	path = third_party/requests
 	url = https://github.com/kennethreitz/requests
-[submodule "third_party/gocode"]
-	path = third_party/gocode
-	url = https://github.com/nsf/gocode.git
 [submodule "third_party/jedi"]
 	path = third_party/jedi
 	url = https://github.com/davidhalter/jedi
@@ -35,3 +32,6 @@
 [submodule "third_party/cregex"]
 	path = third_party/cregex
 	url = https://github.com/micbou/regex.git
+[submodule "third_party/gocode"]
+	path = third_party/gocode
+	url = https://github.com/mdempsky/gocode.git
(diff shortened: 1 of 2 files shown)


bstaletic:
Thanks for the PR. The builds are failing because now, gocode is being cloned from mdempsky's repo, but ycmd is still looking for the hash of the nsf's gocode commit.

You could solve this by running `git submodule sync`, but then all users will have to do the same once they update ycmd, which isn't really nice. A more userfriendly, but uglier option is to completely remove all traces of nsf's gocode and then place mdempsky's gocode in a new directory, such as `third_party/gocode_new`.

shafreeck:
@bstaletic I removed the original submodel and added a new one.

shafreeck:
Unfortunately, it cannot be built now, the new `gocode` repo cannot be built directly, I will try to fix it later

shafreeck:
Unfortunately, it cannot be built now, the new `gocode` repo cannot be built directly, I will try to fix it later

isaachier:
Added fix on @shafreeck's branch (https://github.com/shafreeck/ycmd/pull/1). If/when that is accepted, it should fix the build issues here.

micbou:
See https://github.com/Valloric/ycmd/issues/1097#issuecomment-418200260.

ChoppinBlockParty:
Since go has changed drastically their package system (for the best I think), `gocode` is basically a binary, that could be easily acquired (or updated) by `go get -u github.com/mdempsky/gocode`, you do not need to have any build instruction around it.  This binary could be used solely by `ycmd` or by other tools as well. I do not see a benefit of having it as a submodule in `ycmd` it is runtime requisite rather than a development dependency.

micbou:
The benefit is to have a frozen version that we know is working with ycmd.

ChoppinBlockParty:
All is done in `go_complete` is start/stop of `gocode` and API interaction. If they create a breaking API change, yes, it is a problem, and the user has to be on old version, until `ycmd` is migrated to a new API - it is like with every software, major/breaking version release. I meant that it ycmd does not need gocode source code in it database to be build, it is an external dependency.

puremourning:
Our philosophy is to pin versions that we test. We don't plan to change that.

shafreeck:
@isaachier Thanks a lot for your PR. The gocode cannot be built still, because of lack of GOPATH, would you like to fix this?

bstaletic:
@shafreeck Take a look at #1098.

shafreeck:
@bstaletic Cool, I will close this PR

PR #1095, opened 2018-08-31
[READY] Allow users to configure LSP servers through extra conf

bstaletic:
The LSP configuration, as discussed in gitter.
Users are able to configure jdt.ls with an extra conf like this:
```python
def Settings( **kwargs ):
  if kwargs[ 'language' ] == 'java':
    return { 'ls': {
                     # jdt.ls config, for example:
                     'java.rename.enabled': False
                     # 'java': { 'rename': { 'enabled': False } } is also valid
                   }
           }
```

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1095)


bstaletic:
Resolved conflicts.

bstaletic:
@zzbot r=puremourning

PR #1100, opened 2018-09-11
[READY] Add Python path to debugging information

micbou:
This is particularly useful when writing an extra conf that modify Python's `sys.path`.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1100)


bstaletic:
@zzbot r+

PR #1101, opened 2018-09-11
[READY] Allow completion of the core library in Python code

micbou:
This PR improves ycmd's extra conf by adding the root directory to `sys.path` and by setting the interpreter path to the value stored in the `PYTHON_USED_DURING_BUILDING` file (if it exists). These two changes allow completion of the `ycm_core` module in the Python code.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1101)


bstaletic:
@zzbot r+

PR #1103, opened 2018-09-12
[READY] Support framework headers completion and code navigation

micbou:
On macOS, the compiler searches header paths of the form `A/B/.../Z` as `A.framework/Headers/B/.../Z` in the `/System/Library/Frameworks` and `/Library/Frameworks` directories, e.g. `CoreServices/CoreServices.h` is found as `/System/Library/Frameworks/CoreServices.framework/Headers/CoreServices.h`. More generally, the compiler does that for each path specified through the `-F` or `-iframework` flags. This PR adds support for completing those paths and jumping to them.

Demo:

framework-headers

Fixes https://github.com/Valloric/YouCompleteMe/issues/2647.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1103)

### DIFF
--- a/ycmd/completers/cpp/clang_completer.py
+++ b/ycmd/completers/cpp/clang_completer.py
@@ -32,7 +32,11 @@ from xml.etree.ElementTree import ParseError as XmlParseError
 
 import ycm_core
 from ycmd import responses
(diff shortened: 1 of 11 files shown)


bstaletic:
@zzbot r+

PR #1104, opened 2018-09-17
[READY] Improve filename completer

micbou:
This PR brings the following improvements to the filename completer:
 - allow completion of paths containing spaces;
 - allow completion of multiple paths on the same line;
 - allow completion of relative paths not starting with `.`, `..`, or `./`;
 - allow completion of Windows environment variables (e.g. `%USERPROFILE%`).

The first three improvements are achieved by assuming that if a user types `[something]/` and `[something]/` is an existing path (absolute or relative to the working directory) then he want to complete that path.


erouvin:
Hi,

Not sure this is the right place but I have an issue for path completion when the character '@' is starting a directory name. The path completion allows the directory name to be added but then restarts at the root instead of showing the subsequent files or subdirectories.

vim 8.1.436

Thank you.

erouvin:
Sorry, I am new in this. How can I tell if I get the right release ?
I am using Vundle to install/update the plugins.

micbou:
To try these changes, navigate to the ycmd folder and fetch the PR:
```sh
cd ~/.vim/bundle/YouCompleteMe/third_party/ycmd
git fetch origin pull/1104/head:filepath-completer-improvements
git checkout filepath-completer-improvements
```
then start Vim and check if completion works with paths starting with `@`. Once you are done, go back to the master branch in the ycmd folder:
```sh
git checkout master
```

PR #1107, opened 2018-09-20
[READY] Update Clang to 7.0.0

micbou:
We use our [own LLVM package for Linux](https://github.com/micbou/llvm/releases/tag/7.0.0) to avoid the numerous issues caused by the official binaries being linked to an old version of the terminfo library.

No binaries have been yet uploaded for AArch64 and ARM so we return an error for these platforms. We'll add them later if they become available.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1107)


Valloric:
:lgtm:

Thanks for the PR!

@zzbot r+

PR #1109, opened 2018-09-23
[READY] Update jdt.ls to 0.25.0

micbou:
Some changes were required:
 - the hover response does not return a list with only one item anymore but instead return the item directly e.g. `{ 'language': 'java', 'value': 'some value' }` is returned instead of `[ { 'language': 'java', 'value': 'some value' } ]`;
 - for some reason, an empty list of diagnostics is returned for the project root;
 - renaming only works for valid identifiers (`a-quite-long-string` is not a valid Java identifier);
 - a regression has been introduced: renaming a method does not rename places where the method is overridden. See https://github.com/eclipse/eclipse.jdt.ls/issues/802.

The `PollForMessagesTimeoutException` exception from `Poll_Diagnostics_ProjectWide_Eclipse_test`  is suppressed.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1109)


micbou:
@zzbot r+

PR #1110, opened 2018-09-26
GetHoverResponse should handle null response from server.

kadircet:
As mentioned in [Hover Request](https://microsoft.github.io/language-server-protocol/specification#textDocument_hover)
```
Response:

result: Hover | null
```
Current implementation was failing with None object has no _getitem_ method, with this patch it will provide more useful information.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1110)


kadircet:
Added tests for GetHoverResponse, both for null and non-null case.

kadircet:
Addressed all the comments and squashed commits.

micbou:
@zzbot retry

PR #1115, opened 2018-10-01
[READY] Move GIL release to bindings

bstaletic:
- No #include <pybind11/pybind11.h> in `ClangCompleter.cpp` and `IdentifierCompleter.cpp`.
- Releases GIL just before function call, instead of just after.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1115)


Valloric:
:lgtm:

Thanks for the PR!

@zzbot r+

micbou:
@zzbot retry

PR #1117, opened 2018-10-04
[READY] Add clangd uploader script.

kadircet:
This patch adapts update_clang script to extract and bundle clangd binary. If you guys want I can actually merge this one with the existing update_clang script, but I think this is better for maintenance, if you guys decide to make clangd default only deleting the update_clang script would be enough instead of modifying the script and also getting new sha's etc. Also there is a version conflict since clangd focuses on release v7 whereas current libclang is lookign at v6 release. Would be great if @micbou can run the script and upload to bintray, so that we can spin up the CI for tests of clangd.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1117)

### DIFF
--- a/update_clangd.py
+++ b/update_clangd.py
@@ -1,0 +1,411 @@
+#!/usr/bin/env python
+
+# Passing an environment variable containing unicode literals to a subprocess
+# on Windows and Python2 raises a TypeError. Since there is no unicode
+# string in this script, we don't import unicode_literals to avoid the issue.
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+
+import argparse
+import subprocess
+import contextlib
+import os
+import os.path as p
+import platform
+import shutil
+import sys
+import tempfile
+import tarfile
+import hashlib
+from distutils.spawn import find_executable
+
+try:
+  import lzma
+except:
+  from backports import lzma
+
+DIR_OF_THIS_SCRIPT = p.dirname( p.abspath( __file__ ) )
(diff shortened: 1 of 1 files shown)


kadircet:
- Made changes mentioned in #1107
- Deleted outputdir printing, sorry it was a leftover from debugging.

kadircet:
BTW, it seems like https://github.com/micbou/llvm/releases/download/7.0.0/clang+llvm-7.0.0-x86_64-unknown-linux-gnu.tar.xz doesn't contain clangd.

kadircet:
Actually I was talking about the url that script itself downloads the llvm release, not the url it uploads after creating a smaller bundle.

bstaletic:
That should be downloaded from llvm directly. The github archives replace bintray archives, i.e. both, github and bintray archives are our minimal archives. Github was used because the archive was too big for bintray. So, every clangd archive is downloaded from llvm, repackaged and, assuming bintray will not refuse the archive, uploaded there.

kadircet:
Ok, will revert back to llvm url then. I have changed it to github release url because thats what @micbou did in #1107, and suggested me to do that as well.

bstaletic:
Forget what I said, I mixed something up. @micbou is preparing new archives that will contain clangd.

kadircet:
Updated checksums in #1118 to reflect those changes. Hopefully it should be working now.

Valloric:
Either I'm crazy, or the larger clangd PR #1118 includes this script as well. Are we then dropping this PR in favor of the other one or are we going to remove this file from the other PR?

kadircet:
Removed this file from #1118

micbou:
@zzbot r+

Issue #1114, opened 2018-09-28
Introduction of Clangd Completer


kadircet:
## Hi everyone!
We’ve been implementing a new language server called [clangd](https://clang.llvm.org/extra/clangd.html) that implements LSP and have been using it in our internal system with YCM through a new completer for a while now instead of libclang. We wanted to upstream that completer so that other people can use it with YCM as well.  

### Need your opinions:
clangd comes as an external binary, our completer plugin looks for a clangd binary in the path and uses it as the LSP server. Also the binary has been shipped with clang-tools packages and LLVM releases. So they can be downloaded in numerous ways:
- [Debian packages](https://packages.debian.org/buster/clang-tools) (also [ubuntu](https://packages.ubuntu.com/cosmic/clang-tools) has it in its own repos)
- [LLVM releases ](http://releases.llvm.org/download.html) in the forms of [source code](http://releases.llvm.org/7.0.0/clang-tools-extra-7.0.0.src.tar.xz) and pre-built, [ubuntu](http://releases.llvm.org/7.0.0/clang+llvm-7.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz), [windows](http://releases.llvm.org/7.0.0/LLVM-7.0.0-win64.exe), [macOS](http://releases.llvm.org/7.0.0/clang+llvm-7.0.0-x86_64-apple-darwin.tar.xz) binaries.
- Also available in [homebrew/llvm](https://formulae.brew.sh/formula/llvm).
- Also it is whitelisted in [Travis CI apt sources](https://github.com/travis-ci/apt-source-safelist/blob/master/ubuntu.json#L275).

So, we are wondering on how to make it available for ycmd CI pipelines in all environments in the most maintainable way possible:
- Would you suggest a similar way that has been used for libclang? IIUC, it downloads a precompiled library during tests, we can do the same with clangd download a pre-built version from llvm-releases.
- We can fetch the source code(again from llvm release) and build clangd on each environment.
- Or anything else that can be achieved with the above mentioned packages/releases.

bstaletic:
First of all, I'm excited about moving ycmd from built-in libclang completer to something like clangd. We (ycmd maintainers) have been looking into things like clangd/cquery for a while. And we've had a couple of concerns, but generally speaking would like to switch because that would mean we could drop a lot of complexity. I'll get back to these later.

- Fetching sources and compiling clangd is too time and RAM consuming to be feasible as for CI so for some users, making the approach simply out of the question.
[...]

PR #1118, opened 2018-10-04
[READY] Introduce clangd_completer.

kadircet:
- This PR is supposed to be merged as an experimental alternative to the current libclang completer.
- It is missing some features, like GetDoc or include completion, so it won't be the default until it is at feature parity.
- Include completion will be available as soon as we upgrade to LLVM8.
- There is no support for extra confs. Clangd expects a compilation database with
- There are two configuratoin knobs - one for choosing clangd completer and one for telling ycmd to move out of the way and let clangd do all the filtering, sorting and caching.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1118)


kadircet:
Seems like there is clangd installed on CI but with llvm-5.0.0 release, changing install mechanism to require at least llvm-7.0.0 release.

kadircet:
So travis has the following error
```
/home/travis/build/Valloric/ycmd/third_party/clangd/output/bin/clangd: relocation error:
/home/travis/build/Valloric/ycmd/third_party/clangd/output/bin/clangd: symbol _ZNKSt3_V214error_category10_M_messageB5cxx11Ei,
version GLIBCXX_3.4.21 not defined in file libstdc++.so.6 with link time reference
```

When I searched, I've found such an answer https://askubuntu.com/a/777814, does anybody know how to apply that to travis ?

kadircet:
Tests should be passing now, would be great if someone can cancel pending tests in appveyor.

kadircet:
Yes let's cancel that. There seems to be a different issue going on with windows tests I'll investigate.

kadircet:
Also we have the option to specify resource-dir into clangd binary with '-resource-dir' command line argument. But we still need to point to some directory, which in our case can be clang_includes.

bstaletic:
Pointing to `clang_includes` would be the best option. I didn't know clangd had that flag.

kadircet:
@micbou would be glad if you can cancel https://ci.appveyor.com/project/Valloric/ycmd/builds/19370585

kadircet:
Hey everyone, seems like everything is passing currently. So just wanted to check whether you guys are reviewing or waiting for smaller PRs or anything else. Since no one said anything about

> I know this is huge, but sending this one as more like a meta PR to check whether everything is working on CI. Happy to make smaller PRs that will fail on CI, but would be easier to approve and later on only merge this one.

that part in the first message :D or in the issue #1114

hbostann:
Is there anything we can do to fasten the process, I would love to see this in master.

bstaletic:
Thanks for the interest. As you can see, most of the discussion is resolved. What is missing is another ycmd maintainer to review the PR.

Also, I want to test LLVM trunk thoroughly. I'd like to see how this PR behaves when used on something like Valliric/YouConoleteMe#2668.

sam-mccall:
> Okay, let's say we want to let users play with clangd arguments. Like you said, this is something that would require project level granular configuration. In that case the correct thing to do is going through an extra conf, not g:ycm_foo. Since extra confs are python scripts, it's easy to create very flexible and custom tailored configurations.

Just wanted to give some perspective on clangd flags...
Firstly, agreed everything should work out-of-the-box. The recommended configuration for `clangd` users is zero flags. (Though clangd developers often need to set some).

While not totally consistent, the design is that flags are user-preferences, not per-project configuration. We aim to have good defaults, but some examples of what users want to set:
 - whether overloads are bundled or separated
 - whether signatures are visible
 - number of background threads used
 - whether private members are shown as completions when inaccessible
 - use of disk vs ram for caches
Here users generally want the same settings for all projects, and conversely one project shouldn't use the same settings for different users. So global settings rather than ycm_extra_conf seems like a more appropriate scope, to me.

There's a more practical reason too: these arguments must be passed at clangd startup, so cannot vary per-project in a consistent way (a single clangd/vim session can of course span multiple projects). This matches the way that other LSP servers and clients work in practice.

For completeness, we do have some non-preference flags, too:
 - flags relating to project configuration are basically broken by design and we're eliminating them over time. e.g.
    - the `index-file` flag Kadir mentioned will be replaced by the auto-index feature which is multi-project aware
    - the `compile-commands-dir` flag is a compromise added as part of integration into Theia, which runs clangd with a different lifecycle. It shouldn't be used otherwise (we should make it hidden...)
 - hidden flags guarding experimental/unstable features during development, to enable debug logging, etc
 - flags to support llvm's integration testing driver

sam-mccall:
On Wed, Nov 28, 2018, 13:07 Boris Staletic <notifications@github.com wrote:

> Okay, taking your whole comment into account, I agree that the extra conf
> isn't the right place for this. The question still remains: Do we really
> want to expose all of this?
>
I'm not going to go through the features point by point, because I don't
think it's central. Suffice it to say that all the arguments you made were
known and considered, and there are arguments on the other side too. Not
all the decisions were my personal favorite, but engineering is compromise
:-)

More broadly, clangd team has two goals with this integration:
 - improve c-family support for YCM users (clangd is an implementation
detail: it has technical improvements over libclang). YCM team has a clear
vision for what these users want, we're happy to make the default behavior
reflect what you want (where clangd can reasonably support it). Similarly
installing clangd with YCM.
 - provide a good option for using clangd with vim[1] (YCM is an
implementation detail: it has the best vim plugin of plausible LSP
clients). An important part of this is that clangd's product decisions
aren't entirely second-guessed by the editor plugin. E.g. it's bizarre if
clangd's docs say "feature X works with emacs, vscode and sublime, but not
vim because YCM devs think it's not useful". Especially if it's not a
"frontend feature".

Maybe this framing sheds some light on where we've been talking past each
other about defaults and configurability?

Our hope was that an integration into YCM would achieve both goals. For
that to work, I think YCM should be picking the defaults, clangd's own
options etc should be exposed, and YCM team shouldn't be on the hook for
supporting variants. (How can we achieve this?)

But maybe it's too awkward to make these coexist, or the YCM team is
interested in 1 but not 2. That's important to know too! We don't have a
clear backup plan for another LSP client, and should find one. I think we
should move ahead with this completer regardless, but it certainly affects
scope and priority.

@bstaletic @puremourning @valloric Sorry for the wall of text, but really
want to know what you each see as the best future, here.

[1] our current best advice for clangd+vim is to join google and use our
forked YCM :-/

bstaletic:
Okay, I'm actually thankful for the wall of text. I see nothing wrong with achieving both of clangd goals with YCM. I'd like for both "sides" to be satisfied with the end result. For the clangd arguments, the easiest way would be to actually expose `clangd_args` as a key in `user_options` (global variable in YCM). Like they currently are exposed.

However, I have one concern. What are the chances that some set of clagd arguments goes against some, likely accidental, assumption that ycmd has made and thus prevents some functionality from working?

Specifically, on my mind is the current branch in my fork that implements `signatureHelp` for the libclang completer. It relies on a libclang returning overloads kind set to `OVERLOAD`. Is there a posibility that user, using his preferred flags, breaks something like this? I assuming chances are rather slim, because we're communicating over LSP protocol.

If anyone can tell me why I shouldn't be concerned about this, I'll be completely okay with exposing `clangd_args`. As far as I have seen `clangd -help-list-hidden` displays a ton of possible arguments and it's definitely not feasible for ycmd to test every combination.

> our current best advice for clangd+vim is to join google and use our forked YCM

This makes me wonder if YCM is good or are others bad.

micbou:
@zzbot r+

PR #1119, opened 2018-10-05
[READY] Fix UriToFilePath.

kadircet:
As specified in the [URI](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier)
```
Of the ASCII character set, the characters : / ? # [ ] @ are reserved for use as delimiters of the generic URI components and must be percent-encoded
```

But turns out, implementation in url2pathname doesn't respect that. This patch fixes its behaviour in lsp. Which was the reason all tests related with a filepath(allmost every test) in windows for clangd was failing.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1119)

### DIFF
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -1015,7 +1015,12 @@ class LanguageServerCompleter( Completer ):
       # diagnostics and return them in OnFileReadyToParse. We also need these
       # for correct FixIt handling, as they are part of the FixIt context.
       params = notification[ 'params' ]
-      uri = params[ 'uri' ]
+      # Since percent-encoded strings are not cannonical, they can choose to use
+      # upper case or lower case letters, also there are some characters that
+      # can be encoded or not. Therefore, we convert them back and forth
+      # according to our implementation to make sure they are in a cannonical
+      # form for access later on.
+      uri = lsp.FilePathToUri( lsp.UriToFilePath( params[ 'uri' ] ) )
       with self._server_info_mutex:
         self._latest_diagnostics[ uri ] = params[ 'diagnostics' ]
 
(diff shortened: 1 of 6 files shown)


kadircet:
filepathtourl also needs to be fixed.

kadircet:
No change is not ready yet, hopefully I'll send it out on Monday with tests.

kadircet:
The test I moved from java_completer to language_server_completer was actually checking a behavior in language_server_completer rather than java_completer itself.

It turns out uris are not cannonical they can contian upper/lower case letters in percent encoding or choose to encode characters in a not certain way, as specification says "you can encode those if you want". Therefore I made sure we put a canonical version in _latest_diagnostics so that we can lookup via filepath to uri conversion later on.

Also handling for C:\\ and C%3a\\ was different so I tried to patch that up like this for now, since URI scheme says one may choose to encode or not that colon both should be valid, but looks rest of the world is not encoding so we will also change clangd's behavior to not encode colon. But it will only be available in LLVM8 release.

kadircet:
___
*[ycmd/completers/language_server/language_server_completer.py, line 1023 at r1](https://reviewable.io/reviews/valloric/ycmd/1119#-LON5Hon2Ke6fOEouNRd:-LON61h70uOuEXAeMvM7:bbqltc3) ([raw file](https://github.com/valloric/ycmd/blob/12101fe80d59f6a0ae5558579e5fc9a2bdf7903a/ycmd/completers/language_server/language_server_completer.py#L1023)):*

This looks really ugly, but I guess there's nothing we can do?
</blockquote></details>

at least nothing that comes to my mind, since we store diagnostics with this uri and it is not unique for the reasons I mentioned within the comment we need to at least store a canonical version that can be generated again later on from a filepath so that we can access those.

another approach could be to use filepaths as keys instead of uris, but not sure if there would be cases left out since uris are more generic and handle different schemes.

PR #1121, opened 2018-10-06
[READY] Support completion FixIts for C-family languages

micbou:
Since version 7.0.0, libclang proposes FixIts to change `.` into `->` and vice versa when completing the members of an object. This PR adds support for those FixIts.

Demo:

clang-completion-fixit

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1121)


puremourning:
@zzbot r+

PR #1123, opened 2018-10-13
[READY] Install fixed version of TypeScript in third-party folder

micbou:
Add the `--ts-completer` parameter to the `install.py` script to configure the TypeScript completer. This parameter installs a specific version of TypeScript in the third-party folder which is used by the completer. If not installed this way, the completer still looks for `tsserver` in the `PATH` to ensure backward compatibility. This change has several benefits:
 - tests on the build bots can't be broken by a new version of TypeScript:
 - users are on a known version of TSServer that is tested;
 - users are not instructed to install TypeScript globally. See issue https://github.com/Valloric/YouCompleteMe/issues/3149.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1123)

### DIFF
--- a/.circleci/install_dependencies.sh
+++ b/.circleci/install_dependencies.sh
@@ -112,12 +112,6 @@ cargo -V
 
 echo "export PATH=${CARGO_PATH}:\$PATH" >> $BASH_ENV
 
-##################
-# TypeScript setup
-##################
-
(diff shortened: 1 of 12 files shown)


puremourning:
@zzbot r+

PR #1125, opened 2018-10-17
add "-source" option for gocode autocompletion

kyoh86:
Now, ycmd cannot complete golang external packages.
ycmd must enable source importer with -source flag.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1125)

### DIFF
--- a/ycmd/completers/go/go_completer.py
+++ b/ycmd/completers/go/go_completer.py
@@ -140,6 +140,7 @@ class GoCompleter( Completer ):
 
     stdoutdata = self._ExecuteCommand( [ self._gocode_binary_path,
                                          '-sock', 'tcp',
+                                         '-source',
                                          '-addr', self._gocode_host,
                                          '-f=json', 'autocomplete',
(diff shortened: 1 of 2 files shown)


kyoh86:
for your reference: https://github.com/mdempsky/gocode/issues/32#issuecomment-401988251

bstaletic:
What worries me about this is the performance impact it imposes. However, I have never used go, so I would like to hear from go users if this is something we really want to enable by default.

bstaletic:
Looking at the tests, something is clearly broken here. 3 out of 6 are broken because the expected call in the tests doesn't have `-source`, but the other three say that gocode panicked, even the basic completion test failed because there are syntax errors in the file.

kyoh86:
oh I see, sorry. I'll fix it later.

kyoh86:
> gocode panicked

Sorry, I cannot reproduce the error on local...

micbou:
PR seems abandoned. Closing.

PR #1126, opened 2018-10-26
[READY] Fix Flake8 errors

micbou:
[The new version of pycodestyle adds two warnings affecting our code base](https://github.com/PyCQA/pycodestyle/blob/77ad77278d04db396fcaddbc174b75bf3933a153/CHANGES.txt#L16-L22):
 - W504: warning for checking that a break doesn't happen after a binary operator. We ignore that one because it's perfectly fine to break the line after `+` or `and`;
 - W605: warning for invalid escape sequences in string literals. This one is legit. It will fix the corresponding deprecation warnings we are getting with Python 3.

We also need to ignore E301 because it now triggers for methods preceded by a comment.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1126)


Valloric:
:lgtm:

Thanks for the PR!

@zzbot r+

PR #1129, opened 2018-11-14
[READY] Fixed builds with system libclang in Windows

ElArtista:
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1129)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -277,8 +277,8 @@ if ( EXTERNAL_LIBCLANG_PATH OR USE_SYSTEM_LIBCLANG )
     find_library( TEMP
                   NAMES
                   clang
-                  libclang
                   libclang.so.1
+                  libclang
                   PATHS
                   ${ENV_LIB_PATHS}
                   ${LLVM_CONFIG_PATH}
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -253,14 +253,12 @@ include_directories(
 
 if ( EXTERNAL_LIBCLANG_PATH OR USE_SYSTEM_LIBCLANG )
   if ( USE_SYSTEM_LIBCLANG )
-    set ( LIBCLANG_FILE libclang.so.1 )
     if ( APPLE )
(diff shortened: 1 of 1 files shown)


ElArtista:
Context:
I tried running the post install YCM script with --system-libclang in Windows and it failed. This pull request fixes the problem.
I also applied the changes you proposed, i think its much better now!

ElArtista:
Fixed as requested.

ElArtista:
@puremourning as i said in a previous comment:

> I tried running the post install YCM script with --system-libclang in Windows and it failed. This pull request fixes the problem.

The reason it failed is that libclang dynamic library in Windows is named `libclang.dll`, but we only searched for `clang` and `libclang.so.1` variants. Appending a `libclang` value in the NAMES parameter of the find_library function (the extension `.dll` is not needed because cmake adds it when searching for library files on Windows) fixes this issue.

The second issue is that when we copied the libclang dynamic library to our directory we used this line in cmake script:
```
COMMAND ${CMAKE_COMMAND} -E copy "${PATH_TO_LLVM_ROOT}/bin/libclang.dll" "$<TARGET_FILE_DIR:${PROJECT_NAME}>"
```
that did not work in the --system-libclang case because the variable PATH_TO_LLVM_ROOT was used unset. So we make sure we set it in this case too to fix that.

ElArtista:
@bstaletic Rebased and ready for merge.

Valloric:
The arguments from @micbou are convincing (as they usually are). I retract my LGTM.

ElArtista:
Well, from my perspective, `--system-libclang` compiles ycmd using some existing LLVM installation in the system. In Linux this is some llvm package that came from the package manager of the given distro, that installed the relevant llvm files where the Filesystem Hierarchy Standard or given distro dictated. In Windows these kind of pre-provided dependencies usually come by hand, by putting them in some user-specified location and adding the binary directories to the PATH.
In either case the user acknowledges the existence of the LLVM installation in his system and chooses to use that.

As both a Linux (primarily) and a Windows (secondarily) user I would love to have some consistency in my workstations, and this flag provides it for the ycmd build process.

I don't see what kind of harm could this bugfix PR cause, as this flag is already available and usable in Windows environments, leading for know in broken builds (without this fix) and lots of confusion.

PR #1130, opened 2018-11-17
[READY] Do not default to x86 for libclang download

bstaletic:
If a user is on linux, and not using ARM or Aarch64, we currently assume the user is using x86.
Instead we should check if x86 is right and if not, fail with an error, telling the user to either compile libclang or use his system libclang.

Fixes Valloric/YouCompleteMe#3227

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1130)


bstaletic:
@zzbot r=Valloric

PR #1131, opened 2018-11-19
Add support for gocode's '-builtin' and '-unimported-packages' flags

palsivertsen:
In this pull request:
* Update `gocode` submodule to latest version
  * Add new submodule https://go.googlesource.com/tools (dependency of `gocode`)
* Add parameters to completion command

This forces `-builtin` and `-unimported-packages` flags. Could be improved by adding vim config flags on top.

Fixes Valloric/YouCompleteMe/issues/3212

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1131)

### DIFF
--- a/.circleci/install_dependencies.sh
+++ b/.circleci/install_dependencies.sh
@@ -18,7 +18,6 @@ brew update || brew update
 # with CS7027 signing errors.
 REQUIREMENTS="cmake
               node.js
-              go
               readline
               autoconf
               pkg-config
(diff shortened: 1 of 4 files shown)


palsivertsen:
Seems like builds are failing due to outdated Go version (< 1.9). I tried to do an upgrade in 127db133e69ff373ecb7523db6777192887ccfef with no luck.

palsivertsen:
> How recent is is go 1.9?
[Go1.9](https://golang.org/doc/devel/release.html#go1.9) was released  2017/08/24
> Can we expect that users already have it on their machines?
Hard to say. There have been two more releases since then. Go release versions in 1.X.Y are backwards compatible, so developers shouldn't have an issue upgrading. Also Go's [release policy](https://golang.org/doc/devel/release.html#policy) only supports the last two major releases.

> You tried to unpack it in /usr/local/ even though you don't have permission to write there.
Oh. I'm not familiar with macOS and are not sure how to install go. Binaries are available as archive (`.tar.gx`) and installer (`.pkg`) for macOS.

> We will also need tests that confirm these flags are indeed doing what we believe they are doing.
I'm not a Python dev, but I'll see what I can do

micbou:
PR seems abandoned. Closing.

PR #1132, opened 2018-11-19
[READY] Requests update

bstaletic:
Python 3.7.1 produces a new `DerpecationWarning` for the old `requests` and warns that it will completely stop working starting from version 3.8. Appveyor is the only build bot with python 3.7, but it is 3.7.0 which doesn't have that deprecation warning.

The `idna`, `chardet`, `urllib3` and `certifi` are thre dependencies of `requests` that are no longer bundled.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1132)


bstaletic:
Checked out an actual tag for urllib and chardet to pass requests compatibility checks.

bstaletic:
@zzbot r=micbou

PR #1133, opened 2018-11-20
[WIP] Use dense_hash_map instead of the STL counterparts

bstaletic:
This lowers the time measured in `FilterAndSort` benchmarks by ~30% and time measured in `IdentifierCompleter` benchmarks by 15 to 20%.
Where `google::dense_hash_map` differs from `std::unordered_map` (and similar types) is:
- It requires a sentinel entry that represents "empty" value.
- If one intends to remove entries, a sentinel entry that represents "deleted" value is also needed.
- The sentinel values need to be different.
- Removing an entry doesn't deallocate, to avoid iterator invalidation.
- There is no initializer list constructor, so `LANGUAGE_MAP` remained an `std::unordered_map`.

### Why, oh why does this need GCC 5?

Well, it doesn't need gcc 5, but it needs a newer gcc 4.9.x than the one on Travis. I successfully built this branch on gcc 4.9, though gcc 4.8 is definitely out.

I'd argue that this isn't so bad, because:
- Even Ubuntu 16.04 has gcc 5.4.0 by default.
- RHEL users already need to install dev-toolset to compile ycmd.
- Mac users most likely use clang.
So nothing really changes for either of the largest groups of linux users.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1133)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,8 +36,11 @@ addons:
     packages:
      - cmake-data
      - cmake
-     # 4.8 is the first version of GCC with good enough C++11 support.
-     - g++-4.8
+     # 5.0 is the first one that can compile
+     # dense_hash_map< string, dense_hash_map< string, vector< string > > > foo;
+     # auto& bar = foo[ str ];
+     # *on Travis*
+     - g++-5
      # Install Clang 3.4 and its standard library.
      - clang
      - clang-tidy-3.9
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -60,6 +60,8 @@ flags = [
 '-x',
 'c++',
 '-isystem',
+'cpp/sparsehash',
+'-isystem',
 'cpp/pybind11',
 '-isystem',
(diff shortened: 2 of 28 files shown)


Valloric:
If we're updating our hashtable, we should probably be using Google's new Swiss Table which is meant to replace dense_hash_map: https://abseil.io/blog/20180927-swisstables

bstaletic:
That's the WIP part. I've made benchmarks comparing the three hash table implementations.
Swiss tables sound great, but there are problems:

- My first attempt at using swiss tables resulted in a great performance penalty. After a lot of testing they are really close to `dense_hash_map`. Here's my benchmark:
https://gist.github.com/bstaletic/7d3e31db7fc8d40a12787fab09ff092f
- I'm unable to compile swiss tables in debug mode.

I also opened an issue on the abseil repository to help me with the performance issues and if that works, I'll try to figure out debug mode compile time errors.

bstaletic:
This thing will take some time, so closing for now.

PR #1134, opened 2018-11-21
Fix DeprecationWarning from classes in collections.abc

yan12125:
This patch removes the following warning:
```
/usr/share/nvim/runtime/third_party/ycmd/ycmd/utils.py:499: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  class HashableDict( collections.Mapping ):
```
This warning is originated from https://github.com/python/cpython/pull/5460

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1134)

### DIFF
--- a/ycmd/utils.py
+++ b/ycmd/utils.py
@@ -25,7 +25,6 @@ from __future__ import absolute_import
 from builtins import *  # noqa
 
 from future.utils import PY2, native
-import collections
 import copy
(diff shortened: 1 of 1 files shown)


micbou:
@zzbot r+

PR #1136, opened 2018-11-23
[READY] Enable warnings from clang-tidy in header files

bstaletic:
This explains why clang-tidy used to miss raising a warning. No headers produce warnings by default. There are two options controlling this behaviour.

- `-header-filter=<regex>` matches header file names against the `<regex>` and if a header matches, the header's warnings will be reported. On its own, applies only to user headers.
- `-system-headers` allows `-header-filter` to be applied to system headers as well.

We only need the first one and, since we don't use `-system-headers` a `.*` regex works for us. On top of that, it can be set in the config file, so `CMakeLists.txt` clutter is avoided.


Valloric:
:lgtm:

Thanks for the PR!

@zzbot r+

PR #1138, opened 2018-11-24
[READY] Simplify completer hook mechanism

micbou:
Use `importlib.import_module` to import the `hook.py` file of a completer instead of loading its source. In addition to simplify the code, this is needed if we want to package only the `.pyc` files since source files don't exist in that case.

Changes are required in the tests because of how mocking works (see [Where to patch](https://docs.python.org/3/library/unittest.mock.html#where-to-patch) in the Python docs).

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1138)


Valloric:
Always happy to see more simplicity. Thanks!

:lgtm:

@zzbot r+

PR #1139, opened 2018-11-27
[READY] Remove unused options

micbou:
These options are only used in YCM. Now that PR https://github.com/Valloric/YouCompleteMe/pull/3243 is merged, we can safely remove them.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1139)


micbou:
@zzbot r+

PR #1142, opened 2018-12-05
[READY] Do not rely on unordered_map ordering in tests

bstaletic:
Currently the `IdentifierUtilsTest.ExtractIdentifiersFromTagsFileWorks` relies on the fact that you can compare `std::unordered_map` and its preserved order of insertion. This is not true for much faster hash tables like `absl::flat_hash_map`. Using `absl::flat_hash_map` yield between 15 and 20 percent better performance, but also breaks `IdentifierUtilsTest.ExtractIdentifiersFromTagsFileWorks` on current master.

This PR rewrites `IdentifierUtilsTest.ExtractIdentifiersFromTagsFileWorks` to not depend on ordering. If you pay attention, you'll see that now `"c"` is last in the test, but second in the tag file.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1142)

### DIFF
--- a/cpp/ycm/tests/IdentifierUtils_test.cpp
+++ b/cpp/ycm/tests/IdentifierUtils_test.cpp
@@ -29,6 +29,8 @@ namespace fs = boost::filesystem;
 using ::testing::ElementsAre;
 using ::testing::ContainerEq;
 using ::testing::WhenSorted;
+using ::testing::Pair;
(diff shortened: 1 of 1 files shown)


bstaletic:
Wow, this was quick.
@zzbot r=micbou

PR #1143, opened 2018-12-07
[READY] Move jedi and requests to their deps folders

bstaletic:
Should solve the python confusion about stale pyc files.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1143)


bstaletic:
@zzbot retry

PR #1145, opened 2018-12-10
[READY] Properly stop Clangd in isolated tests

micbou:
Use a `try/finally` block to stop Clangd in isolated tests so that the server is shut down even if the test fails.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1145)


puremourning:
@zzbot r+

PR #1147, opened 2018-12-10
[READY] Do not fall back to default values in Clangd completer

micbou:
We should only rely on the values set in the `ycmd/default_settings.json` file.


puremourning:
@zzbot r+

micbou:
@zzbot retry

PR #1148, opened 2018-12-10
[READY] Add the requests and Jedi modules to test PYTHONPATH

puremourning:
The test framework uses requests to communicate with ycmd. When it was
moved to its own directory, we didn't update run_tests.py to include it
in PYTHONPATH used by nosetests.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1148)

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -20,22 +20,30 @@ import sys
 DIR_OF_THIS_SCRIPT = p.dirname( p.abspath( __file__ ) )
 DIR_OF_THIRD_PARTY = p.join( DIR_OF_THIS_SCRIPT, 'third_party' )
 
-python_path = []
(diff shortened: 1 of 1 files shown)


puremourning:
For the record this manifests as run_tests.py failing on systems where requests is not installed globally.

jedi/parso are not affected because they are not used by the test _framework_.

puremourning:
Wait, that's a lie. Jedi/parso are required too because we don't _always_ do integration tests.

puremourning:
Any reason not to merge this and then subsequently Chang it? It’s just right now I can’t run tests.

puremourning:
Don’t disagree that explicit is better now that it is fiddly.

bstaletic:
@zzbot r+

micbou:
@zzbot retry

PR #1149, opened 2018-12-10
[clangd] Don't include clangd in --all

puremourning:
Clangd is enabled by default when it is installed/configured. Currently,
clangd is a feature regression and breaking change for most users, By
including in --all, we are breaking users by default.

For now, remove it from --all and require that --clangd-complerter is
explicitly requested.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1149)


micbou:
@zzbot r+

PR #1150, opened 2018-12-11
[READY] Ignore the local clangd installation and cache

puremourning:
This prevents annoying untracked files in git status.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1150)


puremourning:
@zzbot r=micbou

PR #1152, opened 2018-12-12
[READY] Boost 1.69 update

bstaletic:
Boost.System is now a header only library, but `error_code.cpp` still exists with one empty function, so that people can still link against it, for backwards compatibility.

In `update_boost.py` I added `system` to `BOOST_LIBS_FOLDERS_TO_REMOVE`, but I have not removed `system` from the required components of system boost in `cpp/ycm/CMakeLists.txt` so that we don't break system boost older than 1.69 and get similar complaints to the ones about `libclang.so.7`.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1152)


Valloric:
:lgtm:

Thanks for the PR!

puremourning:
> Though this SO says it doesn't work on macOS.

I'll try it

puremourning:
```
-- Boost version: 1.68.0
-- Found the following Boost libraries:
--   filesystem
--   regex
--   system
```

bstaletic:
@zzbot r=puremourning

PR #1153, opened 2018-12-14
[READY] Only detail filtered candidates in Python completer

micbou:
Computing the fields of all candidates (`extra_menu_info`, `detailed_info`, etc.) becomes a performance issue when Jedi returns a lot of completions. Instead, we should only compute the fields on the filtered candidates (at most 50 by default). The same strategy could be applied to other completers like the TypeScript and LSP ones.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1153)


Valloric:
This is a really good idea, thanks!

@zzbot r+

PR #1154, opened 2018-12-15
[READY] Include signature to Python completions

micbou:
Thanks to PR #1153, we can now include the signature to the Python candidates without impacting too much performance.


bstaletic:
@zzbot r+

PR #1158, opened 2018-12-16
[READY] Refactor logging

micbou:
Instead of redefining `_logger` in each module, define it once in `ycmd/utils.py` and import it from there. Also, don't use `format()` when logging a message because if the formatting fails then the code is interrupted. Let the `logging` module handle the formatting.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1158)


bstaletic:
@zzbot r+

PR #1159, opened 2018-12-16
[READY] Detail libclang completion candidates after filtering

bstaletic:
Here's another one.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1159)


Valloric:
@micbou @puremourning I think this PR is waiting for another LGTM.

PR #1160, opened 2018-12-19
[READY] Fix exit codes in documentation

micbou:
[The exit codes 5 and 6 are inverted in the docs](https://github.com/Valloric/ycmd/blob/ed8d17c2e392458e197d90d6a9cba3dc94a7e0e9/ycmd/server_utils.py#L56-L59).

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1160)


micbou:
@zzbot r+

PR #1161, opened 2018-12-19
[RFC] Only resolve filtered completion items

puremourning:
This applies similar logic to the typescript and other completer to only call the 'resolve' completion request for the candidates that pass our initial filter on their simple `insertion_text`.

Notable wrinkles:

- some language servers don't require resolving, so we have to ensure we don't double up on work
- we need to identify candidates which have been resolved and ones which haven't as candidates may be passed to `DetailCandidates` from the cache.
- whether we are resolving or not, we still need to calculate the start code point and such the same way we did before.
- we now no longer need to prevent resolve when there are > 100 candidates, as by default ycmd restricts filtered results to 50

### DIFF
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -779,6 +779,23 @@ class LanguageServerCompleter( Completer ):
     else:
       items = response[ 'result' ][ 'items' ]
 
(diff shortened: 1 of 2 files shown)


puremourning:
It's good. @zzbot r=bstaletic

PR #1163, opened 2018-12-21
[build.py] Skip loooking for cmake/cmakeargs if skip_build and no_regex

theli-ua:
Sometimes you just want to fetch jdt or something.
My particular use case is fetching jdt. And ycm_core/clangcompleter are built outside of build.py due to completely unstandard locations for python include/build dirs that build.py just can't detect. (so it'll fail looking for them even though I just want to fetch jdt)

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1163)

### DIFF
--- a/build.py
+++ b/build.py
@@ -872,8 +872,9 @@ def WritePythonUsedDuringBuild():
 
 def Main():
   args = ParseArguments()
-  cmake = FindCmake()
-  cmake_common_args = GetCmakeCommonArgs( args )
+  if not args.skip_build or not args.no_regex:
+      cmake = FindCmake()
+      cmake_common_args = GetCmakeCommonArgs( args )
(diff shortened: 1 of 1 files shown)


puremourning:
FWIW you can pass additional arguments to cmake with:

```
$ env EXTRA_CMAKE_ARGS=" -D....=.... ...." ./build.py ....
```

Commonly:

```
$ env EXTRA_CMAKE_ARGS='-DPATH_TO_LLVM_ROOT=/path -DPYTHON_LIBRARY=....' etc.
```

puremourning:
I did this in https://github.com/Valloric/ycmd/pull/1163

puremourning:
Close in favour of #1164

PR #1164, opened 2018-12-21
[READY] Don't find cmake and python if not needed

puremourning:
This allows builds to just update, say jdt.ls or clangd with
'./build.py --skip-build --no-regex --java-completer' and the like.

On some systems finding python and cmake can be a pain, requiring
specifying lots of arguments, so this simplifies and speeds up that use
case.

Replaces: https://github.com/Valloric/ycmd/pull/1163

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1164)

### DIFF
--- a/build.py
+++ b/build.py
@@ -870,16 +870,24 @@ def WritePythonUsedDuringBuild():
     f.write( sys.executable )
 
 
-def Main():
-  args = ParseArguments()
+def DoCmakeBuilds( args ):
   cmake = FindCmake()
   cmake_common_args = GetCmakeCommonArgs( args )
+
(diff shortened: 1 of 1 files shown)


puremourning:
`  510  ./build.py --skip-build --no-regex --java-completer`

No cmake search. 

`  511  ./build.py --skip-build --no-regex --clangd-completer`

No cmake search.

`  512  ./build.py --skip-build --clangd-completer`

Finds cmake and installs cregex


PASS

puremourning:
@zzbot r=micbou

PR #1166, opened 2018-12-29
[READY] Fix java poll thread abort test by ensuring that the abort actually happens

puremourning:
Throwing an exception in another thread doesn't cause the test to fail, so instead we update some variable.

We use a dict because they are mutable (`aborted = True` doesn't work because of the python scoping rules) and we check the result of that after joining the thread.

We also increase the number of tries because we know we have issues with the current number.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1166)


micbou:
@zzbot r+

PR #1172, opened 2018-12-31
[READY] Fix system headers search on macOS

micbou:
We can't rely on upstream libclang to find the system headers on macOS because it ignores the `-resource-dir` flag and use the current working directory to find the builtin includes instead, which fails since they are in a different place (in the `clang_includes` folder). Furthermore, it's unable to find the framework headers on macOS 10.14 without setting the sysroot. So, we try to reproduce the logic used by Apple Clang to find the system headers by looking at the output of the command
```sh
clang++ -x c++ -E -v -
```
which prints the list of system header directories and by reading the source code of upstream Clang. This code probably differs from the latest version of Apple Clang but still gives an idea of how it works.

The builtin includes have been moved to `clang_includes/lib/clang/version/include` to replicate the structure used by upstream libclang. This could be useful if we decide to include the libc++ standard library which would be copied to `clang_includes/include/c++/v1`. Clangd would then be able to find the libc++ headers with `-resource-dir` sets to the `clang_includes/lib/clang/version` folder (not libclang because of the reason given above).

Fixes https://github.com/Valloric/YouCompleteMe/issues/3159.
Fixes https://github.com/Valloric/YouCompleteMe/issues/3249.
Fixes https://github.com/Valloric/YouCompleteMe/issues/3286.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1172)


bstaletic:
Codecov is complaining because `_SysRootSpecifiedIn()` used to be covered and it isn't in this PR.
Let's take care of that as well.

PR #1174, opened 2019-01-05
[READY] Report java server startup status correctly

puremourning:
Since some recent jdt.ls version, we always seem to receive a
'language/status' message type 'Starting' (100%) _after_ we receive the
'language/status' message with type 'Started'. This makes the server
look like it is still initializing although it is really ready to go.

We work around this dodgy message sequence by ignoring anything received
after 'Started'.


bstaletic:
@zzbot r+

PR #1177, opened 2019-01-19
[READY] Make LSP debug info consistent

puremourning:
Include server startup state and working directory in all LSP
completers. Migrate java completer to use this rather than
including its own.

This is step 1 in refactoring the LSP completer to make it easy to add new ones.


micbou:
@zzbot r+

PR #1178, opened 2019-01-19
[NOT READY] Improve the project directory detection

puremourning:
NOTE: Do not merge yet: depends on #1177 

In LSP's view of the world, everything has a project root directory.
Currently we dumbly use the path of the first file opened as the root
unless the completer implementation does something smarter. For example,
the java completer tries to find a project config file.

We now try and guess a project directory using the following hierarchy:
 - extra conf dir if it is not the global extra conf
 - client working dir
 - dir of first opened file

Other LSP servers legitimately expect a reasonable project directory to
be supplied on initialisation. Clangd presumably doesn't because it uses
the path of the compile_commands.json that _it finds itself_. Therefore,
this change should have no effect on real use cases of existing uses.

This is change 2 in the batch of changes to simplify LSP completer 
additions.

### DIFF
--- a/ycmd/completers/cpp/clangd_completer.py
+++ b/ycmd/completers/cpp/clangd_completer.py
@@ -218,12 +218,12 @@ class ClangdCompleter( language_server_completer.LanguageServerCompleter ):
 
   def DebugInfo( self, request_data ):
     with self._server_state_mutex:
-      clangd = responses.DebugInfoServer(
-        name = 'clangd',
-        handle = self._server_handle,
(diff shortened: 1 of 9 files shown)


puremourning:
Fixed the tests :)

puremourning:
It's READY. I will follow up with a change to tidy this up, as manually calling all this stuff everywhere is a bit messy. I will hoist the start server and initialise calls into the base class.

micbou:
@zzbot r+

PR #1180, opened 2019-01-19
[READY] Trigger semantic completion when instructed by server

puremourning:
LSP initialization exchange includes the characters that should trigger
semantic completion. We now combine those characters with our own set of
regexes and any supplied by the user.

We do this by combining both user-supplied and server-supplied triggers (and of course the  "default" triggers) in the prepared triggers structure, which is already per-completer. the merge is a strict superset per filetype supported by the completer.

The main reason for this is to make adding new completers easier (one less thing to do). However, it's actually strictly worse in some cases (c++ being a specific case). The reason for this is that LSP only supports a single character for triggers, so clangd must send `>` as a trigger character. This is a false-positive in many cases, and triggering semantic completion after, say `template < typename T >` not only makes no sense, but returns no useful completions.

So for we allow the concrete completer to manipulate the server-supplied list. In the case of clangd, we ignore it.

For Java, the list makes sense and adds useful features, such as proper completion of `@Annotations`. Added a test for this.

### DIFF
--- a/ycmd/completers/completer_utils.py
+++ b/ycmd/completers/completer_utils.py
@@ -32,16 +32,38 @@ from ycmd.utils import ( LOGGER, ToCppStringCompatible, ToUnicode, re, ReadFile,
 
 class PreparedTriggers( object ):
   def __init__( self, user_trigger_map = None, filetype_set = None ):
+    self._user_trigger_map = user_trigger_map
+    self._server_trigger_map = None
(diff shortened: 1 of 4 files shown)


puremourning:
This is now ready. Tests in place for both Java and Clangd.

Will check the coverage data as well.

bstaletic:
@zzbot r=micbou

PR #1181, opened 2019-01-20
[READY] Fix LSP server startup when UnknownExtraConf is thrown

puremourning:
The change to only load settings on initialize was flawed. In the case
where UnknownExtraConf was thrown, the server was started, but the
initialize request was not sent. This essentially left the completer
deadlocked, waiting for an initialize response that would never arrive.

The solution is to revert to requesting the settings prior to starting
the server instance, but instead of doing it on every
OnFileReadyToParse, have that concrete StartServer methods do this.

This isn't ideal, but it does solve the problem. In future, the
SimpleLSPServer will make this better, but ultimately we should probably
change it so that concrete completers don't call SendInitialize at all,
and the base class decides when to start the server and send the init.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1181)

### DIFF
--- a/ycmd/completers/cpp/clangd_completer.py
+++ b/ycmd/completers/cpp/clangd_completer.py
@@ -330,6 +330,10 @@ class ClangdCompleter( language_server_completer.LanguageServerCompleter ):
       if self.ServerIsHealthy():
         return
 
+      # We have to get the settings before starting the server, as this call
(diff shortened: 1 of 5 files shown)


bstaletic:
@zzbot r+

PR #1182, opened 2019-01-23
[READY] Update ExtraConfData description in API docs

micbou:
The API docs still mention the `FlagsForFile` function.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1182)


micbou:
@zzbot r+

PR #1183, opened 2019-01-24
[READY] Update parso to 0.3.2

micbou:
This new version of parso supposedly fixes a lot of the `Exception: There's an issue (x != y) with the diff parser` errors when parsing Python code. See https://github.com/davidhalter/jedi/issues/1245#issuecomment-457019206.


puremourning:
@zzbot r+

PR #1184, opened 2019-01-26
[READY] Simplify LSP completer API for starting server

puremourning:
We always need to get the settings, start the server, then send the
initialize request. This change hoists that code into the base class,
reducing boilerplate required by concrete classes and overall
simplifying usage. By doing so, this also prevents a loop when starting
the clangd completer fails, implementing the approach taken by java
completer.

This is the next step in simplifying LSP completer creation.

This is a follow up to #1178.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1184)


micbou:
Let's merge. This line can always be covered in a separate PR.

@zzbot r+

PR #1187, opened 2019-01-30
[READY] Improve Clangd completer initialization

micbou:
This PR brings the following changes to the Clangd completer:
 - do not raise an exception when initializing the Clangd completer since this prevents ycmd from falling back to the libclang completer;
 - never fall back to the third-party Clangd if the one specified in `clangd_binary_path` is not valid. The reason behind that change is that I think most users are going to be confused by such fallback since they'll think that the Clangd from `clangd_binary_path` is used while in fact it's not. I can revert this change if you disagree;
 - expand environment variables and `~` in the path specified by the `clangd_binary_path` option;
 - do not print in the logs that the Clangd library is outdated if the path does not point to an executable;
 - cosmetic changes like capitalizing Clangd in comments and logs.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1187)

### DIFF
--- a/ycmd/completers/cpp/clangd_completer.py
+++ b/ycmd/completers/cpp/clangd_completer.py
@@ -31,14 +31,26 @@ from ycmd import responses, utils
 from ycmd.completers.completer_utils import GetFileLines
 from ycmd.completers.language_server import language_server_completer
 from ycmd.completers.language_server import language_server_protocol as lsp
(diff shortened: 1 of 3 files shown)


Valloric:
:lgtm:

Thanks for the PR!

@zzbot r+

PR #1191, opened 2019-02-17
[READY] Ease clangd_completer initialization

kadircet:
- Makes `use_clangd` option binary.
  - It is no longer needed to set `use_clangd = 'Always'` after setting `clangd_binary_path` if user doesn't have clangd binary in third party folder.
- Gets rid of redundant `GetThirdPartyClangd` calls.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1191)

### DIFF
--- a/ycmd/completers/cpp/clangd_completer.py
+++ b/ycmd/completers/cpp/clangd_completer.py
@@ -108,7 +108,7 @@ def GetThirdPartyClangd():
   return pre_built_clangd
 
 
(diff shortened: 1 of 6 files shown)


kadircet:
There goes the `[READY]` tag. Waiting for #1192 to be merged, I'll do a rebase afterwards

bstaletic:
@zzbot r+

PR #1193, opened 2019-02-22
Fix system header search paths on macOS for Objective-C++, too.

s4y:
I've been running into https://github.com/Valloric/YouCompleteMe/issues/303, even though ycmd has a workaround for it. It looks like the issue is just that I'm writing ObjC++ and that change only affects C++.

So, this change makes `-x objective-c++` eligible, too.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1193)


s4y:
Hmm, I *think* the CI failure isn't related to my change.

micbou:
@zzbot r+

PR #1195, opened 2019-02-25
[READY] Ignore identifiers returned by TSServer in JavaScript

micbou:
TSServer sometimes returns identifiers from the current file as suggestions in JavaScript. These suggestions are not relevant so we ignore them.

Fixes https://github.com/Valloric/YouCompleteMe/issues/3330.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1195)


micbou:
@zzbot r+

PR #1197, opened 2019-03-01
[READY] Add -B to INCLUDE_FLAGS

bstaletic:
This flag tells the compiler frontend (`gcc`/`clang`) where to look for the subprograms (`cpp`, `cc`, `cc1plus`, `as`, `ld` etc.). This flag is needed when bootstrapping a toolchain, so it is unlikely to be found in a custom extra conf, but it can appear in users' compilation databases. While it is not really an include flag, it behaves just like one.

Fixes Valloric/YouCompleteMe#3333

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1197)


micbou:
@zzbot r+

micbou:
@zzbot retry

PR #1200, opened 2019-03-05
[RFC] Remap GoTo commands in Python completer

micbou:
The `GoToDefinition`, `GoToDeclaration`, and `GoTo` commands in the Python completer are currently mapped to Jedi methods as follows:
 - `GoToDefinition` to [`goto_definitions`](https://jedi.readthedocs.io/en/latest/docs/api.html#jedi.Script.goto_definitions);
 - `GoToDeclaration` to [`goto_assignments`](https://jedi.readthedocs.io/en/latest/docs/api.html#jedi.Script.goto_assignments);
 - `GoTo` to `goto_definitions` then `goto_assignments` if `goto_definitions` failed.

These mappings don't really make sense since there is no distinction between a definition and a declaration in Python. All these `GoTo` commands should be identical. Said that, we need to decide to which function these commands should be mapped: `goto_definitions` or `goto_assignments`? My understanding is that `goto_definitions` doesn't jump to the definition but the type definition while `goto_assignments` actually jumps to the definition with the caveat that it doesn't follow imports by default. This is confirmed by looking at some examples (these examples are used in the tests):
```
my_variable = 1
my_variable
```
With the cursor on the second `my_variable`:
 - `goto_definitions` returns the `int` builtin. This is clearly not the definition of `my_variable` but its type;
 - `goto_assignments` returns the first `my_variable` which is where `my_variable` is defined.

```
class MyClass():
  pass

my_instance = MyClass()
my_instance
```
With the cursor on the second `my_instance`:
 - `goto_definitions` returns the first `MyClass`. Again, not the definition of `my_instance` but its type;
 - `goto_assignments` returns the first `my_instance`.

This PR thus proposes to remap the three `GoTo` commands to `goto_assignments` (with some additional work to follow imports and to deal with aliases) and to introduce a new command `GoToType` which is mapped to `goto_definitions`.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1200)


puremourning:
Submitted #1275

PR #1203, opened 2019-03-08
[READY] Reject any server to client requests

puremourning:
Some LSP features allow the language server to request that the language client does things, like [apply edits](https://microsoft.github.io/language-server-protocol/specification#workspace_applyEdit) and such.

We don't currently support these requests, so we should reject them.

Perhaps one day we will support them, but right now the requests just never get a response which makes us a bad LSP citizen.

This is the next patch in the now infamous "yaml" branch. RFC because no tests.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1203)

### DIFF
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -392,6 +392,14 @@ class LanguageServerConnection( threading.Thread ):
     self.WriteData( message )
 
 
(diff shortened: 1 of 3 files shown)


puremourning:
@zzbot r=micbou

bstaletic:
@zzbot retry

PR #1204, opened 2019-03-09
[READY] Ignore any null or empty values and use our defaults

puremourning:
Some servers (heinously) send null when they should omit keys in
message objects. This causes us to break because our "default" values
are not used when uding .get( v, "default ). dict.get returns the value
if the key _exists_. If the key exists and the value is None (null), you
get None back.

Use ( dict.get( v ) or "default" ), instead of dict.get( v, default ).
This returns "default" in the case of dict[ 'v' ] being "falsy".

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1204)


puremourning:
@zzbot r=micbou

micbou:
@zzbot retry

PR #1205, opened 2019-03-12
[READY] Improve cache in LSP completer

micbou:
ycmd is caching completions that are already filtered by Clangd if we send the cursor column (`column_codepoint`) to Clangd when `clangd_uses_ycmd_caching` is set to `1`,  This leads to unexpected results when triggering semantic completion in a middle of a word. Here's a demo of the issue:

missing-completions-before

No `foo` suggestion when deleting the `b` character. If we send the starting column (`start_codepoint`) instead:

missing-completions-after

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1205)

### DIFF
--- a/ycmd/completers/completer.py
+++ b/ycmd/completers/completer.py
@@ -411,7 +411,7 @@ class CompletionsCache( object ):
   """Cache of computed completions for a particular request."""
 
   def __init__( self ):
-    self._access_lock = threading.Lock()
+    self._access_lock = threading.RLock()
     self.Invalidate()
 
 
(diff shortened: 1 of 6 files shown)


bstaletic:
This PR creates a different problem - ycmd could cache incomplete completion candidates.

`coc.nvim` has solved this by caching completions only once `isIncomplete == false`. Here's a hacky implementation of something like that:
https://github.com/bstaletic/ycmd/commit/7487cfb7180dcf42110ab9b649dea87f90bf93ce

bstaletic:
> AFAIK, Clangd always sets isIncomplete to False.

No, it only sets `isIncomplete` when the results are really incomplete. I have tested it before making that commit.

> I don't see how that helps. Even if the list is complete, it's still potentially caching a filtered list.

That's right. I was trying to say that we need to take `isIncomplete` into consideration as well.

kadircet:
Yes, @bstaletic is right about clangd's behavior.

bstaletic:
@zzbot r+

PR #1207, opened 2019-03-16
[READY] Drop python 3.4

bstaletic:
Python 3.4 reached end of life today (2019-03-16). Ubuntu 18.04 (what we officially support) has `python` pointing to Python 2.7 and `python3` pointing to Python 3.6.

Dropping Python 3.4 will allow us to use [abseil](https://github.com/abseil/abseil-cpp)'s flat hash maps and improve [benchmarks](https://gist.github.com/bstaletic/7d3e31db7fc8d40a12787fab09ff092f) by ~20% ~~except for `IdentifierCompleterFixture/CandidatesWithCommonPrefix/65536/10` (still an improvement, but a small one)~~. Turns out we can get ~20% in every benchmark.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1207)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,7 @@ jobs:
       - *save-cache
       - *upload-coverage
     environment:
-      YCMD_PYTHON_VERSION: 3.4
+      YCMD_PYTHON_VERSION: 3.5
   benchmark:
     <<: *common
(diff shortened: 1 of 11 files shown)


micbou:
@zzbot r+

PR #1210, opened 2019-03-18
[READY] Search custom Clangd in PATH

micbou:
This change makes it possible to use a custom Clangd from the PATH by setting the `clangd_binary_path` option to `'clangd'` instead of having to specify an absolute path.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1210)


micbou:
@zzbot r+

micbou:
@zzbot retry

PR #1211, opened 2019-03-19
[READY] Parallel sorting

bstaletic:
C++17 introduced parallel `<algorithm>`, `<numeric>` and `<memory>`. Current state is that MSVC has their `<execution>` header, but GCC and Clang don't and plan to eventually use https://github.com/intel/parallelstl. That is the same repository that is included in this PR.

### Dependencies

- Intel TBB, found at https://github.com/01org/tbb
  - We are using my fork because we need CMake. 
  - The Cmake script has been taken from https://github.com/wjakob/tbb

### Benchmarks

https://gist.github.com/bstaletic/dfa74ef0f72667c8ed41610d28b04e38

The parallel version of sorting algorithms currently kick in once there's at least 2048 candidates to return to the user, because my benchmarks on my laptop with my OS had consistently shown that was the point where parallel sorting actually outweighs the overhead of spinning up threads\*.

\*The implementation is fully allowed to disregard the user's request for a parallel sort if the array is small, however, back when I was running these benchmarks there was a tiny overhead until 2048. It might be worth checking this behaviour again.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1211)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -48,3 +48,6 @@
 [submodule "third_party/idna"]
 	path = third_party/requests_deps/idna
 	url = https://github.com/kjd/idna
+[submodule "cpp/tbb"]
+	path = cpp/tbb
+	url = https://github.com/bstaletic/tbb
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -60,6 +60,8 @@ flags = [
 '-x',
 'c++',
 '-isystem',
+'cpp/pstl/include',
+'-isystem',
 'cpp/pybind11',
 '-isystem',
 'cpp/whereami',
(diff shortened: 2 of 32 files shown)


bstaletic:
So, further benchmarks show the following:

- Below 256 candidates, parallel sorting benchmarks are noisy because on a small data set it might not even run in parallel.
- When there's a large number of candidates, but we only keep a few (say 10/65536), parallel sorting is performed, but turns out slower than sequential.
- In `IdentifierCompleter` benchmark the time was halved.
- In `FilterAndSortUnstoredCandidates` , it's the storing of candidates that takes time, so the performance gain is slightly less than 10%. Abseil will make a larger impact here.
- in `FilterAndSortStoredCandidates` the needed time for 65536 candidates drops from 50ms to 35ms.

All bechmarks have been done on my laptop, which, as far as hardware/OS combinations go, is a data sample of 1.

Issue #1105, opened 2018-09-18
Migrate Rust completer to Rust Language Server


miguelsaldivar:
So it seems that [racerd](https://github.com/jwilm/racerd), has stalled and hasn't been updated in a year. Were there plans to move to a different rust completion server?

puremourning:
Such as?

bstaletic:
Last I heard RLS was in a fairly good shape. As in Rust Language Server.

miguelsaldivar:
Yeah I think [RLS](https://github.com/rust-lang-nursery/rls) would be nice, it seems like it's heavily used and updated regularly.

puremourning:
Do you have specific issues that it would address?

bstaletic:
@puremourning We already have a couple unresolved rust related issues that might go away with RLS. It might be worth a try.

puremourning:
Fair enough. As always provided there is a benefit and no regression and a completely free migration path... :)

Then again I don’t Rust so PRs welcome?

micbou:
There is a rough implementation in [that branch](https://github.com/micbou/ycmd/tree/rust-language-server). If people are interested, I may polish it and submit a PR.

miguelsaldivar:
Oh awesome @micbou :sunglasses:, if you need any help let me know and I'll try to find some time to help. 
@puremourning a bunch of small things like: auto completion on types from [extern blocks](https://github.com/racer-rust/racer/issues/890).

Valloric:
We'll eventually need to migrate to RLS since racer is deadish. Also, RLS is powered by racer internally (for the initial RLS implementation).

@jwilm Thoughts?

bstaletic:
@micbou Says there's not too much work left to polish his branch enough for migration.

jwilm:
RLS capabilities are a super set of Racer's; there's no technical reasons ycmd shouldn't use RLS at this point. There may be some other benefits like eliminating a Rust build step in ycmd in favor of "make sure rls is installed on your system". I believe this comes as a component with rustup today.

puremourning:
We are currently working on [significantly reducing the overhead in adding new LSP based completers](https://github.com/puremourning/ycmd-1/commit/eb915e410d9576d4ae9aef674baf7f8245e14c81).

When this is complete we will review which servers offer value and stability. We'll keep this in mind.

Valloric:
While RLS has its problems, it's the obvious next step for our Rust integration. 

[...]

PR #1212, opened 2019-03-19
Update Racerd

fandahao17:
Hi there!
I know that [ycmd is migrating rust completer to RLS](https://github.com/Valloric/ycmd/issues/1105), but the Racerd version that ycmd currently uses is too old (last update 2 years ago) and doesn't work well under recent Rust versions  (most times no completion pops up after pressing ' . ', not recognizing some Rust 2018 features, etc.). So it's necessary to switch to the most recent Racerd version before we finally migrate to RLS.
  
Since starting from version 2.1, racer needs **nightly rust**, I also added "nightly" in build scripts, test scripts and the documentation.
  
But as was mentioned in the issue above, **Racerd isn't under effective maintenance and its most recent version is still a little bit out of date**. [This pull request there](https://github.com/jwilm/racerd/pull/77) tries to keep it up-to-date.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1212)

### DIFF
--- a/.circleci/install_dependencies.sh
+++ b/.circleci/install_dependencies.sh
@@ -111,7 +111,6 @@ rustup update
 rustc -Vv
 cargo -V
 
-echo "export RUST_SRC_PATH=$(rustc --print sysroot)/lib/rustlib/src/rust/src" >> $BASH_ENV
 echo "export PATH=${CARGO_PATH}:\$PATH" >> $BASH_ENV
 
 #################
--- a/.circleci/install_dependencies.sh
+++ b/.circleci/install_dependencies.sh
@@ -111,6 +111,7 @@ rustup update
 rustc -Vv
 cargo -V
 
+echo "export RUST_SRC_PATH=$(rustc --print sysroot)/lib/rustlib/src/rust/src" >> $BASH_ENV
 echo "export PATH=${CARGO_PATH}:\$PATH" >> $BASH_ENV
 
 #################
--- a/.circleci/install_dependencies.sh
+++ b/.circleci/install_dependencies.sh
@@ -106,6 +106,7 @@ curl https://sh.rustup.rs -sSf | sh -s -- -y
 
 CARGO_PATH="${HOME}/.cargo/bin"
 PATH="${CARGO_PATH}:${PATH}"
+rustup toolchain add nightly
 rustup update
 rustc -Vv
 cargo -V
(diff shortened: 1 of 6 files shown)


fandahao17:
The test shows what I'm worried about. Even the most recent version of racerd can't compile under newest Rust compilers. This can be solved by [updating its dependencies in the *Cargo.toml* file](
https://github.com/jwilm/racerd/pull/77).

fandahao17:
The tests failed because some of racerd's dependencies are too old. @jwilm Could you merge the [PR](https://github.com/jwilm/racerd/pull/77) to update Racerd so that we can provide better support for recent Rust versions?

jwilm:
Merged the PR you linked @fandahao17

fandahao17:
Sorry for so many failed commits. I mistakenly thought the tests failed because of the lack of rust-src.However,  I finally found that there may be some problems with our [tests](https://github.com/Valloric/ycmd/blob/master/ycmd/tests/rust/get_completions_test.py). As the [error message indicates](https://circleci.com/gh/Valloric/ycmd/4977),  the completion requests that were supposed to fail ended up getting a successful completion result. This may be caused by racer automatically detecting Rust sys root location ( discussed [here](https://github.com/racer-rust/racer/pull/863) and [here](https://github.com/racer-rust/emacs-racer/issues/93), which was introduced after these tests were written. As a consequence, the 4 tests with
```
@patch( 'ycmd.completers.rust.rust_completer._GetRustSysroot',
        return_value = '/non/existing/rust/src/path' )
```
will not actually change the Rust sysroot value.
However, removing these tests doesn't seem like a good choice. But I haven't come up a better idea yet.

bstaletic:
So... Appveyor failure is a flake. CircleCI I don't know what it is about, perhaps another flake.

> However, removing these tests doesn't seem like a good choice. But I haven't come up a better idea yet.

Just removing the tests to get a passing build would be wrong. If I understand correctly, latest racerd doesn't need `$RUST_SRC_PATH` for basic completion and can even use `rustup` to figure things out. Sounds great. We should update the `rust_completer.py` to reflect these changes and then update the tests according to any new code we introduce.

fandahao17:
> Sounds great. We should update the `rust_completer.py` to reflect these changes and then update the tests according to any new code we introduce.
Perhaps I'll give it a try.

PR #1214, opened 2019-03-20
[READY] Fix off-by-one error in LSP Initialize request

bstaletic:
- LSP `completionItemKind` has values 1 through 25
- Our `ITEM_KIND` is a python list that has `None` as element 0.
- `len( ITEM_KIND ) + 1 == 27`
- `list( range( 1, len( ITEM_KIND ) + 1 ) ) == [ 1 ... 26 ]`
- We shouldn't add 1, because `ITEM_KIND[0] is None`

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1214)


bstaletic:
@zzbot r+

micbou:
@zzbot r+

PR #1215, opened 2019-03-21
[READY] Update Clang to 8.0.0

micbou:
Update Clang to version 8.0.0 for the libclang and Clangd completers. This requires the following changes:
 - the compilation database heuristic is removed in favor of the libclang one;
 - the issue where protected members were not accessible in derived class is fixed;
 - semantic completion is now automatically triggered on include statements in the Clangd completer;
 - the GoTo tests for the Clangd completer are refactored;
 - diagnostic messages returned by Clangd are now capitalized.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1215)


bstaletic:
@zzbot r+

PR #1219, opened 2019-03-26
[READY] Update Unicode Standard to 12.0.0

micbou:
This new version of the Unicode standard doesn't affect us much except [the removal of the surrogate characters from the control ones](https://www.unicode.org/versions/Unicode12.0.0/#UAX_Changes):

> Surrogate code points were moved from Control to XX for the Grapheme_Cluster_Break property, to eliminate the need to have isolated surrogate code points in the test cases.

The [regex Python module](https://bitbucket.org/mrabarnett/mrab-regex) does not yet support the new standard so it will be updated later.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1219)


bstaletic:
@zzbot r+

PR #1220, opened 2019-03-28
[READY] Do not reset state at server start in Clangd completer

micbou:
Resetting the state at server start is unnecessary (it's already reset at completer initialization and at server shutdown) and even problematic as it prevents passing initialization options to Clangd.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1220)


puremourning:
@zzbot r+

PR #1223, opened 2019-03-30
[READY] Include settings in debug info for LSP

puremourning:
This is useful to check that we actually read the settings.

It's equivalent to listing the flags in the libclang completer.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1223)


micbou:
@zzbot r+

micbou:
@zzbot retry

PR #1225, opened 2019-04-01
[READY] Do not ignore extra conf twice in Clangd tests

micbou:
[Extra confs are already ignored in `IsolatedApp`](https://github.com/Valloric/ycmd/blob/c04647006ba5b1d2dc3053a01e5a966515cf1f59/ycmd/tests/test_utils.py#L241).

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1225)


bstaletic:
@zzbot r+

puremourning:
@zzbot retry

PR #1227, opened 2019-04-06
[READY] Use the right application in Java and Tern tests

micbou:
The shared application is used when waiting for the server to be ready instead of the one passed to the `Start*CompleterServerInDirectory` functions.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1227)


puremourning:
@zzbot r+

PR #1228, opened 2019-04-08
[READY] Add numpydoc to Jedi dependencies

micbou:
[Jedi depends on numpydoc](https://github.com/davidhalter/jedi/blob/8d0c4d3cec4973a2722652bf33f093ac3f1f385a/jedi/evaluate/docstrings.py#L53) to get type hinting from docstrings written in [the Numpy format](https://numpydoc.readthedocs.io/en/latest/). We could tell users to install numpydoc for this use case but it's tricky because we need to tell them to install it for the Python used to build ycmd, not the Python that they are working on. This also defeats the purpose of bundling the Jedi and Parso dependencies in the first place since they could be installed separately as well.

Note that this docstring format is used by other projects than Numpy so we are not adding a dependency just for one Python module.

Fixes https://github.com/Valloric/YouCompleteMe/issues/3361.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1228)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -24,6 +24,9 @@
 [submodule "third_party/jedi"]
 	path = third_party/jedi_deps/jedi
 	url = https://github.com/davidhalter/jedi
+[submodule "third_party/jedi_deps/numpydoc"]
+	path = third_party/jedi_deps/numpydoc
+	url = https://github.com/numpy/numpydoc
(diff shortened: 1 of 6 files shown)


bstaletic:
@zzbot r+

bstaletic:
@zzbot retry

PR #1231, opened 2019-04-10
[Ready] Ignore clang-tidy false positive

gauravjuvekar:
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1231)

### DIFF
--- a/cpp/ycm/ycm_core.cpp
+++ b/cpp/ycm/ycm_core.cpp
@@ -69,7 +69,7 @@ PYBIND11_MODULE( ycm_core, mod )
   mod.def( "YcmCoreVersion", &YcmCoreVersion );
 
   // This is exposed so that we can test it.
-  mod.def( "GetUtf8String", []( py::object o ) -> py::bytes {
+  mod.def( "GetUtf8String", []( py::object o ) -> py::bytes {     // NOLINT
                                   return GetUtf8String( o ); } );
 
   py::class_< IdentifierCompleter >( mod, "IdentifierCompleter" )


gauravjuvekar:
Fixes a compile warning

YouCompleteMe/third_party/ycmd/cpp/ycm/ycm_core.cpp:72:44: error: the parameter 'o' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]
  mod.def( "GetUtf8String", []( py::object o ) -> py::bytes {

bstaletic:
Thanks for the PR.
How did you get that error? It comes from `clang-tidy` and is disabled in our `.clang-tidy` configuration file.

gauravjuvekar:
Got it while compiling latest YCM with 

`$ ./install.py --clang-completer --system-libclang --cs-completer --go-completer --rust-completer --java-completer --clang-tidy`
with 
```
Clang 8.0.0
g++ (GCC) 8.2.1 20181127
```

bstaletic:
Okay, I just checked `sizeof(pybind11::object)` and it is `8`, as expected. That makes the clang-tidy diagnostic a false positive, so instead of passing `const object&`, let's turn `clang-tidy` off on that line by appending `// NOLINT`.

gauravjuvekar:
@bstaletic 
Done

bstaletic:
Only the first line needs `// NOLINT`, that's the line for which the warning is reported. Also, squash the commits.

gauravjuvekar:
@bstaletic Done

bstaletic:
Thanks for the pull request.
:lgtm:

bstaletic:
@micbou I disagree. New versions of clang-tidy can introduce aliases of already existing checks. This thing slipped through because we have disabled an old name of the check, but a newer version of clang-tidy added an alias and that's how it slipped through.

If you really do think we should avoid `// NOLINT`, then we need to get the latest clang-tiy working on Travis.

bstaletic:
Also, this isn't a bad warning to have enable, but it does create a false positive here, so I believe that silencing it here is better than completely disabling the warning..

PR #1232, opened 2019-04-10
[READY] Clangd GetType response bugfix

bstaletic:
Right now, clangd completer returns a string as a `GetType` response instead of `{ 'message': $STRING }`.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1232)


bstaletic:
Surprisingly, returning a string magically works. I've been using clangd's `GetType` a lot and... it worked.

micbou:
@zzbot r+

bstaletic:
@zzbot retry

PR #1234, opened 2019-04-12
[WIP] Update JDT to 0.35

bstaletic:
This updates jdt.ls to the latest release.
JDT now really loves to organize imports and generate getters and setters. Also the "delete, then change" and overlapping edits don't seem to exist any more.

The remaining problem is the `Subcommands_FixIt_InvalidURI_test` for which we get a `ResponseAbortedException`. And once that test fails, the next request raises `ValueError: write to closed file`.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1234)


bstaletic:
Nope, 0.38 works better.

PR #1236, opened 2019-04-16
[READY] Allow newer coverage version

bstaletic:
There's still the problem with paths, but codecov has ["Fixing Paths"](https://docs.codecov.io/docs/fixing-paths) section in the docs.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1236)

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -25,6 +25,9 @@ coverage:
   - .*/pybind11/.*
   - .*/whereami/.*
 
+fixes:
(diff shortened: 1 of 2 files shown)


bstaletic:
This is ready. All that was needed was to explain to codecov that `utils.py` is actually `ycmd/utils.py`.

Valloric:
:lgtm:

Thanks for the PR!

@zzbot r+

PR #1239, opened 2019-04-26
[READY] Print diagnostic code field

kadircet:
Clangd started populating diagnostic's code field with relevant information with https://reviews.llvm.org/D60822 and https://reviews.llvm.org/D58291. 

This patch aims to surface that information to ycmd users, as vscode does.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1239)

### DIFF
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -2263,12 +2263,19 @@ def _BuildDiagnostic( contents, uri, diag ):
     filename = ''
 
   r = _BuildRange( contents, filename, diag[ 'range' ] )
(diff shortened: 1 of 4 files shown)


puremourning:
@kadircet can you rebase and push the PR ? It still has the old CI system on the commit checks !

puremourning:
Tests are failing on RLS

kadircet:
yup just saw. I believe rust was not merged when I've written the patch?

They appear to be reporting codes like 'E0609', I have no idea whether they are meaningful or not. Anyone knows if these are worth keeping ?

puremourning:
Latest master should be more flake-free. So maybe one more rebase ?

PR #1245, opened 2019-05-09
[READY] Generic LSP  completer

bstaletic:
With this PR plugging a new server for testing is as simple as sending `language_server = { $LANGUAGE: [ $SERVER_COMMANDLINE ] }`.

The server that is plugged in this way can do everything the `SimpleLSPCompleter` can *by default*, but also provides `GetHover` subcommand to the user. It doesn't try to be smart about it and just returns `{ 'message': self.GetHoverResponse( request_data ) }`.

The testing was done with Microsoft's sample implementation of the server which does't support the hover request.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1245)


bstaletic:
Merge conflicts solved.

puremourning:
Let's go.

puremourning:
Thanks for your patience!

PR #1250, opened 2019-05-15
[RFC] Add extra conf support to Clangd completer

micbou:
This PR adds support for `.ycm_extra_conf.py` files to the Clangd completer by sending the flags to Clangd through [the `didChangeConfiguration` notification](https://microsoft.github.io/language-server-protocol/specification#workspace_didChangeConfiguration) using [the `compilationDatabaseChanges` setting](https://clangd.github.io/extensions.html#compilation-commands) on the `FileReadyToParse` event.

The value of the `language` field used in extra conf files for the Clangd completer is changed to the same value as the libclang completer (`cfamily`) to allow the use of extra confs that were written for the libclang completer. For instance, [ycmd's own `.ycm_extra_conf.py` file](https://github.com/Valloric/ycmd/blob/master/.ycm_extra_conf.py) is supported with no additional changes.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1250)

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,14 +97,13 @@ non-semantic.
 
 There are also several semantic engines in YCM. There's a libclang-based
 completer and [clangd][clangd]-based completer that both provide semantic
-completion for C-family languages. The [clangd][clangd]-based completer doesn't
-support extra conf; you must have a compilation database. [clangd][clangd]
-support is currently **experimental** and changes in the near future might break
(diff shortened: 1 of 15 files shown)


puremourning:
Rebased at #1267

puremourning:
Superseded by #1267. Thanks to @micbou for the PR!

PR #1252, opened 2019-05-17
[READY] Automatically detect codeAction support

bstaletic:
This PR lets us automatically detect `codeAction` support.

Additionally, the change to `LanguageServerCompleter.HandleServerCommand()` was needed because:

> Because of the `GenericLSPCompleter`. If we tell users "`try g:ycm_language_server`" and the server reports `codeActionProvider`, then, because of return None in `SimpleLSPCompleter`, it would just silently fail.
With this change, users (and we) will get a message "hey, for this completer you don't have a proper `HandleServerCommand`".

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1252)


bstaletic:
Fair enough, closing.

PR #1255, opened 2019-05-31
[RFC] WIP: Signature help

puremourning:
Ref: https://github.com/Valloric/YouCompleteMe/issues/234

# Status

This opens up the discussion for adding signature help based on popup menu support that's being implemented in Vim.

It's just a prototype at the moment, but opening a PR to start off a discussion about implementation details.

# Design

Current implementation is to add the signature help information to the 'completions' request. The advantage of this is that it requires the least client code change. The disadvantage is that it strictly delays completions in the general case.

An alternative would be to add /signature_help endpoint and all the client code to manage that request separately. It's an option, but I wanted something quick.

# API

The ycmd API is to add to the completion request:

* `signature_help_state` the current state of the menu. 'ACTIVE' or 'INACTIVE'. This is necessary to constantly re-trigger signature help while the popup is visible. This is the only way in LSP to know when to close th popup: when the sig help returns empty.

And on the reply:

* `signature_info`: [LSP signature help response](https://microsoft.github.io/language-server-protocol/specification#textDocument_signatureHelp).

Why beat around the bush? We just use the LSP format directly because why not. This decision can change if we so decide.

# Misc

This branch also includes #1250 which I needed to test with clangd.
Only supports LSP right now, because that's easy. Tested with clangd and jdt.ls.

# Demos

Screenshot 2019-05-31 at 23 32 31

Screenshot 2019-05-31 at 23 33 05

Screenshot 2019-05-31 at 23 33 39

image

Longer demo:

YCM-popup-win-arg

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1255)


oblitum:
> The purpose of the `labelOffset` is to say where within the signature's `label` the parameter actually falls, to (presumably) avoid the string matching, which I suppose could be borked.
> 
> It's a really dumb API IMO, but it's simple enough to make it work.

tips: https://github.com/neoclide/coc.nvim/issues/269#issuecomment-449015925

puremourning:
Superseded by individual PRs, specifically #1324

PR #1256, opened 2019-06-02
Allow maintainer to manually merge

puremourning:
This allows adding a "Ship It!" label to the PR to allow mergify to
merge it even if the Reviewable is not happy, but assuming there is at
least 1 approving review and no outstanding blocking comments.

This allows maintainers to agree to merge something with essentially
only 1 LGTM, but without always automatically merging based on that. 2
LGTMs is still a good rule for a lot of things, but is cumbersome for
smaller or more urgent PRs.

I've already created the label on Valloric/ycmd repo. Will push this change to YouCompleteMe repo too later.

Tested on my fork:

Screenshot 2019-06-02 at 14 57 20

Removing the label, it updates about 1 minute later to not be ticked (pass).

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1256)

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -8,3 +8,20 @@ pull_request_rules:
       merge:
         method: merge
         strict: smart
+      comment:
+        message: Thanks for sending a PR!
+  - name: Manual merge on Azure Pipelines and Maintainer Override
+    conditions:
+      - base=master
+      - status-success=Valloric.ycmd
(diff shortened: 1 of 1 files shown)


puremourning:
I sneaked in the change to add a "Thanks for the PR" comment too, because bots should have a voice!

PR #1257, opened 2019-06-03
[WIP] Go completer: find root go.mod

bstaletic:
Rust completer renamed GetProjectDirectory to _GetProjectDirectory,
which I didn't notice when rebasing after merging the Rust completer.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1257)


bstaletic:
Like I mentioned in the gitter room, I don't know how to test this, even though I do remember that tests used to fail without the correctly set `rootUri`.
For the record, RLS tests wouldn't catch this either.

puremourning:
Simple fix. Ship it. Tests can come later.

PR #1258, opened 2019-06-04
[READY] Go tests for diagnostics and go.mod detection

bstaletic:
Adds two thing.

- Diagnostics test which I somehow forgot to commit.
- A test to address the concern from #1257. Needs to be isolated so we are sure it starts in a different directory.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1258)


bstaletic:
Rebased the PR and the tests are passing.

puremourning:
new tests. all passing. what's not to like. ship it.

PR #1259, opened 2019-06-06
[READY] Make RLS subcommands tests more reliable

bstaletic:
On Azure pipelines, the GoTo tests started failing for no real reason,
so we spin those tests until they pass.

This commit also updates RLS to the latest nightly, because of a slight
improvement in codeActions.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1259)


puremourning:
tests are green. LGTM. ship it.

PR #1263, opened 2019-06-15
[READY] Support Java extensions, in particular the java debugger

puremourning:
OK, this is a little cheeky, because it's mainly for using with [vimspector](https://puremourning.github.io/vimspector-web/) and in particular to make java debugging possible when you use Vimspector with ycmd : https://github.com/puremourning/vimspector/issues/3#issuecomment-450359563

There are other plugins too, such as one which in theory allows jdt.ls to support building its own code.

We do not include any extensions by default.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1263)


puremourning:
I will add to my mental todo list to add a test which actually exercises the extension loading

PR #1264, opened 2019-06-15
[READY] Add vimspector config

puremourning:
This allows debugging of ycmd tests directly using vimspector. I use a
mapping which runs the test under the cursor, so one of the variables it
asks you for it "test" which is essentially arguments to nosetests.

NOTE: It doesn't work on windows, and it requires the use of a
virtualenv at the moment. This is mainly due to a limitation of
vimspector that it only has simple variable replacement in its config.

It's actually possible to set variables based on shell commands so if the use of venv is painful for anyone else who actually uses it I can improve that.

Demo:

vimspector-YCM-debug-test

To use:

 - install Vimspector: https://puremourning.github.io/vimspector-web/. enable the `HUMAN` mappings (or the vs ones if you prefer. or write your own)
 - install the python debugger `cd /path/to/vimspector; ./install_gadget.py --enable-python`
 - open a test file and navigate to place to break, hit `<F9>`
- hit `<F5>`
 - `Test` should be set to either empty (all tests) or nosetest arg such as `ycmd/tests/get_completions.test:<function>`
 - `python_version` should be `2` or `3` - this selects the regex module that gets loaded.

I also have a bunch of mappings and a compiler plugin that lets me do this with a single keystroke.

This can all be improved over time if people use it a lot.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1264)

### DIFF
--- a/.vimspector.json
+++ b/.vimspector.json
@@ -1,0 +1,33 @@
+{
+  "configurations": {
+    "python - launch nosetests": {
+      "adapter": "vscode-python",
+      "configuration": {
+        "name": "Python nosetests",
+        "type": "vscode-python",
+        "request": "launch",
+
+        "cwd": "${workspaceRoot}",
+        "stopOnEntry": true,
+        "console": "integratedTerminal",
+        "justMyCode": false,
+
+        "debugOptions": [],
+
+        "program": "$VIRTUAL_ENV/bin/nosetests",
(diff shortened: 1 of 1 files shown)


bstaletic:
The configuration works (not that I doubted it), therefore :lgtm:.

puremourning:
great. let's ship it.

PR #1265, opened 2019-06-15
[READY] Ignore dirty gopls repo

puremourning:
After building, it has the gopls binary in there, but the repo doesn't have a .gitignore.

So we ignore it.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1265)


puremourning:
Trivial change, reviewed OK. Ship it.

PR #1266, opened 2019-06-15
[READY] Update JDT to 0.40.0

bstaletic:
A few significant changes:

- JDT doesn't send overlapping code actions or nonsense like "delete+insert"
- JDT loves getters and setters, so those fixits are now everywhere.
- JDT loves to organize imports, so those fixits are now everywhere.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1266)


puremourning:
ok let's rebase this one too now, shall we ?

PR #1267, opened 2019-06-15
[READY] Add extra conf support to Clangd completer

puremourning:
Rebased and fixed up @micbou's PR #1250 

I have been using this for some time and it works well.

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1267)

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,15 +96,13 @@ non-semantic.
 
 There are also several semantic engines in YCM. There's a libclang-based
 completer and [clangd][clangd]-based completer that both provide semantic
-completion for C-family languages. The [clangd][clangd]-based completer doesn't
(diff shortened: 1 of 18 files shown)


bstaletic:
One more thing. Since we never advertised that `compile_flags.txt` works in ycmd and never had plans to support them, let's change the clangd `testdata` to use extra confs instead of `compile_flags.txt`.

puremourning:
> I believe we should restore the lookup order to be same as libclang-based one.

@kadircet - done as above.


I tested the Mac include thing and it makes clangd work on my computers! \o/

puremourning:
Greeeeeeen!

PR #1209, opened 2019-03-17
[READY] Switch C# completer to Omnisharp-roslyn

mispencer:
Switch to new rosyln version of Omnisharp for C# completion.

Missing features:
* Auto-import
* FixIts
* Completion Forcing
* Reloading (i.e. refreshing without restarting Omnisharp)

---
This change is [Reviewable](https://reviewable.io/reviews/valloric/ycmd/1209)

PR #1274, opened 2019-06-26
[READY] Convert to Omnisharp Roslyn

bstaletic:
Compared to #1209 there are only minor changes, to make this mergable:

- Rebased onto latest master to avoid flakes.
- `Diagnostics_Basic_test()` has a `@WithRetry` decorator, because it was flaky.
- String literals don't have a leading `u`, because we're using `from __future__ import unicode_literals`

Big thanks to @mispencer for the work put into writing this completer.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1274)


puremourning:
LGTM. shall we ship it ?

PR #1276, opened 2019-06-29
[READY] Remove debugging subcommands from C# completer

bstaletic:
This removes `StopServer` and `SolutionFile` subcommands. Both were implemented because @mispencer found it convenient to debug ycmd like that, but we shouldn't really expose these to users, because the first isn't really useful and the information the second brings can be found in the debug info.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1276)


bstaletic:
Yeah, I broke the tests... give me a few minutes.

PR #1294, opened 2019-07-25
[READY] Vimspector: Remove duplication of python path

puremourning:
The `--dump-path` option was there for this precise purpose, but until [now](https://github.com/puremourning/vimspector/pull/44) we couldn't use it in Vimspector. Now we can, so let's use it.

Had to move the building of the test completer, because npm spam broke the --dump-path output .

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1294)

### DIFF
--- a/.vimspector.json
+++ b/.vimspector.json
@@ -2,6 +2,25 @@
   "configurations": {
     "python - launch nosetests": {
       "adapter": "vscode-python",
+      "variables": [
+        {
+          "python": {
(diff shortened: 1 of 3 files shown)


puremourning:
I made it so that it _might_ work if you don't use a virtualenv.
I use pyenv and for some reason it doesn't work unless I use pyenv virtualenv.

PR #1299, opened 2019-07-29
Don't consider clangd 10 out of date.

mingtaoy:
The regex used to parse the clangd version only allows major numbers as
a single digit. Consequently, this results in a parsing failure for the
following version string:

```
    clangd version 10.0.0 (https://github.com/llvm/llvm-project.git bd68a052f292b5df7c5717bd880d796ac7507fc0)
```

This diff changes the regex to allow for two digit major version
numbers, and changes the comparison from a string based comparison to a
tuple/int based comparison to ensure that 10.0.0 > 7.0.0.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1299)

### DIFF
--- a/ycmd/completers/cpp/clangd_completer.py
+++ b/ycmd/completers/cpp/clangd_completer.py
@@ -43,7 +43,7 @@ from ycmd.utils import ( CLANG_RESOURCE_DIR,
                          PathsToAllParentFolders,
                          re )
 
-MIN_SUPPORTED_VERSION = '7.0.0'
+MIN_SUPPORTED_VERSION = ( 7, 0, 0 )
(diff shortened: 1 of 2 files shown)


mingtaoy:
No reason, other than to keep the parsing regex as minimally-capturing as possible.

Addressed comments.

mingtaoy:
Addressed comments from @puremourning 

I factored out the regex parsing into a separate function `ParseClangdVersion` to use for unit testing. I included version strings from a clangd built from trunk, a release clangd, and an old clangd version. Not too familiar on naming/formatting conventions in ycmd, so please advise if unacceptable.

PR #1301, opened 2019-08-10
[READY] Update go completer

bstaletic:
Significant changes:

- Rename support
- References support
- Single file outside `$GOPATH` or a go module support

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1301)

### DIFF
--- a/ycmd/completers/go/go_completer.py
+++ b/ycmd/completers/go/go_completer.py
@@ -86,8 +86,10 @@ class GoCompleter( simple_language_server_completer.SimpleLSPCompleter ):
     try:
       result = self.GetHoverResponse( request_data )[ 'value' ]
       return responses.BuildDisplayMessageResponse( result )
-    except language_server_completer.ResponseFailedException:
-      raise RuntimeError( 'Unknown type.' )
+    except RuntimeError as e:
+      if e.args[ 0 ] == 'No hover information.':
+        raise RuntimeError( 'Unknown type.' )
+      raise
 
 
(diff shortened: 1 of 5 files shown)


bstaletic:
I updated the PR to the latest gopls. Diagnostics are behaving odd:

- Running `ycmd/tests/go/diagnostics_test.py` passes, but only if tests aren't isolated.
- Running the test along with the rest of the test suite fails.
  - Relevant gist: https://gist.github.com/bstaletic/711145d1c663a86b223e8b039d67a4c2
- Spinning the tests using `WithRetry` fixes the `FileReadyToParse` and gopls eventually returns correct diags.
- `WithRetry` didn't help with the message poll diag test. I couldn't figure out how to make it pass, so I used `ExpectedFailure()` to skip it.

preslavmihaylov:
can we revive this?

preslavmihaylov:
We could also fast forward to latest gopls changes as this PR is a bit dated

puremourning:
Boris updated it on Momday

laoshaw:
youcompleteme does clangd etc just fine but not doing any autocomplete for golang, there is gopls installed under ycmd/third_party/go/bin/gopls(newest version 0.4.0), do I need configure it somehow in .vimrc to activate that? all I did is "python3 install.py --all", after that go is not autocompleting at all.

bstaletic:
> do I need configure it somehow

No, but please open an issue report instead of discussing here.

PR #1302, opened 2019-08-11
[READY] Update jedi to 0.15.0 and parso to 0.5.0

bstaletic:
This version of jedi correctly handles python's type hints and I think supports python 3.8.

Jedi now bundles [`typeshed`](https://github.com/python/typeshed) as a submodule, which might throw off some users. This is also why our tests sometimes end up in `typeshed/stdlib/2and3/builtins.pyi`.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1302)

### DIFF
--- a/ycmd/completers/python/python_completer.py
+++ b/ycmd/completers/python/python_completer.py
@@ -248,22 +248,16 @@ class PythonCompleter( Completer ):
   def _BuildGoToResponse( self, definitions ):
     if len( definitions ) == 1:
       definition = definitions[ 0 ]
-      if definition.in_builtin_module():
-        raise RuntimeError( 'Can\'t jump to builtin module.' )
       return responses.BuildGoToResponse( definition.module_path,
                                           definition.line,
                                           definition.column + 1 )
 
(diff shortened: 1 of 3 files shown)


bstaletic:
Finally figured the errors out!

- The `builtins.pyi` is sometimes `__builtin__.pyi` and `2to3` is sometimes `2`, so I only matched `
- The `numpydoc` can't be updated because the new version depends on `sphinx` which has a ton of dependencies and doing it all in `jedi_deps` is insane.
  - Full list of `sphinx` dependencies according to Arch:

```
Depends On      : python-babel  python-docutils  python-imagesize  python-jinja  python-pygments  python-requests  python-setuptools  python-snowballstemmer
                  python-sphinx-alabaster-theme  python-sphinxcontrib-applehelp  python-sphinxcontrib-devhelp  python-sphinxcontrib-htmlhelp
                  python-sphinxcontrib-jsmath  python-sphinxcontrib-qthelp  python-sphinxcontrib-serializinghtml
```

PR #1305, opened 2019-08-30
Update ycm_core.cpp

zwkno1:
fix gcc error:

cpp/ycm/ycm_core.cpp:72:44: error: the parameter 'o' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1305)


bstaletic:
Thanks for the PR, but this is not right.

See https://github.com/ycm-core/ycmd/pull/1246

PR #1310, opened 2019-09-11
Add smart_case option to enable/disable smart case matching

iperry:
Introduce a 'smart_case' option to control query matching behavior.

When disabled, use normal case-sensitive matching, otherwise retain the
existing "smart case" functionality.

Set the default to true to maintain compatibility.

Global option 'g:smart_case' controls this behavior from vim, defaulting to '1'.

### Motivation:

The default smart case behavior may not always be wanted. I've included an example based on some C code that uses a naming convention like so (for subsystem 'foo'):
1) Register contents are #defined like so `#define FOO_CONSTANT`
2) Function names are lowercased `void foo_do_something()`

The default smart case matching can be too noisy, with limited utility:
smart_case_enabled

With smart_case = 0, lowercase completion:
smart_case_disabled_lower

With smart_case = 0, uppercase completion:
smart_case_disabled_upper

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1310)

### DIFF
--- a/cpp/ycm/Candidate.cpp
+++ b/cpp/ycm/Candidate.cpp
@@ -75,7 +75,7 @@ Candidate::Candidate( std::string&& text )
 }
 
 
-Result Candidate::QueryMatchResult( const Word &query ) const {
+Result Candidate::QueryMatchResult( const Word &query, bool smart_case ) const {
   // Check if the query is a subsequence of the candidate and return a result
   // accordingly. This is done by simultaneously going through the characters of
   // the query and the candidate. If both characters match, we move to the next
@@ -112,7 +112,7 @@ Result Candidate::QueryMatchResult( const Word &query ) const {
     const auto &candidate_character = *candidate_character_pos;
     const auto &query_character = *query_character_pos;
 
-    if ( query_character->MatchesSmart( *candidate_character ) ) {
+    if ( query_character->Matches( *candidate_character, smart_case ) ) {
       index_sum += candidate_index;
 
       ++query_character_pos;
--- a/cpp/ycm/Candidate.h
+++ b/cpp/ycm/Candidate.h
@@ -50,7 +50,7 @@ public:
     return text_is_lowercase_;
   }
 
-  YCM_EXPORT Result QueryMatchResult( const Word &query ) const;
+  YCM_EXPORT Result QueryMatchResult( const Word &query, bool smart_case ) const;
 
 private:
   void ComputeCaseSwappedText();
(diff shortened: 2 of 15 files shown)


bstaletic:
Thanks for the pull request. However, I'm not convinced by this feature. For example, say you want to complete `gpio_port_read`. Instead of typing `gpio_`, try [`gpr`](https://i.imgur.com/Fzp6s6J.png) (`gr` also works). And for `GPIO_IDR` macro try [`gi`](https://i.imgur.com/O9nfNB1.png).

The algorithm favours word boundaries over consecutive matches and so should the users.

iperry:
> Thanks for the pull request. However, I'm not convinced by this feature. For example, say you want to complete `gpio_port_read`. Instead of typing `gpio_`, try [`gpr`](https://i.imgur.com/Fzp6s6J.png) (`gr` also works). And for `GPIO_IDR` macro try [`gi`](https://i.imgur.com/O9nfNB1.png).
> 
> The algorithm favours word boundaries over consecutive matches and so should the users.

It's really an issue of discoverability.

This scenario assumes that I know I want to complete `gpio_port_read`. In this situation, I am working with an external API that I do not already know well, completing `gpio_` in this particular scenario allows me to discover the available functions, without the register constants polluting my candidate list with noise.

puremourning:
what's the status of this PR?

@iperry many apologies for not looking at it until now. It honestly totally slipped past me. 

First, thanks so much for sending a PR and putting together a nice comprehensive description!

I would like to give the change a try and see how it feels. We're obviously cautious of changing the filtering and sorting logic on the basis that it really defines the mental model our users have with YCM. At the same time, we're generally against adding more options so far as that it possible.

While I personally don't use smart case matching in the general case, I do prefer it for completion because I have to press fewer keys most of the time. Usually, as Boris says, I'm entering word-boundary cahracters which are typically upper case (though not always), so relying on smart case (or case insensitive) matching is usually more efficient. I think that translates well to a lot of use cases on the basis that we have rarely if ever been asked to change this.

And this comes to the crux of my opinion on this: it's not a commonly requested feature to my knowledge. I would have thought if this was very useful it would have been requested before. I suspect that our current rules are just good enough for most people and most use cases. 

I'm also not convinced that `g:smart_case` is good or appropriate. We certainly don't want to match the `smart case` option in the client, because that would be a breaking change, and we probably don't really want an option at all. I suspect that if we _did_ do something it would be along the lines of having a way to trigger completions specifically in a case-sensitive way. The challenge is there's no obvious or intuitive way to do so.

I'm open to ideas on the latter point.

What's not really negotiable is tests. We absolutely will need a suite of tests that cover this difference in behaviour across the feature set, including things like unicode etc.

puremourning:
This PR seems to be abandoned. Thanks for the contribution anyway!

PR #1316, opened 2019-09-16
[READY] Roslyn update

bstaletic:
A couple of changes:

- `GoToReferences` support
- Tests are cleaned up a lot.
- Roslyn 1.32.19 -> 1.34.2 update, because why not.

The PR can be reviewed by commits, instead of as a whole.

Roslyn also supports `FixIt`s and singature help, but I couldn't find any kind of API documentation to implement those.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1316)


puremourning:
> Roslyn also supports FixIts and singature help, but I couldn't find any kind of API documentation to implement those.


IIRC the documentation is the code. We can add them later if needs.

PR #1318, opened 2019-09-18
[READY] Specify documentation formats supported and preference plaintext

puremourning:
LSP allows a preference order for documentation formats. There has been discussion about adding random ones like RST and HTML etc.

We prefer plain text because that's all that we can display in our clients.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1318)


puremourning:
I realise I should probably add a test for this, but I actually don't know a server which actually implements the priority list correctly, so this is a largely harmless change, in preparation for signature help (1/N)

PR #1324, opened 2019-09-19
[READY] SignatureHelp: Support signature help for c-family using clangd

puremourning:
This implements the LSP signature help support and tests it for clangd.

A new options was added: disable_signature_help. if for some reason you
don't like signature help, this can be set to disable it.

NOTE: No special changes were required to clangd_completer as this is
implemnented completesly in the language_server_completer layer.
Therefore, this change acutally makes signature help work for any LSP
based completer, while tests are only provided (for now) for c-family
using clangd..

This isn't perfect yet: there's no capability mechanism to indicate
that signature help isn't supported for a filetype so the client will
just constantly send requests that return emnpty. For now, we don't
solve that, instead deferring that decision to either send an error
(like we do with the message poll), or add a new capability response,
e.g. to the completer available reqeust (such as adding a version 2 of
that request that returns an objext).

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1324)


bstaletic:
Woohoo!
@puremourning thanks for all the work you put into this.

PR #1336, opened 2019-10-05
added support for typescriptreact (.tsx) files

lincoln-b:
As suggested in https://github.com/ycm-core/YouCompleteMe/issues/3499, this is a pull request for adding support for .tsx files to YouCompleteMe. 

The text of that issue is copied below for convenience.

*******

Since vim has added [two new file types](https://github.com/vim/vim/issues/4830) for Javascript and Typescript files that use JSX syntax, files with extensions .jsx and .tsx now have the filetype "javascriptreact" and "typescriptreact" respectively.

Is it possible to enable the typescript auto completer for those filetypes as well by default? Right now when editing those files I get the error `No semantic completer exists for filetypes: ['typescriptreact']`.

My temporary workaround for now is to put the following line in my .vimrc:

    au BufNewFile,BufRead *.tsx set filetype=typescript

This works, but it would be even better to be able to do this out of the box.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1336)

### DIFF
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -361,7 +361,7 @@ class TypeScriptCompleter( Completer ):
 
 
   def SupportedFiletypes( self ):
-    return [ 'javascript', 'typescript' ]
+    return [ 'javascript', 'typescript', 'typescriptreact' ]
 
 
   def ComputeCandidatesInner( self, request_data ):
--- a/ycmd/completers/typescriptreact/hook.py
+++ b/ycmd/completers/typescriptreact/hook.py
@@ -1,0 +1,33 @@
+# Copyright (C) 2015 Google Inc.
+#
+# This file is part of ycmd.
+#
+# ycmd is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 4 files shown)


lincoln-b:
Didn't even notice the copyright weirdness because I copied straight from `ycmd/completers/typescript/hook.py` which has the same header.

```
C02Z53Y9LVDR:ycmd i510882$ ack -Q 'Copyright (C) 2015 Google Inc.'
ycmd/completers/go/hook.py
1:# Copyright (C) 2015 Google Inc.

ycmd/completers/typescript/hook.py
1:# Copyright (C) 2015 Google Inc.

ycmd/hmac_utils.py
1:# Copyright (C) 2015 Google Inc.
```

Anyway, let me know if the test I added is adequate.

lincoln-b:
Not sure what you mean by completion test. Are you referring to https://github.com/ycm-core/ycmd/blob/master/ycmd/tests/typescript/typescript_completer_test.py? I am using the same TypeScriptCompleter as other tests.

bstaletic:
Sweet, there's even a test there. Thanks for the pull request @lincoln-b  :lgtm:

lincoln-b:
Thanks everyone!

PR #1338, opened 2019-10-09
[READY] Implement `applyEdit` to allow incomplete `FixIts`

bstaletic:
When making an `executeCommand` request, a LSP server might in exchange
send a series of "things", which we expect to be `applyEdit` requests.

We then store those requests in a list until the LSP server responds to
the original `executeCommand` and finally let the specific LSP completer
implementation figure out what to do with those unconsumed `applyEdit`s.

This allows ycmd to forward an incomplete `codeAction` from clangd to a
client in order to save time computing expensive `codeActions`.

The exchange should look like this:

- A client makes a `FixIt` request.
- ycmd makes a `codeAction` request to clangd.
- clangd responds with a mix of resolved and unresolved code actions.
- ycmd transforms resolved code actions in `FixIts` and forwards
unresolved ones.
- A client lets a user choose a fixit from the list.
- If the chosen fixit is unresolved, the client sends a /resolve_fixit
request.
- ycmd executes LSP server specific logic to resolve the fixit and
responds with another `FixItResponse`.
- In case of clangd, the server specific logic is:
  - Make the appropriate `executeCommand` request.
  - Store the received `applyEdit`.
  - Turn the `applyEdit` into a ycmd `FixIt`.
  - Send a `FixItResponse` to the client.

Open questions:

- ~~Should `ResolveFixit` in `ClangdCompleter` be moved to
`SimpleLSPCompleter` or even `LanguageServerCompleter`?~~
- ~~How should "stray" `applyEdit`s be drained?~~
- ~~What if an `executeCommand` triggers server-to-client requests other
than `applyEdit`?~~
- ~~@puremourning mentioned `codeActionLiteralSupport` capability, but I'm not sure what is that supposed to do.~~
- ~~Should we advertise `applyEdit` capability?~~

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1338)

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -220,6 +220,18 @@
                     </tr>
                     <tr>
                             <td class="swagger--summary-path" rowspan="1">
+                                <a href="#path--resolve_fixit">/resolve_fixit</a>
+                            </td>
+                        <td>
+                            <a href="#operation--resolve_fixit-post">POST</a>
+                        </td>
+                        <td>
+                            <p>Resolve an incomplete FixIt.</p>
+        
+                        </td>
+                    </tr>
+                    <tr>
+                            <td class="swagger--summary-path" rowspan="1">
                                 <a href="#path--run_completer_command">/run_completer_command</a>
                             </td>
                         <td>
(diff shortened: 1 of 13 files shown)


puremourning:
> @puremourning mentioned codeActionLiteralSupport capability, but I'm not sure what is that supposed to do.

This would allow us to avoid a round-trip if the server can send the `edit` directly in the response to the `codeAction` request. We can do this in another PR.

PR #1339, opened 2019-10-13
Allow user configuration of Format subcommand for JDT completer

bstaletic:
fixes ycm-core/YouCompleteMe#3505

I also considered just letting every LSP completer do this kind of configuration. For that I'd like to check if and how are other LSP servers configured.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1339)

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,6 +335,23 @@ def Settings( **kwargs ):
   }
 ```
 
+##### Java settings
+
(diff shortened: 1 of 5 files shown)


puremourning:
tests are still not working ._0

PR #1350, opened 2019-10-30
[READY] Implement GetDoc for gopls

bstaletic:
I don't when this happened, but gopls can apparently provide docs, with some curious [initialization options](https://github.com/golang/tools/blob/2b779830f9d33eccacea7d9e741475351b225b1a/internal/lsp/source/hover.go#L20-L36).

We always get `signature` appended to the hover info. That's what `GetType` displays. Besides that, we also get `Synopsis` when available. To get `FullDocumentation` we need to pass that as `HoverKind` in initialization options: https://github.com/golang/tools/blob/2b544e3f2db1e740d696ff91381987f6357e6bda/internal/lsp/source/options.go#L74-L107.

Speaking of initialization options, [the default completion options](https://github.com/golang/tools/blob/2b544e3f2db1e740d696ff91381987f6357e6bda/internal/lsp/source/options.go#L63-L69) are just as interesting. Then again, maybe it doesn't matter, since we never send any query for gopls to filter completions.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1350)

### DIFF
--- a/ycmd/completers/go/go_completer.py
+++ b/ycmd/completers/go/go_completer.py
@@ -82,10 +82,21 @@ class GoCompleter( simple_language_server_completer.SimpleLSPCompleter ):
     return [ 'go' ]
 
 
-  def GetType( self, request_data ):
+  def GetDoc( self, request_data ):
     try:
       result = self.GetHoverResponse( request_data )[ 'value' ]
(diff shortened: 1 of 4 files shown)


bstaletic:
I'll reopen when it's actually ready.

PR #1354, opened 2019-11-07
[READY] Make LSP codeActions generic

bstaletic:
We can finally move all the FixIt LSP mess into the `LanguageServerCompleter`. It should even work for `GenericLSPCompleter`.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1354)


puremourning:
mergify seems to have given up on this. merging manually.

PR #1362, opened 2019-11-16
Add detection for python 3.8 and 3.9

morguldir:
Arch now uses python3.8

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1362)


morguldir:
Glad to help, might have to do with me having python36 installed.

PR #1363, opened 2019-11-17
[READY] Merge SimpleLSPCompleter into LanguageServerCompleter

bstaletic:
Also moves RestartServer to the base class.

RFC because I'm sure it can be polished more. For example, the `_server_state_mutex` and `_server_info_mutex` still exist.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1363)


bstaletic:
>  For that matter the restriction with the `Simple` completer was that it had to be stdio IIRC. is there still possibility of having a custom startup/connection type ?

Sure. A configurable connection can be added if the need arises. the `StartServer()` didn't change at all.

> IIRC there was an intention to keep the health check lock free as it is called a lot. Maybe that's no longer relevant or needed.

The health check isn't lock free on master. I don't know what the intention was, but yeah. With this pull request, I tried my best to keep the existing behaviour.

> Sorry it will take some head scratching to work out the locking approach now. it should be simpler I guess, but I will need to page in how it all works again.

Absolutely understandable. You definitely understand this part of the code better than I do. Current situation on master is that the `LanguageServerCompleter` base class uses `server_info_mutex`, the `Simple` derived class uses `server_state_mutex` and the `StandardIOConnection` uses two mutexes for stdin/stdout.

This PR only removed `server_state_mutex` and did not touch the `StandardIOConnection`.

bstaletic:
Once again codecov is red because the test for `if not self._use_clean_workspace and wipe_workspace:` in java's `StartServer()` doesn't exist.

bstaletic:
Newest new merge conflicts are resolved.

PR #1367, opened 2019-11-24
[READY] Remove frozendict

bstaletic:
Python deprecated accessing abstract collections through `collections` module in favour of `collections.abc`. Frozendict was never fixed, but ycmd has had `HashableDict` for a while which is also an immutable dictionary and a more robust\* one than `frozendict`. Thus the fix for the broken `frozendict` dependency is to just use our own `HashableDict` even for user options.

\* `frozendict.__getitem__()` returned values by reference, so it was still possible to mutate the dictionary. `HashableDict.__getitem__()` calls `deepcopy()` instead.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1367)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
 [submodule "third_party/bottle"]
 	path = third_party/bottle
 	url = https://github.com/defnull/bottle
-[submodule "third_party/frozendict"]
-	path = third_party/frozendict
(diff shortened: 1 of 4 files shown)


bstaletic:
There was the same problem with openssl@1.0 here as well. Fixed in the same commit. Should I split them in two commits?

PR #1368, opened 2019-11-28
[READY] Implement WorkspaceClientCapabilities.workspaceEdit.documentChanges

bstaletic:
As discussed on gitter, this implements `documentChanges` because some servers ignore thte capabilities and always send `WorkspaceEdits` containing versioned `documentChanges`. Speaking of "versioned", we are just ignoring the versions, which seemed to work for typescript-language-server and rust-analyzer.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1368)


bstaletic:
:lgtml:

PR #1369, opened 2019-12-01
[READY] Simplify LSP completer subcommand discovery

bstaletic:
All LSP completers were already using the exact same `_RestartSever()`, so this PR moves it into the base class.
When `GetType` and/or `GetDoc` is defined for an LSP completer, it is automatically discovered in the base class, so we don't have to repeat it everywhere in the `GetCustomSubcommands`.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1369)


bstaletic:
Fixed the tests...

PR #1370, opened 2019-12-01
[READY] Make textDocument/sigantureHelp response handling more robust

bstaletic:
Ycmd couldn't cope with the following:

- `result:null` - happens with gopls on `non_existent_func(`
- `result:{'signature':[]}` - i.e. no parameters; happens with gopls for every function that has no parameters.

Both of those are valid responses.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1370)


puremourning:
Then we need to include it even if empty.

PR #1372, opened 2019-12-02
[READY] GoToImplementation for TSServer

bstaletic:
This is the only thing that typescript-language-server could do and we couldn't.

There's also a removed `try` and `catch` statement, since the exception is impossible to trigger as far as I can tell and isn't covered by tests.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1372)


bstaletic:
:lgtm:

PR #1373, opened 2019-12-04
[RFC] Clangd GetType format

bstaletic:
The `GetType` response from clangd, in the current master, takes two lines. This causes "hit <Enter> to continue" message in vim, which can be annoying.

The first line is `Declared in namespace foo` and the second is something like `struct Foo{}`.

After some discussion with @maximebl, we thought the most readable way of putting all the information on the same line is `struct Foo{}; Declared in namespace foo.`

RFC because popups are an options, even though I haven't been too fond of that idea so far, though I can be convinced that they might be better.

&nbsp;

The other change in this PR is actually a bug fix - the `GetDoc` response should have been using the detailed info response, not the message response. This goes for clangd as well as gopls. Feel free to tell me that this should be in a different PR.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1373)


bstaletic:
:lgtm:

PR #1375, opened 2019-12-09
Feature Request: Add build.py CMake executable location argument

jgehrig:
I run YouCompleteMe on Windows at work. I frequently run into issues with the CMake detection logic in `build.py`. 

In these cases CMake is always installed, in `%PATH%`, and callable from `cmd.exe`:
``` bash
> where python
{ Prints path to cmake.exe }
> cmake --version
{ Prints CMake version }
> build.py
ERROR: Unable to find executable 'cmake'. CMake is required to build ycmd
```

I usually work around this issue temporarily by hard-coding the problematic line:
```patch
-return FindExecutableOrDie( 'cmake' , 'CMake is required to build ycmd' )
+return '''C:\Program Files\CMake\bin\cmake.exe'''
```

It would be nice to have an argument where the CMake executable can be quickly specified. Any thoughts on adding such an argument?

Here is a simple patch... I'm happy to change anything.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1375)

### DIFF
--- a/build.py
+++ b/build.py
@@ -444,6 +444,10 @@ def ParseArguments():
                               'on the ycm_core code itself.' )
   parser.add_argument( '--core-tests', nargs = '?', const = '*',
                        help = 'Run core tests and optionally filter them.' )
+  parser.add_argument( '--cmake-path',
+                       help = 'For developers: specify the cmake executable. '
+                              'Useful for testing with specific versions, or '
+                              'if the system is unable to find cmake.' )
 
   # These options are deprecated.
   parser.add_argument( '--omnisharp-completer', action = 'store_true',
@@ -480,8 +484,13 @@ def ParseArguments():
   return args
 
 
-def FindCmake():
-  return FindExecutableOrDie( 'cmake', 'CMake is required to build ycmd' )
+def FindCmake( args ):
+  cmake_exe = 'cmake'
+
+  if args.cmake_path:
+    cmake_exe = args.cmake_path
+
+  return FindExecutableOrDie( cmake_exe, 'CMake is required to build ycmd' )
 
 
 def GetCmakeCommonArgs( args ):
@@ -1106,7 +1115,7 @@ def WritePythonUsedDuringBuild():
 
 
 def DoCmakeBuilds( args ):
-  cmake = FindCmake()
+  cmake = FindCmake( args )
   cmake_common_args = GetCmakeCommonArgs( args )
 
   if not args.skip_build:


puremourning:
Thanks for sending a patch!

I'm actually not against this in general, however...

> In these cases CMake is always installed, in %PATH%, and callable from cmd.exe:

I would much rather understand why this isn't working for you. The logic has worked for me and other YCM maintainers for years. While we don't regularly run on windows, it does work in our CI.

Can you debug why it can't find it ?

bstaletic:
@puremourning said everything I wanted to say. If `FindExecutable()` isn't working, you might have much bigger issues than just finding `cmake.exe`.

jgehrig:
For me, it happens on different machines for different reasons. I can usually massage PATH, and get things working. However, Sometimes it is easier to manually specify the path rather that diagnose a one-off PATH problem in Windows... Windows is a mess.

I know `build.py` **should** compile without this option, and that this scenario indicates there is some sort of Windows + PATH + Python issues going on with my systems. After all, `FindExecutableOrDie` simply inspects `os.environ[ 'PATH' ]` for `cmake.exe`.

I perform this hack often enough, and it seems this could be a useful convenience feature.

jgehrig:
>Can you debug why it can't find it ?

Sure, I can debug the real root cause.

I currently have one of my Windows machines in this state...

bstaletic:
The important thing to understand here is if it is a bug in `FindExecutable()` or just a user error. If it's the former we should fix the actual bug.

jgehrig:
> The important thing to understand here is if it is a bug in FindExecutable() or just a user error. If it's the former we should fix the actual bug.

For this specific machine, the issue is the presence of `""` in `%PATH%`.

On this machine `%PATH%` is specified as:
`C:\Program Files\ConEmu; ... "C:\Program Files\CMake\bin"`

The syntax is not correct, but it is tolerated inconsistently:
```
> cmake --version
cmake version 3.16.0

CMake suite maintained and supported by Kitware (kitware.com/cmake).

> where cmake
INFO: Could not find files for the given pattern(s).
```

jgehrig:
However, my primary point with this Pull Request was:
> I perform this hack often enough, and it seems this could be a useful convenience feature.

I would like to be able to manually specify `cmake.exe` on Windows. Currently I do that by editing build.py.

On Windows there can easily be multiple versions of CMake (installed or in PATH). There are also common configurations for which CMake is installed without a PATH entry. Specifying the path is easier than messing with Window's PATH.

Also... Thanks for the awesome software and quick responses! 👍

puremourning:
> Alright, since FindExecutable isn't buggy, I'll be fine with this PR once the CI is green.

Me too

bstaletic:
@jgehrig Thanks for the PR.

:lgtm:

jgehrig:
It looks like the codecov report is marked as failing... I'm not sure exactly how the report is related to my changes.

Is there anything I should do to address this?

bstaletic:
It's fine. Codecov is just flaky.

This PR is waiting @puremourning to give his official blessing (LGTM) before we merge.

jgehrig:
> Can you briefly describe how you tested this ?

Sure. I just performed a basic build with and without the new `--cmake-path` argument on two machines.

In my case:
`build.py --clang-completer --clang-tidy --cmake-path "C:\Program Files\CMake\bin\cmake.exe"`
`build.py --clang-completer --clang-tidy`

Both of these builds were performed on Windows, one with `cmake` in `PATH` and one without. Everything behaved as expected: on the machine without `cmake.exe` or `--cmake-path` the build hits the expected CMake not found failure . Every other build passes and YCM for Vim works as expected.

I could also do some quick checks on a Linux machine this evening if you feel that would be useful.

If the user provides an invalid executable:
```
# build.py --clang-completer --clang-tidy --cmake-path "C:\Program Files\CMake\bin\badpath.exe"
ERROR: Unable to find executable 'C:\Program Files\CMake\bin\badpath.exe'. CMake is required to build ycmd
```
Currently, there is no guard to prevent a user from selecting a valid, but unrelated executable:
```
# build.py --clang-completer --clang-tidy --cmake-path "C:\Program Files\CMake\bin\cmake-gui.exe"
```

PR #1378, opened 2019-12-10
[READY] Add typescriptreact default semantic triggers

### LINKED ISSUE #1377: TypeScriptReact not working properly
Hello! Thank you for your work on this project 🙏

I was very happy to see that #1336 was merged which would make it easier to do TS development in vim, however it does not seem to work. I have spent quite some time the last few days trying to get it to work but I am stuck.

I am editing the same `.tsx` file in the examples.

This is the resulting completion when the type is `typescriptreact` (default):
Skärmavbild 2019-12-10 kl  20 48 37

If I force the filetype to `typescript` (`:set filetype=typescript`) I get the following:
Skärmavbild 2019-12-10 kl  20 47 11

Another difference is that in the second example the completion appears instantly after I type `.` but in the first example I have to type a few characters.

Any ideas what it might be/how to solve this?

bstaletic:
Fixes #1377 

The pull reequest that added typescriptreact support did not add the default semantic completion triggers.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1378)

### DIFF
--- a/ycmd/completers/completer_utils.py
+++ b/ycmd/completers/completer_utils.py
@@ -226,6 +226,7 @@ DEFAULT_FILETYPE_TRIGGERS = {
     'python,'
     'scala,'
     'typescript,'
+    'typescriptreact,'
     'vb' ) : [ '.' ],
   'ruby,rust' : [ '.', '::' ],
   'lua' : [ '.', ':' ],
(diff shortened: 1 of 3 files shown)


bstaletic:
Would you prefer me to add tests for diagnostics and all the other things that were left out in the original PR?

puremourning:
Not unless you really want to. I think this is fine.

bstaletic:
Eh... let's merge.
:lgtm:

PR #1380, opened 2019-12-17
[READY] Implement `Format` subcommand for the C# completer

bstaletic:
Just what the title says. There are two things I need to point out. 

There *is* a [`/formatRange`](https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Abstractions/OmniSharpEndpoints.cs#L12) request, but it is only concerned with the end codepoint. Without the start codepoint, ranged formatting request seems completely broken. https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Abstractions/Models/v1/Format/FormatRangeRequest.cs

The `/codeformat` request formats the entire file and has no format options: https://github.com/OmniSharp/omnisharp-roslyn/blob/master/src/OmniSharp.Abstractions/Models/v1/CodeFormat/CodeFormatRequest.cs

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1380)

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -33,7 +33,10 @@ import threading
 from ycmd.completers.completer import Completer
 from ycmd.completers.completer_utils import GetFileLines
 from ycmd.completers.cs import solutiondetection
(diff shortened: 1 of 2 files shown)


bstaletic:
:lgtm:

PR #1381, opened 2019-12-21
[READY] Support user-specified project directory in LSP servers

puremourning:
Store all settings returned from the Settings function

Allow extra conf to supply project directory. This is useful at least
for java with complex workspaces containing multiple .project files

For the output in debug info, we _could_ change to show all settings,
but for now, let's keep it as it was before.

TODO:
* [ ] Tests for other language server backends

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1381)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -19,7 +19,8 @@ import ycm_core
 import os
 import inspect
 from ycmd import extra_conf_store
-from ycmd.utils import ( OnMac,
+from ycmd.utils import ( AbsoluatePath,
+                         OnMac,
                          OnWindows,
                          PathsToAllParentFolders,
                          re,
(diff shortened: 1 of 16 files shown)


puremourning:
> but rather making the workspace from extra conf a list of initially known workspaces and sending those in the initialize request?

I guess that's an option. But in my case I would return the one path which is the root of the tree (`src/` in my example). We would also then have to implement multiple workspace folders, which I don't know if we want to do. 

Come to think of ot I don't even know if jdt.ls treats multiple workspace folders as the same or different eclipse workspaces. I would actually expect them to be _different_ eclipse workspaces for this exact reason - that it searches recursively for projects. It's possible with multiple "folders" you could have colliding projects (such as the same source code  in different branches).

PR #1382, opened 2019-12-24
[READY] Drop python2 support

### LINKED ISSUE #1365: Plan to drop python 2 support
Same story as  ycm-core/YouCompleteMe#3538

- [ ] Drop `python-future` dependency.
- [ ] Remove python 2 from `build.py` and `CMakeLists.txt`.
- [ ] Remove all `PY2` conditional code and `__future__` imports.
- [ ] Replace `super( Class, self )` with `super()`.
- [ ] Remove `# coding: utf-8` headers from files.
- [ ] Use `class C:` instead of `class C( object ):`.
- [ ] Use `str.encode()` instead of `str.encode( 'utf-8' )`.
- [ ] Replace the single use of `inspect.getargspec().keywords` with `inspect.getfullargspec().varkw`

Did I miss anything?

bstaletic:
The commit history is a mess, but this is a complete set of changes for dropping python2.

Fixes #1365

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1382)


bstaletic:
@iogf @abingham @ptrv @Qusic @LuckyGeck @akshayganeshen @mawww @richard1122 @jakeanq @orsonteodoro 

We're dropping python2 support. For clients not written in python, there will be a new exit code that indicates that ycmd has been started with python2. For clients written in python, that also depend on things in ycmd module (as far as I know that's only YCM), some utilities were dropped too.

bstaletic:
:lgtm:

PR #1383, opened 2019-12-24
[READY] Run valgrind in CI

bstaletic:
Python 3.8 is finally nice enough to not get in the way of running valgrind with `PYMALLOC=malloc`. The question is what should we run through valgrind? Here are the options:

- `ycm_core_tests`, which embeds the python interpreter. Currently valgrind detects [a leak in pybind](https://github.com/pybind/pybind11/pull/2020)
- `ycm_benchmarks`, same comment as the above.
- nose tests? I haven't tried this, but valgrind will detect [another leak in pybind](https://github.com/pybind/pybind11/pull/2024)
- Some combination of the above?

Currently the CI is running only a single benchmark, from `ycm_benchmarks` binary.

CI is also complaining about a missing `libsqlite3-dev` package. We probably should fix it, even though the tests aren't broken.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1383)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,7 +20,7 @@ jobs:
         YCM_BENCHMARK: true
         COVERAGE: false
       'C++ linting':
-        YCM_PYTHON_VERSION: '3.6.3'
+        YCM_PYTHON_VERSION: '3.8.0'
         YCM_CLANG_TIDY: true
         COVERAGE: false
     maxParallel: 6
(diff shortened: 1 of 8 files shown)


bstaletic:
Ugh... I just realized that I'll have to redo suppressions when we move to pytest/python3.6 because some internal functions, visible in debug build and therefore in the suppressions, have changed.

I'll wait until we merge pytest. Marking as WIP.

bstaletic:
Just a note. `--ignore` doesn't work when you explicitly select a file via glob, so `--ignore=ycmd/tests/shutdown_test.py` in `run_tests.py` doesn't work. The solution is to use `--deselect` instead. I'll push the change once we fix CI in #1400.

PR #1386, opened 2020-01-07
[READY] Raise exception on empty C# /typelookup

### LINKED ISSUE #1385: CSharp Completer - "Type" response field is sometimes empty
https://github.com/ycm-core/ycmd/blob/5091b462e7e03759e6aa8c6eb3b8ed05df50e4a8/ycmd/completers/cs/cs_completer.py#L701

I've logged the request's response and found out that, sometimes, the [ "Type" ] field comes out as `None`. _Sometimes_ means it works reliably for some cases, but not for all of them.

When it's missing, it causes YCM to thrown an exception, although it does not seem like the fault of it being absent is in our end - this is simply a unexpected failure case not being handled.

---

I can reproduce the empty response 100% of the time in the following situations
- When trying to get the documentation for a generic Class
- When trying to get the documentation for a [Field](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/fields) in an Interface

Those are situations to be reported in-depth in [omnisharp-roslyn](https://github.com/OmniSharp/omnisharp-roslyn) perhaps, but I think YCM needs to handle the case where the response comes as `None` and not have it blow up inadvertently. It's happening very often in a fairly robust project.

bstaletic:
Fixes #1385 
Returning `{ 'detailed_info': None }` and `{ 'message': None }` causes problems for clients, including YCM. Instead, we should raise an exception and tell the user that there's no type info/docs.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1386)

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -602,6 +602,8 @@ class CsharpSolutionCompleter( object ):
     result = self._GetResponse( '/typelookup', request )
     message = result[ "Type" ]
 
+    if not message:
+      raise RuntimeError( 'No type info available.' )
     return responses.BuildDisplayMessageResponse( message )
 
 
@@ -698,11 +700,14 @@ class CsharpSolutionCompleter( object ):
     request[ "IncludeDocumentation" ] = True
 
     result = self._GetResponse( '/typelookup', request )
-    message = result[ "Type" ]
+    message = result.get( 'Type' ) or ''
+
     if ( result[ "Documentation" ] ):
       message += "\n" + result[ "Documentation" ]
 
-    return responses.BuildDetailedInfoResponse( message )
+    if not message:
+      raise RuntimeError( 'No documentation available.' )
+    return responses.BuildDetailedInfoResponse( message.strip() )
 
 
   def _DefaultParameters( self, request_data ):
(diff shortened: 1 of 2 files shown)


ycmd/completers/cs/cs_completer.py
```diff
@@ -702,6 +704,8 @@ def _GetDoc( self, request_data ):
     if ( result[ "Documentation" ] ):
       message += "\n" + result[ "Documentation" ]
 
+    if not message:
```
resolritter: @bstaletic I think you should do this before trying to concatenate with `result[ "Documentation" ]` (in the lines above), as you can't assume both should always come out as `None` in this situation - that's an implementation detail if it's the case, not a invariant.

bstaletic: I did consider that. The reason I didn't do it the way you suggest is because, logically, how can there be a documentation for something that has an unknown type? I was also unable to get Roslyn to return non-null `Documentation`, but null `Type`.

To be honest, I mostly didn't want to mess with mocking a response that I can't actually get from the server.

bstaletic: Actually, I've made some changes. Instead of doing `message = result[ 'Type' ]`, now we're doing `message = result.get( 'Type' ) or ''`. This way the concatenation will always be safe and in the off chance that there's no type, but there is documentation, `.strip()` will strip the leading new line.

resolritter: That's nice! Thank you

PR #1387, opened 2020-01-16
[READY] Drop `ToCppStringCompatible()`

bstaletic:
Pybind can handle being passed either `bytes` or `str`, so we don't need to convert before calling a C++ API from python.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1387)


bstaletic:
Sure! :lgtm:

PR #1389, opened 2020-01-16
[READY] A giant test suite rewrite

bstaletic:
Apologies for +-3000 lines diff.  Nose has not been maintained for years now and it's currently broken. Nose2 is out of the question because it [doesn't support our test suite layout](https://nose2.readthedocs.io/en/latest/differences.html). That only leaves raw unittest and pytest as alternatives. Pytest was chosen, mostly because it was easier to get it to work.

Once again my suggestion is to review each commit separately.

- The first commit is just a simple switch from standalone `mock` to `unittest.mock`. We're testing on python 3.6 because python3.5 doesn't have all that we need in its `unittest.mock`. Namely [`assert_called()`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called) is missing.
  - I didn't bump the requirement for python to 3.6 in general, because [it's not dead yet](https://devguide.python.org/#status-of-python-branches). I know @puremourning wishes to drop python 3.5 completely, so I'm open for discussion.
- The second commit is pretty boring and just replaces `nose.tools.eq_` and `nose.tools.ok_` with the equivalent hamcrest utilities.
- 3rd commit is the big one. It drops `nose` and introduces the dependency on `pytest`. The differences include:
  - There's no test yielding functionality. Instead [`@pytest.mark.parametrize`](https://docs.pytest.org/en/latest/parametrize.html#parametrize) is used.
  - `@patch`-introduced arguments are at the front, not the back, so we can't just catch them with `*args` when we don't care.
  - Decorators need to be attached to `foo_test()`, not to functions that used to be yielded.
  - Decorators aren't allowed to mess with arguments (like `@SharedYcmd`), so they needed to be rewritten as `pytest.mark.parametrize` calling a [`pytest.fixture`](https://docs.pytest.org/en/latest/fixture.html).
  - `WithRetry` and `ExpectedFailure` worked fine without any change.
- Finally, pytest complained about two deprecation warnings:
  - `threading.Thread.IsAlive()` is deprecated in 3.8 and removed in 3.9 in favour of `threading.Thread.is_alive()`.
  - Hamcrest deprecated `contains` and introduced `contains_exactly`.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1389)


bstaletic:
So the only thing that's missing right now is "why was `cs/get_completions_test.py` renamed to `cs/1_completions_test.py`?"

I'm not sure and I'll take another look at it tonight, but I don't promise I'll find a satisfying answer. As with the `RefactorRename_Multi` test, there are probably two things at play here:

- A different order of tests compared to nose.
- Roslyn being moody about references it decides to see, affected by order of tests but not completely solvable with `IsolatedYcmd`.

bstaletic:
Flake8 errors fixed.

puremourning:
woooo


`\o/`

puremourning:
now we just have to do YCM too :/

PR #1391, opened 2020-02-01
fix 'index out of range' in DownloadClangd

tungpd:
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1391)

### DIFF
--- a/build.py
+++ b/build.py
@@ -1039,7 +1039,7 @@ def DownloadClangd( printer ):
   CLANGD_OUTPUT_DIR = p.join( CLANGD_DIR, 'output' )
 
   target = GetClangdTarget()
-  target_name, check_sum = target[ not IS_64BIT ]
+  target_name, check_sum = target[ not IS_64BIT if len(target) > 1 else 0 ]
   target_name = target_name.format( version = CLANGD_VERSION )
   file_name = '{}.tar.bz2'.format( target_name )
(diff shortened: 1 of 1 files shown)


bstaletic:
Besides the flake8 error, please provide more details. Which OS/architecture have you encountered this error on? Right now I don't see how you could have ran into the error unless you are on an OS/architecture for which there just isn't a precompiled binary.

tungpd:
> Besides the flake8 error, please provide more details. Which OS/architecture have you encountered this error on? Right now I don't see how you could have ran into the error unless you are on an OS/architecture for which there just isn't a precompiled binary.

build on armv7 32bits processor should emit an 'index out of range' error
tungpd — ssh pi — 275×66 2020-02-01 18-53-08

puremourning:
Thanks for sending a PR. As this is a bug fix, Please could you raise an issue in the tracker covering the problem, steps to repro etc. Then we can link it to this pr

bstaletic:
> build on armv7 32bits processor

So what is `sys.maxsize` on ARMv7?

bstaletic:
Oh, I get it now. I think I'd prefer a dummy value as the first element in the list on line 1044, with a comment. Doing `targe[ not IS_64BIT if len( target ) > 1 else 0 ]` feels like using a chainsaw where a scalpel would do.

PR #1393, opened 2020-02-09
[READY] Update python dependencies, including jedi

### LINKED ISSUE #1359: GoToDefinition/GoToDeclaration behaviour in python
At present after #1343, both `GoToDefinition` and `GoToDeclaration` does `jedi`'s `Script.goto_definitions()`, which jumps, for example, from
```python
@classmethod
def f(cls):
    pass
f|
```
to some `builtins.pyi`. And from
```python
@some_decorator
def f():
    pass
f|
```
to where `some_decorator` is defined.
I understand the second case is useful sometimes, though there are when one doesn't care the decorators (which I believe is even most cases), but I couldn't really think of some practical use case of the first one. 
Notice that `jedi` has a `follow_imports` parameter for `goto_assignments` which jumps to `f` in both cases so acts differently and is really useful.
So could we just switch `GoToDeclaration` back to `goto_assignments(follow_imports=True)`?

bstaletic:
This PR depends on #1389 and currently includes all the changes from that PR. I'll rebase after we merge #1389.

The pull request updates jedi to 0.16.0 and switches to the new API. We're still hitting one `DeprecationWarning` and 4 tests are marked with `xfail` as they seem to be a jedi regression.

For now, I wanted to just submit the pull request so people can try it out.

Fixes #1359 
Pinging users who have complained about the way we use jedi on master:

@yangyangxcf @apporc @vanzi @f-squirrel @ApolloBian

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)

### DIFF
--- a/.vimspector.json
+++ b/.vimspector.json
@@ -1,14 +1,7 @@
 {
   "configurations": {
     "python - launch pytest": {
-      "adapter": {
-        "command": [
-          "python",
-          "-m",
-          "debugpy.adapter"
-        ],
-        "name": "debugpy"
-      },
+      "adapter": "debugpy",
       "variables": [
         {
           "python": {
--- a/.vimspector.json
+++ b/.vimspector.json
@@ -1,14 +1,18 @@
 {
   "configurations": {
-    "python - launch nosetests": {
-      "adapter": "vscode-python",
+    "python - launch pytest": {
+      "adapter": {
+        "command": [
+          "python",
+          "-m",
+          "debugpy.adapter"
+        ],
+        "name": "debugpy"
+      },
       "variables": [
         {
           "python": {
             "shell": "/bin/bash -c 'if [ -z \"${dollar}VIRTUAL_ENV\" ]; then echo $$(which python); else echo \"${dollar}VIRTUAL_ENV/bin/python\"; fi'"
-          },
-          "nosetests": {
-            "shell": "/bin/bash -c 'if [ -z \"${dollar}VIRTUAL_ENV\" ]; then echo $$(which nosetests); else echo \"${dollar}VIRTUAL_ENV/bin/nosetests\"; fi'"
           }
         },
         {
@@ -29,16 +33,15 @@
         "cwd": "${workspaceRoot}",
         "stopOnEntry": true,
         "console": "integratedTerminal",
-        "justMyCode": false,
+        "justMyCode": true,
 
         "debugOptions": [],
 
-        "program": "${nosetests}",
-        "pythonPath": "${python}",
+        "module": "pytest",
+        "python": "${python}",
         "args": [
           "-v",
-          "--with-id",
-          "--ignore-files=(^\\.|^setup\\.py$$)",
+          "--ignore=ycmd/tests/python/testdata",
           "${Test}"
         ],
         "env": {
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -7,6 +7,5 @@ WebTest               >= 2.0.20
 psutil                >= 3.3.0, != 5.0.1
 coverage              >= 4.2
 codecov               >= 2.0.5
-debugpy
 pytest
 pytest-cov
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -7,6 +7,6 @@ WebTest               >= 2.0.20
 psutil                >= 3.3.0, != 5.0.1
 coverage              >= 4.2
 codecov               >= 2.0.5
-ptvsd
+debugpy
 pytest
 pytest-cov
--- a/ycmd/completers/python/python_completer.py
+++ b/ycmd/completers/python/python_completer.py
@@ -156,15 +156,9 @@ class PythonCompleter( Completer ):
   def _GetJediScript( self, request_data ):
     path = request_data[ 'filepath' ]
     source = request_data[ 'file_data' ][ path ][ 'contents' ]
-    line = request_data[ 'line_num' ]
-    # Jedi expects columns to start at 0, not 1, and for them to be Unicode
-    # codepoint offsets.
-    column = request_data[ 'start_codepoint' ] - 1
     environment = self._EnvironmentForRequest( request_data )
     sys_path = self._SysPathForFile( request_data, environment )
     return jedi.Script( source,
-                        line,
-                        column,
                         path,
                         sys_path = sys_path,
                         environment = environment )
@@ -185,7 +179,11 @@ class PythonCompleter( Completer ):
 
   def ComputeCandidatesInner( self, request_data ):
     with self._jedi_lock:
-      completions = self._GetJediScript( request_data ).completions()
+      line = request_data[ 'line_num' ]
+      # Jedi expects columns to start at 0, not 1, and for them to be Unicode
+      # codepoint offsets.
+      column = request_data[ 'start_codepoint' ] - 1
+      completions = self._GetJediScript( request_data ).complete( line, column )
       return [ responses.BuildCompletionData(
         insertion_text = completion.name,
         # We store the Completion object returned by Jedi in the extra_data
@@ -200,7 +198,12 @@ class PythonCompleter( Completer ):
 
   def ComputeSignaturesInner( self, request_data ):
     with self._jedi_lock:
-      signatures = self._GetJediScript( request_data ).call_signatures()
+      line = request_data[ 'line_num' ]
+      # Jedi expects columns to start at 0, not 1, and for them to be Unicode
+      # codepoint offsets.
+      column = request_data[ 'start_codepoint' ] - 1
+      signatures = self._GetJediScript( request_data ).get_signatures( line,
+                                                                       column )
       # Sorting by the number or arguments makes the order stable for the tests
       # and isn't harmful. The order returned by jedi seems to be arbitrary.
       signatures.sort( key=lambda s: len( s.params ) )
@@ -295,7 +298,11 @@ class PythonCompleter( Completer ):
 
   def _GoToDefinition( self, request_data ):
     with self._jedi_lock:
-      definitions = self._GetJediScript( request_data ).goto_definitions()
+      line = request_data[ 'line_num' ]
+      # Jedi expects columns to start at 0, not 1, and for them to be Unicode
+      # codepoint offsets.
+      column = request_data[ 'start_codepoint' ] - 1
+      definitions = self._GetJediScript( request_data ).infer( line, column )
       if definitions:
         return self._BuildGoToResponse( definitions )
     raise RuntimeError( 'Can\'t jump to type definition.' )
@@ -303,7 +310,12 @@ class PythonCompleter( Completer ):
 
   def _GoToReferences( self, request_data ):
     with self._jedi_lock:
-      definitions = self._GetJediScript( request_data ).usages()
+      line = request_data[ 'line_num' ]
+      # Jedi expects columns to start at 0, not 1, and for them to be Unicode
+      # codepoint offsets.
+      column = request_data[ 'start_codepoint' ] - 1
+      definitions = self._GetJediScript( request_data ).get_references( line,
+                                                                        column )
       if definitions:
         return self._BuildGoToResponse( definitions )
     raise RuntimeError( 'Can\'t find references.' )
@@ -325,7 +337,11 @@ class PythonCompleter( Completer ):
 
   def _GetType( self, request_data ):
     with self._jedi_lock:
-      definitions = self._GetJediScript( request_data ).goto_definitions()
+      line = request_data[ 'line_num' ]
+      # Jedi expects columns to start at 0, not 1, and for them to be Unicode
+      # codepoint offsets.
+      column = request_data[ 'start_codepoint' ] - 1
+      definitions = self._GetJediScript( request_data ).infer( line, column )
       type_info = [ self._BuildTypeInfo( definition )
                     for definition in definitions ]
     type_info = ', '.join( type_info )
@@ -336,7 +352,11 @@ class PythonCompleter( Completer ):
 
   def _GetDoc( self, request_data ):
     with self._jedi_lock:
-      definitions = self._GetJediScript( request_data ).goto_definitions()
+      line = request_data[ 'line_num' ]
+      # Jedi expects columns to start at 0, not 1, and for them to be Unicode
+      # codepoint offsets.
+      column = request_data[ 'start_codepoint' ] - 1
+      definitions = self._GetJediScript( request_data ).infer( line, column )
       documentation = [ definition.docstring() for definition in definitions ]
     documentation = '\n---\n'.join( documentation )
     if documentation:
--- a/ycmd/completers/python/python_completer.py
+++ b/ycmd/completers/python/python_completer.py
@@ -159,7 +159,7 @@ class PythonCompleter( Completer ):
     environment = self._EnvironmentForRequest( request_data )
     sys_path = self._SysPathForFile( request_data, environment )
     return jedi.Script( source,
-                        path,
+                        path = path,
                         sys_path = sys_path,
                         environment = environment )
 
@@ -185,7 +185,7 @@ class PythonCompleter( Completer ):
       column = request_data[ 'start_codepoint' ] - 1
       completions = self._GetJediScript( request_data ).complete( line, column )
       return [ responses.BuildCompletionData(
-        insertion_text = completion.name,
+        insertion_text = completion.complete,
         # We store the Completion object returned by Jedi in the extra_data
         # field to detail the candidates once the filtering is done.
         extra_data = completion
@@ -273,6 +273,8 @@ class PythonCompleter( Completer ):
                            self._GoToDefinition( request_data ) ),
       'GoToReferences' : ( lambda self, request_data, args:
                            self._GoToReferences( request_data ) ),
+      'GoToType'       : ( lambda self, request_data, args:
+                           self._GoToType( request_data ) ),
       'GetType'        : ( lambda self, request_data, args:
                            self._GetType( request_data ) ),
       'GetDoc'         : ( lambda self, request_data, args:
@@ -297,11 +299,25 @@ class PythonCompleter( Completer ):
       if definition.column is not None:
         column += definition.column
       filepath = definition.module_path or request_data[ 'filepath' ]
-      gotos.append( responses.BuildGoToResponse( definition.module_path,
+      gotos.append( responses.BuildGoToResponse( filepath,
                                                  definition.line,
                                                  column,
                                                  definition.description ) )
     return gotos
+
+
+  def _GoToType( self, request_data ):
+    with self._jedi_lock:
+      line = request_data[ 'line_num' ]
+      # Jedi expects columns to start at 0, not 1, and for them to be Unicode
+      # codepoint offsets.
+      column = request_data[ 'start_codepoint' ] - 1
+      script = self._GetJediScript( request_data )
+      definitions = script.infer( line, column )
+      if definitions:
+        return self._BuildGoToResponse( definitions, request_data )
+
+    raise RuntimeError( 'Can\'t jump to type definition.' )
 
 
   def _GoToDefinition( self, request_data ):
@@ -315,7 +331,7 @@ class PythonCompleter( Completer ):
       if definitions:
         return self._BuildGoToResponse( definitions, request_data )
 
-    raise RuntimeError( 'Can\'t jump to type definition.' )
+    raise RuntimeError( 'Can\'t jump to definition.' )
 
 
   def _GoToReferences( self, request_data ):
@@ -351,7 +367,7 @@ class PythonCompleter( Completer ):
       # Jedi expects columns to start at 0, not 1, and for them to be Unicode
       # codepoint offsets.
       column = request_data[ 'start_codepoint' ] - 1
-      definitions = self._GetJediScript( request_data ).goto( line, column )
+      definitions = self._GetJediScript( request_data ).infer( line, column )
       type_info = [ self._BuildTypeInfo( definition )
                     for definition in definitions ]
     type_info = ', '.join( type_info )
--- a/ycmd/completers/python/python_completer.py
+++ b/ycmd/completers/python/python_completer.py
@@ -280,18 +280,26 @@ class PythonCompleter( Completer ):
     }
 
 
-  def _BuildGoToResponse( self, definitions ):
+  def _BuildGoToResponse( self, definitions, request_data ):
     if len( definitions ) == 1:
       definition = definitions[ 0 ]
-      return responses.BuildGoToResponse( definition.module_path,
+      column = 1
+      if definition.column is not None:
+        column += definition.column
+      filepath = definition.module_path or request_data[ 'filepath' ]
+      return responses.BuildGoToResponse( filepath,
                                           definition.line,
-                                          definition.column + 1 )
+                                          column )
 
     gotos = []
     for definition in definitions:
+      column = 1
+      if definition.column is not None:
+        column += definition.column
+      filepath = definition.module_path or request_data[ 'filepath' ]
       gotos.append( responses.BuildGoToResponse( definition.module_path,
                                                  definition.line,
-                                                 definition.column + 1,
+                                                 column,
                                                  definition.description ) )
     return gotos
 
@@ -302,9 +310,11 @@ class PythonCompleter( Completer ):
       # Jedi expects columns to start at 0, not 1, and for them to be Unicode
       # codepoint offsets.
       column = request_data[ 'start_codepoint' ] - 1
-      definitions = self._GetJediScript( request_data ).infer( line, column )
+      script = self._GetJediScript( request_data )
+      definitions = script.goto( line, column )
       if definitions:
-        return self._BuildGoToResponse( definitions )
+        return self._BuildGoToResponse( definitions, request_data )
+
     raise RuntimeError( 'Can\'t jump to type definition.' )
 
 
@@ -317,7 +327,7 @@ class PythonCompleter( Completer ):
       definitions = self._GetJediScript( request_data ).get_references( line,
                                                                         column )
       if definitions:
-        return self._BuildGoToResponse( definitions )
+        return self._BuildGoToResponse( definitions, request_data )
     raise RuntimeError( 'Can\'t find references.' )
 
 
@@ -341,7 +351,7 @@ class PythonCompleter( Completer ):
       # Jedi expects columns to start at 0, not 1, and for them to be Unicode
       # codepoint offsets.
       column = request_data[ 'start_codepoint' ] - 1
-      definitions = self._GetJediScript( request_data ).infer( line, column )
+      definitions = self._GetJediScript( request_data ).goto( line, column )
       type_info = [ self._BuildTypeInfo( definition )
                     for definition in definitions ]
     type_info = ', '.join( type_info )
@@ -356,7 +366,7 @@ class PythonCompleter( Completer ):
       # Jedi expects columns to start at 0, not 1, and for them to be Unicode
       # codepoint offsets.
       column = request_data[ 'start_codepoint' ] - 1
-      definitions = self._GetJediScript( request_data ).infer( line, column )
+      definitions = self._GetJediScript( request_data ).goto( line, column )
       documentation = [ definition.docstring() for definition in definitions ]
     documentation = '\n---\n'.join( documentation )
     if documentation:
--- a/ycmd/completers/python/python_completer.py
+++ b/ycmd/completers/python/python_completer.py
@@ -354,9 +354,10 @@ class PythonCompleter( Completer ):
     # from the params field.
     try:
       # Remove the "param " prefix from the description.
+      parameters = definition.get_signatures()[ 0 ].params
       type_info += '(' + ', '.join(
-        [ param.description[ 6: ] for param in definition.params ] ) + ')'
-    except AttributeError:
+        [ param.description[ 6: ] for param in parameters ] ) + ')'
+    except IndexError:
       pass
     return type_info
 
--- a/ycmd/tests/python/get_completions_test.py
+++ b/ycmd/tests/python/get_completions_test.py
@@ -86,24 +86,14 @@ def GetCompletions_Basic_test( app ):
                  CompletionEntryMatcher( 'a',
                                          'self.a = 1',
                                          {
-                                           'extra_data': has_entry(
-                                             'location', has_entries( {
-                                               'line_num': 3,
-                                               'column_num': 10,
-                                               'filepath': filepath
-                                             } )
-                                           )
+                                           'detailed_info': '',
+                                           'kind': 'statement'
                                          } ),
                  CompletionEntryMatcher( 'b',
                                          'self.b = 2',
                                          {
-                                           'extra_data': has_entry(
-                                             'location', has_entries( {
-                                               'line_num': 4,
-                                               'column_num': 10,
-                                               'filepath': filepath
-                                             } )
-                                           )
+                                           'detailed_info': '',
+                                           'kind': 'statement'
                                          } )
                ) )
 
@@ -121,14 +111,10 @@ def GetCompletions_Basic_test( app ):
                    CompletionEntryMatcher( 'a',
                                            'self.a = 1',
                                            {
-                                             'extra_data': has_entry(
-                                               'location', has_entries( {
-                                                 'line_num': 3,
-                                                 'column_num': 10,
-                                                 'filepath': filepath
-                                               } )
-                                             )
-                                           } ) ),
+                                             'detailed_info': '',
+                                             'kind': 'statement'
+                                           }
+                                         ) ),
                  is_not( has_item( CompletionEntryMatcher( 'b' ) ) )
                ) )
 
--- a/ycmd/tests/python/get_completions_test.py
+++ b/ycmd/tests/python/get_completions_test.py
@@ -113,8 +113,7 @@ def GetCompletions_Basic_test( app ):
                                            {
                                              'detailed_info': '',
                                              'kind': 'statement'
-                                           }
-                                         ) ),
+                                           } ) ),
                  is_not( has_item( CompletionEntryMatcher( 'b' ) ) )
                ) )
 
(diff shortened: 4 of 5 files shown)


bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained

___
*[ycmd/completers/python/python_completer.py, line 358 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cKk5-9rRjKPjjg55Z:-M-cKk5-9rRjKPjjg55_:bmr8r9p) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/completers/python/python_completer.py#L358)):*
> ```Python
>       # Remove the "param " prefix from the description.
>       type_info += '(' + ', '.join(
>         [ param.description[ 6: ] for param in definition.params ] ) + ')'
> ```

This line raises a deprecation warning. It's used in two places:

- `_GetType()` which basically does:

```python
definitions = jedi.Script( source, path ).goto( line, column )
type_info = ', '.join( [ _BuildTypeInfo( d ) for d in defs ] )
```

- `DetailCandidates`

```python
completions = jed.Script( source, path ).complete( line, column )
_BuildTypeInfo( completions[ 0 ] )
```

The deprecation warning says that we should be using `get_signatures()[N].params`, but I'm not sure how, considering the second use case.
___
*[ycmd/tests/python/subcommands_test.py, line 139 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNSqH8nAy7lX_IeI2:-M-cNSqIAbuOov7EDjhb:b-hugj5v) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L139)):*
> ```Python
>     pytest.param( { 'request':  ( 'multifile1.py', 2, 47 ),
>                     'response': ( 'multifile2.py', 1, 52 ) },
>                   marks = pytest.mark.xfail( reason = 'jedi regression' ) ),
> ```

These 4 tests seem to be a jedi regression.
___
*[ycmd/tests/python/subcommands_test.py, line 171 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNaoRCyzAZFQ-JWls:-M-cNaoRCyzAZFQ-JWlt:bdh99ib) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L171)):*
> ```Python
>     ( ( 13,  8 ), 'some_class = SomeClass()' ),
>     ( ( 13, 17 ), 'def SomeMethod(first_param, second_param)' ),
>     ( ( 19,  4 ), "variable = 'test', variable = some_function()" )
> ```

`GetType` on a variable doesn't return `instance Foo`, rather `var_name = bar()` which is much less informative.
___
*[ycmd/tests/python/subcommands_test.py, line 265 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cOLwb3uwiOdf2j5E1:-M-cOLwb3uwiOdf2j5E2:b-ak4pp) ([raw file](https://github.com/ycm-core/ycmd/blob/79788c78d64ac7a696e4e06fd48d44fddc4d9489/ycmd/tests/python/subcommands_test.py#L265)):*
> ```Python
>   ] )
> @SharedYcmd
> def Subcommands_GoToType_test( app, test ):
> ```

This is weird. If these tests are placed above the `GetType` tests, jedi gets confused about retrieving type info for member functions.
It seems that this is not the matter of test order, but rather something more obscure with jedi, since I could reproduce the behaviour in vim and isolating the tests didn't help at all.

puremourning (commented):
I'll review fully once the pytest change I emerged so that the diff is a bit easier. We've discussed this and while it's a lot better, the regression for current-directory is a no-go for me. It breaks completions in too many use-cases and asking all our users to add `.ycm_extra_conf.py` specifying the `sys_path` is too much IMO.

In our tests, simple cases are broken, but I tried an absolute minimal case:

* cd /some/dir
* vi `test.py`
* vi `test2.py`

from test2.py, test.py module should be implicitly available for completion and goto, and it is so there must be a subtlety the bug.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393#-:-M-dO0JtA5y_MVLcEoT8:brhd8lk)* status: 0 of 2 LGTMs obtained

bstaletic:
Yeah, I didn't think you'd review this right now. The pytest changes would get in the way too much.

bstaletic:
This has been rebased. The diff is now quite manageable.

Should we ping David Halter?

puremourning (changes requested):
> Reviewed 9 of 118 files at r1, 2 of 2 files at r6, 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[.vimspector.json, line 5 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-dZjcp1PHlJAEiAkQL:-M-dZjcp1PHlJAEiAkQM:b-e8k52s) ([raw file](https://github.com/ycm-core/ycmd/blob/d8be58bf5d78d501414348408f724d7b25b34bba/.vimspector.json#L5)):*
<details><summary><i alt="summary">Quoted 8 lines of code…</i></summary>

> ```
>       "adapter": {
>         "command": [
>           "python",
>           "-m",
>           "debugpy.adapter"
>         ],
>         "name": "debugpy"
>       },
> ```
</details>


As of https://github.com/puremourning/vimspector/pull/115 we can remove this and replace with `"adapter": "debugpy"`
___
*[test_requirements.txt, line 10 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-dZumBDvl8nHchTGAg:-M-dZumBDvl8nHchTGAh:b-l09ren) ([raw file](https://github.com/ycm-core/ycmd/blob/d8be58bf5d78d501414348408f724d7b25b34bba/test_requirements.txt#L10)):*
> ```
> coverage              >= 4.2
> codecov               >= 2.0.5
> debugpy
> ```

as of https://github.com/puremourning/vimspector/pull/115 we don't need a local version of `debugpy`
___
*[ycmd/completers/python/python_completer.py, line 358 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cKk5-9rRjKPjjg55Z:-M-daR3y-2HHXSmBSUDj:bjxxqft) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/completers/python/python_completer.py#L358)):*

This line raises a deprecation warning. It's used in two places:

- `_GetType()` which basically does:

```python
definitions = jedi.Script( source, path ).goto( line, column )
type_info = ', '.join( [ _BuildTypeInfo( d ) for d in defs ] )
```

- `DetailCandidates`

```python
completions = jed.Script( source, path ).complete( line, column )
_BuildTypeInfo( completions[ 0 ] )
```

The deprecation warning says that we should be using `get_signatures()[N].params`, but I'm not sure how, considering the second use case.
</blockquote></details>

what's worse is that `get_signatures` is not documented, so what it returns is a mystery:



I suspect that the `get_signatures` is going to return a list, where there are multiple definitions of a function, e.g. in the `signature_help` tests:

```
if X:
   def Y( a, b, c ):
    pass
else:
  def Y( x `0;
    pass
```

Basically, this looks like overloads.

Maybe just pick `get_signatures()[ 0 ]` if `len( get_signatures() ) > 0` rather than catching the `AttributeError` - we can then raise a issue or something to say that it doens't correctly handle multiple signatures? Maybe for `GetType` we could list all of them ? Shrug.
___
*[ycmd/completers/python/python_completer.py, line 370 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-dbaJfASg29BN27weF:-M-dbaJfASg29BN27weG:b-p75n0k) ([raw file](https://github.com/ycm-core/ycmd/blob/d8be58bf5d78d501414348408f724d7b25b34bba/ycmd/completers/python/python_completer.py#L370)):*
If we switched this to `infer` would it fix this:

> GetType on a variable doesn't return instance Foo, rather var_name = bar() which is much less informative.

?
___
*[ycmd/tests/python/subcommands_test.py, line 139 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNSqH8nAy7lX_IeI2:-M-dbLfG-14M5YfOog4i:b6ggj5f) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L139)):*

These 4 tests seem to be a jedi regression.
</blockquote></details>

Yeah. We'll need to see if we can minimum-repro this. I might have some time to try today.... no promises though ;)
___
*[ycmd/tests/python/subcommands_test.py, line 265 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cOLwb3uwiOdf2j5E1:-M-dbqB-EkQX6fW9BPxT:b-c2l7en) ([raw file](https://github.com/ycm-core/ycmd/blob/79788c78d64ac7a696e4e06fd48d44fddc4d9489/ycmd/tests/python/subcommands_test.py#L265)):*

This is weird. If these tests are placed above the `GetType` tests, jedi gets confused about retrieving type info for member functions.
It seems that this is not the matter of test order, but rather something more obscure with jedi, since I could reproduce the behaviour in vim and isolating the tests didn't help at all.
</blockquote></details>

that's ,... sigh.

puremourning (commented):
I think we should try and make minimal repro cases using pure jedi api and our test directory. our usage of the API is pretty clean, but if we can minmimze it then it stands a chance of being fixed.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393#-:-M-dbyYuCndeOOLJUjWz:b-d336ik)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/tests/python/subcommands_test.py, line 171 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNaoRCyzAZFQ-JWls:-M-dcCH3ABVjnNlFm4Rr:b-srazba) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L171)):*

`GetType` on a variable doesn't return `instance Foo`, rather `var_name = bar()` which is much less informative.
</blockquote></details>

perhaps if we use `infer()` for `get type` ? If the type goes through an import (like `from xyzzy import SOME_VARIABLE`), I'd expect GetType to return the type of `SOME_VARIABLE`, not the module?. Caveat - haven't tried this.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[.vimspector.json, line 5 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-dZjcp1PHlJAEiAkQL:-M-dclW-18aMpyuDlZTL:bxzcak5) ([raw file](https://github.com/ycm-core/ycmd/blob/d8be58bf5d78d501414348408f724d7b25b34bba/.vimspector.json#L5)):*

> ```
>       "adapter": {
>         "command": [
>           "python",
>           "-m",
>           "debugpy.adapter"
>         ],
>         "name": "debugpy"
>       },
> ```

As of https://github.com/puremourning/vimspector/pull/115 we can remove this and replace with `"adapter": "debugpy"`
</blockquote></details>

Just to be clear, because reviewable is messing up the display with your quoted code...

You're saying "replace `"adapter": { ... }` with just `"adapter": "debugpy"`?
___
*[test_requirements.txt, line 10 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-dZumBDvl8nHchTGAg:-M-ddMLcD0-plUvn19l1:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/d8be58bf5d78d501414348408f724d7b25b34bba/test_requirements.txt#L10)):*

as of https://github.com/puremourning/vimspector/pull/115 we don't need a local version of `debugpy`
</blockquote></details>

Done.
___
*[ycmd/completers/python/python_completer.py, line 358 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cKk5-9rRjKPjjg55Z:-M-ddVVa0bCPPrAbsKxG:bq6s0ti) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/completers/python/python_completer.py#L358)):*

what's worse is that `get_signatures` is not documented, so what it returns is a mystery:



I suspect that the `get_signatures` is going to return a list, where there are multiple definitions of a function, e.g. in the `signature_help` tests:

```
if X:
   def Y( a, b, c ):
    pass
else:
  def Y( x `0;
    pass
```

Basically, this looks like overloads.

Maybe just pick `get_signatures()[ 0 ]` if `len( get_signatures() ) > 0` rather than catching the `AttributeError` - we can then raise a issue or something to say that it doens't correctly handle multiple signatures? Maybe for `GetType` we could list all of them ? Shrug.


</blockquote></details>

We currently are using `get_signatures()` for `ComputeSignaturesInner()`.
___
*[ycmd/completers/python/python_completer.py, line 370 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-dbaJfASg29BN27weF:-M-de7kz2STzalJ9MClD:b-xd3vmj) ([raw file](https://github.com/ycm-core/ycmd/blob/d8be58bf5d78d501414348408f724d7b25b34bba/ycmd/completers/python/python_completer.py#L370)):*

If we switched this to `infer` would it fix this:

> GetType on a variable doesn't return instance Foo, rather var_name = bar() which is much less informative.

?
</blockquote></details>

You're right!
___
*[ycmd/tests/python/subcommands_test.py, line 139 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNSqH8nAy7lX_IeI2:-M-dgZJh0M3SdyvHWOGg:b-1n0kda) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L139)):*

Yeah. We'll need to see if we can minimum-repro this. I might have some time to try today.... no promises though ;)
</blockquote></details>

You mentioned that using relative paths fixes the issue, but... Does it?

It does fix the issue, at least for me, if I first `cd ycmd/tests/python/testdata/goto/`.
It does not fix the issue if I stay in the project root.

Isn't the project root the cwd when running the jedi tests?
___
*[ycmd/tests/python/subcommands_test.py, line 171 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNaoRCyzAZFQ-JWls:-M-denHrC7lBnycFDuBI:b-w0joa3) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L171)):*

perhaps if we use `infer()` for `get type` ? If the type goes through an import (like `from xyzzy import SOME_VARIABLE`), I'd expect GetType to return the type of `SOME_VARIABLE`, not the module?. Caveat - haven't tried this.
</blockquote></details>

`infer()` works, as far as our tests go.

Do we want `infer()` for `GetDoc` as well? Or maybe that `help()` thing?
___
*[ycmd/tests/python/subcommands_test.py, line 265 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cOLwb3uwiOdf2j5E1:-M-dgKd-2Egv7b8ZRJ2-:bu3u1lu) ([raw file](https://github.com/ycm-core/ycmd/blob/79788c78d64ac7a696e4e06fd48d44fddc4d9489/ycmd/tests/python/subcommands_test.py#L265)):*

that's ,... sigh.
</blockquote></details>

Do we pretend I didn't notice this and treat it as "discussion resolved"?

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[ycmd/tests/python/subcommands_test.py, line 139 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNSqH8nAy7lX_IeI2:-M-drVZB0bCiwBAIImbv:blsu7m4) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L139)):*

You mentioned that using relative paths fixes the issue, but... Does it?

It does fix the issue, at least for me, if I first `cd ycmd/tests/python/testdata/goto/`.
It does not fix the issue if I stay in the project root.

Isn't the project root the cwd when running the jedi tests?
</blockquote></details>

I think so. yes when I tried manually from the goto dir, the relative import worked, but when running the test, it didn't

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[ycmd/tests/python/subcommands_test.py, line 171 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNaoRCyzAZFQ-JWls:-M-drdZI9Gy6uSO45A1F:b-wsvid) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L171)):*

`infer()` works, as far as our tests go.

Do we want `infer()` for `GetDoc` as well? Or maybe that `help()` thing?
</blockquote></details>

I think `infer` for `GetType` 

for GetDoc, it does seem like `help()` might be the right thing. worth a try anyway.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/tests/python/subcommands_test.py, line 139 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNSqH8nAy7lX_IeI2:-M-dtwsh1nj8LenIK5RE:b-yvjsbz) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L139)):*

I think so. yes when I tried manually from the goto dir, the relative import worked, but when running the test, it didn't
</blockquote></details>

So we see the same behaviour. I'm gonna try to repro manually.
___
*[ycmd/tests/python/subcommands_test.py, line 171 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNaoRCyzAZFQ-JWls:-M-du-a-964psv_3xFL8:b-fx7whf) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L171)):*

I think `infer` for `GetType` 

for GetDoc, it does seem like `help()` might be the right thing. worth a try anyway.
</blockquote></details>

`help()` returns the following on *any* member variable:

```
Attribute references
********************

An attribute reference is a primary followed by a period and a name:

   attributeref ::= primary "." identifier

The primary must evaluate to an object of a type that supports
attribute references, which most objects do.  This object is then
asked to produce the attribute whose name is the identifier.  This
production can be customized by overriding the "__getattr__()" method.
If this attribute is not available, the exception "AttributeError" is
raised.  Otherwise, the type and value of the object produced is
determined by the object.  Multiple evaluations of the same attribute
reference may yield different objects.
```

It really does act like calling builtin `help()`. I'm not sure we want that, but maybe...

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/tests/python/subcommands_test.py, line 139 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNSqH8nAy7lX_IeI2:-M-dxci3DXw3zhfY81K_:b-nl74k7) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L139)):*
Repro'd on the first try.

```
>>> import jedi
>>> child_path = 'child.py'
>>> parent_path = 'parent.py'
>>> with open(child_path) as f: child_source = f.read()
...
>>> with open(parent_path) as f: parent_source = f.read()
...
>>> child_source
'from parent import P\nclass C( P ):\n  def c( self ):\n    self.from_base()\n'
>>> parent_source
'class P:\n  def from_base( self ): pass\n'
>>> environment = jedi.get_default_environment()
>>> sys_path = environment.get_sys_path()
>>> script = jedi.Script(child_source, child_path, sys_path = sys_path, environment = environment)
>>> line = 4
>>> column = 12 - 1
>>> definitions = script.g
script.get_context(       script.get_references(    script.goto(              script.goto_definitions(
script.get_names(         script.get_signatures(    script.goto_assignments(
>>> definitions = script.goto(line, column)
>>> definitions
[]
```

And with a relative import:

```
>>> import jedi
>>> child_path = 'child.py'
>>> parent_path = 'parent.py'
>>> with open(child_path) as f: child_source = f.read()
...
>>> child_source
'from .parent import P\nclass C( P ):\n  def c( self ):\n    self.from_base()\n'
>>> with open(parent_path) as f: parent_source = f.read()
...
>>> parent_source
'class P:\n  def from_base( self ): pass\n'
>>> environment = jedi.get_default_environment()
>>> x.get_sys_path(
KeyboardInterrupt
>>> sys_path = environment.get_sys_path()
>>> script = jedi.Script(child_source, child_path, sys_path = sys_path, environment = environment)
>>> line = 4
>>> column = 12 - 1
>>> definitions = script.goto(line, column)
>>> definitions
[<Definition full_name='ycmd.tests.python.testdata.goto.parent.P.from_base', description='def from_base'>]
```

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/completers/python/python_completer.py, line 370 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-dbaJfASg29BN27weF:-M-eIBUb59ns1XwsmGZY:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/d8be58bf5d78d501414348408f724d7b25b34bba/ycmd/completers/python/python_completer.py#L370)):*

You're right!
</blockquote></details>

Done.
___
*[ycmd/tests/python/subcommands_test.py, line 139 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNSqH8nAy7lX_IeI2:-M-eICpF3jXpmLdxM5wR:bths9hg) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L139)):*

Repro'd on the first try.

```
>>> import jedi
>>> child_path = 'child.py'
>>> parent_path = 'parent.py'
>>> with open(child_path) as f: child_source = f.read()
...
>>> with open(parent_path) as f: parent_source = f.read()
...
>>> child_source
'from parent import P\nclass C( P ):\n  def c( self ):\n    self.from_base()\n'
>>> parent_source
'class P:\n  def from_base( self ): pass\n'
>>> environment = jedi.get_default_environment()
>>> sys_path = environment.get_sys_path()
>>> script = jedi.Script(child_source, child_path, sys_path = sys_path, environment = environment)
>>> line = 4
>>> column = 12 - 1
>>> definitions = script.g
script.get_context(       script.get_references(    script.goto(              script.goto_definitions(
script.get_names(         script.get_signatures(    script.goto_assignments(
>>> definitions = script.goto(line, column)
>>> definitions
[]
```

And with a relative import:

```
>>> import jedi
>>> child_path = 'child.py'
>>> parent_path = 'parent.py'
>>> with open(child_path) as f: child_source = f.read()
...
>>> child_source
'from .parent import P\nclass C( P ):\n  def c( self ):\n    self.from_base()\n'
>>> with open(parent_path) as f: parent_source = f.read()
...
>>> parent_source
'class P:\n  def from_base( self ): pass\n'
>>> environment = jedi.get_default_environment()
>>> x.get_sys_path(
KeyboardInterrupt
>>> sys_path = environment.get_sys_path()
>>> script = jedi.Script(child_source, child_path, sys_path = sys_path, environment = environment)
>>> line = 4
>>> column = 12 - 1
>>> definitions = script.goto(line, column)
>>> definitions
[<Definition full_name='ycmd.tests.python.testdata.goto.parent.P.from_base', description='def from_base'>]
```
</blockquote></details>

Solved.
___
*[ycmd/tests/python/subcommands_test.py, line 265 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cOLwb3uwiOdf2j5E1:-M-eIHNK7jHVmuWXCafF:b-2vh4zq) ([raw file](https://github.com/ycm-core/ycmd/blob/79788c78d64ac7a696e4e06fd48d44fddc4d9489/ycmd/tests/python/subcommands_test.py#L265)):*

Do we pretend I didn't notice this and treat it as "discussion resolved"?
</blockquote></details>

This was caused by the same bug that caused the weird goto behaviour, so... solved.

puremourning (changes requested):
> Reviewed 4 of 4 files at r8.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[.vimspector.json, line 5 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-dZjcp1PHlJAEiAkQL:-M-eIW-b9rjJYgdcK2Sx:b2nign3) ([raw file](https://github.com/ycm-core/ycmd/blob/d8be58bf5d78d501414348408f724d7b25b34bba/.vimspector.json#L5)):*

Just to be clear, because reviewable is messing up the display with your quoted code...

You're saying "replace `"adapter": { ... }` with just `"adapter": "debugpy"`?
</blockquote></details>

yep
___
*[ycmd/completers/python/python_completer.py, line 358 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cKk5-9rRjKPjjg55Z:-M-eIbsDEDewzo_3MDP_:bcwcv8a) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/completers/python/python_completer.py#L358)):*

We currently are using `get_signatures()` for `ComputeSignaturesInner()`.
</blockquote></details>

yeah I realise that, but what I said still applies.
___
*[ycmd/tests/python/subcommands_test.py, line 171 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNaoRCyzAZFQ-JWls:-M-eInrP-nyIH533CniE:b-3iu32m) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L171)):*

`help()` returns the following on *any* member variable:

```
Attribute references
********************

An attribute reference is a primary followed by a period and a name:

   attributeref ::= primary "." identifier

The primary must evaluate to an object of a type that supports
attribute references, which most objects do.  This object is then
asked to produce the attribute whose name is the identifier.  This
production can be customized by overriding the "__getattr__()" method.
If this attribute is not available, the exception "AttributeError" is
raised.  Otherwise, the type and value of the object produced is
determined by the object.  Multiple evaluations of the same attribute
reference may yield different objects.
```

It really does act like calling builtin `help()`. I'm not sure we want that, but maybe...
</blockquote></details>

let's leave it and if we need to change it based on feedback, let's do that.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[.vimspector.json, line 5 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-dZjcp1PHlJAEiAkQL:-M-eK0_x7IltAfCMyCF0:b-75jb7x) ([raw file](https://github.com/ycm-core/ycmd/blob/d8be58bf5d78d501414348408f724d7b25b34bba/.vimspector.json#L5)):*

yep
</blockquote></details>

Done. I'm just not pushing just this change, I'll push after we resolve the other two discussions.
___
*[ycmd/completers/python/python_completer.py, line 358 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cKk5-9rRjKPjjg55Z:-M-eKBJU6_SgaiK1cJvS:blc66mx) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/completers/python/python_completer.py#L358)):*

yeah I realise that, but what I said still applies.
</blockquote></details>

`get_signatures()` *is* documented.

```
 get_signatures(line=None, column=None, *args, **kwargs)[source]

    Return the function object of the call you’re currently in.

    E.g. if the cursor is here:

    abs(# <-- cursor is here

    This would return the abs function. On the other hand:

    abs()# <-- cursor is here

    This would return an empty list..
    Return type:	list of classes.Signature
```

https://jedi.readthedocs.io/en/latest/docs/api.html#jedi.Script.get_signatures

I'm apparently stupid and still don't understand what am I supposed to do to solve this deprecation warning.
___
*[ycmd/tests/python/subcommands_test.py, line 171 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cNaoRCyzAZFQ-JWls:-M-eKXdF65o32Ihkku-r:b-hbjqdq) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/tests/python/subcommands_test.py#L171)):*

let's leave it and if we need to change it based on feedback, let's do that.
</blockquote></details>

Alright. For what it's worth, `help()` is documented here:

https://jedi.readthedocs.io/en/latest/docs/api.html#jedi.Script.help

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/completers/python/python_completer.py, line 358 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cKk5-9rRjKPjjg55Z:-M-eM5UM6crnm3v92xNA:bhy6uib) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/completers/python/python_completer.py#L358)):*
Followup, after playing with `get_signatures()` in the shell:

```
>>> import jedi
>>> src = 'def f():pass\n\nf('
>>> script = jedi.Script(src, path = '')
>>> script.get_signatures(3, 0)
[]
>>> script.get_signatures(3, 1)
[]
>>> script.get_signatures(3, 2)
[<Signature: index=None f()>]
```

It only returns signatures when cursor column is inside parens.

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/completers/python/python_completer.py, line 358 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cKk5-9rRjKPjjg55Z:-M-eMvs4AMYK4xOEBQ1h:bxz1fh0) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/completers/python/python_completer.py#L358)):*

Followup, after playing with `get_signatures()` in the shell:

```
>>> import jedi
>>> src = 'def f():pass\n\nf('
>>> script = jedi.Script(src, path = '')
>>> script.get_signatures(3, 0)
[]
>>> script.get_signatures(3, 1)
[]
>>> script.get_signatures(3, 2)
[<Signature: index=None f()>]
```

It only returns signatures when cursor column is inside parens.
</blockquote></details>

I'm referring to `Declaration.get_signatures` which is what replaces `Declaration.params`, not `Script.get_signatures`

https://jedi.readthedocs.io/en/latest/docs/api-classes.html#jedi.api.classes.BaseDefinition.get_signatures

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/completers/python/python_completer.py, line 358 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1393#-M-cKk5-9rRjKPjjg55Z:-M-eRz_20YDGUvZmMZ8G:buao2q5) ([raw file](https://github.com/ycm-core/ycmd/blob/b69117cf88511ca80b93f214bf905f85f0e8616b/ycmd/completers/python/python_completer.py#L358)):*
To quote myself...

> I am apparently stupid

Your suggestion worked. I decided to just assume there will be a signature and if not, catch `IntexError`.
It feels a little dirty to do it like this, but it was also the tersest solution. Objections are welcom.

puremourning (approved):
:lgtm:

> Reviewed 2 of 2 files at r9.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393#-:-M-eTuq80Cb6FomFqiRt:bnfp4nl)* status: 1 of 2 LGTMs obtained

puremourning (approved):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393)* status: 1 of 2 LGTMs obtained

puremourning (approved):
:lgtm:

> Reviewed 1 of 1 files at r10.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393#-:-M-eWdf55WnMT1V82LNc:bnfp4nl)* status: 1 of 2 LGTMs obtained

bstaletic (commented):
:label: assuming the tests pass.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1393#-:-M-eWZbs-ZZ8okwE1fuS:bdji4al)* status: 1 of 2 LGTMs obtained

bstaletic:
:lgtm:

PR #1394, opened 2020-02-09
[READY] Fix IndexError in build.py script

### LINKED ISSUE #1392: Index out of range error from DownloadClangd when try to build on a system with armv7 32bits processor
When build on a system with **armv7 32bits** processor this line of code should emit an **Index out of range** error due to [1]
My proposed solution in [2]
https://github.com/ycm-core/ycmd/blob/d3378ca3a3103535c14b104cb916dcbcdaf93eeb/build.py#L1042
[1]. https://github.com/ycm-core/ycmd/blob/d3378ca3a3103535c14b104cb916dcbcdaf93eeb/build.py#L1024-L1027
[2]. #1391

bstaletic:
Fixes #1392 
Closes #1391 

I think this would be a cleaner solution than the one proposed in #1391.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1394)

### DIFF
--- a/build.py
+++ b/build.py
@@ -1023,6 +1023,7 @@ def GetClangdTarget():
         'e593f7d036434db023c1c323756d6630bb4a2f868c45d682ba967846061f5fa9' ) ]
   if OnArm():
     return [
+      None, # First list index is for 64bit archives. ARMv7 is 32bit only.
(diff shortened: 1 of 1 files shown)


bstaletic:
:lgtm:

PR #1395, opened 2020-02-09
[READY] Allow users to specify paths to external executables

bstaletic:
Closes https://github.com/ycm-core/ycmd/issues/1289

JDT is intentionally left out for the sake of complexity.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1395)

### DIFF
--- a/ycmd/utils.py
+++ b/ycmd/utils.py
@@ -268,6 +268,18 @@ def FindExecutable( executable ):
   return None
 
 
+def FindExecutableWithFallback( executable_path, fallback ):
+  if executable_path:
+    executable_path = FindExecutable( ExpandVariablesInPath( executable_path ) )
+    if not executable_path:
+      # If the user told us to use a non-existing path, report an error.
+      # Don't attempt to be too clever about the fallback.
+      return None
+    return executable_path
+  else:
+    return fallback
+
+
 def ExecutableName( executable ):
   return executable + ( '.exe' if OnWindows() else '' )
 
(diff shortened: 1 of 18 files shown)


puremourning:
> I realized. I clearly should have tested this better before pushing. Sorry...

haha. no problem at all.

ycmd/utils.py
```diff
@@ -266,6 +266,15 @@ def FindExecutable( executable ):
   return None
 
 
+def FindExecutableWithFallback( executable_path, fallback ):
+  if executable_path:
+    executable_path = FindExecutable( ExpandVariablesInPath( executable_path ) )
+    if not executable_path:
+      return None
```
matteomaser: Shouldn't this return `executable_path` when `FindExecutable` actually returned something?

bstaletic: Ah! This pull request is getting quite embarrassing.
Thanks for catching that. I pushed an update.

I should write a test for this thing, but I'm not sure how without mocking things like `os.path.isfile()` and if I mock everything, I'm not sure how useful the test would be.

@puremourning Thoughts?

@matteomaser Please use Reviewable.

matteomaser:
Except for the missing else issue I commented, I confirm this is a better workaround for https://github.com/ycm-core/YouCompleteMe/issues/3587

Until I delved inside the code to fix that simple issue, I had some trouble understanding what I had to put inside let `g:ycm_rls_binary_path`. I was pleasantly surprised when I found out my PATH was searched and I could just use `g:ycm_rls_binary_path = 'rls'`. Maybe this could be documented in the readme.

Ps. thanks for your work on YCM!

PR #1396, opened 2020-02-11
[READY] Clean up python's sys path

bstaletic:
Removing python-future from the extra conf was missed in the python2
support drop pull request.

Numpydoc is now appended to the sys.path to allow python to find the
system-wide numpydoc instead. This is because we're unlikely to update
numpydoc as the v0.9.0+ depend on sphinx and that would be rather insane
to keep around as a submodule.

This way, for users that have a newer numpydoc and use the new features of it, we allow jedi to use that new numpydoc, but for other users the bundled version still works.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1396)

### DIFF
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -189,13 +189,6 @@ def Settings( **kwargs ):
   return {}
 
 
-def GetStandardLibraryIndexInSysPath( sys_path ):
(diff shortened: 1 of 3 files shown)


bstaletic:
:lgtm:

PR #1401, opened 2020-02-16
[READY]  Stop using regex to parse tags files

bstaletic:
Boost.Regex is definitely more convenient for parsing, but manual parsing has two advantages: performance and no need for Boost.Regex.

Changes in this PR:

- `ReadUtf8File()`  now returns a vector of strings. This was done for convenience. Manually parsing tags and taking different actions depending on whether the delimiter happened to be `\t` or `\n` got hairy pretty quickly.
- Boost.Regex is removed from `cpp/BoostParts`.
- Since I ran `update_boost.py`, the bundled boost is now 1.72.

To measure performance, I used @micbou's benchmark from [here](https://github.com/micbou/ycmd/blob/benchmark-extract-identifiers-from-tags-file/cpp/ycm/benchmarks/IdentifierUtils_bench.cpp). I can add that file to our benchmarks if requested. Benchmark results:

```
PR
====================================================================================================================
IdentifierUtilsFixture/ExtractIdentifiersFromTagsFileBench/1                       13705 ns      13635 ns      51396
IdentifierUtilsFixture/ExtractIdentifiersFromTagsFileBench/16                     106535 ns     106368 ns       6602
IdentifierUtilsFixture/ExtractIdentifiersFromTagsFileBench/256                   1560440 ns    1558896 ns        452
IdentifierUtilsFixture/ExtractIdentifiersFromTagsFileBench/4096                 24717922 ns   24695089 ns         28
IdentifierUtilsFixture/ExtractIdentifiersFromTagsFileBench/65536               395063591 ns  394762222 ns          2
IdentifierUtilsFixture_BigO                                                      6028.22 N    6023.62 N
IdentifierUtilsFixture_RMS                                                             0 %          0 %

Master
====================================================================================================================
IdentifierUtilsFixture/ExtractIdentifiersFromTagsFileBench/1                       25794 ns      25697 ns      27442
IdentifierUtilsFixture/ExtractIdentifiersFromTagsFileBench/16                     128557 ns     128315 ns       5460
IdentifierUtilsFixture/ExtractIdentifiersFromTagsFileBench/256                   1773793 ns    1771079 ns        401
IdentifierUtilsFixture/ExtractIdentifiersFromTagsFileBench/4096                 27627206 ns   27583347 ns         25
IdentifierUtilsFixture/ExtractIdentifiersFromTagsFileBench/65536               442732872 ns  442207172 ns          2
IdentifierUtilsFixture_BigO                                                      6755.53 N    6747.50 N
IdentifierUtilsFixture_RMS                                                             0 %          0 %
```

The performance boost is slightly more than 10%.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1401)


bstaletic:
`string_view` is C++17. Yes, we can improve this code with `string_view`, but not until we move to C++17.
We could grab a `const auto`, but since `lines` is already `const`, so `auto&&` is already `const`-correct.

PR #1403, opened 2020-02-20
Update Omnisharp to 1.34.12

mispencer:
Are not using 1.34.13 because it is missing the http-win-x86 package.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1403)

### DIFF
--- a/build.py
+++ b/build.py
@@ -770,46 +770,44 @@ def GetCsCompleterDataForPlatform():
   ####################################
   DATA = {
     'win32': {
+      'version': 'v1.34.12',
+      'download_url': ( 'https://github.com/OmniSharp/omnisharp-roslyn/release'
+                        's/download/v1.34.12/omnisharp.http-win-x86.zip' ),
       'file_name': 'omnisharp.http-win-x86.zip',
-      'version': 'v1.34.2',
-      'download_url': ( 'https://github.com/OmniSharp/omnisharp-roslyn/relea'
-                        'ses/download/v1.34.2/omnisharp.http-win-x86.zip' ),
-      'check_sum': ( 'd66ee6ce347bba58de06a585bff63e8f42178c8b212883be0700919'
-                     '61c3c63d6' ),
+      'check_sum': ( '3413b635c224043d16d1fd4ccf7eb3b67a6bfe8b7fdf810938ab13b2'
+                     'caab9ecc' ),
     },
     'win64': {
+      'version': 'v1.34.12',
+      'download_url': ( 'https://github.com/OmniSharp/omnisharp-roslyn/release'
+                        's/download/v1.34.12/omnisharp.http-win-x64.zip' ),
       'file_name': 'omnisharp.http-win-x64.zip',
-      'version': 'v1.34.2',
-      'download_url': ( 'https://github.com/OmniSharp/omnisharp-roslyn/relea'
-                        'ses/download/v1.34.2/omnisharp.http-win-x64.zip' ),
-      'check_sum': ( 'ab6bdac04b7225a69de11a0bdf0777facbe7d9895e9b6b4c8ebe8b5'
-                     '4b51412e5' ),
+      'check_sum': ( '3b22b11f97b8d482ae92f1063434d1ec50ccfa326c16440ec2d79b1b'
+                     '505658cc' ),
     },
     'macos': {
+      'version': 'v1.34.12',
+      'download_url': ( 'https://github.com/OmniSharp/omnisharp-roslyn/release'
+                        's/download/v1.34.12/omnisharp.http-osx.tar.gz' ),
       'file_name': 'omnisharp.http-osx.tar.gz',
-      'version': 'v1.34.2',
-      'download_url': ( 'https://github.com/OmniSharp/omnisharp-roslyn/relea'
-                        'ses/download/v1.34.2/omnisharp.http-osx.tar.gz' ),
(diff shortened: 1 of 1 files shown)


bstaletic:
Thanks for the pull request. The tests are broken.

- On Windows two fixit tests fail with the same error.
  - Seems like roslyn started returning more than 1 fixit, so ycmd isn't automatically resolving them.
- On macOS the `Diagnostics_Basic_test` never receives diagnostics.
- On linux the `WaitUntilCsCompleterIsReady()` is raising a `RuntimeError`.

bstaletic:
I can explain the behaviour on linux and mac. When a user opens a file for which a roslyn subserver has not been started, ycmd receives a `FileReadyToParse` event. Since the server has not been started, `_StartServer()` is called and then we return. `/codecheck` request isn't sent until the second `FileReadyToParse`.

bstaletic:
A quick patch for the above:

```diff
diff --git a/ycmd/completers/cs/cs_completer.py b/ycmd/completers/cs/cs_completer.py
index e640853e..afb392e1 100644
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -221,12 +221,11 @@ class CsharpCompleter( Completer ):
     if ( self.user_options[ 'auto_start_csharp_server' ]
          and not solutioncompleter._ServerIsRunning() ):
       solutioncompleter._StartServer()
-      return

     # Bail out if the server is unresponsive. We don't start or restart the
     # server in this case because current one may still be warming up.
-    if not solutioncompleter.ServerIsHealthy():
-      return
+    while not solutioncompleter.ServerIsHealthy():
+      pass

     errors = solutioncompleter.CodeCheck( request_data )

```

bstaletic:
Applying my hack isn't the end of troubles. In some cases the new Roslyn doesn't report errors for a file. 

- Diagnostics:  On linux/macOS diagnostics never arrive for `MaxDiagnostics.cs` `DiagnosticRange.cs`. For `Program.cs` the line and description of the error differs between Windows and others.
- Completions: Everything works, except... our tests are always waiting on diagnostics before running the test. The diagnostics just don't come for `Unicode.cs` and therefore the test errors before it even started.
- Subcommands:
  - Fixits:
    - `FixIt_Range` and `FixIt_Single` are actually broken and there's the same error on both Windows and others.
    - `NoFixitsFound` errors on mac and linux because there are no diagnostics in that file. It would have worked if our tests weren't waiting for diagnostics.
  - `RefactorRename_MultiFile` is just unpredictable. I thought I got it to work stably, but in the new version, on mac and linux, it finds only 7 files for the rename, instead of 8.
  - `GetDoc_Invalid` - same thing as completion and `NoFixitsFound`.
  - `GetType_EmptyMessage` - same thing as `GetDoc_Invalid`.
  - `GoTo_Basic` and `GoToImplementation_Basic` - ditto.

So most of the test failures can be solved with a more robust test setup and two more tests actually need to be fixed. The diagnostics tests, well 3 of them, are actually going to be a problem. Thoughts?

bstaletic:
```diff
diff --git a/ycmd/tests/cs/subcommands_test.py b/ycmd/tests/cs/subcommands_test.py
index 9fbb0c27..e85f6bf4 100644
--- a/ycmd/tests/cs/subcommands_test.py
+++ b/ycmd/tests/cs/subcommands_test.py
@@ -37,6 +37,7 @@ from ycmd.tests.test_utils import ( BuildRequest,
 from ycmd.utils import ReadFile, LOGGER


+@WithRetry
 @SharedYcmd
 def Subcommands_FixIt_NoFixitsFound_test( app ):
   fixit_test = PathToTestFile( 'testy', 'FixItTestCase.cs' )
@@ -113,6 +114,13 @@ def Subcommands_FixIt_Range_test( app ):
       'end': { 'line_num': 4, 'column_num': 27 }
     } } )
     response = app.post_json( '/run_completer_command', request ).json
+    request = BuildRequest( line_num = 4,
+                            column_num = 23,
+                            contents = contents,
+                            filetype = 'cs',
+                            filepath = fixit_test,
+                            fixit = response[ 'fixits' ][ 0 ] )
+    response = app.post_json( '/resolve_fixit', request ).json
     assert_that( response, has_entries( {
       'fixits': contains_exactly( has_entries( {
         'location': LocationMatcher( fixit_test, 4, 23 ),
@@ -139,7 +147,13 @@ def Subcommands_FixIt_Single_test( app ):
                             filetype = 'cs',
                             filepath = fixit_test )
     response = app.post_json( '/run_completer_command', request ).json
-    LOGGER.debug( 'r = %s', response )
+    request = BuildRequest( line_num = 4,
+                            column_num = 23,
+                            contents = contents,
+                            filetype = 'cs',
+                            filepath = fixit_test,
+                            fixit = response[ 'fixits' ][ 0 ] )
+    response = app.post_json( '/resolve_fixit', request ).json
     assert_that( response, has_entries( {
       'fixits': contains_exactly( has_entries( {
         'location': LocationMatcher( fixit_test, 4, 23 ),
```

That patch should fix Winblows. [I tried to fix everything](https://github.com/bstaletic/ycmd/tree/new-roslyn), but somehow [this commit](https://github.com/bstaletic/ycmd/commit/028dadf8fcc583f4ef0031fc6659344da07f5a9c#diff-1cbf298f83fd90033fd0a2f4ebd25add) breaks a ton of seemingly unrelated stuff.

Good luck to whoever tries to make Roslyn work.

puremourning:
Closing this for now as it seems abandoned.

PR #1405, opened 2020-03-04
[READY] Update gopls to 0.3.3

icholy:
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1405)


icholy:
@bstaletic no fixits were being returned for that case, and `Subcommands_FixIt_NullResponse_test` already handles that. So I figured it was useless.

puremourning:
would say Vim exceptions are a big no no. 

Can we report the gopls issue upstream?

bstaletic:
> Can we report the gopls issue upstream?

The minimal repro is:

```go
package main
import ( "fmt" )
```

- Cursor on line 2
- `YcmCompleter FixIt`
- Choose either of the fixits. There's no "resolve" step, since gopls responds with code action literals.

I'm looking at the docs to see if I can figure out how to use `gopls query` to repro without a client.

puremourning:
> I think you guys are taking on too much responsibility here. This is an extremely narrow edge case which no-one will ever hit.

In my 20 years of experience as a software engineer, things that "nobody will ever hit" are hit on a surprisingly regular basis :)

Moreover, in this case it's us (YCM team) to whom users report issues and it's our product which they installed, so it's on us to make a good product even if we rely on (ropey) third party engines.

icholy:
Do I need to rebase?

bstaletic:
No, this is a new problem. I'm completely confused by it. Azure seems unable to reach http://security.ubuntu.com/ubuntu/ even though the link works. This makes installing the dependencies error and linux builds fail...

PR #1406, opened 2020-03-10
[READY] Update CI images

bstaletic:
Use latest macOS and linux images and drop testing on VS2015.

The macOS and Windows changes are made because azure is dropping VS2015 and macos 10.13 images. Linux was just convenient.

There's also a bump to clang 3.9 and that's because Ubuntu 18.04 doesn't ship clang older than that.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1406)


bstaletic:
:lgtm:

PR #1411, opened 2020-03-15
[READY] Use non-reentrant locks everywhere

bstaletic:
We can actually use `threading.Lock` everywhere.

Where possible, we avoid taking and releasing the lock multiple times, by introducing `Foo()` and `FooNoLock()`, where the former takes a lock and calls the latter. The `LanguageServerCompleter` shutdown procedure is complex enough that we can't avoid multiple locks, so two places where this `FooNoLock()` thing isn't possible are:

- `_RestartServer()` in `LanguageServerCompleter`.
- `StartServer()` in `JavaCompleter` when it throws a `LanguageServerConnectionTimeout`.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1411)

### DIFF
--- a/ycmd/completers/completer.py
+++ b/ycmd/completers/completer.py
@@ -475,24 +475,36 @@ class CompletionsCache:
   """Cache of computed completions for a particular request."""
 
   def __init__( self ):
-    self._access_lock = threading.RLock()
+    self._access_lock = threading.Lock()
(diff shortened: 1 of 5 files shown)


bstaletic:
:lgtm_strong:

PR #1412, opened 2020-03-15
[READY] Update psutil to silence github alert

bstaletic:
Also adds `requirements.txt` so that we can get alerts for runtime dependencies too.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1412)


bstaletic:
:lgtm:

PR #1414, opened 2020-03-16
[READY] Improve hover formatting for the GenericLSPCompleter

bstaletic:
Since people seem to be actually using the generic LSP completer, it is high time we fixed the way we format the hover response. It would be nice if we could just forget about the deprecated MarkedString, but even among the 4 officially supported LSP servers, we're exercising every possible hover response. Hints of what I think of `interface Hover` is perhaps visible in one of the tests.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1414)


bstaletic:
The `HoverIsList` test doesn't have a `_test` suffix, so it hasn't been run. I'll fix it in a sec.

bstaletic:
Fixed.

bstaletic:
:lgtm:

PR #1415, opened 2020-03-16
[READY] Update jdtls to 0.52.0

bstaletic:
JDT has mostly been squashing bugs, except...

- In `diagnostics_test.py`, we don't see the ghost `test.java` at the root of the project.
- In `get_completions_test.py`, the `extra_menu_info` is much more detailed, which showed another... peculiar JDT behaviour. Every once in a while, jdt.ls will respond with `Object.equal(Object obj) : boolean` instead of `Object.equal(Object arg0) : boolean` (note the name of the argument). To avoid this problem in two tests that exhibited it, I chose to use `has_items()` instead of `contains_exactly()` and thus avoid matching the `Object` API.
- In `subcommands_test.py`, there's a new code action available and the `Subcommands_Format_Range_Spaces_test` surprisingly changed a bit.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1415)


bstaletic:
:lgtm:

PR #1417, opened 2020-03-17
[READY] Update rust to nightly-2020-03-12

bstaletic:
It's been 6 months since the last update. Diagnostics got a little slower, so two tests needed `@WithRetry`. The third test needed to be isolated in order to work. Otherwise it reports some weird diagnostic, saying that `main()` is missing. Maybe we should take the opportunity to look into the diagnostic issue?

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1417)


bstaletic:
:lgtm:

Issue #1413, opened 2020-03-16
Feature - Use proxy or ask for file's path


PeterlitsZo:
I live in China and by the way the net do not work well when the build.py want to use `requests.get` to some file on the web.

I kown the `requests` can read shell's env variable to get the proxy's addrass and port, but I also hope you please can make build.py to can work with args like `--proxy=sock5://127.0.0.1:1080` or ask for file and let me download by some way else.

I do not kown a lot about git, but this task looks ok for me. If you like, I can help to with a new source file.

bstaletic:
Hi. We've had similar issues reported previously. For example: https://github.com/ycm-core/ycmd/issues/1280

Mind telling us what exactly is causing you troubles?

puremourning:
> I kown the requests can read shell's env variable to get the proxy's addrass and port, but I also hope you please can make build.py to can work with args like --proxy=sock5://127.0.0.1:1080

Why is that better than `env http_proxy=sock5://127.0.0.1:1080 ./install.py` ? It's a lot more work and testing for us, so i would rather not do it unless there's a good reason.

PeterlitsZo:
I do love my country, but to be honest, because of [GFW](https://en.wikipedia.org/wiki/Great_Firewall), it is slowly to download some file your build.py need. I try to use ssr to communicate with proxy to get something I want, such as google's severve or github's file(we can get github but it is so slowly to get file). when I use wsl and I find your build.py need to download a lot of thing which is hard to get in China...... 

I have a way that use env variable `ALL_PROXY` or something like that to make `requests` get file by the proxy, but it will be better that has a thing to get the thing......

I kown that it's easy to add something but hard to remove. But if you still have interset to help me, then I think I can join you to touch the aim.

PeterlitsZo:
> > I kown the requests can read shell's env variable to get the proxy's addrass and port, but I also hope you please can make build.py to can work with args like --proxy=sock5://127.0.0.1:1080
> 
> Why is that better than `env http_proxy=sock5://127.0.0.1:1080 ./install.py` ? It's a lot more work and testing for us, so i would rather not do it unless there's a good reason.

thanks for your advises!

it will be a nice way, I think.
by the way I just give my advice and it is true that there are some ways to make it.
I do response your thinking.

Thanks (again)

p.s. Are you a Chinese? I think I have watched your answer about `debugging on vim` on zhihu, is that you?

puremourning:
> p.s. Are you a Chinese?

[...]

PR #1418, opened 2020-03-19
change build.py to make it suppost http[s]-proxy

PeterlitsZo:
1. It is ok to use --http-proxy=xxxx://xxxxx:xxxxx
                or --https-proxy=xxxx://xxxxx:xxxxx
2. Change the behave of function `DownloadFileTo` to make it
   download under proxy
3. Add a varible named PROXY, which default to be empty dict

---[ for issue #1413 ]---

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1418)

### DIFF
--- a/build.py
+++ b/build.py
@@ -108,6 +108,9 @@ CLANGD_BINARIES_ERROR_MESSAGE = (
   'or use your system Clangd. '
   'See the YCM docs for details on how to use a custom Clangd.' )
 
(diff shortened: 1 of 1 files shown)


puremourning:
Seems abandoned, and `http_proxy` environment vars work, so closing.

Thanks for the PR though!

PR #1422, opened 2020-04-15
[READY] Do not assume that users have setuptools available

bstaletic:
Because some users really don't. Seems like we can't even touch anything remotely related to pip without getting bug reports...

This means that we won't always be able to compile watchdog's fsevents extension module.

This won't affect Windows and Linux users, but on macOS, without fsevents, watchdog will fall back to the slower kqueue API.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1422)

### DIFF
--- a/build.py
+++ b/build.py
@@ -1125,14 +1125,25 @@ def CompileWatchdog( script_args ):
     RemoveDirectoryIfExists( build_dir )
     RemoveDirectoryIfExists( lib_dir )
 
-    CheckCall( [ sys.executable,
-                 'setup.py',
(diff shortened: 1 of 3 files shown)


bstaletic:
The reasons the build is failing:

1. `import setuptools` triggers flake8 "imported bun unused warning".
2. Valgrind is reporting something and since the only thing that changed is the update to `watchdog`, the error has to come from them.

bstaletic:
Alright, I fixed the CI here. :lgm:

PR #1429, opened 2020-04-28
Various typo fixes

DonKult:
Hi,

a bunch of typo fixes mostly found by `codespell` after I stumbled manually over some. The first commit bundles most of them in documentation and code comments, so they shouldn't have an impact.

The second changes a function name, so technically an internal API break.

The third is still documentation, but it looks like a json scheme which might or might not be used to verify instances which would need to be fixed as well (or are finally valid now).

Note that while these change `docs/openapi.yml` I have not regenerated the html files as the README advises as I have no working `npm` setup and don't particular fancy one for this right now.

(You may squash, edit & discard to your hearts content – I do not claim copyright ownership for typo fixes.)

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1429)

### DIFF
--- a/JAVA_SUPPORT.md
+++ b/JAVA_SUPPORT.md
@@ -35,7 +35,7 @@ card][design]. In short:
 
 * 2 classes implement the language server protocol in the
   `language_server_completer.py` module:
- * `LanguageServerConnection` - an abstraction of the comminication with the
+ * `LanguageServerConnection` - an abstraction of the communication with the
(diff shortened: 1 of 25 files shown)


DonKult:
Sure, here you go. Just wanted to highlight that these other two are not technically documentation-only changes.
As before: I have not regenerated the html files.

bstaletic:
@DonKult Thanks for the pull request.

:lgtm:

PR #1430, opened 2020-04-29
[READY] Use `os.path.abspath` in ycmd.responses.Location

bstaletic:
This addresses the ycmd side of https://github.com/ycm-core/YouCompleteMe/issues/3662

My concern is whether or not this is a subtle API break. Is it possible that some other ycmd client, at this point, depends on `Location` holding the "real" path?

Note that we're changing [`realpath`](https://docs.python.org/3/library/os.path.html#os.path.normpath) to [`abspath`](https://docs.python.org/3/library/os.path.html#os.path.abspath). Neither of those are [`normpath`](https://docs.python.org/3/library/os.path.html#os.path.normpath).

I'm not sure if `os.path.realpath()` is supposed to behave like ycm-core/YouCompleteMe#3662 describes, even on Winblows.

@puremourning Please test the complete solution, outlined in the bug report.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1430)

### DIFF
--- a/ycmd/responses.py
+++ b/ycmd/responses.py
@@ -228,7 +228,7 @@ class Location:
     self.line_number_ = line
     self.column_number_ = column
     if filename:
-      self.filename_ = os.path.realpath( filename )
+      self.filename_ = os.path.abspath( filename )
     else:
       # When the filename passed (e.g. by a server) can't be recognized or
       # parsed, we send an empty filename. This at least allows the client to


puremourning:
This slipped down my pile. I did start playing with this in windows a while back. I haven't forgotten about it; sorry for the delay.

The short version is that I'm not sure that this change is needed in the end. There was some other part in the client that needed changing. But I started to write some tests and got distracted.

bstaletic:
> This slipped down my pile. I did start playing with this in windows a while back. I haven't forgotten about it; sorry for the delay.

No worries.

> The short version is that I'm not sure that this change is needed in the end. There was some other part in the client that needed changing. But I started to write some tests and got distracted.

While debugging with the user, the following was the solution we came up with:

> In YouCompleteMe\python\ycm\youcompleteme.py the UpdateWithNewDiagnosticsForFile( self, filepath, diagnostics ) function checks to see if the error found is present in an open buffer using the vimsupport.GetBufferNumberForFilename( filepath ) function. However this function uses the realpath function on the filepath passed, and causes the check to fail even when the buffer is open and the filepath is 'W:\test.cpp'.

> The filepath for the Location class in third_party\ycmd\ycmd\responses.py has a member variable called filename_ which is also set using the same realpath function, and as such it will get the incorrect filepath if 'W:\test.cpp' is passed.

bstaletic:
:lgtm:

puremourning:
Tested this on windows:

turns out the change is only required since python 3.8

https://docs.python.org/3/library/os.path.html#os.path.realpath

> Changed in version 3.8: Symbolic links and junctions are now resolved on Windows.


Confirmed without this change, using python 3.8 we return the underlying real path. with this change we return the mapped path. when combined with the client change, this fixes the issue reported by the user.

:lgtm:

bstaletic:
Rebased.

PR #1434, opened 2020-05-07
[READY] Do not return False from the message poll attempting to start the server

### LINKED ISSUE #1433: Extra confs can make `/message_poll` wrongfully return `False`
```python
def Settings(**kwargs):
    __import__("time").sleep(10)
```

Steps to repro:

1. Open a LSP supported file type.
2. Confirm loading the extra conf.
3. Wait ~10 seconds and confirm that `:YcmDebugInfo` says that the server is running.

Expected: Async diagnostics work.
Actual: Long poll, by now, has returned `False` and thus no new polling is being done. What's weird is that [this line](https://github.com/ycm-core/ycmd/blob/master/ycmd/handlers.py#L308) is the one causing the problem. This is the log, with extra debugging sprinkled in relevant places:

```
2020-05-06 21:36:43,551 - DEBUG - BOOM! No semantic completer exists for filetypes: ['']
2020-05-06 21:36:53,986 - DEBUG - No Settings function defined in /home/bstaletic/work/ycmd/ycmd/tests/rust/testdata/common/.ycm_extra_conf.py
2020-05-06 21:36:53,987 - DEBUG - Using path /home/bstaletic/work/ycmd/ycmd/tests/rust/testdata/common for extra_conf_dir
2020-05-06 21:36:53,987 - INFO - Starting rastCompleter: ['/usr/sbin/rust-analyzer']
```

bstaletic:
Fixes #1433

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1434)

### DIFF
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -849,6 +849,7 @@ class LanguageServerCompleter( Completer ):
 
     self._server_keep_logfiles = user_options[ 'server_keep_logfiles' ]
     self._stderr_file = None
+    self._server_started = False
 
     self._Reset()
 
@@ -878,7 +879,6 @@ class LanguageServerCompleter( Completer ):
     self._project_directory = None
     self._settings = {}
     self._extra_conf_dir = None
-    self._server_started = False
 
 
   def GetCompleterName( self ):
@@ -1524,6 +1524,7 @@ class LanguageServerCompleter( Completer ):
     StartServer. In general, completers don't need to call this as it is called
     automatically in OnFileReadyToParse, but this may be used in completer
     subcommands that require restarting the underlying server."""
+    self._server_started = False
     self._extra_conf_dir = self._GetSettingsFromExtraConf( request_data )
 
     # Only attempt to start the server once. Set this after above call as it may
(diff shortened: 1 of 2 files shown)


theli-ua:
:lgtm: :D

bstaletic:
Not LGTM.

Two tests are broken:

```
=========================== short test summary info ===========================
FAILED ycmd/tests/java/diagnostics_test.py::PollForMessages_ServerNotRunning_test[]
FAILED ycmd/tests/java/diagnostics_test.py::PollForMessages_AbortedWhenServerDies_test[]
===== 2 failed, 1703 passed, 22 skipped, 6 warnings in 912.29s (0:15:12) ======
```

Both return `True` when the test expects `False`.

The `ServerNotRunning` one is expected - before the server is running, we don't want to kill the message poll. I just changed that one to expect `True` locally. Note that the java completer has a ton of setup to do before it gets to the `super().StartServer()`.

The `AbortedWhenServerDies` one is properly broken. After shutting down the server, the `_initialize_event` is not set, but the `_server_started` is also `False`.

Maybe we need a new bool? Something like `self._kill_message_poll`?

bstaletic:
Another note... Are we trying to differentiate a shutdown where the server won't be immediately started again and a restart?

bstaletic:
Fixed it!

The solution was:

1. Don't reset `self._server_started` in `ServerReset()`.
2. Reset  to `False` in `__init__()`.
3. Reset to `False` as soon as we enter `_StartAndInitializeServer()`
3.1. Next step is loading the extra conf, followed by setting `_server_started` to `True`.

PR #1436, opened 2020-05-15
[READY] Upgrade gopls to version 0.4.1

icholy:
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1436)

### DIFF
--- a/azure/windows/install_dependencies.bat
+++ b/azure/windows/install_dependencies.bat
@@ -13,5 +13,5 @@ C:\Python\python -c "with open('C:\Python\Lib\site-packages\sitecustomize.py', '
 :: Go configuration
 ::
 
-curl https://dl.google.com/go/go1.12.4.windows-amd64.msi -o C:\go-installer.msi
+curl https://dl.google.com/go/go1.14.3.windows-amd64.msi -o C:\go-installer.msi
 msiexec /i C:\go-installer.msi TARGETDIR=C:\Go /qn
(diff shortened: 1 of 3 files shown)


icholy:
The latest version of gopls uses [errors.As](https://golang.org/pkg/errors/#As) which was introduced in Go 1.13.

puremourning:
So this update would increase our minimum go requirement? Hmmm. I'm not sure if I love that.

How big is the update ? What are the killer features for this release ?

puremourning:
To answer the question I think we currently use the version of go installed on the azure VM: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops

We can change the linux/macOS `install_dependencies.sh` scripts to install a newer go if we need to (for macOS, `brew install goling`, for linux maybe we can also use linuxbrew - we use that for a couple of other things).

puremourning:
Actually it looks like there are multiple go versions installed 👍 

Go 1.11 (go version go1.11.13 linux/amd64)
Go 1.12 (go version go1.12.17 linux/amd64)
Go 1.13 (go version go1.13.10 linux/amd64)
Go 1.14 (go version go1.14.2 linux/amd64)

Looks like we can [pick a version](https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/go?view=azure-devops&tabs=go-current) like this:

```
steps: 
- task: GoTool@0
  inputs:
    version: '1.13.5'
```

bstaletic:
@icholy Thanks for the pull request.

:lgtm:

PR #1438, opened 2020-05-22
Add the __syncthreads() intrinsic to __clang_cuda_device_functions.h

JesseLu:
YCM currently does not recognize the `__syncthreads()` intrinsic that is used by CUDA. It seems like a small oversight since the `__syncthreads_{and|count|or}()` intrinsics are recognized by YCM.

With this PR, YCM will no longer complain about `__syncthreads()` in CUDA code with `__syncthreads() undeclared identifier`.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1438)

### DIFF
--- a/third_party/clang/lib/clang/10.0.0/include/__clang_cuda_device_functions.h
+++ b/third_party/clang/lib/clang/10.0.0/include/__clang_cuda_device_functions.h
@@ -539,6 +539,7 @@ __DEVICE__ void __sincosf(float __a, float *__s, float *__c) {
   return __nv_fast_sincosf(__a, __s, __c);
 }
 __DEVICE__ float __sinf(float __a) { return __nv_fast_sinf(__a); }
+__DEVICE__ int __syncthreads(void) { return __nvvm_bar0(); }  
 __DEVICE__ int __syncthreads_and(int __a) { return __nvvm_bar0_and(__a); }
 __DEVICE__ int __syncthreads_count(int __a) { return __nvvm_bar0_popc(__a); }
 __DEVICE__ int __syncthreads_or(int __a) { return __nvvm_bar0_or(__a); }


bstaletic:
Thanks for the pull request. However, neither me nor @puremourning know much about CUDA.

It does seem reasonable, BUT these headers are taken straight from the LLVM archives. They will be overwritten when LLVM11 comes out.

@kadircet Any thoughts?

puremourning:
Thanks for making a PR but this needs to be fixed or changed upstream in clang. These headers are taken directly from the clang release and were not in the business of customising them.

kadircet:
As Boris and Ben pointed out these should be fixed in clang. I am not familiar with how these intrinsics are maintained, but looks like there were some missing functions that have been added afterwards.

You can send out a patch upstream, i suppose something similar to https://reviews.llvm.org/D43602 should do. For more details on sending a patch to LLVM see https://llvm.org/docs/Contributing.html

bstaletic:
I created https://reviews.llvm.org/D80464.

JesseLu:
Thank you!

Artem-B:
`__syncthreads` is a built-in and does not belong to any headers. If YCM does not see it, it is probably is not compiling the CUDA sources correctly. Can you post complete command line of the options passed to the clang driver and the options of the resulting `cc1` invocation?

puremourning:
Closing as correctly specifying compiler intrinsics is not a YCM issue, but an upstream library issue.

The headers included are taken directly from the upstream llvm+clang release so any changes need to be made there.

@JesseLu if you could provide the clangd log with debug logging enabled for @Artem-B that might help, but for the record, it's clangd that's invoking the compiler, and the clang driver is not invoked at any point.

Artem-B:
@puremourning clangd creates internal instance of the driver which in turn creates compilation jobs, command line and all. While the clang binary is indeed not executed, the code that does the job in clangd is largely the same one used by clang. 

I think you can get clangd to print relevant info with `clangd --log=verbose` and, maybe, also manually changing command line used in the compilation database and adding `-v` to the compiler arguments.

bstaletic:
> `clangd --log=verbose`

This is controlled with `let g:ycm_log_level='debug'`

The `-v` in the driver arguments needs an update to the extra conf or the compilation database.

kadircet:
verbose log should be enough for clangd to log cc1 invocation commands.
Thanks all for taking a look at this.
also feel free to continue the discussion in github.com/clangd/clangd as
this is either a clang or clangd issue.

PR #1447, opened 2020-06-17
[READY] Fix CI and OmniSharp - Don't use standalone mode for OmniSharp

puremourning:
It seems that something about our test setup just broken and the `setup_module` calls are not happening, or at least, not always happening.

These unit-style setup functions are pretty much deprecated and pytest docs say "use fixtures".

So we convert our setup_module and teardown_module to pytest fixtures that are "autouse=True" which is essentially the equivalent of `setup_module`.

The `teardown_module` part is done using a `yield` from the fixture function.

WIP: to see if the tests pass

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1447)


bstaletic:
Instead of `-Werror`, try to export `PYTHONWARNINGS=error`.

```
-W argument
       Warning control.  Python sometimes prints warning message to sys.stderr.  A typical warning message has the following form: file:line:  cate‐
       gory:  message.   By default, each warning is printed once for each source line where it occurs.  This option controls how often warnings are
       printed.  Multiple -W options may be given; when a warning matches more than one option, the action for the  last  matching  option  is  per‐
       formed.   Invalid -W options are ignored (a warning message is printed about invalid options when the first warning is issued).  Warnings can
       also be controlled from within a Python program using the warnings module.

       The simplest form of argument is one of the following action strings (or a unique abbreviation): ignore to ignore all  warnings;  default  to
       explicitly  request  the  default behavior (printing each warning once per source line); all to print a warning each time it occurs (this may
       generate many messages if a warning is triggered repeatedly for the same source line, such as inside a loop); module to  print  each  warning
       only  the  first time it occurs in each module; once to print each warning only the first time it occurs in the program; or error to raise an
       exception instead of printing a warning message.

       The full form of argument is action:message:category:module:line.  Here, action is as explained above but only applies to messages that match
       the remaining fields.  Empty fields match all values; trailing empty fields may be omitted.  The message field matches the start of the warn‐
       ing message printed; this match is case-insensitive.  The category field matches the warning category.  This must be a class name; the  match
       test  whether the actual warning category of the message is a subclass of the specified warning category.  The full class name must be given.
       The module field matches the (fully-qualified) module name; this match is case-sensitive.  The line field matches the line number, where zero
       matches all line numbers and is thus equivalent to an omitted line number.

```

puremourning:
> Instead of `-Werror`, try to export `PYTHONWARNINGS=error`.
> 
> ```
> -W argument
>        Warning control.  Python sometimes prints warning message to sys.stderr.  A typical warning message has the following form: file:line:  cate‐
>        gory:  message.   By default, each warning is printed once for each source line where it occurs.  This option controls how often warnings are
>        printed.  Multiple -W options may be given; when a warning matches more than one option, the action for the  last  matching  option  is  per‐
>        formed.   Invalid -W options are ignored (a warning message is printed about invalid options when the first warning is issued).  Warnings can
>        also be controlled from within a Python program using the warnings module.
> 
>        The simplest form of argument is one of the following action strings (or a unique abbreviation): ignore to ignore all  warnings;  default  to
>        explicitly  request  the  default behavior (printing each warning once per source line); all to print a warning each time it occurs (this may
>        generate many messages if a warning is triggered repeatedly for the same source line, such as inside a loop); module to  print  each  warning
>        only  the  first time it occurs in each module; once to print each warning only the first time it occurs in the program; or error to raise an
>        exception instead of printing a warning message.
> 
>        The full form of argument is action:message:category:module:line.  Here, action is as explained above but only applies to messages that match
>        the remaining fields.  Empty fields match all values; trailing empty fields may be omitted.  The message field matches the start of the warn‐
>        ing message printed; this match is case-insensitive.  The category field matches the warning category.  This must be a class name; the  match
>        test  whether the actual warning category of the message is a subclass of the specified warning category.  The full class name must be given.
>        The module field matches the (fully-qualified) module name; this match is case-sensitive.  The line field matches the line number, where zero
>        matches all line numbers and is thus equivalent to an omitted line number.
> ```

Yeah, that, or I could pass the argument in the right place. I was passing it for the Valgrind run.....

PR #1454, opened 2020-07-08
[READY] Migrate from rls to rust-analyzer

bstaletic:
We can finally migrate to rust-analyzer. It works pretty good, as far as I could tell.

Two questions though:

1. If we're letting users override `rustc` that RA should be using, should we allow them to specify `cargo` too?
2. Should we rename `rls_binary_path` to `ra_binary_path`?

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1454)


puremourning:
> Should we rename rls_binary_path to ra_binary_path?

Yes. We could even issue a warning if we detect a `rls_binary_path` but no `ra_binary_path` ?

> If we're letting users override rustc that RA should be using, should we allow them to specify cargo too?

I dno't have a good enough feel for this, but I would suggest no, until someone asks for it.

PR #1456, opened 2020-07-22
Fix issue with readonly go modules

Felixoid:
Hello.

As mentioned in issue https://github.com/ycm-core/YouCompleteMe/issues/3721, `--go-completer` causes the state, when the YCM directory couldn't be deleted.

This is the intended behavior, see https://github.com/golang/go/issues/27161. But in https://github.com/golang/go/issues/31481 the flag `-modcacherw` was proposed, and in https://golang.org/cl/202079 it was implemented.

To keep backward compatibility, the `GOFLAGS` ENV is used.

This fix https://github.com/ycm-core/YouCompleteMe/issues/3721

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1456)

### DIFF
--- a/build.py
+++ b/build.py
@@ -844,6 +844,7 @@ def EnableGoCompleter( args ):
   new_env[ 'GO111MODULE' ] = 'on'
   new_env[ 'GOPATH' ] = p.join( DIR_OF_THIS_SCRIPT, 'third_party', 'go' )
   new_env[ 'GOBIN' ] = p.join( new_env[ 'GOPATH' ], 'bin' )
+  new_env[ 'GOFLAGS' ] = new_env.get( 'GOFLAGS', '' ) + ' -modcacherw'
   CheckCall( [ go, 'get', 'golang.org/x/tools/gopls@v0.4.2' ],
              env = new_env,
              quiet = args.quiet,


Felixoid:
Hey @bstaletic.

I've decided to use `GOFLAGS` because of its description:

> A space-separated list of -flag=value settings to apply to go commands by default, when the given flag is known by the current command. 

Unfortunately, even when GOFLAGS is used, go elder than 1.14 returns an error:

```
docker run -it golang:1.12
root@fd1127a7af78:/go# GO111MODULE=on GOFLAGS="$GOFLAGS -modcacherw" go get golang.org/x/tools/gopls@v0.4.2
go: parsing $GOFLAGS: unknown flag -modcacherw
```

So, the usage of `GOFLAGS` here is equal to your proposal `[ go, 'get', '-modcacherw', 'golang.org/x/tools/gopls@v0.4.2' ]`

But now I have a question. How to implement version dependent logic for this part? The flag `-modcacherw` was implemented only in 1.14, and I don't feel good about breaking older versions.

puremourning:
> The flag -modcacherw was implemented only in 1.14, and I don't feel good about breaking older versions.

Nor me. I don’t think we should do this then. Faffing with go versions is no it worth this change. 

Maybe just document the solution (setenv GOFLAGS ...) in the issue for anyone else hitting it.

Felixoid:
Maybe it worths updating YCM README for go support as well? And `--go-completer` help string?

I like the idea of leaving this optional, and with `GOFLAGS` it's possible.

puremourning:
> Maybe it worths updating YCM README for go support as well? And --go-completer help string?

Maybe put something on the faq (wiki) pointing to the GH issue and our suggested resolution

bstaletic:
I mean, it should be possible to read the output of `go version`, parse it with a regex and compare it with `1.14`. I wouldn't mind the extra complexity, but @puremourning seems to think that's too much.

puremourning:
Yeah it seems a slippery slope to me. The underlying problem is trivial to fix with chmod so I’m not going to lose sleep over it.  But look, if the code is simple and tested then I probably won’t say no ...

Felixoid:
I think the logic around getting a version and so on is the unnecessary complexity increase. Since the local fix for vim-plug is trivial:

```diff
-Plug 'ycm-core/YouCompleteMe', { 'do': 'git submodule update --init --recursive; ./install.py --clangd-completer --go-completer' }
+Plug 'ycm-core/YouCompleteMe', { 'do': 'git submodule update --init --recursive && GOFLAGS=-modcacherw ./install.py --clangd-completer --go-completer' }
```

I'll add the note to https://github.com/ycm-core/YouCompleteMe/wiki/FAQ about the way to fix it. This PR could be reopened in some time, when versions `<1.14` will be old enough.

puremourning:
Thanks for your contribution 🙏

Felixoid:
Here's a note in FAQ https://github.com/ycm-core/YouCompleteMe/wiki/FAQ#go-completion-read-only-files-and-directories-in-third_partyycmdthird_partygopkg

PR #1459, opened 2020-08-03
[READY] Use full path to check java version

puremourning:
This fixes the check when running `install.py` in YCM, or when running
from a different dir.

Before this change you _always_ received the warning when running install.py from YCM's directory. Now it works properly, using the absoute path to the check script.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1459)

### DIFF
--- a/build.py
+++ b/build.py
@@ -967,8 +967,9 @@ def CheckJavaVersion( required_version ):
   java_version = None
   try:
     java_version = int(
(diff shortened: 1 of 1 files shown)


bstaletic:
For the record, this PR was hit with the `GoToSymbol` flake.

:lgtm_strong:

PR #1462, opened 2020-08-05
[READY] Add javascriptreact support

bstaletic:
Looks like this weird thing can actually work. WIP because I haven't written any tests.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1462)


bstaletic:
Added one test for completion and one for GoToReferences. Both are, I believe, javascriptreact specific.

bstaletic:
Java flaaaakes...

PR #1464, opened 2020-08-07
[READY] Support signature parameter offsets in LSP signatureHelp responses

bstaletic:
According to my testing, this works. Latest rust-analyzer exercises its freedom to use offsets, instead of strings and it also omits `activeSignature` when there's only one.

No tests, because [the missing `rustfmt`](https://rust-lang.github.io/rustup-components-history/) is preventing us from upgrading to the newer nightly.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1464)


bstaletic:
Actually, clangd seems to be using offsets. :lgtm:

PR #1468, opened 2020-08-19
[READY] Selectively implement workspace/configuration for LSP completers

bstaletic:
*Selectively* because this is server-specific beyond any reason. Or a chance of repair. Thank you MS/LSP.

This is implementing a few things:

- An option to specify server-specific capabilities, to opt-into `workspace/configuration`.
- `workspace/configuration`  for rust-analyzer, gopls and generic LSP completers.
  - The implementation for the generic LSP completers has a horribly ugly API. Suggestions are very welcome, including "just skip `GenericLSPCompleter`.
- Moves the capability specific to rust-analyzer to rust_completer.py

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468)


bstaletic:
Turns out rust-analyzer has a new problem.

https://github.com/rust-analyzer/rust-analyzer/issues/5829

This shouldn't really affect users, but it was a problem for our tests. We *could* add a `Cargo.toml` to `ycmd/tests` instead of parsing a `showMessage`.

puremourning (changes requested):
> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[build.py, line 973 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MFCbwrv8qVJXExPrTrG:-MFCbwrw5kyO8sZROX2m:b-gkhxot) ([raw file](https://github.com/ycm-core/ycmd/blob/a982ae8914898acca466cc231ced56ce4ca50eea/build.py#L973)):*
> ```Python
>       new_env.pop( 'JAVA_TOOL_OPTIONS' )
>     except KeyError:
>       print( new_env )
> ```

not sure we need this. it's going to print the env for _most_ users ? this should probably be `pass`
___
*[ycmd/completers/go/go_completer.py, line 112 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MFCcECsBsBd3fGd8Bfo:-MFCcECsBsBd3fGd8Bfp:b-oav3n9) ([raw file](https://github.com/ycm-core/ycmd/blob/a982ae8914898acca466cc231ced56ce4ca50eea/ycmd/completers/go/go_completer.py#L112)):*
> ```Python
>   def WorkspaceConfigurationResponse( self, request ):
>     return [ self._settings.get( 'ls', {} )
>              for i in request[ 'params' ][ 'items' ] ]
> ```

probably worth a quick comment explaining that we return the whole dict for every requested "section" because we don't have any concept of "section" and it isn't actually defined by the protocol.
___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 83 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MFCcTyG9llb_OFswhVq:-MFCcTyG9llb_OFswhVr:b12u5iy) ([raw file](https://github.com/ycm-core/ycmd/blob/a982ae8914898acca466cc231ced56ce4ca50eea/ycmd/completers/language_server/generic_lsp_completer.py#L83)):*
> ```Python
> 
>   def ExtraCapabilities( self ):
>     return self._settings.get( 'ls_capabilities', {} )
> ```

in my fork I had a similar dilema (about snippets). I initially [implemented it as a extra conf](https://github.com/puremourning/ycmd-1/commit/db52a8a64b5be332907e84a5dda1321c560f0f35#diff-7f58211aa2e5ab22db96c1c3e3e20ed7R1959) entry that allowed you to specify capabilities in the extra conf dict, but this turned out to be a right pain, because you actually want it to be based on the _definition_ of the generic completer, _not_ the per-project settings.

In the end, I [added something to the g:ycm_language_server](https://github.com/puremourning/ycmd-1/commit/069320c989eb6b72c44e42dba57af91515fbfa06) and this was much better so that it _jsut worked_ for the JSON/Yaml completer and I didn't have to add extra conf files everywhere.

So:

* If we're going to use a extra conf entry, can/should we use the one in https://github.com/puremourning/ycmd-1/commit/db52a8a64b5be332907e84a5dda1321c560f0f35
* In any case, I think it's worth at least considering for the generic LSP completer to use something similar to https://github.com/puremourning/ycmd-1/commit/069320c989eb6b72c44e42dba57af91515fbfa06

---

I see that the approach here is actually to allow the extra conf to provide a _callable_ which returns the settings. that was a bit of a surprise to me and may be a bit sophisticated for most users. i would be tempted to require a dict mapping the requested "section" to the config ?
___
*[ycmd/completers/language_server/language_server_completer.py, line 601 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MFCeY_s-TqjzUGNEA0d:-MFCeY_s-TqjzUGNEA0e:b-uijiyg) ([raw file](https://github.com/ycm-core/ycmd/blob/a982ae8914898acca466cc231ced56ce4ca50eea/ycmd/completers/language_server/language_server_completer.py#L601)):*
> ```Python
>         self.SendResponse( lsp.Reject( request, lsp.Errors.MethodNotFound ) )
>     elif method == 'client/registerCapability':
>       self._TriggerWatchdog( request )
> ```

I think this method should be called `_HandleRegisterCapability` or something. it just happens that right now it sets up watchdog for the one registration we currently support.
___
*[ycmd/completers/language_server/language_server_protocol.py, line 328 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MFCewPI-NbuH6-3mcOs:-MFCewPI-NbuH6-3mcOt:blbb0e0) ([raw file](https://github.com/ycm-core/ycmd/blob/a982ae8914898acca466cc231ced56ce4ca50eea/ycmd/completers/language_server/language_server_protocol.py#L328)):*
> ```Python
>     },
>   }
>   capabilities.update( extra_capabilities )
> ```

IIRC the `update` here is not 'recursive' in the sense that if a completer uses "ExtraCapabiities" to try to set 'textDocument:  completion: completionItem:  snippetSuppor`, then it will replace the entire `textDocument` entry with just that?

In my implementation I added this thing: https://github.com/puremourning/ycmd-1/commit/db52a8a64b5be332907e84a5dda1321c560f0f35#diff-c66ed66c39b4b5e32c5d069b525e898dR541-R581 which does the "overlay" approach. is that more like what we want?
___
*[ycmd/tests/rust/signature_help_test.py, line 110 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MFCfzMnEf_RGfUPcWyQ:-MFCfzMnEf_RGfUPcWyR:bb7dwwg) ([raw file](https://github.com/ycm-core/ycmd/blob/a982ae8914898acca466cc231ced56ce4ca50eea/ycmd/tests/rust/signature_help_test.py#L110)):*
> ```Python
>           'activeParameter': 0,
>           'signatures': contains_exactly(
>             SignatureMatcher( 'fn build_rocket(&self)', [] )
> ```

bug?

puremourning (changes requested):
> Reviewed 1 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 83 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MFCcTyG9llb_OFswhVq:-MFpYhzN77krNodKZnkN:bsx4i2m) ([raw file](https://github.com/ycm-core/ycmd/blob/a982ae8914898acca466cc231ced56ce4ca50eea/ycmd/completers/language_server/generic_lsp_completer.py#L83)):*

> I added something to the g:ycm_language_server and this was much better so that it jsut worked for the JSON/Yaml completer and I didn't have to add extra conf files everywhere.

I think I like this better than the extra conf route, for the exact reason you mention.

> I see that the approach here is actually to allow the extra conf to provide a callable which returns the settings. that was a bit of a surprise to me and may be a bit sophisticated for most users. i would be tempted to require a dict mapping the requested "section" to the config ?

That doesn't sound like a bad idea. Though you've seen how weird the configuration things are. Do you think a plain old dictionary is flexible enough?
</blockquote></details>

I mean, we have 3 options(;

- always use a dict (this would be my inclination, as you can always write code to set up the dict)
- always use a callable ( think this is probably the worse because it's a bit too much 'code' )
- do both: `if hasattr( thing, '__call__' )` call it else it's a dict

The last one is a compromise, but I wonder if we really need it. You'd be the best judge because you originally thought a callable was necessary/useful so probably thought of some use cases. That said, none of the above preclude doing any of the others in future.
___
*[ycmd/tests/rust/signature_help_test.py, line 110 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MFCfzMnEf_RGfUPcWyQ:-MFpZaOf8Nswg3fl6KYR:b-p51031) ([raw file](https://github.com/ycm-core/ycmd/blob/a982ae8914898acca466cc231ced56ce4ca50eea/ycmd/tests/rust/signature_help_test.py#L110)):*

I'm not sure. Both of these are valid:

```rust
self.build_rocket(); // 1
Builder::build_rocket(self); // 2
```

We're using the first syntax in our tests, in which case there really are no arguments. I'd argue that the correct signature help is `fn build_rocket()` for syntax 1 and `fn build_rocket(&self)` for syntax 2.
</blockquote></details>

You're right, this is a fix not a regression.

puremourning (approved):
> Reviewed 4 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 83 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MFCcTyG9llb_OFswhVq:-MGhvhvpClXd7yRW7Hsb:byfg1vh) ([raw file](https://github.com/ycm-core/ycmd/blob/a982ae8914898acca466cc231ced56ce4ca50eea/ycmd/completers/language_server/generic_lsp_completer.py#L83)):*

I mean, we have 3 options(;

- always use a dict (this would be my inclination, as you can always write code to set up the dict)
- always use a callable ( think this is probably the worse because it's a bit too much 'code' )
- do both: `if hasattr( thing, '__call__' )` call it else it's a dict

The last one is a compromise, but I wonder if we really need it. You'd be the best judge because you originally thought a callable was necessary/useful so probably thought of some use cases. That said, none of the above preclude doing any of the others in future.
</blockquote></details>

Turns out I couldn't convince myself that the callable is needed, so I changed this to a dictionary.

bstaletic:
Rebased after #1472.

puremourning (changes requested):
> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 89 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGocEOC3qLU5d2b32Dl:-MGocEOC3qLU5d2b32Dm:b-rhnh4m) ([raw file](https://github.com/ycm-core/ycmd/blob/597d8d8ddc9430ded7d3d690817d482c3f0c363b/ycmd/completers/language_server/generic_lsp_completer.py#L89)):*
> ```Python
>     if self.ExtraCapabilities().get( 'workspace', {} ).get( 'configuration' ):
>       sections_to_config_map = self._settings.get( 'ls_config', {} )
>       return [ sections_to_config_map[ i ]
> ```

should we catch the possible KeyError here and ... return some default ? or reject the request ?

what happens now if there's a KeyError here? We most certainly don't want to kill the LS connection.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 89 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGocEOC3qLU5d2b32Dl:-MGopMkx3gIHTc_QDNYO:bffjh0q) ([raw file](https://github.com/ycm-core/ycmd/blob/597d8d8ddc9430ded7d3d690817d482c3f0c363b/ycmd/completers/language_server/generic_lsp_completer.py#L89)):*

should we catch the possible KeyError here and ... return some default ? or reject the request ?

what happens now if there's a KeyError here? We most certainly don't want to kill the LS connection. 
</blockquote></details>

Good catch! Changed to `.get( i, None )`, which should be a perfectly valid response. My initial thought was "it's user's responsibility not to miss a key".

puremourning (approved):
:lgtm:

> Reviewed 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468#-:-MGoqGnPBhgQ_LNbe5bR:bnfp4nl)* status: 1 of 2 LGTMs obtained

bstaletic (commented):
Don't merge yet. I want to take a look at writing the tests for this, since currently there are none.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468#-:-MGorOWe004-X562KePc:b-6r4x7s)* status: 1 of 2 LGTMs obtained

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 89 at r5](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGp2DgW6ClnbMlFLR0e:-MGp2DgW6ClnbMlFLR0f:b-sjzaww) ([raw file](https://github.com/ycm-core/ycmd/blob/dd4e3e9910dcd08247883cd511d02d5dddf6eed7/ycmd/completers/language_server/generic_lsp_completer.py#L89)):*
> ```Python
>     if self.ExtraCapabilities().get( 'workspace', {} ).get( 'configuration' ):
>       sections_to_config_map = self._settings.get( 'ls_config', {} )
>       return [ sections_to_config_map.get( i, None )
> ```

We need to document this somewhere.

bstaletic (commented):
I don't know if codecov will be happy with the tests, but I think it will. I still did not cover the two lines that either mean the completer isn't written well, or the server violates the protocol. Either way, there will be a `lsp.Reject` call, so we can still see what's going on in the logs. Maybe some assert would be a good idea?

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468#-:-MGtd8EtEj2rwIsJnN67:b-8gc5ge)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 89 at r5](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGp2DgW6ClnbMlFLR0e:-MGsJWDF0Bv3ELDjmjBS:b-pns998) ([raw file](https://github.com/ycm-core/ycmd/blob/dd4e3e9910dcd08247883cd511d02d5dddf6eed7/ycmd/completers/language_server/generic_lsp_completer.py#L89)):*

We need to document this somewhere. 
</blockquote></details>

Added to the README. Please check the wording.
___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 27 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGtch031Y0WB4ze0RG8:-MGtch031Y0WB4ze0RG9:borzlco) ([raw file](https://github.com/ycm-core/ycmd/blob/6441a6a302ba682e4a0f5f5f0576300287cf94b7/ycmd/completers/language_server/generic_lsp_completer.py#L27)):*
> ```Python
>     self._supported_filetypes = server_settings[ 'filetypes' ]
>     self._project_root_files = server_settings.get( 'project_root_files', [] )
>     self._capabilities = server_settings.get( 'capabilities', {} )
> ```

Calling it just `capabilities`, since the user is already inside a dictionary called `language_server`.
___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 99 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGtcoTVBqW74WfkEdok:-MGtcoTVBqW74WfkEdol:bcnqd3w) ([raw file](https://github.com/ycm-core/ycmd/blob/6441a6a302ba682e4a0f5f5f0576300287cf94b7/ycmd/completers/language_server/generic_lsp_completer.py#L99)):*
> ```Python
>   def WorkspaceConfigurationResponse( self, request ):
>     if self._capabilities.get( 'workspace', {} ).get( 'configuration' ):
>       sections_to_config_map = self._settings.get( 'ls_config', {} )
> ```

I really don't like the `ls_config` name. How about `sections`, to make it sound really different and (hopefully) avoid confusion?

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic and @puremourning)

___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 99 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGtcoTVBqW74WfkEdok:-MGtuIFMEkJpGIpx6Tas:b-pdphgh) ([raw file](https://github.com/ycm-core/ycmd/blob/6441a6a302ba682e4a0f5f5f0576300287cf94b7/ycmd/completers/language_server/generic_lsp_completer.py#L99)):*

I really don't like the `ls_config` name. How about `sections`, to make it sound really different and (hopefully) avoid confusion?
</blockquote></details>

agree `ls_config` isn't nice. `config_sections` works for me. I've never liked `ls` as we use it now tbh,  but that's not really changeable :)
___
*[ycmd/tests/language_server/generic_completer_test.py, line 537 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGtvJr-37Jwva4rIu3f:-MGtvJr-37Jwva4rIu3g:bkwwgr) ([raw file](https://github.com/ycm-core/ycmd/blob/89282549c1b10169dbe8698b6ad2988d5982e82c/ycmd/tests/language_server/generic_completer_test.py#L537)):*
> ```Python
> @patch( 'ycmd.completers.completer.Completer.ShouldUseSignatureHelpNow',
>         return_value = True )
> def GenericLSPCompleter_WorkspaceConfiguration_test( should_use_sig, app ):
> ```

am I being dumb? how is this signature help test related to workspace configuration ?
___
*[ycmd/tests/language_server/generic_completer_test.py, line 562 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGtvcLw0Ppf1MoYAAne:-MGtvcLw0Ppf1MoYAAnf:b-ecj5e8) ([raw file](https://github.com/ycm-core/ycmd/blob/89282549c1b10169dbe8698b6ad2988d5982e82c/ycmd/tests/language_server/generic_completer_test.py#L562)):*
> ```Python
> 
> @IsolatedYcmd( {
>   'global_ycm_extra_conf': PathToTestFile( 'generic_server', 'single_diag.py' ),
> ```

did you forget to submit this file?
___
*[ycmd/tests/language_server/generic_completer_test.py, line 568 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGtvSIC5yvCA5xBweyz:-MGtvSIC5yvCA5xBwez-:bu18kok) ([raw file](https://github.com/ycm-core/ycmd/blob/89282549c1b10169dbe8698b6ad2988d5982e82c/ycmd/tests/language_server/generic_completer_test.py#L568)):*
> ```Python
>         'capabilities': { 'workspace': { 'configuration': True } },
>         'cmdline': [ 'node', PATH_TO_GENERIC_COMPLETER, '--stdio' ] } ] } )
> def GenericLSPCompleter_SingleDiagnostics_test( app ):
> ```

should we add a corroborating test that proves we get 4 diagnostics in the case where we don't override it ?

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 99 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGtcoTVBqW74WfkEdok:-MGtyS7R9MnDKPr8HtyI:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/6441a6a302ba682e4a0f5f5f0576300287cf94b7/ycmd/completers/language_server/generic_lsp_completer.py#L99)):*

agree `ls_config` isn't nice. `config_sections` works for me. I've never liked `ls` as we use it now tbh,  but that's not really changeable :)
</blockquote></details>

Done.
___
*[ycmd/tests/language_server/generic_completer_test.py, line 537 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGtvJr-37Jwva4rIu3f:-MGtyjTHEEkQfht1B8d3:b-8s77bd) ([raw file](https://github.com/ycm-core/ycmd/blob/89282549c1b10169dbe8698b6ad2988d5982e82c/ycmd/tests/language_server/generic_completer_test.py#L537)):*

am I being dumb? how is this signature help test related to workspace configuration ?
</blockquote></details>

I thought I removed that one! My bad...
___
*[ycmd/tests/language_server/generic_completer_test.py, line 562 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGtvcLw0Ppf1MoYAAne:-MGtys2MDhLv1voHYuAb:bulbe3g) ([raw file](https://github.com/ycm-core/ycmd/blob/89282549c1b10169dbe8698b6ad2988d5982e82c/ycmd/tests/language_server/generic_completer_test.py#L562)):*

did you forget to submit this file?
</blockquote></details>

Yes, I did.
___
*[ycmd/tests/language_server/generic_completer_test.py, line 568 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1468#-MGtvSIC5yvCA5xBweyz:-MGu-Gn08EvIE17ZtwbS:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/89282549c1b10169dbe8698b6ad2988d5982e82c/ycmd/tests/language_server/generic_completer_test.py#L568)):*

should we add a corroborating test that proves we get 4 diagnostics in the case where we don't override it ?
</blockquote></details>

Done.

bstaletic:
Got coverage to 100%. This should be ready.

puremourning (approved):
:lgtm:

> Reviewed 6 of 11 files at r6, 2 of 5 files at r7, 6 of 6 files at r8.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468#-:-MGwN3RN4HhI_3hivztH:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

puremourning (approved):
:lgtm:

> Reviewed 11 of 11 files at r10.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1468#-:-MGyKOQV3btyQUZMvic3:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic:
It only took me a month to finish this... :lgtm:

PR #1471, opened 2020-09-08
[READY] Support connecting to TCP server

puremourning:
In the generic LSP completer, we allow you to specify a new key `port` which is the TCP port on localhost to connect to instead of using stdio.

This is required for Godot.

The quirk of the tests is that _most_ LSP servers actually want to be the client of the TCP connection, not the server. As they _also_ support stdio, there's no good reason to actually support that use case right now, so we don't (it would be trivial to add as an extension to this). In order to test this, we use the "demo" LS that MS provide. I've replaced the pre-built code with the source and built it all, as this allows us flexibility to change it in the future.

As the standard LSP requires being a client, and so does this implementation in ycmd, I wrote a little script using `nc` to bridge the two tcp links. This script listens on 2 ports and pipes all the data from one socket to the other using a couple of FIFOs. This won't work on windows, and I'm not sure if it will work on linux. The only other alternative is to implement a "server" socket in the demo LS (hence why I switched to including the source).

Sorry, not the best explanation.

Review by commits is best.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1471)


bstaletic (changes requested):
Am I supposed to review the source of the MS example LSP server? I'll try the PR on my laptop.

The tests are failing on linux and windows.

> Reviewed 4 of 4 files at r1, 1 of 1 files at r2, 2 of 32 files at r3, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1471#-:-MGhowneDOIR20auWtNu:bop72qv)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 28 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhjuBEE0MhQKzR8F-M:-MGhjuBEE0MhQKzR8F-N:b-m5jrzr) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/generic_lsp_completer.py#L28)):*
> ```Python
>     self._project_root_files = server_settings.get( 'project_root_files', [] )
> 
>     self._command_line = server_settings.get( 'cmdline' )
> ```

Why is this a `.get()` and not a simple `[]`?
___
*[ycmd/completers/language_server/language_server_completer.py, line 308 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhkGfEEu4lJtPhZyp_:-MGhkGfEEu4lJtPhZypa:b7lk1r) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L308)):*
> ```Python
>   @abc.abstractmethod
>   def IsConnected( self ):
>     pass
> ```

Should this really be an `abstractmethod`? The stdio connection doesn't use it.
___
*[ycmd/completers/language_server/language_server_completer.py, line 774 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhly4Y6Ec40SlRaiJx:-MGhly4Y6Ec40SlRaiJy:besy5ms) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L774)):*
> ```Python
> 
>   def IsConnected( self ):
>     return self._client_socket
> ```

Do we need an explicit cast to `bool`?
___
*[ycmd/completers/language_server/language_server_completer.py, line 793 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhnU6rE1PUssAQKiLk:-MGhnU6rE1PUssAQKiLl:b54wqoh) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L793)):*
> ```Python
> 
>       if sent == 0:
>         raise RuntimeError( 'write socket failed' )
> ```

`'Writing to socket failed'`?
___
*[ycmd/completers/language_server/language_server_completer.py, line 818 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhncx2AIFDJ_zIinIm:-MGhncx35jqRGqPowH0Q:b-ftj9n2) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L818)):*
> ```Python
>           raise LanguageServerConnectionStopped()
> 
>         raise RuntimeError( 'read socket failed' )
> ```

`'Reading from socket failed'`?
___
*[ycmd/completers/language_server/language_server_completer.py, line 828 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhnibp4nDtk09JHqjH:-MGhnibp4nDtk09JHqjI:b-vj4lre) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L828)):*
> ```Python
>       bytes_read += len( chunk )
> 
>     return utils.ToBytes( '' ).join( chunks )
> ```

Why not `return b''.join( chunks )`?
___
*[ycmd/completers/language_server/language_server_completer.py, line 1027 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhoGMq55C_M02kc3Bi:-MGhoGMq55C_M02kc3Bj:bxljsky) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L1027)):*
> ```Python
>                  self.GetCommandLine() )
> 
>     if self.GetCommandLine():
> ```

Should we check this or the type of connection?
___
*[ycmd/completers/language_server/language_server_completer.py, line 1194 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhoL01El5-eujnnBl7:-MGhoL01El5-eujnnBl8:bs4q9op) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L1194)):*
> ```Python
> 
>   def ServerIsHealthy( self ):
>     if not self.GetCommandLine():
> ```

Ditto.

bstaletic (commented):
This is the traceback that I get, when I run the new tests.

```
Exception in thread Thread-1:
  File "/home/bstaletic/.vim/pack/bundle/start/YouCompleteMe/third_party/ycmd/ycmd/completers/language_server/language_server_completer.py", line 357, in run
    self.TryServerConnectionBlocking()
  File "/home/bstaletic/.vim/pack/bundle/start/YouCompleteMe/third_party/ycmd/ycmd/completers/language_server/language_server_completer.py", line 759, in TryServerConnectionBlocking
    raise reason
  File "/home/bstaletic/.vim/pack/bundle/start/YouCompleteMe/third_party/ycmd/ycmd/completers/language_server/language_server_completer.py", line 764, in TryServerConnectionBlocking
    self._client_socket = socket.create_connection( ( '', self.port ) )
  File "/usr/lib/python3.8/socket.py", line 787, in create_connection
    for res in getaddrinfo(host, port, 0, SOCK_STREAM):
  File "/usr/lib/python3.8/socket.py", line 918, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -2] Name or service not known

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/home/bstaletic/.vim/pack/bundle/start/YouCompleteMe/third_party/ycmd/ycmd/completers/language_server/language_server_completer.py", line 382, in run
    self.Shutdown()
  File "/home/bstaletic/.vim/pack/bundle/start/YouCompleteMe/third_party/ycmd/ycmd/completers/language_server/language_server_completer.py", line 778, in Shutdown
    self._client_socket.close()
AttributeError: 'NoneType' object has no attribute 'close'
```

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1471#-:-MGhtlB04VUdLr6I6UtF:b-k0olx7)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

puremourning:
yeah I figured the test tcp bridge thing wouldn't work everywhere. bit annoying. I think I forgot some @UnixOnly too

puremourning:
I've removed the dodgy bridge process and just implemented the listen socket in the generic_server.

This was trivial enough using the vscode-languageserver-protocol APIs and some reading of the code.

The diff will look massive, but it's basically just adding this:

```typescript
const idx = process.argv.indexOf( '--listen' )
if ( idx >= 0 && idx + 1 < process.argv.length ) {
  const port = parseInt( process.argv[ idx + 1 ] );
  createClientSocketTransport( port ).then( (transport) => {
    return transport.onConnected().then( (protocol) => {
      let connection = createConnection( protocol[ 0 ], protocol[ 1 ] )
      RunServer( connection );
    } );
  } );

} else {
  let connection = createConnection(ProposedFeatures.all);
  RunServer( connection );
}
```

And wrapping the whole of the rest of the code in a function so it can be called from a callback.

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1471)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 28 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhjuBEE0MhQKzR8F-M:-MGruMH1BZsa58D_vMpI:b-3bwn21) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/generic_lsp_completer.py#L28)):*

Why is this a `.get()` and not a simple `[]`?
</blockquote></details>

because in the case of Godot, we don't start the sever, we just connect to it.

I;ve essentially made starting a server process optional for the generic completer. If GetCommandLine() returns None, then the language_server_completer won't attempt to start any server.
___
*[ycmd/completers/language_server/language_server_completer.py, line 308 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhkGfEEu4lJtPhZyp_:-MGruZfA95JlUEvfuj6k:bj7zo46) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L308)):*

Should this really be an `abstractmethod`? The stdio connection doesn't use it.
</blockquote></details>

I think it's preferable to make it abstract so that any implementations have to at least think about what the method should return otherwise it's easy to just end up using the default and not thinking about it.
___
*[ycmd/completers/language_server/language_server_completer.py, line 774 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhly4Y6Ec40SlRaiJx:-MGrutps2gFrm7qnFv25:b-vv582c) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L774)):*

Do we need an explicit cast to `bool`?
</blockquote></details>

yes, good shout
___
*[ycmd/completers/language_server/language_server_completer.py, line 793 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhnU6rE1PUssAQKiLk:-MGrv3so3pnaiHOjHRaY:bn9sbx0) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L793)):*

`'Writing to socket failed'`?
</blockquote></details>

improved this message
___
*[ycmd/completers/language_server/language_server_completer.py, line 818 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhncx2AIFDJ_zIinIm:-MGrv5CxDDXDGvasHlEn:bn9sbx0) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L818)):*

`'Reading from socket failed'`?
</blockquote></details>

improved this message
___
*[ycmd/completers/language_server/language_server_completer.py, line 828 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhnibp4nDtk09JHqjH:-MGrvAUKEh2Da3csxK4Z:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L828)):*

Why not `return b''.join( chunks )`?
</blockquote></details>

Done.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1027 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhoGMq55C_M02kc3Bi:-MGrvBdJ7vvdwIhDcfrv:b-gp25vu) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L1027)):*

Should we check this or the type of connection?
</blockquote></details>

We check this, because in the tests we actually supply a cmdline _and_ a port. So it's a "TCP" type connection, but we started the server.

In the Godot case, it's a TCP type connection, but we _didn't_ start the server.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1194 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhoL01El5-eujnnBl7:-MGrvN729HJAQ1uWLPsl:b-z3p5ge) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L1194)):*

Ditto.
</blockquote></details>

ditto above

bstaletic (commented):
> most LSP servers actually want to be the client of the TCP connection, not the server. 

Can you convince me that patching the example server is easier than allowing LSP servers to act as TCP clients?

> Reviewed 1 of 35 files at r6, 1 of 1 files at r7, 1 of 32 files at r8, 1 of 3 files at r11, 1 of 1 files at r13.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1471#-:-MGs3EYV0dfO_AhvOioU:bo0lnv2)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/completers/language_server/generic_lsp_completer.py, line 28 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhjuBEE0MhQKzR8F-M:-MGs18cgAurQPuIHvtXw:bdqt6yx) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/generic_lsp_completer.py#L28)):*

because in the case of Godot, we don't start the sever, we just connect to it.

I;ve essentially made starting a server process optional for the generic completer. If GetCommandLine() returns None, then the language_server_completer won't attempt to start any server.
</blockquote></details>

Yeah, figured it out.
___
*[ycmd/completers/language_server/language_server_completer.py, line 308 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhkGfEEu4lJtPhZyp_:-MGs1E0T8Zexsz-VUVai:b-gxztsp) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L308)):*

I think it's preferable to make it abstract so that any implementations have to at least think about what the method should return otherwise it's easy to just end up using the default and not thinking about it.
</blockquote></details>

Fair enough.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1027 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhoGMq55C_M02kc3Bi:-MGs0ycb-aaB6KzVlneA:b-c4t93p) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L1027)):*

We check this, because in the tests we actually supply a cmdline _and_ a port. So it's a "TCP" type connection, but we started the server.

In the Godot case, it's a TCP type connection, but we _didn't_ start the server.
</blockquote></details>

That makes sense, but... Who does start the godot server?

bstaletic (changes requested):
> Reviewed 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1471)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[third_party/generic_server/server/src/server.ts, line 1 at r13](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGs6UmP6LgfRAafzdrZ:-MGs6UmP6LgfRAafzdr_:b-2qx9ki) ([raw file](https://github.com/ycm-core/ycmd/blob/c766cb8db1e87c8bdf932c4f4d8713042d51fbc9/third_party/generic_server/server/src/server.ts#L1)):*
> ```XML
> /* --------------------------------------------------------------------------------------------
> ```

Indentation is completely messed up, because tabs vs spaces.

puremourning (commented):
(from later) 
> That makes sense, but... Who does start the godot server?

Good question, I should have explained this. Godot is a game editor, and there's a checkbox in its configuration "start language server". This makes the Godot editor open a TCP port on localhost which speaks LSP and your editor can connect to it. So the _user_ starts the Godot server by starting Godot and opening a project. I think this is the only plausible way this could work. I used https://github.com/habamax/vim-godot to get Godot in my vim, as it was @habamax that was requesting YCM supported this model.

> Can you convince me that patching the example server is easier than allowing LSP servers to act as TCP clients

I think so... What I can say is that we have distinct use-cases for:

* ycmd starts LS and communicates using stdio (most of them)
* ycmd connects to a port opened by an external LS (Godot)

We don't have use case for:

* ycmd starts LS, listens on port and LS connects to ycmd's port (no use case as all servers that do this also support stdio)

So, in the tests, ycmd starts LS and communicates with LS using TCP. But, the LS wants to be the server (the case we don't have a use-case for), so rather than implementing that and thus testing something we don't have a use-case for, I made the example server act like the Godot server so we can test a scenario we actually have use for.

Got me?

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1471#-:-MGsLdtkBW1YnMo8Vf8A:b-l6t4eh)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[third_party/generic_server/server/src/server.ts, line 1 at r13](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGs6UmP6LgfRAafzdrZ:-MGsOJNX6LY3YWBoikD8:b6icc0o) ([raw file](https://github.com/ycm-core/ycmd/blob/c766cb8db1e87c8bdf932c4f4d8713042d51fbc9/third_party/generic_server/server/src/server.ts#L1)):*

Indentation is completely messed up, because tabs vs spaces.
</blockquote></details>

hmm. I've run YcmCompleter Format on the whole file.

This makes it diverge a lot more from the original source, but I think that's OK.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1027 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGhoGMq55C_M02kc3Bi:-MGsP2-M1YtEdYhiqSLd:b-49kr2q) ([raw file](https://github.com/ycm-core/ycmd/blob/89c02f5d8c7a9aac06511d0ac5916f18f44f0abe/ycmd/completers/language_server/language_server_completer.py#L1027)):*

That makes sense, but... Who does start the godot server?
</blockquote></details>

The user (it's a game engine/editor so the user opens it and sets the start language server option in the project)

bstaletic (changes requested):
> Got me?

Loud and clear.

There's one more question remaining, but I'm pretty sure it will be fine, so :lgtm: 

> Reviewed 1 of 1 files at r15, 1 of 32 files at r16, 1 of 1 files at r18, 2 of 2 files at r19.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1471#-:-MGspHXeB7BeM9GMDiOf:bpu8u7g)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[third_party/generic_server/server/src/server.ts, line 1 at r13](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGs6UmP6LgfRAafzdrZ:-MGsnUPC1JqfwmrJB-3g:b-2vhq7) ([raw file](https://github.com/ycm-core/ycmd/blob/c766cb8db1e87c8bdf932c4f4d8713042d51fbc9/third_party/generic_server/server/src/server.ts#L1)):*

hmm. I've run YcmCompleter Format on the whole file.

This makes it diverge a lot more from the original source, but I think that's OK.
</blockquote></details>

You can run `%s/  /\t/g` now, to fix it if you want. I'm not blocking on this though. I also think this is fine.
___
*[ycmd/completers/language_server/language_server_completer.py, line 790 at r19](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGspQqL0oKU8N5ci0Bm:-MGspQqMA-fs-y0GShV7:b-tegao0) ([raw file](https://github.com/ycm-core/ycmd/blob/b9fbcf4762e13e62aa293152f5f45692cc16ee73/ycmd/completers/language_server/language_server_completer.py#L790)):*
> ```Python
>       try:
>         sent = self._client_socket.send( data[ total_sent: ] )
>       except OSError:
> ```

Where did this `socket.error` -> `OSError` change come from?

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1471)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/completers/language_server/language_server_completer.py, line 790 at r19](https://reviewable.io/reviews/ycm-core/ycmd/1471#-MGspQqL0oKU8N5ci0Bm:-MGst9jVCwSnC4KO_Snw:boytczv) ([raw file](https://github.com/ycm-core/ycmd/blob/b9fbcf4762e13e62aa293152f5f45692cc16ee73/ycmd/completers/language_server/language_server_completer.py#L790)):*
when reviewing the docs, I found that [`socket.error` is deprecated](https://docs.python.org/3/library/socket.html#socket.error)

> A deprecated alias of OSError.

> Changed in version 3.3: Following PEP 3151, this class was made an alias of OSError.

puremourning (commented):
:lgtm:

> Reviewed 2 of 4 files at r1, 1 of 35 files at r6, 1 of 1 files at r15, 30 of 32 files at r16, 1 of 1 files at r18, 2 of 2 files at r19.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1471#-:-MGsu-Ck1e_9FVk4zLbj:bnfp4nl)* status: :shipit: complete! 2 of 2 LGTMs obtained (waiting on @bstaletic)

habamax:
Hi, dumb question :)

I have updated YouCompleteMe and ... it doesn't work. I guess server is not updated automatically and I should pull it separately? If so, when YouCompleteMe would be in sync with ycmd?

bstaletic:
Hi, not a dumb question at all. Yes, right now ycmd in YCM is behind master. I want to get #1468 in before updating the submodule in YCM and I think #1473 would be a good idea too. You can pull the latest master with `cd YCM/third_party/ycmd; git pull origin master; git submodule update --init --recursive`, which ensures that the submodules after ycmd's update are correctly checked out.

puremourning:
Yeah @habamax we will need to merge https://github.com/ycm-core/ycmd/pull/1473 and a Corresponding YCM client change for Godot to work properly in master. 

For now my fork’s resolve branch should work for you.

habamax:
got it, thx!

PR #1473, opened 2020-09-10
[READY] Resolve completion items on demand rather than upfront

puremourning:
Again, opening this for thoughts while testing.

Previously we would do completion/resolve for every matching completion candidate up front. This was also embarrassingly serial. For 100 candidates, we'd do 100 synchronous resolve requests. This can lead to timeouts.

This leads to unusable performance on the Godot server (because for some reason there are a lot of candidates and they take a _long_ time to resolve serially).

So we implement an endpoint to resolve a completion item on-demand, and change the resolve behaviour so that we only ever resolve 10 candidates up-front.

The approach is to add a `'resolve'` key to the `extra_data` dictionary which is an ID telling the client that this item needs to be resolved to fill in the extra data documentation. If this key is not present, resolve is not required.

Couple of things to note:

1. we resolve up to 10 _matching_ candidates (i.e. we still do FilterAndSortCandidates _first_)
2. we therefore still have a mixture of resolved and un-resolved candidates in the cache
3. we only allow resolving a candidate if there is a cache HIT for the request data. this means we can use the index into the full set of items as the "key" for what to resolve.
4. we don't resolve things multiple times. for candidates that are already resolved, we don't include the 'resolve' key which tells the client that it might need resolving.

Client changes are here: https://github.com/puremourning/YouCompleteMe/tree/resolve (also very much WIP right now)

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -220,6 +220,18 @@
                     </tr>
                     <tr>
                             <td class="swagger--summary-path" rowspan="1">
+                                <a href="#path--resolve_completion">/resolve_completion</a>
+                            </td>
+                        <td>
+                            <a href="#operation--resolve_completion-post">POST</a>
+                        </td>
+                        <td>
+                            <p>Resolve detailed_info for a completion item</p>
+        
+                        </td>
+                    </tr>
+                    <tr>
+                            <td class="swagger--summary-path" rowspan="1">
                                 <a href="#path--resolve_fixit">/resolve_fixit</a>
                             </td>
                         <td>
@@ -1652,6 +1664,214 @@
                 </div>
             </div>
         
+        <span id="path--resolve_completion"></span>
+            <div id="operation--resolve_completion-post" class="swagger--panel-operation-post panel">
+                <div class="panel-heading">
+                    <div class="operation-summary">Resolve detailed_info for a completion item</div>
+                    <h3 class="panel-title"><span class="operation-name">POST</span> <strong>/resolve_completion</strong></h3>
+                </div>
+                <div class="panel-body">
+                    <section class="sw-operation-description">
+                        <p>Given the <code>resolve</code> ID for a completion item, returns the item with
+            additional data filled in.</p>
+            <p>Note, the supplied <code>request_data</code> must match exactly the <code>request_data</code>
+            for the corresponding completion request, except with the <code>resolve</code> key
+            filled in with the <code>resolve</code> ID from the completion item&#39;s <code>extra_data</code>.</p>
+            
+                    </section>
+            
+                        
+                            <section class="sw-request-body">
+                                
+                                        <div class="row">
+                                            <div class="col-md-6">
+                                                <p><p>The context data, including the current cursor position, and details
+                        of dirty buffers.</p>
+                        </p>
+                                            </div>
+                                            <div class="col-md-6 sw-request-model">
+                            <div  class="panel panel-definition">
+                                <div class="panel-body">
+                                        
+                                                <section class="json-schema-properties">
+                                                    <dl>
+                                                            <dt data-property-name="line_num">
+                                                                <span class="json-property-name">line_num:</span>
+                                                                
+                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/LineNumber">LineNumber</a>
+                                                                </span>
+                                                                <span class="json-property-range" title="Value limits"></span>
+                                                                
+                                                                    <span class="json-property-required"></span>
+                                                            </dt>
+                                                            <dd>
+                                                                
+                                                                <div class="json-inner-schema">
+                                                                    
+                                                                </div>
+                                                            </dd>
+                                                            <dt data-property-name="column_num">
+                                                                <span class="json-property-name">column_num:</span>
+                                                                
+                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/ColumnNumber">ColumnNumber</a>
+                                                                </span>
+                                                                <span class="json-property-range" title="Value limits"></span>
+                                                                
+                                                                    <span class="json-property-required"></span>
+                                                            </dt>
+                                                            <dd>
+                                                                
+                                                                <div class="json-inner-schema">
+                                                                    
+                                                                </div>
+                                                            </dd>
+                                                            <dt data-property-name="filepath">
+                                                                <span class="json-property-name">filepath:</span>
+                                                                
+                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/FilePath">FilePath</a>
+                                                                </span>
+                                                                <span class="json-property-range" title="Value limits"></span>
+                                                                
+                                                                    <span class="json-property-required"></span>
+                                                            </dt>
+                                                            <dd>
+                                                                
+                                                                <div class="json-inner-schema">
+                                                                    
+                                                                </div>
+                                                            </dd>
+                                                            <dt data-property-name="file_data">
+                                                                <span class="json-property-name">file_data:</span>
+                                                                
+                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/FileDataMap">FileDataMap</a>
+                                                                </span>
+                                                                <span class="json-property-range" title="Value limits"></span>
+                                                                
+                                                                    <span class="json-property-required"></span>
+                                                            </dt>
+                                                            <dd>
+                                                                
+                                                                <div class="json-inner-schema">
+                                                                    
+                                                                </div>
+                                                            </dd>
+                                                            <dt data-property-name="completer_target">
+                                                                <span class="json-property-name">completer_target:</span>
+                                                                
+                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/CompleterTarget">CompleterTarget</a>
+                                                                </span>
+                                                                <span class="json-property-range" title="Value limits"></span>
+                                                                
+                                                            </dt>
+                                                            <dd>
+                                                                
+                                                                <div class="json-inner-schema">
+                                                                    
+                                                                </div>
+                                                            </dd>
+                                                            <dt data-property-name="working_dir">
+                                                                <span class="json-property-name">working_dir:</span>
+                                                                
+                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/WorkingDirectory">WorkingDirectory</a>
+                                                                </span>
+                                                                <span class="json-property-range" title="Value limits"></span>
+                                                                
+                                                            </dt>
+                                                            <dd>
+                                                                
+                                                                <div class="json-inner-schema">
+                                                                    
+                                                                </div>
+                                                            </dd>
+                                                            <dt data-property-name="extra_conf_data">
+                                                                <span class="json-property-name">extra_conf_data:</span>
+                                                                
+                                                                <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/ExtraConfData">ExtraConfData</a>
+                                                                </span>
+                                                                <span class="json-property-range" title="Value limits"></span>
+                                                                
+                                                            </dt>
+                                                            <dd>
+                                                                
+                                                                <div class="json-inner-schema">
+                                                                    
+                                                                </div>
+                                                            </dd>
+                                                            <dt data-property-name="resolve">
+                                                                <span class="json-property-name">resolve:</span>
+                                                                
+                                                                <span class="json-property-type">object</span>
+                                                                <span class="json-property-range" title="Value limits"></span>
+                                                                
+                                                                    <span class="json-property-required"></span>
+                                                            </dt>
+                                                            <dd>
+                                                                <p>The &#39;resolve&#39; key from the &#39;extra_data&#39; in the completion item
+                                                to resolve.</p>
+                                                
+                                                                <div class="json-inner-schema">
+                                                                    
+                                                                </div>
+                                                            </dd>
+                                                    </dl>
+                                                </section>
+                                </div>
+                            </div></div>
+                                        </div>
+                            </section>        
+                    
+                        <section class="sw-responses">
+                                <p><span class="label label-default">application/json</span> 
+                    </p>
+                    
+                            <dl>
+                                    <dt class="sw-response-200">
+                                        200 OK
+                                    
+                                    </dt>
+                                    <dd class="sw-response-200">
+                                            <div class="row">
+                                                <div class="col-md-12">
+                                                    <p>The list of completion suggestions.</p>
+                                            
+                                                </div>
+                                            </div>
+                                            <div class="row">
+                                                
+                                                <div class="col-md-6 sw-response-model">
+                                                <div  class="panel panel-definition">
+                                                    <div class="panel-body">
+                                                                <a class="json-schema-ref" href="#/definitions/ResolveCompletionResponse">ResolveCompletionResponse</a>
+                                                    </div>
+                                                </div></div>
+                                            
+                                            </div>                </dd>
+                                    <dt class="sw-response-500">
+                                        500 Internal Server Error
+                                    
+                                    </dt>
+                                    <dd class="sw-response-500">
+                                            <div class="row">
+                                                <div class="col-md-12">
+                                                    <p>An error occurred.</p>
+                                            
+                                                </div>
+                                            </div>
+                                            <div class="row">
+                                                
+                                                <div class="col-md-6 sw-response-model">
+                                                <div  class="panel panel-definition">
+                                                    <div class="panel-body">
+                                                                <a class="json-schema-ref" href="#/definitions/ExceptionResponse">ExceptionResponse</a>
+                                                    </div>
+                                                </div></div>
+                                            
+                                            </div>                </dd>
+                            </dl>
+                        </section>
+                </div>
+            </div>
+        
         <span id="path--resolve_fixit"></span>
             <div id="operation--resolve_fixit-post" class="swagger--panel-operation-post panel">
                 <div class="panel-heading">
@@ -2863,6 +3083,45 @@
                                                                                     
                                                                                 </div>
                                                                             </dd>
+                                                                            <dt data-property-name="fixits">
+                                                                                <span class="json-property-name">fixits:</span>
+                                                                                
+                                                                                <span class="json-property-type">object[]</span>
+                                                                                <span class="json-property-range" title="Value limits"></span>
+                                                                                
+                                                                            </dt>
+                                                                            <dd>
+                                                                                <p>Any editional edits to apply when selecting this completion</p>
+                                                                
+                                                                                <div class="json-inner-schema">
+                                                                                    
+                                                                                                    <section class="json-schema-array-items">
+                                                                                                        
+                                                                                                        <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/FixIt">FixIt</a>
+                                                                                                        </span>
+                                                                                                        <span class="json-property-range" title="Value limits"></span>
+                                                                                                        
+                                                                                                        <div class="json-inner-schema">
+                                                                                                            
+                                                                                                        </div>
+                                                                                                    </section>                </div>
+                                                                            </dd>
+                                                                            <dt data-property-name="resolve">
+                                                                                <span class="json-property-name">resolve:</span>
+                                                                                
+                                                                                <span class="json-property-type">number</span>
+                                                                                <span class="json-property-range" title="Value limits"></span>
+                                                                                
+                                                                            </dt>
+                                                                            <dd>
+                                                                                <p>Identifier used to resolve this completion item using
+                                                                /resolve_completion. If not supplied, the item is already
+                                                                fully resolved.</p>
+                                                                
+                                                                                <div class="json-inner-schema">
+                                                                                    
+                                                                                </div>
+                                                                            </dd>
                                                                     </dl>
                                                                 </section>
                                                     </div>
@@ -3979,6 +4238,63 @@
                                                     <div class="json-inner-schema">
                                                         
                                                     </div>
+                                                </dd>
+                                        </dl>
+                                    </section>
+                    </div>
+                </div>        
+                <div id="definition-ResolveCompletionResponse" class="panel panel-definition">
+                        <div class="panel-heading">
+                                <h3 class="panel-title"><a name="/definitions/ResolveCompletionResponse"></a>ResolveCompletionResponse:
+                                    <span class="json-property-type">
+                <span class="json-property-type">object</span>
+                <span class="json-property-range" title="Value limits"></span>
+                
+                </span>
+                                </h3>
+                        </div>
+                    <div class="panel-body">
+                            
+                                    <section class="json-schema-properties">
+                                        <dl>
+                                                <dt data-property-name="completion">
+                                                    <span class="json-property-name">completion:</span>
+                                                    
+                                                    <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/Candidate">Candidate</a>
+                                                    </span>
+                                                    <span class="json-property-range" title="Value limits"></span>
+                                                    
+                                                        <span class="json-property-required"></span>
+                                                </dt>
+                                                <dd>
+                                                    
+                                                    <div class="json-inner-schema">
+                                                        
+                                                    </div>
+                                                </dd>
+                                                <dt data-property-name="errors">
+                                                    <span class="json-property-name">errors:</span>
+                                                    
+                                                    <span class="json-property-type">object[]</span>
+                                                    <span class="json-property-range" title="Value limits"></span>
+                                                    
+                                                        <span class="json-property-required"></span>
+                                                </dt>
+                                                <dd>
+                                                    <p>Any errors reported by the semantic completion engine.</p>
+                                    
+                                                    <div class="json-inner-schema">
+                                                        
+                                                                        <section class="json-schema-array-items">
+                                                                            
+                                                                            <span class="json-property-type">    <a class="json-schema-ref" href="#/definitions/ExceptionResponse">ExceptionResponse</a>
+                                                                            </span>
+                                                                            <span class="json-property-range" title="Value limits"></span>
+                                                                            
+                                                                            <div class="json-inner-schema">
+                                                                                
+                                                                            </div>
+                                                                        </section>                </div>
                                                 </dd>
                                         </dl>
                                     </section>
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -480,6 +480,19 @@ definitions:
             description: |-
               Additional documentation/information to be displayed
               alongside (after) information in `detailed_info`.
+          fixits:
+            type: array
+            description: |-
+              Any editional edits to apply when selecting this completion
+            items:
+              $ref: "#/definitions/FixIt"
+          resolve:
+            type: number
+            description: |-
+              Identifier used to resolve this completion item using
+              /resolve_completion. If not supplied, the item is already
+              fully resolved.
+
 
   CompletionResponse:
     type: object
@@ -498,6 +511,20 @@ definitions:
           1-based byte index of the column from which the completion should be
           applied. This is the column in which the words suggested in
           `completions` should be placed.
+      errors:
+        type: array
+        description: Any errors reported by the semantic completion engine.
+        items:
+          $ref: "#/definitions/ExceptionResponse"
+
+  ResolveCompletionResponse:
+    type: object
+    required:
+      - completion
+      - errors
+    properties:
+      completion:
+        $ref: "#/definitions/Candidate"
       errors:
         type: array
         description: Any errors reported by the semantic completion engine.
@@ -931,6 +958,62 @@ paths:
           description: An error occurred.
           schema:
             $ref: "#/definitions/ExceptionResponse"
+  /resolve_completion:
+    post:
+      summary: Resolve detailed_info for a completion item
+      description: |-
+        Given the `resolve` ID for a completion item, returns the item with
+        additional data filled in.
+
+        Note, the supplied `request_data` must match exactly the `request_data`
+        for the corresponding completion request, except with the `resolve` key
+        filled in with the `resolve` ID from the completion item's `extra_data`.
+      produces:
+        - application/json
+      parameters:
+        - name: request_data
+          in: body
+          description: |-
+            The context data, including the current cursor position, and details
+            of dirty buffers.
+          required: true
+          schema:
+            type: object
+            required:
+              - line_num
+              - column_num
+              - filepath
+              - file_data
+              - resolve
+            properties:
+              line_num:
+                $ref: "#/definitions/LineNumber"
+              column_num:
+                $ref: "#/definitions/ColumnNumber"
+              filepath:
+                $ref: "#/definitions/FilePath"
+              file_data:
+                $ref: "#/definitions/FileDataMap"
+              completer_target:
+                $ref: "#/definitions/CompleterTarget"
+              working_dir:
+                $ref: "#/definitions/WorkingDirectory"
+              extra_conf_data:
+                $ref: "#/definitions/ExtraConfData"
+              resolve:
+                description: |-
+                  The 'resolve' key from the 'extra_data' in the completion item
+                  to resolve.
+      responses:
+        200:
+          description: The list of completion suggestions.
+          schema:
+            $ref: "#/definitions/ResolveCompletionResponse"
+        500:
+          description: An error occurred.
+          schema:
+            $ref: "#/definitions/ExceptionResponse"
+
 
   /signature_help_available:
     get:
--- a/ycmd/completers/completer.py
+++ b/ycmd/completers/completer.py
@@ -28,6 +28,10 @@ NO_USER_COMMANDS = 'This completer does not define any commands.'
 MESSAGE_POLL_TIMEOUT = 10
 
 
+class CompletionsChanged( Exception ):
+  pass # pragma: no cover
+
+
 class Completer( metaclass = abc.ABCMeta ):
   """A base class for all Completers in YCM.
 
@@ -121,6 +125,14 @@ class Completer( metaclass = abc.ABCMeta ):
   fields in DetailCandidates() which is called after the filtering is done. See
   python_completer.py for an example.
 
+  You can also support 'delayed' detailing of candidates. The way this works is
+  that you must add a 'resolve' key to the candidate's 'extra_data' which will
+  be round-tripped to the client. The client passes the 'resolve' key back to us
+  in the /resolve_completion request and your completer must override the
+  DetailSingleCandidate() method, which is passed the full list of candidates
+  from the cache and the resolve key supplied. This method must return the
+  candidate, fully detailed. See the LanguageServerCompleter for an example.
+
   If the completer wants to use extra confs, it should implement Language()
   function as well, which returns a string that identifies the language in
   user's .ycmd_extra_conf.py file.
@@ -198,6 +210,11 @@ class Completer( metaclass = abc.ABCMeta ):
 
     self._completions_cache = CompletionsCache()
     self._max_candidates = user_options[ 'max_num_candidates' ]
+    self._max_candidates_to_detail = user_options[
+      'max_num_candidates_to_detail' ]
+
+    LOGGER.info( f"Completion config: { self._max_candidates }, detailing "
+                 f"{ self._max_candidates_to_detail } candiates" )
 
 
   # It's highly likely you DON'T want to override this function but the *Inner
@@ -286,7 +303,31 @@ class Completer( metaclass = abc.ABCMeta ):
     candidates = self._GetCandidatesFromSubclass( request_data )
     candidates = self.FilterAndSortCandidates( candidates,
                                                request_data[ 'query' ] )
+
     return self.DetailCandidates( request_data, candidates )
+
+
+  def ShouldDetailCandidateList( self, candidates ):
+    if self._max_candidates_to_detail < 0:
+      return True
+
+    if len( candidates ) < self._max_candidates_to_detail:
+      return True
+
+    return False
+
+
+  def ResolveCompletionItem( self, request_data ):
+    candidates = self._completions_cache.GetCompletionsIfCacheValid(
+      request_data )
+
+    if not candidates:
+      raise CompletionsChanged( 'Resolve request must not change request data' )
+
+    return self.DetailSingleCandidate( request_data,
+                                       candidates,
+                                       request_data[ 'resolve' ] )
+
 
 
   def _GetCandidatesFromSubclass( self, request_data ):
@@ -299,6 +340,13 @@ class Completer( metaclass = abc.ABCMeta ):
     raw_completions = self.ComputeCandidatesInner( request_data )
     self._completions_cache.Update( request_data, raw_completions )
     return raw_completions
+
+
+  def DetailSingleCandidate( self,
+                             request_data,
+                             candidates,
+                             to_resolve ): # pragma: no cover
+    raise RuntimeError( "Delayed detail candidate not implemented" )
 
 
   def DetailCandidates( self, request_data, candidates ):
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -937,6 +937,13 @@ class LanguageServerCompleter( Completer ):
     pass # pragma: no cover
 
 
+  # Resolve all completion items up-front
+  RESOLVE_ALL = True
+
+  # Don't resolve any completion items, but prepare them for resolve
+  RESOLVE_NONE = False
+
+
   def __init__( self, user_options, connection_type = 'stdio' ):
     super().__init__( user_options )
 
@@ -1263,10 +1270,11 @@ class LanguageServerCompleter( Completer ):
     # should be based on ycmd's version of the insertion_text. Fortunately it's
     # likely much quicker to do the simple calculations inline rather than a
     # series of potentially many blocking server round trips.
-    return ( self._CandidatesFromCompletionItems( items,
-                                                  False, # don't do resolve
-                                                  request_data ),
-             is_incomplete )
+    return ( self._CandidatesFromCompletionItems(
+              items,
+              LanguageServerCompleter.RESOLVE_NONE,
+              request_data ),
+            is_incomplete )
 
 
   def _GetCandidatesFromSubclass( self, request_data ):
@@ -1287,7 +1295,9 @@ class LanguageServerCompleter( Completer ):
 
   def DetailCandidates( self, request_data, completions ):
     if not self._resolve_completion_items:
-      # We already did all of the work.
+      return completions
+
+    if not self.ShouldDetailCandidateList( completions ):
       return completions
 
     # Note: _CandidatesFromCompletionItems does a lot of work on the actual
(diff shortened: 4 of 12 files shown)


puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 0 of 2 LGTMs obtained

___
*[ycmd/completers/completer.py, line 321 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGsxZtu9SHh6MBgD6Pl:-MGsxZtu9SHh6MBgD6Pm:b-bth7nv) ([raw file](https://github.com/ycm-core/ycmd/blob/d9d2415bbd68d364c743a0fc2dadbba36234b4cb/ycmd/completers/completer.py#L321)):*
> ```Python
> 
> 
>   def DetailCandidate( self, request_data, candidates, candidate ):
> ```

last parameter is `to_resolve`. document this method in the headers comment.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1275 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGsy3eJ25w9QfeHM8cH:-MGsy3eJ25w9QfeHM8cI:b-5jrn9y) ([raw file](https://github.com/ycm-core/ycmd/blob/d9d2415bbd68d364c743a0fc2dadbba36234b4cb/ycmd/completers/language_server/language_server_completer.py#L1275)):*
> ```Python
>         # Store the actual item in the extra_data area of the completion item.
>         # We'll use this later to do the resolve.
>         extra_data[ 'item' ] = item
> ```

need to consider popping this off when sending to the actual client. this is currently only needed when retrieving data from the cache, but we're currently actually round-tripping this data all the way to the client and back for unresolved candidates. Strictly we only need to round-trip the index ('resolve').

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/handlers.py, line 154 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGt-7HDA8zh6mPMT1nv:-MGt-7HDA8zh6mPMT1nw:b-a1i2no) ([raw file](https://github.com/ycm-core/ycmd/blob/d9d2415bbd68d364c743a0fc2dadbba36234b4cb/ycmd/handlers.py#L154)):*
> ```Python
>   completion = []
>   try:
>     completion = [ completer.ResolveCompletionItem( request_data ) ]
> ```

First question, how does the API look like? Is it "I give you *a candidate* and you return me a list of resolved candidates"?

I assume you wanted to reuse as much code as possible on the client side, so you made this return the same thing as `/completion` request?

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/handlers.py, line 154 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGt-7HDA8zh6mPMT1nv:-MGtPBil8YDWOkxqTGuB:bqk41q8) ([raw file](https://github.com/ycm-core/ycmd/blob/d9d2415bbd68d364c743a0fc2dadbba36234b4cb/ycmd/handlers.py#L154)):*

First question, how does the API look like? Is it "I give you *a candidate* and you return me a list of resolved candidates"?

I assume you wanted to reuse as much code as possible on the client side, so you made this return the same thing as `/completion` request?
</blockquote></details>

You know me so well....

So initially what I wanted to do was have the /resolve_completion request return a FULL set of candidates with the requested candidate now resolved. I implemented that but there was just no useful way to use that with Vim (you'll see why). So in the end I decided to have the resolve_completion request return a _single_ (now resolved) candidate but using the exact same response API as the /completion request. This allows the client to basically re-use it's CompletionRequest object and just pull out the data it needs from the guaranteed-to-be-1 returned candidate. We can also re-use the mechanism for reporting and handling errors, etc.

We redundantly return the (wrong) completion_start_column, but I think we can live with that.

puremourning (commented):
added a few tests. I'll check the coverage report after 

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473#-:-MGte_JX2_rL1HvlCEaW:b-khp0h2)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/handlers.py, line 154 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGt-7HDA8zh6mPMT1nv:-MGtevjG3fRa5w0kwcrS:b-m0xysq) ([raw file](https://github.com/ycm-core/ycmd/blob/d9d2415bbd68d364c743a0fc2dadbba36234b4cb/ycmd/handlers.py#L154)):*
I have to argue that the API is confusing, from a very high level perspective.

> I have you *one* candidate. ONE. Why are you returning a string now? Is there some very weird circumstance where you can return more than one? No? Then why the hell is it a list??

I'm not saying that your arguments are wrong. I'm also not sure if the hypothetical confusion is a valid concern. I do know that the above line of thinking is what made me stop reviewing code until I get a confirmation that I'm correctly ~~confused~~ reading the code.

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/handlers.py, line 154 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGt-7HDA8zh6mPMT1nv:-MGtwBWgCZRy46qHi1Hg:b-8b2p7s) ([raw file](https://github.com/ycm-core/ycmd/blob/d9d2415bbd68d364c743a0fc2dadbba36234b4cb/ycmd/handlers.py#L154)):*

I have to argue that the API is confusing, from a very high level perspective.

> I have you *one* candidate. ONE. Why are you returning a string now? Is there some very weird circumstance where you can return more than one? No? Then why the hell is it a list??

I'm not saying that your arguments are wrong. I'm also not sure if the hypothetical confusion is a valid concern. I do know that the above line of thinking is what made me stop reviewing code until I get a confirmation that I'm correctly ~~confused~~ reading the code.
</blockquote></details>

sure. we can change it easily. I've already set `completion_start_column` to `None` so it's going to be easy to factor out the redundant info and the redundant list into something more sensible.

My process is always make the smallest change, the jankiest dumbest change possible to get something working, then iterate from there.

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGtxCNp0PP2-ibO4Vt4:-MGtxCNp0PP2-ibO4Vt5:b-uygy0n) (no related file):*
Need to update the API docs and (maybe) the example client

---

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGtxIgG2zIIyt8dM63Z:-MGtxIgG2zIIyt8dM63_:b-3peawy) (no related file):*
API discussion. I'm thinking:

```
post: resolve_completion
request_data - must be the exact same as the /completion request with the 'resolve' key added (as copied from the item to resolve's `extra_data`)

result: {
  'completion': { completion item },
  'errors': [ ... errors ]
}
```

This result structure is very similar to the /completion structure except:

* single completion not a list is returned
* no start column returned

Seem ok?

---

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGtxIgG2zIIyt8dM63Z:-MGu-m5k-3U_LHg14xqp:b-h5qq0v) (no related file):*

API discussion. I'm thinking:

```
post: resolve_completion
request_data - must be the exact same as the /completion request with the 'resolve' key added (as copied from the item to resolve's `extra_data`)

result: {
  'completion': { completion item },
  'errors': [ ... errors ]
}
```

This result structure is very similar to the /completion structure except:

* single completion not a list is returned
* no start column returned

Seem ok?
</blockquote></details>

That does seem fine.

---

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 0 of 2 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGtxIgG2zIIyt8dM63Z:-MGu3oKv4EMJdhgYR0zC:bb6un1j) (no related file):*

That does seem fine.
</blockquote></details>

Done

---
___
*[ycmd/completers/language_server/language_server_completer.py, line 1275 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGsy3eJ25w9QfeHM8cH:-MGu53nqDJi1NRIClzX9:b-6bql1c) ([raw file](https://github.com/ycm-core/ycmd/blob/d9d2415bbd68d364c743a0fc2dadbba36234b4cb/ycmd/completers/language_server/language_server_completer.py#L1275)):*

need to consider popping this off when sending to the actual client. this is currently only needed when retrieving data from the cache, but we're currently actually round-tripping this data all the way to the client and back for unresolved candidates. Strictly we only need to round-trip the index ('resolve').
</blockquote></details>

This is annoying to do. It requires making a new copy of all the candidates but popping this off the copy. It might be simpler to just leave it and optimise the data transfer later

Ideally, now that we have this setup with resolving from the cache, we should allow the completer implentations to store completer-specific data alongside the completion results in the cache, e.g. the cache actually stores tuple of `( candidate, internal_data )` rather than just `candidate`.

Maybe I'll do that, but it might be a bit too much refactoring for this patch.

bstaletic (changes requested):
Windows doesn't seem to like your tests.

```
FAILED ycmd/tests/java/get_completions_test.py::GetCompletions_MoreThan10_NoResolve_ThenResolve_test[]
FAILED ycmd/tests/java/get_completions_test.py::GetCompletions_FewerThan10_Resolved_test[]
```

Both are missing the docstring part of `detailed_info`.

> Reviewed 5 of 8 files at r2, 8 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473#-:-MGx2iT2C9z4FZlpmUzl:b50rm4t)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[docs/openapi.yml, line 483 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGwuEK4AnwRuNdtX3Ra:-MGwuEK4AnwRuNdtX3Rb:b-rgtgjl) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/docs/openapi.yml#L483)):*
> ```YAML
>               Additional documentation/information to be displayed
>               alongside (after) information in `detailed_info`.
>           fixits:
> ```

Wow, was this really not documented??
___
*[ycmd/completers/completer.py, line 329 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGwwdFT1vkuDquV1TU4:-MGwwdFT1vkuDquV1TU5:b52k01s) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/completer.py#L329)):*
> ```Python
> 
> 
>   def DetailCandidate( self, request_data, candidates, to_resolve ):
> ```

We already have `DetailCandidates()` (plural). Introducing `DetailCandidate()` (singular) could be confusing.

Maybe `ResoliveCandidate()`? Naming is hard.
___
*[ycmd/completers/completer.py, line 331 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx28hf9x3LJeJ8PLIm:-MGx28hf9x3LJeJ8PLIn:b-e4zux8) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/completer.py#L331)):*
> ```Python
>   def DetailCandidate( self, request_data, candidates, to_resolve ):
>     raise RuntimeError( # pragma: no cover
>       "Delayed detail candidate not implemented" ) # pragma: no cover
> ```

I think you should be able to put `# pragma: no cover` on the `def DetailCandidate()` line and apply it to the whole block.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1275 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGsy3eJ25w9QfeHM8cH:-MGwydF-3H_yofzdDxWu:bl6huz9) ([raw file](https://github.com/ycm-core/ycmd/blob/d9d2415bbd68d364c743a0fc2dadbba36234b4cb/ycmd/completers/language_server/language_server_completer.py#L1275)):*

This is annoying to do. It requires making a new copy of all the candidates but popping this off the copy. It might be simpler to just leave it and optimise the data transfer later

Ideally, now that we have this setup with resolving from the cache, we should allow the completer implentations to store completer-specific data alongside the completion results in the cache, e.g. the cache actually stores tuple of `( candidate, internal_data )` rather than just `candidate`.

Maybe I'll do that, but it might be a bit too much refactoring for this patch.
</blockquote></details>

I'm fine with this being a separate pull request.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1251 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx-H251yRQg7TC8tby:-MGx-H251yRQg7TC8tbz:b-vbusoe) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1251)):*
> ```Python
>     # likely much quicker to do the simple calculations inline rather than a
>     # series of potentially many blocking server round trips.
>     return ( self._CandidatesFromCompletionItems(
> ```

Why is there a pair of parenthesis around `self._Candidates...`?
___
*[ycmd/completers/language_server/language_server_completer.py, line 1334 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx-bspCTtq6Z95Z26w:-MGx-bspCTtq6Z95Z26x:b3k039o) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1334)):*
> ```Python
> 
>   RESOLVE_ALL = True
>   RESOLVE_NONE = False
> ```

Why place this, supposedly, "global" and const data in this place? "Vicinity of `_CandidatesFromCompletionItems()` is obvious, but these constants are passed in as a parameter. Once I start reading `_CandidatesFromCompletionItems()`, I don't care about `RESOLVE_ALL`/`RESOLVE_NONE`, but about `to_resolve` instead.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1338 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx0THEABDT1A50xIgC:-MGx0THEABDT1A50xIgD:b-pipnv7) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1338)):*
> ```Python
>   def _CandidatesFromCompletionItems( self,
>                                       items,
>                                       to_resolve,
> ```

While we're at it, maybe `to_resolve` isn't the best name either. At first, I was expecting this to be the item that needs to be resolved, which was odd.

Maybe that's just me though. I won't block on this.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1383 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx1B6JEY0QdnSQeglK:-MGx1B6JEY0QdnSQeglL:b-mit6f0) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1383)):*
> ```Python
>     for idx, item in enumerate( items ):
>       this_tem_is_resolved = item.get( '_resolved', False )
>       resolve_this_item = to_resolve == LanguageServerCompleter.RESOLVE_ALL
> ```

I don't see a reason for `resolve_this_item`, since it will always have the same value as `to_resolve`.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1400 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx1YCw45XJDHufKp1q:-MGx1YCw45XJDHufKp1r:bjwm0ve) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1400)):*
> ```Python
>         continue
> 
>       if ( to_resolve == LanguageServerCompleter.RESOLVE_NONE and
> ```

Isn't this just `not to_resolve`? I still think that a better name for `to_resolve` would avoid the need for these explicit checks and improve readability.
___
*[ycmd/tests/test_utils.py, line 402 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGwv5rkC8mUUWOaPbUb:-MGwv5rkC8mUUWOaPbUc:bb9liy0) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/tests/test_utils.py#L402)):*
> ```Python
>     while True:
>       try:
>         return test( *args, **kwargs )
> ```

Why was this change needed?
___
*[ycmd/tests/java/get_completions_test.py, line 555 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGwvjCJ7wDQEE4o8sQy:-MGwvjCJ7wDQEE4o8sQz:b2lh9tq) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/tests/java/get_completions_test.py#L555)):*
> ```Python
> @SharedYcmd
> def GetCompletions_MoreThan10_NoResolve_ThenResolve_test( app ):
>   request, response = RunTest( app, {
> ```

Why not (conditionally?) extend `RunTest()` instead of returning a request-response tuple and continuing the test below?

puremourning (commented):
> Windows doesn't seem to like your tests.

May be missing some retries. I'll check.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473#-:-MGxG0RF0lQRl29snXbu:b-iqj3ou)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[docs/openapi.yml, line 483 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGwuEK4AnwRuNdtX3Ra:-MGxDY1k5J98qAfOCa6t:bp2st3h) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/docs/openapi.yml#L483)):*

Wow, was this really not documented??
</blockquote></details>

it wasn't :)
___
*[ycmd/completers/completer.py, line 329 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGwwdFT1vkuDquV1TU4:-MGxD_b76xtZudaQIh08:b-qkjh5g) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/completer.py#L329)):*

We already have `DetailCandidates()` (plural). Introducing `DetailCandidate()` (singular) could be confusing.

Maybe `ResoliveCandidate()`? Naming is hard.
</blockquote></details>

let's use `DetailSingleCandidate()` that's clear.
___
*[ycmd/completers/completer.py, line 331 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx28hf9x3LJeJ8PLIm:-MGxDeogBQY7MmNWWJot:b-sith4z) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/completer.py#L331)):*

I think you should be able to put `# pragma: no cover` on the `def DetailCandidate()` line and apply it to the whole block.
</blockquote></details>

ah ok ta
___
*[ycmd/completers/language_server/language_server_completer.py, line 1334 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx-bspCTtq6Z95Z26w:-MGxDpdZEjOsmB76ifsa:b-cr3xr2) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1334)):*

Why place this, supposedly, "global" and const data in this place? "Vicinity of `_CandidatesFromCompletionItems()` is obvious, but these constants are passed in as a parameter. Once I start reading `_CandidatesFromCompletionItems()`, I don't care about `RESOLVE_ALL`/`RESOLVE_NONE`, but about `to_resolve` instead.
</blockquote></details>

this is a bit of a hangover from the original implementation that I used, but I preferred the self-documenting nature of "RESOLVE_ALL" vs "True". I found the original code `if not resolve and ...` confusing and spend 10 minutes or so going over this code and stepping through in the debugger to understand how it worked (I realise that's entirely my own fault!), so I found this a clearer way to read the code.

I can certainly revert it if you prefer.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1338 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx0THEABDT1A50xIgC:-MGxELCn9wER8r5XQbZp:b1jnm2) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1338)):*

While we're at it, maybe `to_resolve` isn't the best name either. At first, I was expecting this to be the item that needs to be resolved, which was odd.

Maybe that's just me though. I won't block on this.
</blockquote></details>

originally I had `to_resolve`:

* `True` = resolve_all
* `False` = resolve_none
* `an int()` = the index to resolve.

Again this is no longer used as we're only resolving one, so I can revert if you prefer.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1383 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx1B6JEY0QdnSQeglK:-MGxE_x-1oHNDMBDFCJK:bt1nrvg) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1383)):*

I don't see a reason for `resolve_this_item`, since it will always have the same value as `to_resolve`.
</blockquote></details>

as above, this is a refactor of a refactor, so we can take it back
___
*[ycmd/completers/language_server/language_server_completer.py, line 1400 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx1YCw45XJDHufKp1q:-MGxEfWH4JqrFSndEF7s:bb7oryy) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1400)):*

Isn't this just `not to_resolve`? I still think that a better name for `to_resolve` would avoid the need for these explicit checks and improve readability.
</blockquote></details>

sure
___
*[ycmd/tests/test_utils.py, line 402 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGwv5rkC8mUUWOaPbUb:-MGxEibY8or_84iIxbS7:byxei0w) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/tests/test_utils.py#L402)):*

Why was this change needed?
</blockquote></details>

In the completion tests I'm having `RunTest` return `request, response` so that it can pull out the `resolve` item from a completion test and resolve it using the same request object.

This makes that actually work when the `RunTest` has this decorator applied.

i.e. this makes the decorator work properly, returning the function's return value rather than `None` 
___
*[ycmd/tests/java/get_completions_test.py, line 555 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGwvjCJ7wDQEE4o8sQy:-MGxF0OE8NuglQO32RNS:b-5g91c) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/tests/java/get_completions_test.py#L555)):*

Why not (conditionally?) extend `RunTest()` instead of returning a request-response tuple and continuing the test below?
</blockquote></details>

That just seemed a bit messy to me (and a bit of an inner-system-effect). We get close to defining an internal language within the dict passed to `RunTest` to say `if this then that` etc. which I'm against.

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[ycmd/completers/language_server/language_server_completer.py, line 1251 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx-H251yRQg7TC8tby:-MGxGQk-BVRqw-L0h0-s:b19kxca) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1251)):*

Why is there a pair of parenthesis around `self._Candidates...`?
</blockquote></details>

this method returns a tuple `( candidates, is_incomplete )`

bstaletic (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[ycmd/completers/completer.py, line 329 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGwwdFT1vkuDquV1TU4:-MGxHG9IEPxtLtKkcWrS:b4ruo3t) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/completer.py#L329)):*

let's use `DetailSingleCandidate()` that's clear.
</blockquote></details>

That sounds good.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1334 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx-bspCTtq6Z95Z26w:-MGxIzYa9Fys24OACNFN:blp3sgd) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1334)):*
> I preferred the self-documenting nature of "RESOLVE_ALL" vs "True".

On the side of the called of `_CandidatesFromCompletionItems()`, I completely agree. I'm not convinced that this explicit comparison of `to_resolve` with `LanguageServerCompleter.RESOLVE_{ALL,NONE}` is easier to read. To me `if not resolve_candidates` (yeah, not just `resolve`) reads find and you don't end up splitting every line into two rows.

It's less code, it fits on one line and I don't think `if not resolve_candidates` is unclear.

If you still disagree with that last point, which is definitely the most important, let's leave it alone.

However, I still find the location of the `RESOLVE_ALL` an d`RESOLVE_NONE` variables odd. I'd still keep them, for the sake of the caller side, but I'd place them near `__init__`, where other stuff is initialized.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1383 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx1B6JEY0QdnSQeglK:-MGxKoSg0M3MG3-d2GFk:btdkd8d) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1383)):*

as above, this is a refactor of a refactor, so we can take it back
</blockquote></details>

`resolve_this_item` is really just an alias, so I'd remove it and use `to_resolve`.

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[ycmd/completers/language_server/language_server_completer.py, line 1334 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx-bspCTtq6Z95Z26w:-MGxNwVj8gN3uT-Kccq7:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1334)):*

I do agree :)
</blockquote></details>

Done.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1383 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx1B6JEY0QdnSQeglK:-MGxNslL-8XXR3XiKDeH:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1383)):*

sure, agreed.
</blockquote></details>

Done.
___
*[ycmd/completers/language_server/language_server_completer.py, line 1400 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx1YCw45XJDHufKp1q:-MGxNti1De6z6Mu5sXGT:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1400)):*

sure
</blockquote></details>

Done.

bstaletic (commented):
Mostly LGTM, with the two unresolved comments in `completer.py` remaining.

After the last change, I definitely like more how the `language_server_completer.py` changes look now.

> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473#-:-MGxRjUl-I848x5pElT1:b-w5huqr)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[ycmd/completers/language_server/language_server_completer.py, line 1251 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx-H251yRQg7TC8tby:-MGxQeCpCf33UaS7OwMg:b766aqb) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/language_server/language_server_completer.py#L1251)):*

this method returns a tuple `( candidates, is_incomplete )`
</blockquote></details>

My bad. I didn't catch that, even thought I remember debating with Michel about this thing.
___
*[ycmd/completers/language_server/language_server_completer.py, line 930 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGxRSzo5pReH7BcUE6y:-MGxRSzo5pReH7BcUE6z:b-u9lqwt) ([raw file](https://github.com/ycm-core/ycmd/blob/29b50206a63e94e0c9c6a5ae0295e095b2e31fc9/ycmd/completers/language_server/language_server_completer.py#L930)):*
> ```Python
>   RESOLVE_ALL = True
> 
>   # Don't resolve any completion items, but prepare them for resolve
> ```

... for resolving?

... for resolution?

Bleh... it's fine!

bstaletic (commented):
Winblows still borked. Look at all those CARRIAGE RETURNS, Ben!

```
completer response: {'completion_start_column': 7,
 'completions': [{'detailed_info': 'useAString(String s) : void\n'
                                   '\n'
                                   'Multiple lines of description here.\r\n'
                                   '\r\n'
                                   ' *  **Parameters:**\r\n'
                                   '    \r\n'
                                   '     *  **s** a string',
                  'extra_menu_info': 'MethodsWithDocumentation.useAString(String '
                                     's) : void',
                  'insertion_text': 'useAString',
                  'kind': 'Method',
                  'menu_text': 'useAString(String s) : void'}],
 'errors': []}
```

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473#-:-MGxVrcSBBoeIlUJwgyF:bl15ins)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/completers/completer.py, line 329 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGwwdFT1vkuDquV1TU4:-MGxmMPn9qLkNa2Llo3j:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/completer.py#L329)):*

That sounds good.
</blockquote></details>

Done.
___
*[ycmd/completers/completer.py, line 331 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGx28hf9x3LJeJ8PLIm:-MGxmMn3BEjLE5Rwrqlj:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/1f9827666b7f7554874dd1b06f3d495fc1a7ba54/ycmd/completers/completer.py#L331)):*

ah ok ta
</blockquote></details>

Done.

bstaletic (approved):
:lgtm_strong:

> Reviewed 1 of 1 files at r5, 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473#-:-MGxpa_mCmi9QcTUDEfp:b-jbsf8z)* status: 1 of 2 LGTMs obtained

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/tests/java/get_completions_test.py, line 631 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGyEyvC1T1QPaNngMmF:-MGyEyvD3r8PgYcDNofZ:b-7tjwgb) ([raw file](https://github.com/ycm-core/ycmd/blob/47c211dce185a3c9b3fc5987cee53f2a4af82a4c/ycmd/tests/java/get_completions_test.py#L631)):*
> ```Python
> 
> @SharedYcmd
> @WithRetry
> ```

This has no effect. It's the shared ycmd, so the decorator only returns a reference to the already set up `shared_app`.

Even with isolated ycmd, I'm pretty sure it would have made zero difference, because pytest magic. I once tested what it is doing and whether the following will keep reinitializing the `isolated_app`:

```python
@WithRetry
@IsolatedYcmd()
def some_test():
```

To my surprise, the above only initialized the `isolated_app` *once*.

bstaletic (approved):
Woohoo! That worked. I love this PR. :lgtm: 

> Reviewed 1 of 1 files at r7, 3 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473#-:-MGy_yb59PG3YiZfauc9:b-ony6r5)* status: 1 of 2 LGTMs obtained

bstaletic (approved):
> Reviewed 4 of 6 files at r9.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/completers/completer.py, line 214 at r9](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MH-GIJL84T2eyKl7ffl:-MH-GIJMAvXzw6fDoUda:byxbgj8) ([raw file](https://github.com/ycm-core/ycmd/blob/f5ca16b82e9c6ed8a27bad2f79b511e7507e5ffd/ycmd/completers/completer.py#L214)):*
> ```Python
>     self._max_candidates = user_options[ 'max_num_candidates' ]
>     self._max_candidates_to_detail = user_options.get(
>       'max_num_candidates_to_detail', -1 )
> ```

Any particular reason for introducing `max_num_candidates_to_detail`?

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/completers/completer.py, line 214 at r9](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MH-GIJL84T2eyKl7ffl:-MH0QKtw4rTzMjRiTBKm:b-194157) ([raw file](https://github.com/ycm-core/ycmd/blob/f5ca16b82e9c6ed8a27bad2f79b511e7507e5ffd/ycmd/completers/completer.py#L214)):*

Any particular reason for introducing `max_num_candidates_to_detail`?
</blockquote></details>

Yah. It’s used by the client to enable this functionality. 

We have to leave it off by default for backward compatibility, and the client can only use this feature if completeopt contains popup and if vim supports it. 

So this is like a capability. But it’s also a tunable.

bstaletic (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/completers/completer.py, line 214 at r9](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MH1fOUN6ck6ybDOaao-:-MH1fOUN6ck6ybDOaao0:b-rmoa3h) ([raw file](https://github.com/ycm-core/ycmd/blob/f5ca16b82e9c6ed8a27bad2f79b511e7507e5ffd/ycmd/completers/completer.py#L214)):*
> ```Python
>     self._max_candidates = user_options[ 'max_num_candidates' ]
>     self._max_candidates_to_detail = user_options.get(
>       'max_num_candidates_to_detail', -1 )
> ```

One last thing. Why is this a `.get()` instead of existing in the `default_settings.json`?

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/completers/completer.py, line 214 at r9](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MH1fOUN6ck6ybDOaao-:-MH2o3k2EgSoAwYBatg0:br2oxmh) ([raw file](https://github.com/ycm-core/ycmd/blob/f5ca16b82e9c6ed8a27bad2f79b511e7507e5ffd/ycmd/completers/completer.py#L214)):*

One last thing. Why is this a `.get()` instead of existing in the `default_settings.json`?
</blockquote></details>

ugh erm. *ahem* I was half hoping you wouldn't notice that.... 

so the reason is that the client reads default_settings.json and if we put it in there, it ends up always using it in preference to the value the client calculated (i.e. depending on completeopt).

It's probably better to refactor the client to not do that, but for now this works well enough.

bstaletic (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/completers/completer.py, line 214 at r9](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MH1fOUN6ck6ybDOaao-:-MH5GBOY3JpbQIQ7zvDE:b-gxv9yq) ([raw file](https://github.com/ycm-core/ycmd/blob/f5ca16b82e9c6ed8a27bad2f79b511e7507e5ffd/ycmd/completers/completer.py#L214)):*

ugh erm. *ahem* I was half hoping you wouldn't notice that.... 

so the reason is that the client reads default_settings.json and if we put it in there, it ends up always using it in preference to the value the client calculated (i.e. depending on completeopt).

It's probably better to refactor the client to not do that, but for now this works well enough.
</blockquote></details>

You know I don't like that. If we start hiding `user_options`,  we'll end up in a huge mess one day and we won't even know it.

puremourning (commented):
:lgtm:

> Reviewed 1 of 8 files at r2, 5 of 8 files at r3, 2 of 6 files at r9, 1 of 1 files at r10, 2 of 2 files at r11, 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473#-:-MHcQeD_7znY8B5QZlz1:bnfp4nl)* status: :shipit: complete! 2 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/completers/completer.py, line 214 at r9](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MH1fOUN6ck6ybDOaao-:-MHcPWb644soPLyu2nia:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/f5ca16b82e9c6ed8a27bad2f79b511e7507e5ffd/ycmd/completers/completer.py#L214)):*

You know I don't like that. If we start hiding `user_options`,  we'll end up in a huge mess one day and we won't even know it.
</blockquote></details>

Done.
___
*[ycmd/tests/java/get_completions_test.py, line 631 at r7](https://reviewable.io/reviews/ycm-core/ycmd/1473#-MGyEyvC1T1QPaNngMmF:-MHcPl9f-jaWRQJE2ICK:b5tm368) ([raw file](https://github.com/ycm-core/ycmd/blob/47c211dce185a3c9b3fc5987cee53f2a4af82a4c/ycmd/tests/java/get_completions_test.py#L631)):*

This has no effect. It's the shared ycmd, so the decorator only returns a reference to the already set up `shared_app`.

Even with isolated ycmd, I'm pretty sure it would have made zero difference, because pytest magic. I once tested what it is doing and whether the following will keep reinitializing the `isolated_app`:

```python
@WithRetry
@IsolatedYcmd()
def some_test():
```

To my surprise, the above only initialized the `isolated_app` *once*.
</blockquote></details>

yep, good point.
Done.

puremourning (commented):
:lgtm:

> Reviewed 1 of 1 files at r13.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473#-:-MHcwI5xBgTuGAimbKnG:bnfp4nl)* status: :shipit: complete! 2 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic (approved):
> Reviewed 1 of 1 files at r10, 2 of 2 files at r11, 1 of 1 files at r13, 1 of 1 files at r14.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1473)* status: 1 of 2 LGTMs obtained (and 1 stale)

pawansingh00:
Is there any config that we can use if we still want to Resolve the completions upfront?

puremourning:
https://github.com/ycm-core/YouCompleteMe#the-gycm_max_num_candidates_to_detail-option

pawansingh00:
@puremourning  This is the blazing fast reply I have ever got, Thank you so much.
Tried this with **-1** also only getting this 

  "insertion_text": "arraycopy",
  "extra_menu_info": "System.arraycopy(Object arg0, int arg1, Object arg2, int arg3, int arg4) : void",
  "menu_text": "arraycopy(Object arg0, int arg1, Object arg2, int arg3, int arg4) : void",
  "detailed_info": "arraycopy(Object arg0, int arg1, Object arg2, int arg3, int arg4) : void\n\n",
  "kind": "Method"
  
  
 With older version I used to get more details in -- `detailed_info`
 
 I am using it outside of - vim

puremourning:
Please raise a new issue with steps to repro, expected and actual output, and an example project , including the full message trace. comments on very old PRs aren't a good place to discuss issues.

pawansingh00:
ok sure, will raise a new issue, wanted to be sure before raising that if I am not doing anything wrong.
Thanks for the response.

PR #1475, opened 2020-09-17
[READY] Do not set architecture on Windows when using Ninja

### LINKED ISSUE #1467: Bug using ninja on Windows
Building ycm with ninja on Windows is failing due to unrecognized platform `x64`.

To reproduce bug, run `python build.py --ninja` on Windows.

Seems, that issue is [here](https://github.com/ycm-core/ycmd/blob/a24204e8382d0660a519f88b59c67026f453c085/build.py#L498). Since default value of `args.msvc` is 16, that `if` statment will be true when called in a way I mentioned before. I suppose, that additonal arguments `-A x64` should be only added if ninja is not selected as a build tool. Current workaround is to build ycmd with `python build.py --ninja --msvc 14`, since there `-A x64` won't be added to arguments.

bstaletic:
Passing `-A` flag to CMake, in combination with `-G Ninja` errors with:

```
CMake Error at CMakeLists.txt:26 (project):
  Generator

    Ninja

  does not support platform specification, but platform

    x64

  was specified.
```

Fixes #1467

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1475)

### DIFF
--- a/build.py
+++ b/build.py
@@ -495,7 +495,7 @@ def GetCmakeCommonArgs( args ):
   cmake_args = [ '-G', GetGenerator( args ) ]
 
   # Set the architecture for the Visual Studio 16 generator.
-  if OnWindows() and args.msvc == 16:
+  if OnWindows() and args.msvc == 16 and not args.ninja:
     arch = 'x64' if IS_64BIT else 'Win32'
     cmake_args.extend( [ '-A', arch ] )
 


puremourning (approved):
were you able to test this ? honestly, I'm not that bothered. just curious.

:lgtm_strong: 

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1475#-:-MHQaJgS02VSRygwndKG:b-d9h33z)* status: 1 of 2 LGTMs obtained

bstaletic:
I ran it on azure a few times. It does seem to work. That's where I've copied the error from.

:lgtm:

PR #1476, opened 2020-09-17
[READY] Resolve a possible deadlock in LSP poll thread

### LINKED ISSUE #1458: HandleNotificationInPollThread can block, leading to deadlocks
@theli-ua's gist with all the threads shown: https://gist.github.com/theli-ua/13bef2a778252eb6c9dd35d44aa6b3cb
Reader thread: https://gist.github.com/theli-ua/13bef2a778252eb6c9dd35d44aa6b3cb#file-gistfile1-txt-L207-L213
The thread that entered _ReadMessages() is the one stuck at HandleNotificationInPollThread on line 1676
```
      with self._server_info_mutex:
        self._latest_diagnostics[ uri ] = params[ 'diagnostics' ]
```

The writes are stuck at `_server_stdin.write( data )`: https://gist.github.com/theli-ua/13bef2a778252eb6c9dd35d44aa6b3cb#file-gistfile1-txt-L99-L117

Final comment from @puremourning:

> Right. That’s the deadlock.
Handle notification in poll thread should not block. Good sleuthing

bstaletic:
This is the first experiment trying to fix #1458.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1476)

### DIFF
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -949,15 +949,19 @@ class LanguageServerCompleter( Completer ):
     #     separate thread and might call methods requiring us to synchronise the
     #     server's view of file state with our own. We protect from clobbering
     #     by doing all server-file-state operations under this mutex.
-    #   - There are certain events that we handle in the message pump thread.
-    #     These include diagnostics and some parts of initialization. We must
-    #     protect against concurrent access to our internal state (such as the
-    #     server file state, and stored data about the server itself) when we
-    #     are calling methods on this object from the message pump). We
-    #     synchronise on this mutex for that.
+    #   - There are certain events that we handle in the message pump thread,
+    #     like some parts of initialization. We must protect against concurrent
+    #     access to our internal state (such as the server file state, and
+    #     stored data about the server itself) when we are calling methods on
+    #     this object from the message pump). We synchronise on this mutex for
+    #     that.
     #   - We need to make sure that multiple client requests don't try to start
     #     or stop the server simultaneously, so we also do all server
     #     start/stop/etc. operations under this mutex
+    #   - Acquiring this mutex from the poll thread can lead to deadlocks.
+    #     Currently, this is avoided by using _latest_diagnostics_mutex to
+    #     access _latest_diagnostics, as that is the only resource shared with
+    #     the poll thread.
     self._server_info_mutex = threading.Lock()
     self.ServerReset()
 
@@ -1004,6 +1008,7 @@ class LanguageServerCompleter( Completer ):
     Implementations are required to call this after disconnection and killing
     the downstream server."""
     self._server_file_state = lsp.ServerFileStateStore()
+    self._latest_diagnostics_mutex = threading.Lock()
     self._latest_diagnostics = collections.defaultdict( list )
     self._sync_type = 'Full'
     self._initialize_response = None
@@ -1477,7 +1482,7 @@ class LanguageServerCompleter( Completer ):
       return responses.BuildDisplayMessageResponse(
           'Diagnostics are not ready yet.' )
 
-    with self._server_info_mutex:
+    with self._latest_diagnostics_mutex:
       diagnostics = list( self._latest_diagnostics[
           lsp.FilePathToUri( current_file ) ] )
 
@@ -1733,7 +1738,7 @@ class LanguageServerCompleter( Completer ):
     filepath = request_data[ 'filepath' ]
     uri = lsp.FilePathToUri( filepath )
     contents = GetFileLines( request_data, filepath )
-    with self._server_info_mutex:
+    with self._latest_diagnostics_mutex:
       if uri in self._latest_diagnostics:
         diagnostics = [ _BuildDiagnostic( contents, uri, diag )
                         for diag in self._latest_diagnostics[ uri ] ]
@@ -1845,7 +1850,7 @@ class LanguageServerCompleter( Completer ):
         # Ignore diagnostics for URIs we don't recognise
         LOGGER.exception( 'Ignoring diagnostics for unrecognized URI' )
         return
-      with self._server_info_mutex:
+      with self._latest_diagnostics_mutex:
         self._latest_diagnostics[ uri ] = params[ 'diagnostics' ]
 
 
@@ -2358,7 +2363,7 @@ class LanguageServerCompleter( Completer ):
 
     cursor_range_ls = lsp.Range( request_data )
 
-    with self._server_info_mutex:
+    with self._latest_diagnostics_mutex:
       # _latest_diagnostics contains LSP rnages, _not_ YCM ranges
       file_diagnostics = list( self._latest_diagnostics[
           lsp.FilePathToUri( request_data[ 'filepath' ] ) ] )


bstaletic (commented):
Quoting @theli-ua

```
Anton Romanov
@theli-ua
19:20
So what exactly should I test? Try repro before, make sure it deadlocks then try repro again?
Boris Staletic
@bstaletic
19:21
Yes and if you can't repro after pulling my PR in, I guess we're good.
Boris Staletic
@bstaletic
19:33
The tests on linux are passing at least, so I guess this isn't worse than it was.
Anton Romanov
@theli-ua
19:38
I think it still deadlocks. I get perpetual timeout trying to do ycmdebuginfo
Boris Staletic
@bstaletic
19:39
Is that with my pull request? If so, I can safely close it, because it didn't help.
Anton Romanov
@theli-ua
19:39
Actually, no, it did work, just took like 10 seconds
Boris Staletic
@bstaletic
19:40
Commands time out after 30 seconds. If it worked after 10, there's a performance problem somewhere, but it's not ycmd deadlocking.
Anton Romanov
@theli-ua
19:41
Yeah, it doesn't deadlock, thats correct
```

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1476#-:-MHSYCoP4wMSLNRJtCBM:bsetudo)* status: 0 of 2 LGTMs obtained

___
*[ycmd/completers/language_server/language_server_completer.py, line 953 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1476#-MHSY16q-LQ9gaNzQFG2:-MHSY16q-LQ9gaNzQFG3:b48qd80) ([raw file](https://github.com/ycm-core/ycmd/blob/02241459a77df3395788598d92e2258bf923610c/ycmd/completers/language_server/language_server_completer.py#L953)):*
> ```Python
>     #     by doing all server-file-state operations under this mutex.
>     #   - There are certain events that we handle in the message pump thread.
>     #     These include diagnostics and some parts of initialization. We must
> ```

This comment can be misleading, if we introduce a new mutex for diagnostics.

Ideas on how to change it are welcome.

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1476)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1476#-MHXpuKh2FL7sfod4fZw:-MHXpuKh2FL7sfod4fZx:btgx0gn) (no related file):*
what do we think? I mean tbh I'm happy to merge just in case it fixes it (I don't think it can make things worse assuming we don't hold 2 locks at once in any thread - we don't right ?). or is there some way we can gain more confidence?

---

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1476)* status: 0 of 2 LGTMs obtained (waiting on @theli-ua)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1476#-MHXpuKh2FL7sfod4fZw:-MHXzdQJBVCHrkFFeGTs:b-ck8cmr) (no related file):*
>  is there some way we can gain more confidence?

I have no way to confirm, as only @theli-ua could repro this deadlock.

> assuming we don't hold 2 locks at once in any thread - we don't right ?

Good question. I have just went back and traced all codepaths that end up taking the new mutex:

1. `GetDetailedDiagnostics()` is called directly from the `ycmd.handlers` and only ever takes one mutex to fetch the `_latest_diagnostics[uri]`.
2. `OnFileReadyToParse()`, again called directly. It does its job and then takes the new mutex to return the latest diagnostics.
3. `GetCodeActions()`, also called directly. Once again, safe use of the mutex at top level.
4. `HandleNotificationInPollThread()` is the only slightly tricky one. This ends up being the `_notification_handler` in the connection classes. The call stack looks something like this: `run()` -> `_ReadMessages()` -> `_DispatchMessage()` -> `HandleNotificationInPollThread()`

Therefore, I believe this won't make anything worse and it might solve the deadlock.

---

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1476)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @theli-ua)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1476#-MHXpuKh2FL7sfod4fZw:-MH_sD348KGntrnbkrEB:brxz26k) (no related file):*

>  is there some way we can gain more confidence?

I have no way to confirm, as only @theli-ua could repro this deadlock.

> assuming we don't hold 2 locks at once in any thread - we don't right ?

Good question. I have just went back and traced all codepaths that end up taking the new mutex:

1. `GetDetailedDiagnostics()` is called directly from the `ycmd.handlers` and only ever takes one mutex to fetch the `_latest_diagnostics[uri]`.
2. `OnFileReadyToParse()`, again called directly. It does its job and then takes the new mutex to return the latest diagnostics.
3. `GetCodeActions()`, also called directly. Once again, safe use of the mutex at top level.
4. `HandleNotificationInPollThread()` is the only slightly tricky one. This ends up being the `_notification_handler` in the connection classes. The call stack looks something like this: `run()` -> `_ReadMessages()` -> `_DispatchMessage()` -> `HandleNotificationInPollThread()`

Therefore, I believe this won't make anything worse and it might solve the deadlock.
</blockquote></details>

let's ship it then

---
___
*[ycmd/completers/language_server/language_server_completer.py, line 953 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1476#-MHSY16q-LQ9gaNzQFG2:-MH_sGBK5kYtsWZ8CQZ2:b-5vxcso) ([raw file](https://github.com/ycm-core/ycmd/blob/02241459a77df3395788598d92e2258bf923610c/ycmd/completers/language_server/language_server_completer.py#L953)):*

This comment can be misleading, if we introduce a new mutex for diagnostics.

Ideas on how to change it are welcome.
</blockquote></details>

just note that we must not hold the server_info_mutex in the poll thread as this can lead to deadlock ?

bstaletic (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1476#-:-MHb6CZ98PThWlMqy0nh:bnfp4nl)* status: 1 of 2 LGTMs obtained

___
*[ycmd/completers/language_server/language_server_completer.py, line 953 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1476#-MHSY16q-LQ9gaNzQFG2:-MHb6Ax6ErdZkWhDagp1:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/02241459a77df3395788598d92e2258bf923610c/ycmd/completers/language_server/language_server_completer.py#L953)):*

just note that we must not hold the server_info_mutex in the poll thread as this can lead to deadlock ?
</blockquote></details>

Done.

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1476#-:-MHbAi0vCj6b5pQFT4bX:bnfp4nl)* status: :shipit: complete! 2 of 2 LGTMs obtained

PR #1479, opened 2020-09-21
@Wentao[download-faster]: update download faster - sansna

sansna:
take advantage of pget to download several times faster

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1479)

### DIFF
--- a/build.py
+++ b/build.py
@@ -17,7 +17,9 @@ import sysconfig
 import tarfile
 from zipfile import ZipFile
 import tempfile
+from pget.down import Downloader
 
+DOWNLOAD_SPLIT = 8
 IS_64BIT = sys.maxsize > 2**32
 PY_MAJOR, PY_MINOR = sys.version_info[ 0 : 2 ]
 PY_VERSION = sys.version_info[ 0 : 3 ]
@@ -141,10 +143,9 @@ def CheckFileIntegrity( file_path, check_sum ):
 
 
 def DownloadFileTo( download_url, file_path ):
-  request = requests.get( download_url, stream = True )
-  with open( file_path, 'wb' ) as package_file:
-    package_file.write( request.content )
-  request.close()
+    downloader = Downloader(download_url, file_path, DOWNLOAD_SPLIT)
+    downloader.start()
+    downloader.wait_for_finish()
 
 
 def OnMac():


bstaletic:
Thanks for the pull request. I don't know what `pget` is, but it doesn't come from the standard library.

sansna:
> Thanks for the pull request. I don't know what `pget` is, but it doesn't come from the standard library.

So what should I do next?

puremourning (commented):
Thanks for sending a PR, but we don't want the installer to have dependencies outside of the standard libary. indeed we would love to ditch any of our third-party python dependencies.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1479#-:-MHk2jMa5cpM0f8Jt2t3:bd9zazi)* status: 0 of 2 LGTMs obtained

bstaletic:
I don't know what @puremourning thinks, but it could work with something like

```python
try:
  import pget
  has_pget = True
except ImportError:
  has_pget = False
```

puremourning (commented):
I guess I would have to understand the real benefits and what the actual problem is here. The additional complexity (and non-testability) worry me for something like this.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1479#-:-MHk3c472BUg1HF8YY-n:bshoxlb)* status: 0 of 2 LGTMs obtained

bstaletic:
> I guess I would have to understand the real benefits and what the actual problem is here.

I completely agree with this.

sansna:
> Thanks for sending a PR, but we don't want the installer to have dependencies outside of the standard libary. indeed we would love to ditch any of our third-party python dependencies.
> 
> > _[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1479#-:-MHk2jMa5cpM0f8Jt2t3:bd9zazi)_ status: 0 of 2 LGTMs obtained

Reminds me of the third_party folder in the repo.
So if I put the whole pget src code inside third_party should work.

sansna:
> I guess I would have to understand the real benefits and what the actual problem is here. The additional complexity (and non-testability) worry me for something like this.
> 
> > _[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1479#-:-MHk3c472BUg1HF8YY-n:bshoxlb)_ status: 0 of 2 LGTMs obtained

Real test of downloading speed with requests.get, about tens of kilobytes per sec.
Using pget of 8 splits, speed goes up to 600 kilobytes per sec.

puremourning:
Hmm, that is quite a big difference - do you know what the bottleneck is ? Seems odd.

sansna:
> Hmm, that is quite a big difference - do you know what the bottleneck is ? Seems odd.

Please do tell me where is the bottleneck.

puremourning:
Thanks for the PR, but it seems this has stalled. We're not keen on relying on non-standard library things particularly in the installer.

PR #1480, opened 2020-09-22
[WIP/RFC] Allow incremental file content updates

bstaletic:
This should allow clients to send incremental updates to files, in case a subserver supports that.

Missing features:

- [x] Report, somehow, to the client whether a subserver supports incremental updates.
- [x] Document the API
- [x] Remove hacks specific to vim.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1480)

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -100,6 +100,18 @@
             <tbody>
                     <tr>
                             <td class="swagger--summary-path" rowspan="1">
+                                <a href="#path--buffer_update_type">/buffer_update_type</a>
+                            </td>
+                        <td>
+                            <a href="#operation--buffer_update_type-get">GET</a>
+                        </td>
+                        <td>
+                            <p>Are incremental buffer updates supported for some filetype</p>
+        
+                        </td>
+                    </tr>
+                    <tr>
+                            <td class="swagger--summary-path" rowspan="1">
                                 <a href="#path--completions">/completions</a>
                             </td>
                         <td>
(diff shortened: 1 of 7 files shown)


puremourning (commented):
also crazy small change. Nice! 

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1480#-:-MHonArYCHz9BgLPQb4l:b-d5qv46)* status: 0 of 2 LGTMs obtained

bstaletic (commented):
While only LSP servers can truly support incremental updates, python, typescript and c# completers never access `file_data`, except for `file_data[ specified_buffer ]`. That means that these filetypes can benefit too, if we design the API correctly.

The tern and libclang completers do require complete `file_data`.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1480#-:-MHu9PbgAU8B5keT8Yvl:bxcss54)* status: 0 of 2 LGTMs obtained

___
*[ycmd/completers/language_server/language_server_completer.py, line 2004 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1480#-MHu95oX81S6mvKPtI3z:-MHu95oX81S6mvKPtI4-:bh9gqz3) ([raw file](https://github.com/ycm-core/ycmd/blob/51727204dd986f86d56534111aac5c4331cbeabf/ycmd/completers/language_server/language_server_completer.py#L2004)):*
> ```Python
>           LOGGER.debug( 'First change' )
>           msg = lsp.DidChangeTextDocument( file_state, file_data[ 'changes' ] )
>           self.GetConnection().SendNotification( msg )
> ```

Is this reachable in practice? Can we expect a client that sends `changes` for a buffer that has not been opened yet?
___
*[ycmd/completers/language_server/language_server_protocol.py, line 420 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1480#-MHu8B1VDGFM97R4mB3h:-MHu8B1VDGFM97R4mB3i:b4k22b1) ([raw file](https://github.com/ycm-core/ycmd/blob/51727204dd986f86d56534111aac5c4331cbeabf/ycmd/completers/language_server/language_server_protocol.py#L420)):*
> ```Python
>           'end':   { 'line': change[ 'range' ][ 'end' ][ 'line' ] - 1,
>                      'character': change[ 'range' ][ 'end' ][ 'col' ] - 1 }
>         } } for change in file_contents ]
> ```

I really don't like this part.

1. We are only changing `replacement_text` to `text` and making all offsets 0 based. This is just transforming a ycmd Range to a LSP Range.
2. These are byte offsets, but LSP expects UTF16 offsets.
2.1. If we require clients to send whole lines, then we can just assume column 0.
2.2. If we allow column offsets, we need to re-encode lines, which requires having up-to-date buffer contents and thus goes against the point of incremental updates in the first place.

puremourning (commented):
So I guess the key question is how do we address the fact that a lot of our code assumes it can do `GetFileContents` and that will be the 'latest' version of the file ? Problem being that right now it's probably reading the file from disk which is probably wrong. For example, right now we're probably relying on the file on disk to calculate things like byte-to-unicode offsets.

If we can address the above, then we don't need to send the full contents of the current buffer. I think the fact that we're having to send just one-special-file is an alert that we've missed something. I suspect that if we solve the point above, then we'll end up with a document model inside ycmd where it always has a cache of the buffer state anyway.

Getting to that point, whenever I've thought about this design in the past, I've ended up with ycmd having to have a document model, as in an in-memory representation of the buffer which it can use for things like offset calculations and for getting specific lines of text. They way I see this working is _always_ send incremental updates _to ycmd_, then the individual completers can send full or incremental updates to the downstream servers depending on what they need.

Let's say clangd requires full updates:

* On first request, YCM has to send full buffer contents and ycmd stores that in its document store
* ycmd needs to send full updates to clangd so asks for full contents from doc store, and gets what YCM sent.
* YCM sends "incremental" updates on next request
* ycmd applies diffs to its internal doc store (making a `file_data` that is up-to-date) and stores the incremental changes since last request from completion instance
* ycmd needs to send full updates to clangd, so it asks the doc store for full contents of document (returned in the form that we currently have in `file_data`)

Let's say jdt requires incremental updates

* On first request, YCM has to send full buffer contents and ycmd stores that in its document store
* ycmd needs to send full updates (didOpen) to jdt so asks for full contents from doc store, and gets what YCM sent.
* YCM sends "incremental" updates on next request
* ycmd applies diffs to its internal doc store (making a `file_data` that is up-to-date) and stores the incremental changes since last request from completion instance
* ycmd needs to send incremental updates (didChange) to jdt, so it asks the doc store for incremental changes, and gets back a list of changes stored from YCM.

Could worK ? Not art though.

Maybe we should go the whole hog and take all the content updates out of the request objects and send them in a different request `/update_file_contents` or something ? 

One thing that's tricky is that ycmd is annoyingly parallel, but document content updates are annoyingly serial, so we'll need to make sure that we present a consistent view of any document contents and that updates are serialised properly.

I realise that's a big scope increase, but I think it's necessary to some extend for correctness. Perhaps we can borrow ideas from the vscode-languageserver-node implementation of LSP's document model and/or find a python implementation that will reduce the workload. Still kind of interesting thing to play with.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1480#-:-MIV7G7cEERcvUUaNDCQ:b-vu54gr)* status: 0 of 2 LGTMs obtained

bstaletic (commented):
> So I guess the key question is how do we address the fact that a lot of our code assumes it can do `GetFileContents`

Or `request_data[ 'filetypes' ]`...

> Problem being that right now it's probably reading the file from disk which is probably wrong.

Not necessarily. In this PR, if the client has sent `changes`, then ycmd will call `GetFileContents()`, which will use `request_data[ 'contents' ]` if it is available. This is why I am still sending contents of "current" buffer.

> If we can address the above, then we don't need to send the full contents of the current buffer. I think the fact that we're having to send just one-special-file is an alert that we've missed something. I suspect that if we solve the point above, then we'll end up with a document model inside ycmd where it always has a cache of the buffer state anyway.

My thoughts exactly. I was hesitant to go down that road right away. An argument against that is... ycmd "just" relays messages from its client to a LSP server. Turns out that "just" is a big understatement.

> They way I see this working is always send incremental updates to ycmd, then the individual completers can send full or incremental updates to the downstream servers depending on what they need.

That's an interesting thought. I'm guessing you're not suggesting a complete removal of the "old" protocol, because of other clients and backwards compatibility?

> Could worK ? Not art though.

From this distance? Sure. I'm not immediately sure how to handle closed buffers, though.

> Maybe we should go the whole hog and take all the content updates out of the request objects and send them in a different request `/update_file_contents` or something ?

Hmmm... not sure, precisely because of your next paragraph.

> One thing that's tricky is that ycmd is annoyingly parallel, but document content updates are annoyingly serial, so we'll need to make sure that we present a consistent view of any document contents and that updates are serialised properly.

Uhmmm... another GET request that simply asks `/have_all_file_updates_been_applied`?

> Perhaps we can borrow ideas from the vscode-languageserver-node

I don't know about them, but I know that vim-go (or was it govim) does have its own representation of buffers, not unlike what you're suggesting.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1480#-:-MJIvkUgBZoKyFkF0a5o:buc5o20)* status: 0 of 2 LGTMs obtained

bstaletic:
Superseded by #1506.

PR #1489, opened 2020-10-24
[READY] Switch to C++17

bstaletic:
List of changes:

- Switch to C++17
- Use `string_view` where appropriate
- Allow shared lock ownership
- Avoid raw new, with `make_unique`
- Avoid `shared_ptr` in the identifier database
- Do not show valgrind 'possibly lost' leak errors
- Avoid AST deserialization errors in libclang/clangd tests

The BoostParts removal is left out because it kills Reviewable.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)

### DIFF
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -86,10 +86,12 @@ get_python_inc(),
 ]
 
 # Clang automatically sets the '-std=' flag to 'c++14' for MSVC 2015 or later,
-# which is required for compiling the standard library, and to 'c++11' for older
+# which is required for compiling the standard library, and to 'c++17' for older
 # versions.
 if platform.system() != 'Windows':
-  flags.append( '-std=c++11' )
+  flags.append( '-std=c++17' )
+else:
+  flags.append( '/std:c++17' )
 
 
 # Set this to the absolute path to the folder (NOT the file!) containing the
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -27,10 +27,10 @@ jobs:
         YCM_PYTHON_VERSION: '3.6.3'
         YCM_BENCHMARK: true
         COVERAGE: false
-      # 'C++ linting':
-      #   YCM_PYTHON_VERSION: '3.9.0'
-      #   YCM_CLANG_TIDY: true
-      #   COVERAGE: false
+      'C++ linting':
+        YCM_PYTHON_VERSION: '3.9.0'
+        YCM_CLANG_TIDY: true
+        COVERAGE: false
     maxParallel: 5
   variables:
     COVERAGE: true
--- a/azure/linux/install_dependencies.sh
+++ b/azure/linux/install_dependencies.sh
@@ -7,11 +7,13 @@ set -e
 sudo apt-get update
 sudo apt-get install libsqlite3-dev
 if [ "${YCM_COMPILER}" == "clang" ]; then
-  sudo apt-get install clang-8
-  sudo update-alternatives --install /usr/bin/cc cc /usr/bin/clang-8 100
-  sudo update-alternatives --install /usr/bin/c++ c++ /usr/bin/clang++-8 100
+  sudo apt-get install clang-7
+  sudo update-alternatives --install /usr/bin/cc cc /usr/bin/clang-7 100
+  sudo update-alternatives --install /usr/bin/c++ c++ /usr/bin/clang++-7 100
 else
-  sudo apt-get install gcc g++
+  sudo apt-get install gcc-8 g++-8
+  sudo update-alternatives --install /usr/bin/cc cc /usr/bin/gcc-8 100
+  sudo update-alternatives --install /usr/bin/c++ c++ /usr/bin/g++-8 100
 fi
 
 if [ "${YCM_CLANG_TIDY}" ]; then
--- a/build.py
+++ b/build.py
@@ -403,7 +403,7 @@ def ParseArguments():
   parser.add_argument( '--system-libclang', action = 'store_true',
                        help = 'Use system libclang instead of downloading one '
                        'from llvm.org. NOT RECOMMENDED OR SUPPORTED!' )
-  parser.add_argument( '--msvc', type = int, choices = [ 14, 15, 16 ],
+  parser.add_argument( '--msvc', type = int, choices = [ 15, 16 ],
                        default = 16, help = 'Choose the Microsoft Visual '
                        'Studio version (default: %(default)s).' )
   parser.add_argument( '--ninja', action = 'store_true',
--- a/build.py
+++ b/build.py
@@ -560,8 +560,8 @@ def RunYcmdTests( args, build_dir ):
             '--gen-suppressions=all',
             '--error-exitcode=1',
             '--leak-check=full',
-            '--show-leak-kinds=all',
-            '--show-reachable=no',
+            '--show-leak-kinds=definite,indirect',
+            '--errors-for-leak-kinds=definite,indirect',
             '--suppressions=' + p.join( DIR_OF_THIS_SCRIPT,
                                         'valgrind.suppressions' ),
             p.join( tests_dir, 'ycm_core_tests' ) ]
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -87,9 +87,6 @@ set( CMAKE_EXPORT_COMPILE_COMMANDS 1 )
 # This is needed so that on macs, the library is built in both 32 bit and 64 bit
 # versions. Without this python might refuse to load the module, depending on
 # how python was built.
-# On Mac, boost needs to be compiled universal as well, if used instead of the
-# included BoostParts lib. For brew, that's
-# "brew install boost --universal"
 # If the user chose to use the system libclang.dylib (or the libclang.dylib
 # binary downloaded from llvm.org) on a mac, then we don't specify universal
 # binary building since the system libclang on macs is not  universal (and thus
@@ -150,7 +147,6 @@ endif()
 #############################################################################
 
 # MSVC has snprintf since version 14 which is newer than what we support.
-# *NOT* defining HAVE_SNPRINTF will break compiling the bundled boost libraries.
 if ( MSVC )
   add_definitions( -DHAVE_SNPRINTF )
 endif()
@@ -164,21 +160,21 @@ endif()
 
 #############################################################################
 
-# Determining the presence of C++11 support in the compiler
-set( CPP11_AVAILABLE false )
+# Determining the presence of C++17 support in the compiler
+set( CPP17_AVAILABLE false )
 if ( CMAKE_COMPILER_IS_GNUCXX )
-  if ( NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.8 )
-    set( CPP11_AVAILABLE true )
+  if ( NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 8 )
+    set( CPP17_AVAILABLE true )
   endif()
 elseif( COMPILER_IS_CLANG )
-  if ( NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.3 )
-    set( CPP11_AVAILABLE true )
-    set( CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LANGUAGE_STANDARD "c++11" )
+  if ( NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7 )
+    set( CPP17_AVAILABLE true )
+    set( CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LANGUAGE_STANDARD "c++17" )
     set( CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LIBRARY "libc++" )
   endif()
 elseif( MSVC )
-  if ( NOT MSVC_VERSION LESS 1900 )
-    set( CPP11_AVAILABLE true )
+  if ( NOT MSVC_VERSION LESS 1914 )
+    set( CPP17_AVAILABLE true )
   endif()
 endif()
 
@@ -188,13 +184,8 @@ endif()
 # number of sections that an object file can hold. This is needed for storing
 # the Unicode table. Also, force MSVC to treat source files as UTF-8 encoded.
 if ( MSVC )
-  add_definitions( /DUNICODE /D_UNICODE /Zc:wchar_t- )
+  add_definitions( /DUNICODE /D_UNICODE )
   set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP /bigobj /utf-8" )
-endif()
-
-if( MSVC OR CYGWIN OR MSYS)
-  # BOOST_ALL_NO_LIB turns off MSVC library autolinking
-  add_definitions( -DBOOST_ALL_NO_LIB )
 endif()
 
 # Enables Python compilation for 64-bit Windows. Already defined by Python
@@ -205,49 +196,45 @@ endif()
 
 #############################################################################
 
-if ( CPP11_AVAILABLE )
+if ( CPP17_AVAILABLE )
   # Cygwin needs its hand held a bit; see issue #473
   if ( CYGWIN AND CMAKE_COMPILER_IS_GNUCXX )
-    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11" )
+    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++17" )
   elseif( NOT MSVC )
-    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11" )
+    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17" )
+  elseif( MSVC )
+    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /std:c++17" )
   endif()
 else()
   # Platform-specific advice goes here. In particular, we have plenty of users
   # in corporate environments on RHEL/CentOS, where the default system compiler
   # is too old.
 
-  # on RHEL and CentOS, users require the devtoolset-3 or greater. However, we
-  # recommended the devtoolset-6 because it is the newest at the time of
+  # on RHEL and CentOS, users require the devtoolset-8 or greater. However, we
+  # recommended the devtoolset-8 because it is the newest at the time of
   # writing. And why not.
   if ( DISTRIBUTION STREQUAL "CentOS" OR DISTRIBUTION STREQUAL "Red Hat" )
     message( STATUS "NOTE: You appear to be on ${DISTRIBUTION}. "
     "In order to use this application, you require a more modern compiler "
     "than the default compiler on this platform. "
-    "Please install the devtoolset-6 or greater. For example, see this link: "
-    "https://www.softwarecollections.org/en/scls/rhscl/devtoolset-6/" )
+    "Please install the devtoolset-8 or greater. For example, see this link: "
+    "https://www.softwarecollections.org/en/scls/rhscl/devtoolset-8/" )
 
     # Finally, just check if it is installed and they just need to activate it.
-    if ( EXISTS /opt/rh/devtoolset-6 )
-      message( STATUS "NOTE: It looks like you have the devtoolset-6 "
-                      "installed in /opt/rh/devtoolset-6, so you probably "
+    if ( EXISTS /opt/rh/devtoolset-8 )
+      message( STATUS "NOTE: It looks like you have the devtoolset-8 "
+                      "installed in /opt/rh/devtoolset-8, so you probably "
                       "just need to activate it and re-run the installation. "
-                      "For example: source /opt/rh/devtoolset-6/enable")
+                      "For example: source /opt/rh/devtoolset-8/enable")
     endif()
   endif()
 
-  message( FATAL_ERROR "Your C++ compiler does NOT fully support C++11." )
+  message( FATAL_ERROR "Your C++ compiler does NOT fully support C++17." )
 endif()
 
-set( Python_ADDITIONAL_VERSIONS 3.9 3.8 3.7 3.6 3.5 )
-find_package( PythonLibs 3.5 REQUIRED )  # 3.5 is ONLY the minimum
+set( Python_ADDITIONAL_VERSIONS 3.10 3.9 3.8 3.7 3.6 )
+find_package( PythonLibs 3.6 REQUIRED )  # 3.6 is ONLY the minimum
 
 #############################################################################
-
-option( USE_SYSTEM_BOOST "Set to ON to use the system boost libraries" OFF )
-
-if (NOT USE_SYSTEM_BOOST)
-  add_subdirectory( BoostParts )
-endif()
 
 add_subdirectory( ycm )
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -191,15 +191,12 @@ if ( APPLE )
   set( CMAKE_INCLUDE_SYSTEM_FLAG_CXX "-isystem " )
 endif()
 
-if ( USE_SYSTEM_BOOST )
-  set( Boost_COMPONENTS filesystem )
-  find_package( Boost REQUIRED COMPONENTS ${Boost_COMPONENTS} )
-else()
-  set( Boost_INCLUDE_DIR ${BoostParts_SOURCE_DIR} )
-  set( Boost_LIBRARIES BoostParts )
-endif()
-
 set( PYBIND11_INCLUDES_DIR "${CMAKE_SOURCE_DIR}/pybind11" )
+
+# Makes MSVC conform to the standard
+if ( MSVC )
+  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /permissive-" )
+endif()
 
 file( GLOB_RECURSE SERVER_SOURCES *.h *.cpp )
 
@@ -228,14 +225,13 @@ endif()
 
 # The SYSTEM flag makes sure that -isystem[header path] is passed to the
 # compiler instead of the standard -I[header path]. Headers included with
-# -isystem do not generate warnings (and they shouldn't; e.g. boost warnings are
-# just noise for us since we won't be changing them).
+# -isystem do not generate warnings (and they shouldn't; e.g. pybind11 warnings
+# are just noise for us since we won't be changing them).
 # Since there is no -isystem flag equivalent on Windows, headers from external
 # projects may conflict with our headers and override them. We prevent that by
 # including these directories after ours.
 include_directories(
   SYSTEM
-  ${Boost_INCLUDE_DIR}
   ${PYBIND11_INCLUDES_DIR}
   ${PYTHON_INCLUDE_DIRS}
   ${CLANG_INCLUDES_DIR}
@@ -318,6 +314,36 @@ if ( UNIX AND NOT ( APPLE OR SYSTEM_IS_OPENBSD OR HAIKU ) )
   set( EXTRA_LIBS rt )
 endif()
 
+# Check if we need to add -lstdc++fs or -lc++fs or nothing
+if( MSVC )
+  set( STD_FS_NO_LIB_NEEDED TRUE )
+else()
+  file( WRITE ${CMAKE_CURRENT_BINARY_DIR}/main.cpp "#include <filesystem>\nint main( int argc, char ** argv ) {\n  std::filesystem::path p( argv[ 0 ] );\n  return p.string().length();\n}" )
+  try_compile( STD_FS_NO_LIB_NEEDED ${CMAKE_CURRENT_BINARY_DIR}
+               SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
+               COMPILE_DEFINITIONS -std=c++17 )
+  try_compile( STD_FS_NEEDS_STDCXXFS ${CMAKE_CURRENT_BINARY_DIR}
+               SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
+               COMPILE_DEFINITIONS -std=c++17
+               LINK_LIBRARIES stdc++fs )
+  try_compile( STD_FS_NEEDS_CXXFS ${CMAKE_CURRENT_BINARY_DIR}
+               SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
+               COMPILE_DEFINITIONS -std=c++17
+               LINK_LIBRARIES c++fs )
+  file( REMOVE ${CMAKE_CURRENT_BINARY_DIR}/main.cpp )
+endif()
+
+if( ${STD_FS_NEEDS_STDCXXFS} )
+  set( STD_FS_LIB stdc++fs )
+elseif( ${STD_FS_NEEDS_CXXFS} )
+  set( STD_FS_LIB c++fs )
+elseif( ${STD_FS_NO_LIB_NEEDED} )
+  set( STD_FS_LIB "" )
+else()
+  message( WARNING "Unknown compiler - not passing -lstdc++fs" )
+  set( STD_FS_LIB "" )
+endif()
+
 #############################################################################
 
 add_library( ${PROJECT_NAME} SHARED
@@ -331,9 +357,9 @@ if ( USE_CLANG_COMPLETER AND NOT LIBCLANG_TARGET )
 endif()
 
 target_link_libraries( ${PROJECT_NAME}
-                       ${Boost_LIBRARIES}
                        ${PYTHON_LIBRARIES}
                        ${LIBCLANG_TARGET}
+                       ${STD_FS_LIB}
                        ${EXTRA_LIBS}
                      )
 
@@ -398,11 +424,7 @@ set_target_properties( ${PROJECT_NAME} PROPERTIES
 
 #############################################################################
 
-
-# For some reason, Xcode is too dumb to understand the -isystem flag and thus
-# borks on warnings in Boost.
-if ( USE_DEV_FLAGS AND ( CMAKE_COMPILER_IS_GNUCXX OR COMPILER_IS_CLANG ) AND
-     NOT CMAKE_GENERATOR_IS_XCODE )
+if ( USE_DEV_FLAGS AND ( CMAKE_COMPILER_IS_GNUCXX OR COMPILER_IS_CLANG ) )
   # We want all warnings, and warnings should be treated as errors
   set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror" )
 endif()
--- a/cpp/ycm/CandidateRepository.cpp
+++ b/cpp/ycm/CandidateRepository.cpp
@@ -64,7 +64,7 @@ std::vector< const Candidate * > CandidateRepository::GetCandidatesForStrings(
                                                   nullptr );
 
       if ( !candidate ) {
-        candidate.reset( new Candidate( std::move( candidate_text ) ) );
+        candidate = std::make_unique< Candidate >( std::move( candidate_text ) );
       }
 
       candidates.push_back( candidate.get() );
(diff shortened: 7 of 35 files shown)


puremourning (changes requested):
> Reviewed 34 of 34 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[cpp/CMakeLists.txt, line 166 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPdx-838AL6gm0EPbr:-MKPdx-838AL6gm0EPbs:b-dl8cx3) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L166)):*
> ```CMake
> set( CPP17_AVAILABLE false )
> if ( CMAKE_COMPILER_IS_GNUCXX )
>   if ( NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9 )
> ```

gcc 8 supports c++17. that's the compiler I use. so we need to support that _for certain_ :)

looks like gcc 7 is enough: https://gcc.gnu.org/projects/cxx-status.html
___
*[cpp/CMakeLists.txt, line 220 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPeLW4CnKbBs0Erjga:-MKPeLW4CnKbBs0Erjgb:b-nwjv23) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L220)):*
> ```CMake
>     "In order to use this application, you require a more modern compiler "
>     "than the default compiler on this platform. "
>     "Please install the devtoolset-6 or greater. For example, see this link: "
> ```

devtoolset-8
___
*[cpp/CMakeLists.txt, line 221 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPeQfz0aEJ33kzcaDD:-MKPeQfz0aEJ33kzcaDE:b-y5u62q) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L221)):*
> ```CMake
>     "than the default compiler on this platform. "
>     "Please install the devtoolset-6 or greater. For example, see this link: "
>     "https://www.softwarecollections.org/en/scls/rhscl/devtoolset-6/" )
> ```

https://www.softwarecollections.org/en/scls/rhscl/devtoolset-8/
___
*[cpp/CMakeLists.txt, line 224 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPeT1V3R7IGD1TSbpN:-MKPeT1V3R7IGD1TSbpO:bjoazir) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L224)):*
> ```CMake
> 
>     # Finally, just check if it is installed and they just need to activate it.
>     if ( EXISTS /opt/rh/devtoolset-6 )
> ```

/opt/rh/devtoolset-8
___
*[cpp/CMakeLists.txt, line 226 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPeYWVACpdi6SOJASI:-MKPeYWVACpdi6SOJASJ:bz64udt) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L226)):*
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>

> ```
>       message( STATUS "NOTE: It looks like you have the devtoolset-6 "
>                       "installed in /opt/rh/devtoolset-6, so you probably "
>                       "just need to activate it and re-run the installation. "
>                       "For example: source /opt/rh/devtoolset-6/enable")
> ```
</details>


-6 -> -8
___
*[cpp/ycm/CandidateRepository.cpp, line 79 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPfxi-6UtXrgzaLVe6:-MKPfxi-6UtXrgzaLVe7:b-31d7fd) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/CandidateRepository.cpp#L79)):*
> ```cpp
> 
> void CandidateRepository::ClearCandidates() {
>   candidate_holder_.clear();
> ```

shouldn't this be done under the lock ?
___
*[cpp/ycm/CharacterRepository.cpp, line 64 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPgC6mDf_LEjegFb1b:-MKPgC6mDf_LEjegFb1c:bwyr7xz) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/CharacterRepository.cpp#L64)):*
> ```cpp
> 
> void CharacterRepository::ClearCharacters() {
>   character_holder_.clear();
> ```

also here, are we missing a lock acquire?
___
*[cpp/ycm/CodePoint.cpp, line 84 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPggSt8NRvnh_k3tBI:-MKPggSt8NRvnh_k3tBJ:b8612r) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/CodePoint.cpp#L84)):*
> ```cpp
> 
> CodePoint::CodePoint( std::string_view code_point )
>   : CodePoint( FindCodePoint( code_point.data() ) ) {
> ```

do we need to pass len to FindCodePoint now ? Or should we change it to take a string_view ?

I mean, we _know_ that the incoming string is null terminated right, but in future, it might not be now that we're using a range.
___
*[cpp/ycm/CodePointRepository.cpp, line 63 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPh1Ba7029nn0fEwq3:-MKPh1Ba7029nn0fEwq4:b-yq3btr) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/CodePointRepository.cpp#L63)):*
> ```cpp
> 
> void CodePointRepository::ClearCodePoints() {
>   code_point_holder_.clear();
> ```

and here?
___
*[cpp/ycm/IdentifierCompleter.cpp, line 58 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPhQhiEHMnidHqw7DW:-MKPhQhj5ZN0xgiuXShC:b-oj84wu) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/IdentifierCompleter.cpp#L58)):*
> ```cpp
>   std::string& filetype,
>   std::string& filepath ) {
>   identifier_database_.ClearCandidatesStoredForFile( std::string( filetype ),
> ```

seems such a shame to copy these then move them :/
___
*[cpp/ycm/IdentifierCompleter.cpp, line 70 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPhkOQ8xL38I5S6HmJ:-MKPhkOQ8xL38I5S6HmK:blscqy2) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/IdentifierCompleter.cpp#L70)):*
> ```cpp
>   for( auto&& path : absolute_paths_to_tag_files ) {
>     identifier_database_.AddIdentifiers(
>       ExtractIdentifiersFromTagsFile( std::move( path ) ) );
> ```

are we really ok to take ownership and reset these string s?

It looks like they come from the python layer:


```
  def _AddIdentifiersFromTagFiles( self, tag_files ):
    absolute_paths_to_tag_files = ycm_core.StringVector()
    for tag_file in self._FilterUnchangedTagFiles( tag_files ):
      absolute_paths_to_tag_files.append( tag_file )

    if not absolute_paths_to_tag_files:
      return

    self._completer.AddIdentifiersToDatabaseFromTagFiles(
      absolute_paths_to_tag_files )

```

are we sure this is safe ? what would happen if we used `absolute_paths_to_tag_files` in the python code after this? safe ?
___
*[cpp/ycm/IdentifierDatabase.cpp, line 42 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPiGKf7V8VmGGcC0qT:-MKPiGKf7V8VmGGcC0qU:b-q1jbbb) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/IdentifierDatabase.cpp#L42)):*
> ```cpp
>   for ( auto&& filetype_and_map : filetype_identifier_map ) {
>     for ( auto&& filepath_and_identifiers : filetype_and_map.second ) {
>       auto filetype = filetype_and_map.first;
> ```

again are we sure that we can move this ?

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (waiting on @micbou and @puremourning)

___
*[cpp/CMakeLists.txt, line 166 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPdx-838AL6gm0EPbr:-MKPkhZNAb8PIPOCkDAU:b-pwlp6w) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L166)):*

gcc 8 supports c++17. that's the compiler I use. so we need to support that _for certain_ :)

looks like gcc 7 is enough: https://gcc.gnu.org/projects/cxx-status.html
</blockquote></details>

`<filesystem>` is a problem on gcc 8. It works, but requires `-lstdc++fs`.

We could also support clang 7, but that also requires a special flag, this time `-lstdc++fs` if using libstdc++, else `-lc++fs` if using libc++...

I'm just not sure how to do this correctly.
___
*[cpp/CMakeLists.txt, line 220 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPeLW4CnKbBs0Erjga:-MKPlPzK2vbpi2LL6V2q:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L220)):*

devtoolset-8
</blockquote></details>

Done.
___
*[cpp/CMakeLists.txt, line 221 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPeQfz0aEJ33kzcaDD:-MKPlQVuBXUqBr6xLHzw:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L221)):*

https://www.softwarecollections.org/en/scls/rhscl/devtoolset-8/
</blockquote></details>

Done.
___
*[cpp/CMakeLists.txt, line 224 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPeT1V3R7IGD1TSbpN:-MKPlRpp1F8OsqKzP6Ut:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L224)):*

/opt/rh/devtoolset-8
</blockquote></details>

Done.
___
*[cpp/CMakeLists.txt, line 226 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPeYWVACpdi6SOJASI:-MKPlSXl9wwTnm8cbtyO:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L226)):*

> ```
>       message( STATUS "NOTE: It looks like you have the devtoolset-6 "
>                       "installed in /opt/rh/devtoolset-6, so you probably "
>                       "just need to activate it and re-run the installation. "
>                       "For example: source /opt/rh/devtoolset-6/enable")
> ```

-6 -> -8
</blockquote></details>

Done.
___
*[cpp/ycm/CandidateRepository.cpp, line 79 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPfxi-6UtXrgzaLVe6:-MKPlUTH8KrvEsaSmavu:b-enov5f) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/CandidateRepository.cpp#L79)):*

shouldn't this be done under the lock ?
</blockquote></details>

Check the header. `ClearFoo()` is exposed only for C++ tests and C++ tests are single threaded.

@micbou was against taking a lock here, but I'm fine with it. Do you want me to change it?
___
*[cpp/ycm/CharacterRepository.cpp, line 64 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPgC6mDf_LEjegFb1b:-MKPlijlCzjzs1a1VANE:b9s4eza) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/CharacterRepository.cpp#L64)):*

also here, are we missing a lock acquire?
</blockquote></details>

Same thing. Only for tests, which are single-threadded.
___
*[cpp/ycm/CodePoint.cpp, line 84 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPggSt8NRvnh_k3tBI:-MKPm0vq89dx_-XehmG5:bmkhzx) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/CodePoint.cpp#L84)):*

do we need to pass len to FindCodePoint now ? Or should we change it to take a string_view ?

I mean, we _know_ that the incoming string is null terminated right, but in future, it might not be now that we're using a range.
</blockquote></details>

We're not using the length anywhere. We're using the C API that expects NULL terminated strings. I agree that this is questionable, so I'm fine with whatever you think is most appropriate.
___
*[cpp/ycm/CodePointRepository.cpp, line 63 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPh1Ba7029nn0fEwq3:-MKPmLI-0Y9uzXF_fB_y:b-ex0u0z) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/CodePointRepository.cpp#L63)):*

and here?
</blockquote></details>

Yeah, tests, single thread, not quite necessary. 
___
*[cpp/ycm/IdentifierCompleter.cpp, line 58 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPhQhiEHMnidHqw7DW:-MKPmTYE9rNyXWOdOybH:b-mi0r40) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/IdentifierCompleter.cpp#L58)):*

seems such a shame to copy these then move them :/
</blockquote></details>

I should look into this...
___
*[cpp/ycm/IdentifierCompleter.cpp, line 70 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPhkOQ8xL38I5S6HmJ:-MKPmffB39Q2wWdLUGiM:b-3ln36n) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/IdentifierCompleter.cpp#L70)):*

are we really ok to take ownership and reset these string s?

It looks like they come from the python layer:


```
  def _AddIdentifiersFromTagFiles( self, tag_files ):
    absolute_paths_to_tag_files = ycm_core.StringVector()
    for tag_file in self._FilterUnchangedTagFiles( tag_files ):
      absolute_paths_to_tag_files.append( tag_file )

    if not absolute_paths_to_tag_files:
      return

    self._completer.AddIdentifiersToDatabaseFromTagFiles(
      absolute_paths_to_tag_files )

```

are we sure this is safe ? what would happen if we used `absolute_paths_to_tag_files` in the python code after this? safe ?
</blockquote></details>

Pybind11 copies the python string into `std::string`. So we're stealing a copy python is providing us with.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 42 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPiGKf7V8VmGGcC0qT:-MKPmrc309TpV-05U_vx:bmg5qpj) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/IdentifierDatabase.cpp#L42)):*

again are we sure that we can move this ?
</blockquote></details>

Same thing. Pybind makes a copy for us.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (waiting on @micbou and @puremourning)

___
*[cpp/ycm/IdentifierCompleter.cpp, line 58 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPhQhiEHMnidHqw7DW:-MKPoc1i8LGdnfoEVyJz:b-jnrenw) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/IdentifierCompleter.cpp#L58)):*

I should look into this...
</blockquote></details>

Looked. Here's the thing - this just makes the copy explicit. It was there previously.

We can't move directly, because we need to use path twice.

I would love to be able to take `std::string&&` from pybind and move that, but... that is for a future version of pybind11.

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic, @micbou, and @puremourning)

___
*[cpp/CMakeLists.txt, line 166 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPdx-838AL6gm0EPbr:-MKPpQz-72Y99KVMBwOz:b3i0k91) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L166)):*

`<filesystem>` is a problem on gcc 8. It works, but requires `-lstdc++fs`.

We could also support clang 7, but that also requires a special flag, this time `-lstdc++fs` if using libstdc++, else `-lc++fs` if using libc++...

I'm just not sure how to do this correctly.
</blockquote></details>

```
if ( CMAKE_CXX_COMPILER_VERSION VERSION_LESS 8 )
   // error
else's( CMAKE_CSS_COMPILER_VERSION VERSION_LESS 9 )
   // ok
   // add link flag `-lstdc++fs`
else()
  // OK
endif()
```

can this work ? we use `-lstdc++fs` at work with devtoolset-8.

I'm fine with requiring clang 8. with PRs welcome to support clang 7.

lots of RHEL/centos users will have gcc8 as the devtoolset-9 is very new (like this year I think). gcc 9 is also pretty new (like 1.5y or something).

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (waiting on @micbou and @puremourning)

___
*[cpp/CMakeLists.txt, line 166 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPdx-838AL6gm0EPbr:-MKPrlFK4_MsLjcAnB4V:bs28xna) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L166)):*

```
if ( CMAKE_CXX_COMPILER_VERSION VERSION_LESS 8 )
   // error
else's( CMAKE_CSS_COMPILER_VERSION VERSION_LESS 9 )
   // ok
   // add link flag `-lstdc++fs`
else()
  // OK
endif()
```

can this work ? we use `-lstdc++fs` at work with devtoolset-8.

I'm fine with requiring clang 8. with PRs welcome to support clang 7.

lots of RHEL/centos users will have gcc8 as the devtoolset-9 is very new (like this year I think). gcc 9 is also pretty new (like 1.5y or something).
</blockquote></details>

What concerns me is the combination of clang + libstdc++.

If clang is new, but libstdc++ came with gcc 8, we still need to add `-lstdc++fs`. Can we check the standard library name?

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic, @micbou, and @puremourning)

___
*[cpp/CMakeLists.txt, line 166 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPdx-838AL6gm0EPbr:-MKPsccq58AvubuY62eQ:buu6j39) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L166)):*

What concerns me is the combination of clang + libstdc++.

If clang is new, but libstdc++ came with gcc 8, we still need to add `-lstdc++fs`. Can we check the standard library name?
</blockquote></details>

hmmm good point. that's true anyway with the clang version.

with auto tools we would do a test compile. does cmake have an equivalent ?

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic, @micbou, and @puremourning)

___
*[cpp/CMakeLists.txt, line 166 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPdx-838AL6gm0EPbr:-MKPt-EYEgBEiraO-rkw:b-u5nhg8) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L166)):*

hmmm good point. that's true anyway with the clang version.

with auto tools we would do a test compile. does cmake have an equivalent ?
</blockquote></details>

https://stackoverflow.com/questions/54290254/problem-adding-stdfilesystem-to-cmake-project

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (waiting on @micbou and @puremourning)

___
*[cpp/CMakeLists.txt, line 166 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPdx-838AL6gm0EPbr:-MKPtV_Q4w_2MzXLYxvU:b-5a15r1) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L166)):*

https://stackoverflow.com/questions/54290254/problem-adding-stdfilesystem-to-cmake-project
</blockquote></details>

CMake does have a test compile feature and I've done it before for Cquery, while it was still actively developed.

I still don't know what I should check for.

As for that SO answer, it's only concerned with libstdc++'s `-lstdc++fs` and they are using cmake 3.12, while our cmake still requires 2.8.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (waiting on @micbou and @puremourning)

___
*[cpp/CMakeLists.txt, line 166 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPdx-838AL6gm0EPbr:-MKPuQgK4wBCE0LhWYOH:b-j3riwr) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/CMakeLists.txt#L166)):*

CMake does have a test compile feature and I've done it before for Cquery, while it was still actively developed.

I still don't know what I should check for.

As for that SO answer, it's only concerned with libstdc++'s `-lstdc++fs` and they are using cmake 3.12, while our cmake still requires 2.8.
</blockquote></details>

https://github.com/jacobdufault/cquery/pull/775/files

puremourning (changes requested):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[cpp/ycm/CandidateRepository.cpp, line 79 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPfxi-6UtXrgzaLVe6:-MKRcJ3t2YGEAL646o6X:b2ni8ds) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/CandidateRepository.cpp#L79)):*

Check the header. `ClearFoo()` is exposed only for C++ tests and C++ tests are single threaded.

@micbou was against taking a lock here, but I'm fine with it. Do you want me to change it?
</blockquote></details>

nah
___
*[cpp/ycm/CMakeLists.txt, line 320 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKRblvQ8d1u-BgPvE3D:-MKRblvRCblLWjfrqOdo:bdnkj4h) ([raw file](https://github.com/ycm-core/ycmd/blob/bc9c759ad05f4416457cfb4ae16e7562aae2c8cf/cpp/ycm/CMakeLists.txt#L320)):*
> ```CMake
> if( NOT MSVC )
>   file( WRITE ${CMAKE_CURRENT_BINARY_DIR}/main.cpp "#include <filesystem>\nint main( int argc, char ** argv ) {\n  std::filesystem::path p( argv[ 0 ] );\n  return p.string().length();\n}" )
>   try_compile( STD_FS_NO_LIB_NEEDED ${CMAKE_CURRENT_BINARY_DIR} SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp COMPILE_DEFINITIONS -std=c++17 )
> ```

can we break this and the other lines at 80 characters so it is a bit more readable ?
___
*[cpp/ycm/CMakeLists.txt, line 352 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKRbrIT8B5qF8UKJQxe:-MKRbrIT8B5qF8UKJQxf:b-qq5nuq) ([raw file](https://github.com/ycm-core/ycmd/blob/bc9c759ad05f4416457cfb4ae16e7562aae2c8cf/cpp/ycm/CMakeLists.txt#L352)):*
> ```CMake
>                        ${PYTHON_LIBRARIES}
>                        ${LIBCLANG_TARGET}
> 		       ${STD_FS_LIB}
> ```

is that... a .... *pukes* .. is that a TAB I see before me ?
___
*[cpp/ycm/CodePoint.cpp, line 84 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPggSt8NRvnh_k3tBI:-MKRbygZ4CRQ7rl6NJkj:biw2p1b) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/CodePoint.cpp#L84)):*

We're not using the length anywhere. We're using the C API that expects NULL terminated strings. I agree that this is questionable, so I'm fine with whatever you think is most appropriate.
</blockquote></details>

it does seem wonky to me to just use the data elem of a string_view, but I think this is probably fine. we'll know soon enough is we break it.

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[cpp/ycm/CMakeLists.txt, line 325 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKRhBQU2WWs_STABt7c:-MKRhBQU2WWs_STABt7d:b19gfk5) ([raw file](https://github.com/ycm-core/ycmd/blob/5ef271360dbc61ad9b49c5b06535442030134fd6/cpp/ycm/CMakeLists.txt#L325)):*
> ```CMake
>   file( REMOVE ${CMAKE_CURRENT_BINARY_DIR}/main.cpp )
> else()
>   set( STD_FS_NO_LIB_NEEDED TRUE )
> ```

invert this if()

```
if( MSVC )
  set( STD_FS_NO_LIB_NEEDED TRUE )
else()

endif()
```

puremourning (approved):
:lgtm: with a couple of minor prettifications.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489#-:-MKRhSyk5MlUacNiuLR3:b-5oz8p8)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

___
*[cpp/ycm/CMakeLists.txt, line 320 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKRblvQ8d1u-BgPvE3D:-MKRintP0TOXuxWg4ENw:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/bc9c759ad05f4416457cfb4ae16e7562aae2c8cf/cpp/ycm/CMakeLists.txt#L320)):*

can we break this and the other lines at 80 characters so it is a bit more readable ?
</blockquote></details>

Done.
___
*[cpp/ycm/CMakeLists.txt, line 352 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKRbrIT8B5qF8UKJQxe:-MKRiAT07JLJx4dhKZel:b-91igkd) ([raw file](https://github.com/ycm-core/ycmd/blob/bc9c759ad05f4416457cfb4ae16e7562aae2c8cf/cpp/ycm/CMakeLists.txt#L352)):*

is that... a .... *pukes* .. is that a TAB I see before me ?
</blockquote></details>

Two... Two tabs.
___
*[cpp/ycm/CMakeLists.txt, line 325 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKRhBQU2WWs_STABt7c:-MKRixo2-w6fjhCozMjB:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/5ef271360dbc61ad9b49c5b06535442030134fd6/cpp/ycm/CMakeLists.txt#L325)):*

invert this if()

```
if( MSVC )
  set( STD_FS_NO_LIB_NEEDED TRUE )
else()

endif()
```
</blockquote></details>

Done.
___
*[cpp/ycm/CodePoint.cpp, line 84 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPggSt8NRvnh_k3tBI:-MKRjKtj95MjRFFF-LLx:boea3qq) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/CodePoint.cpp#L84)):*

it does seem wonky to me to just use the data elem of a string_view, but I think this is probably fine. we'll know soon enough is we break it.
</blockquote></details>

Last chance to tell me to change it?

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

___
*[cpp/ycm/CodePoint.cpp, line 84 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPggSt8NRvnh_k3tBI:-MKTrolyAhkoGGHZpAG9:biaump) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/CodePoint.cpp#L84)):*

Last chance to tell me to change it?
</blockquote></details>

How complex is it to change? Like if it’s a rabbit hole don’t bother but if it’s trivial let’s do it.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

___
*[cpp/ycm/CodePoint.cpp, line 84 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKPggSt8NRvnh_k3tBI:-MKTwp7p2MH77DEVyJv8:b-l8qkmt) ([raw file](https://github.com/ycm-core/ycmd/blob/9d1daf142855a6495eef49afbedb4553f2a8b765/cpp/ycm/CodePoint.cpp#L84)):*

How complex is it to change? Like if it’s a rabbit hole don’t bother but if it’s trivial let’s do it. 
</blockquote></details>

Not very complex. I think I got the code to be pretty clean. It does result in more assembly, but it wasn't reflected in our benchmarks beyong a margin of error.
___
*[cpp/ycm/CodePoint.cpp, line 63 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKU8HCT7WJObANfRRhR:-MKU8HCT7WJObANfRRhS:b5vs9e0) ([raw file](https://github.com/ycm-core/ycmd/blob/f8f3a71ad248a6207fbb8ed21bb780270b19d776/cpp/ycm/CodePoint.cpp#L63)):*
> ```cpp
>     text,
>     []( const char* cp, std::string_view text ) {
>       return text.compare( cp ) > 0;
> ```

Note that we used to do `cp <=> text < 0`. Now we're doing `text <=> cp > 0`.

The obvious first thought is "that should obviously be `>=`", but no. This change isn't about negating the comparison, it's about multiplying the inequasion by `-1`. Besides math, I do have tests to back this up.

Why did I write it like this? Because I thought the following was even more confusing:

    return -1 * text.compare( cp ) < 0;

We can't do `cp.compare( text ) < 0;` because `cp` is `const char*` and comes from that giant Unicode table. We can't change `code_points.original` to hold `std::string_view[]`, because it would blow up the binary size, the size on HDD *and* would make every known compiler stall. We *can* can change `cp` to `std::string_view` in the lambda, but I'm not sure what would be the benefit.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

___
*[cpp/ycm/CodePoint.cpp, line 63 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKU8HCT7WJObANfRRhR:-MKUGs-K1jDaiht_E-FC:bfmut0o) ([raw file](https://github.com/ycm-core/ycmd/blob/f8f3a71ad248a6207fbb8ed21bb780270b19d776/cpp/ycm/CodePoint.cpp#L63)):*

Note that we used to do `cp <=> text < 0`. Now we're doing `text <=> cp > 0`.

The obvious first thought is "that should obviously be `>=`", but no. This change isn't about negating the comparison, it's about multiplying the inequasion by `-1`. Besides math, I do have tests to back this up.

Why did I write it like this? Because I thought the following was even more confusing:

    return -1 * text.compare( cp ) < 0;

We can't do `cp.compare( text ) < 0;` because `cp` is `const char*` and comes from that giant Unicode table. We can't change `code_points.original` to hold `std::string_view[]`, because it would blow up the binary size, the size on HDD *and* would make every known compiler stall. We *can* can change `cp` to `std::string_view` in the lambda, but I'm not sure what would be the benefit.
</blockquote></details>

Looking at the code for `string_view::compare(CharT*)`:

```c++
      compare(const _CharT* __str) const noexcept
      { return this->compare(basic_string_view{__str}); }
```

So we can use `std::string_view cp` here. It will be exactly the same.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

___
*[cpp/ycm/CodePoint.cpp, line 63 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1489#-MKU8HCT7WJObANfRRhR:-MKUvBsd3xJkep17AnDm:b-ssu0a) ([raw file](https://github.com/ycm-core/ycmd/blob/f8f3a71ad248a6207fbb8ed21bb780270b19d776/cpp/ycm/CodePoint.cpp#L63)):*

Looking at the code for `string_view::compare(CharT*)`:

```c++
      compare(const _CharT* __str) const noexcept
      { return this->compare(basic_string_view{__str}); }
```

So we can use `std::string_view cp` here. It will be exactly the same.
</blockquote></details>

Wait... not that we have a `std::string_view`, we can just compare `cp < text`, which is just `std::less{}`, which is also the default comparison for `lower_bound`, so we can completely avoid the lambda. I'll do this after my compiler finishes with llvm.

puremourning (approved):
:lgtm:

> Reviewed 3 of 4 files at r4, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1489#-:-MKVDVP0C0gRL1ZQ2g8p:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic:
:lgtm:

PR #1495, opened 2020-10-29
[READY] Clean up the vlagrind run

bstaletic:
Not ready, because this pull request requires python 3.9.1.
Currently we have 5 suppression rules:

1. The first two rules are valgrind misinterpreting a compiler optimization. Ubuntu 20.04 ships with valgrind 3.15.0 that still reports these as errors. That's why this pull request compiles valgrind 3.16.1.
2. The third was an intentional leak, to prevent crashes, caused by a CPython 3.9.0 bug. Hence the requirement of 3.9.1.
3. The leak reaching into `dlope()` is interesting. From what I gather, a few stars have to line up to leak this memory.
3.1. We're adding the libclang.so path to `LD_LIBRARY_PATH`, making the loader aware of it once `import ycm_core` happens, even without us calling `ctypes.cdll.LoadLibrary()`.
3.2. We run `ycmd/tests/bindings` tests that reach into libclang API.
3.3. We then run `utils_test.py`, which contains tests that call `ctypes.cdll.LoadLibrary()` from `ycmd.utils`.
3.4. Using `LoadLibrary()` to load an already loaded *and used* shared object seems to result in a leak. At least that's 100% reproducible with ycmd and libclang. It seems to also depend on the shared object that is being loaded to not be in `sys.path`.
3.5. Finally, the workaround was the change in `ycmd/utils.py` - try to load `ycm_core` without `LoadLibrary()` and if that fails, try again, but first call `LoadLibrary()`. Also, I had to skip the two tests that would still end up calling `LoadLibrary()`, when running in valgrind.
4. Latest error, is a libclang leak when a compiler invocation fails. Since we never encounter that case in our tests, we can remove the suppression rule.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1495)


puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1495#-:-MOwVH_S3sWXJ0WogJ6B:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[azure/linux/install_dependencies.sh, line 27 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1495#-MOwUsEf2UfIQ_0BdAV3:-MOwUsEg72mgnyTIm_j_:b-abo81w) ([raw file](https://github.com/ycm-core/ycmd/blob/84cd3cba791f4789f1201b6b62d2f80f4dd0f4e2/azure/linux/install_dependencies.sh#L27)):*
> ```Shell
>   cd valgrind-3.16.1
>   ./configure && make -j3 && sudo make install
>   cd -
> ```

pushd/popd ?

bstaletic (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1495#-:-MOwmJz99AjMicx4K9fu:bnfp4nl)* status: 1 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1495#-:-MPo_gPV312CfzByJwiM:bnfp4nl)* status: :shipit: complete! 2 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1505, opened 2020-11-21
[RFC] Implement LSP workspaces support

bstaletic:
This should allow workspaces/nested projects with rust-analyzer and gopls. It should also allow us to switch C# to LSP.

Unfinished busyness:

- No tests, because I didn't know how to set up a test case.
- I'm not sure what should be the exact algorithm for discovering the workspace directory.
- Some parts of the page are horribly inefficient.

So yes, this is a pretty crude pull request, but I think it works.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1505)

### DIFF
--- a/ycmd/completers/go/go_completer.py
+++ b/ycmd/completers/go/go_completer.py
@@ -58,10 +58,6 @@ class GoCompleter( language_server_completer.LanguageServerCompleter ):
 
 
   def GetProjectRootFiles( self ):
-    # Without LSP workspaces support, GOPLS relies on the rootUri to detect a
-    # project.
-    # TODO: add support for LSP workspaces to allow users to change project
-    # without having to restart GOPLS.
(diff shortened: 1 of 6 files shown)


bstaletic:
Am I doing something wrong in this PR? When I test it, I see expected messages marshalled, but no change in behaviour for ra/gopls regarding juggling multiple projects in a single sessions.

puremourning:
wow sorry I can't believe I never looked at this. Shall we keep it around and refresh for later?

bstaletic:
Well, this one seemed to work locally, as in "I can see the server reacting to the message in the logs". It still has not made gopls work with multiple projects. So, again, I don't know what to do with this. If we get it working, there's a chance of getting Omnisharp-Roslyn working as a LSP server. Which would be nice.

I've rebased it, if you want to take a look.

PR #1506, opened 2020-11-23
[RFC] Allow clients to send incremental file updates

bstaletic:
As discussed in #1480, this time ycmd keeps the state of all files, on a per-completer bases. It definitely needs more testing and, like I mentioned in gitter, this doesn't really take care of "what's the filetype of current buffer" - that's still fetched from `request_data[ 'file_data' ]`.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1506)

### DIFF
--- a/ycmd/completers/completer.py
+++ b/ycmd/completers/completer.py
@@ -212,6 +212,7 @@ class Completer( metaclass = abc.ABCMeta ):
     self._max_candidates = user_options[ 'max_num_candidates' ]
     self._max_candidates_to_detail = user_options[
       'max_num_candidates_to_detail' ]
+    self._server_file_state = completer_utils.ServerFileStateStore()
 
     LOGGER.info( f"Completion config: { self._max_candidates }, detailing "
                  f"{ self._max_candidates_to_detail } candiates" )
@@ -448,7 +449,10 @@ class Completer( metaclass = abc.ABCMeta ):
 
 
   def OnBufferUnload( self, request_data ):
-    pass # pragma: no cover
+    try:
+      del self._server_file_state[ request_data[ 'filepath' ] ]
+    except KeyError:
+      pass
 
 
   def OnInsertLeave( self, request_data ):
@@ -555,6 +559,60 @@ class Completer( metaclass = abc.ABCMeta ):
     LOGGER.debug( 'No Settings function defined in %s', module.__file__ )
 
     return {}
+
+
+  def Reset( self ):
+    self._server_file_state = completer_utils.ServerFileStateStore()
+
+
+  def OpenFileHandler( self, file_state, filetypes, changes ):
+    pass # pragma: no cover
+
+
+  def ChangeFileHandler( self, file_state, filetypes, changes ):
+    pass # pragma: no cover
+
+
+  def _AnySupportedFileType( self, file_types ):
+    for supported in self.SupportedFiletypes():
+      if supported in file_types:
+        return True
+    return False
+
+
+  def _UpdateDirtyFilesUnderLock( self, request_data ):
+    for file_name, file_data in request_data[ 'file_data' ].items():
+      if not self._AnySupportedFileType( file_data[ 'filetypes' ] ):
+        LOGGER.debug( 'Not updating file %s, it is not a supported filetype: '
+                       '%s not in %s',
+                       file_name,
+                       file_data[ 'filetypes' ],
+                       self.SupportedFiletypes() )
+        continue
+
+      file_state = self._server_file_state[ file_name ]
+      if 'contents' in file_data:
+        action = file_state.GetDirtyFileAction( file_data[ 'contents' ] )
+      elif 'changes' in file_data:
+        changes = file_data[ 'changes' ]
+        changes_start = changes[ 'start' ][ 'line_num' ]
+        changes_end = changes[ 'end' ][ 'line_num' ]
+        file_state[ changes_start :
+                    changes_end ] = changes[ 'replacement_text' ]
+        action = completer_utils.ServerFileState.CHANGE_FILE
+
+      LOGGER.debug( 'Refreshing file %s: State is %s/action %s',
+                    file_name,
+                    file_state.state,
+                    action )
+      if action == completer_utils.ServerFileState.OPEN_FILE:
+        self.OpenFileHandler( file_state,
+                              file_data[ 'filetypes' ],
+                              file_data.get( 'changes' ) )
+      elif action == completer_utils.ServerFileState.CHANGE_FILE:
+        self.ChangeFileHandler( file_state,
+                                file_data[ 'filetypes' ],
+                                file_data.get( 'changes' ) )
 
 
 class CompletionsCache:
--- a/ycmd/completers/language_server/language_server_protocol.py
+++ b/ycmd/completers/language_server/language_server_protocol.py
@@ -18,7 +18,6 @@
 import collections
 import os
 import json
-import hashlib
 from urllib.parse import urljoin, urlparse, unquote
 from urllib.request import pathname2url, url2pathname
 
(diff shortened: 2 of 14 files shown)


bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1506)* status: 0 of 2 LGTMs obtained

___
*[ycmd/completers/completer.py, line 215 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1506#-MMpdTdeEyBDn0vQpQC7:-MMpdTdeEyBDn0vQpQC8:b776ukg) ([raw file](https://github.com/ycm-core/ycmd/blob/3c2df329b41742d95d47cebfacddc6fefabbd017/ycmd/completers/completer.py#L215)):*
> ```Python
>     self._max_candidates_to_detail = user_options[
>       'max_num_candidates_to_detail' ]
>     self._server_file_state = completer_utils.ServerFileStateStore()
> ```

This line is repeated in `self.Reset()`.
___
*[ycmd/completers/completer.py, line 452 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1506#-MMpda9Q7ZotRPor6Vs2:-MMpda9Q7ZotRPor6Vs3:bat559f) ([raw file](https://github.com/ycm-core/ycmd/blob/3c2df329b41742d95d47cebfacddc6fefabbd017/ycmd/completers/completer.py#L452)):*
> ```Python
>   def OnBufferUnload( self, request_data ):
>     try:
>       del self._server_file_state[ request_data[ 'filepath' ] ]
> ```

Am I supposed to do this? Should we instead keep the file around and just set the state to closed?
___
*[ycmd/completers/completer.py, line 572 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1506#-MMpdmQrBYyyb59vy7tn:-MMpdmQrBYyyb59vy7to:b3sw65s) ([raw file](https://github.com/ycm-core/ycmd/blob/3c2df329b41742d95d47cebfacddc6fefabbd017/ycmd/completers/completer.py#L572)):*
> ```Python
> 
>   def ChangeFileHandler( self, file_state, filetypes, changes ):
>     pass # pragma: no cover
> ```

Since we have `OpenFileHandler` and `ChangeFileHandler`, do we need one for closing?
___
*[ycmd/completers/completer_utils.py, line 328 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1506#-MMpdx_e-r_okjcGNNb7:-MMpdx_e-r_okjcGNNb8:b5la33s) ([raw file](https://github.com/ycm-core/ycmd/blob/3c2df329b41742d95d47cebfacddc6fefabbd017/ycmd/completers/completer_utils.py#L328)):*
> ```Python
> 
> 
>   def _CalculateCheckSum( self, contents ):
> ```

We're still using checksums. Using file versions would be more efficient.
___
*[ycmd/completers/cpp/clang_completer.py, line 66 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1506#-MMpe3Cm94hmxxljqGyP:-MMpe3Cm94hmxxljqGyQ:btq5xkr) ([raw file](https://github.com/ycm-core/ycmd/blob/3c2df329b41742d95d47cebfacddc6fefabbd017/ycmd/completers/cpp/clang_completer.py#L66)):*
> ```Python
>     contents = file_state.contents
>     if not contents or not file_state.filename:
>       return
> ```

This block is kept from before the refactoring, but it looks completely useless, hence the `TODO: WAT?` comment. On top of that, it was never hit by our tests. Should I remove it?
___
*[ycmd/completers/language_server/language_server_completer.py, line 2002 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1506#-MMpfW9IBjTi48U2Zd1o:-MMpfW9IBjTi48U2Zd1p:b-84xj2e) ([raw file](https://github.com/ycm-core/ycmd/blob/3c2df329b41742d95d47cebfacddc6fefabbd017/ycmd/completers/language_server/language_server_completer.py#L2002)):*
> ```Python
>     with self._server_info_mutex:
>       self._UpdateDirtyFilesUnderLock( request_data )
>       # TODO: Consider doing this in self.OnBufferUnload()
> ```

"This" as in:

```python
      files_to_purge = self._UpdateSavedFilesUnderLock( request_data )
      self._PurgeMissingFilesUnderLock( files_to_purge )
```

I still feel like we should be able to do this only on `BufferUnload`, but @puremourning says there were problems with that approach.

bstaletic:
I figured out what's wrong with this. Here's the diff.

```diff
diff --git a/ycmd/completers/language_server/language_server_completer.py b/ycmd/completers/language_server/language_server_completer.py
index c87ca7e1b..2ffef7094 100644
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -1981,7 +1981,8 @@ class LanguageServerCompleter( Completer ):
         'text': fix_it_chunk[ 'replacement_text' ],
         'range': lsp_range
       }
-    if self._sync_type == 'Incremental' and changes is not None:
+    # TODO: Figure out what's wrong with this.
+    if False and self._sync_type == 'Incremental' and changes is not None:
       diff = [ FixItChunkToContentChangeEvent( change )
                for change in changes ]
     else:
```

So why does it not work?

ycmd/completers/language_server/language_server_protocol.py
```diff
@@ -391,38 +301,36 @@ def DidChangeConfiguration( config ):
...
   params = {
     'textDocument': {
       'uri': FilePathToUri( file_state.filename ),
       'version': file_state.version,
     },
   }
-  if file_contents is not None:
-    params.update( { 'text': file_contents } )
+  if file_state.contents is not None:
+    params.update( { 'text': file_state.contents } )
```
puremourning, line 333: when we do this, should we set the file_state.changes to [] ?

bstaletic: I'm honestly not sure what that would do, but I can test that.

ycmd/completers/completer.py
```diff
@@ -556,6 +560,60 @@ def GetSettings( self, module, request_data ):
...
+        continue
+
+      file_state = self._server_file_state[ file_name ]
+      if 'contents' in file_data:
+        action = file_state.GetDirtyFileAction( file_data[ 'contents' ] )
+      elif 'changes' in file_data:
+        changes = file_data[ 'changes' ]
+        changes_start = changes[ 'start' ][ 'line_num' ]
+        changes_end = changes[ 'end' ][ 'line_num' ]
+        file_state[ changes_start :
```
puremourning, line 599: I'm confused by this. `file_state` is a `ServerFileState` object... surely that's not subscriptable?

ycmd/completers/completer.py
```diff
@@ -556,6 +560,60 @@ def GetSettings( self, module, request_data ):
...
+                    changes_end ] = changes[ 'replacement_text' ]
+        action = completer_utils.ServerFileState.CHANGE_FILE
```
puremourning, line 601: how do we know this is the right action ? I mean if the client sent changes without contents, we're assuming we already sent a didOpen ? Is that safe?

ycmd/completers/language_server/language_server_protocol.py
bstaletic: Nope. That didn't solve anything. Any reason for that change?

puremourning: I was wondering if sending the "full" text in the save notification could make the sequence out-of-sync, e.g. consider a series of events where the file is opened, 3 changes are made, then the file is saved. Because Ycmd requests are not sequenced, the events can overtake each other, e.g.

* `didOpen` (full contents state = `A`)
* `didChange` (first delta `A->B`, state = `B`)
* `didChange` (second delta `B->C`, state = `C`)
* `didSave` (full contents of state `D`, state = `D`)
* `didChange` (third delta `C->D`, state = `error (C->D is not valid when applied to last updated state D)`)


But I think that the text in didSave is not allowed to change the server file state (i.e. `didSave (text of state D, state remains C)`

ycmd/completers/completer.py
puremourning: Also, when we do this, what's updating the version number in `ServerFileState` ?

bstaletic:
Rebased. Let's see what the tests will say.

puremourning:
Seems abandoned. We can always re-open this in future.

PR #1509, opened 2020-12-01
[READY] Drop waitress in favour of the standard library

bstaletic:
I'll make some more benchmarks tomorrow, but even if there's no difference between waitress and the standard library, dropping an external dependency would be nice.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)

### DIFF
--- a/ycmd/wsgi_server.py
+++ b/ycmd/wsgi_server.py
@@ -15,44 +15,13 @@
 # You should have received a copy of the GNU General Public License
 # along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
 
-from waitress.server import TcpWSGIServer
-import select
-import sys
+from wsgiref.simple_server import WSGIServer, WSGIRequestHandler
+from socketserver import ThreadingMixIn
 
 
-class StoppableWSGIServer( TcpWSGIServer ):
-  """StoppableWSGIServer is a subclass of the TcpWSGIServer Waitress server
-  with a shutdown method. It is based on StopableWSGIServer class from webtest:
-  https://github.com/Pylons/webtest/blob/master/webtest/http.py"""
+class StoppableWSGIServer( ThreadingMixIn, WSGIServer ):
+  daemon_threads = False
 
-  shutdown_requested = False
-
-  def Run( self ):
-    """Wrapper of TcpWSGIServer run method. It prevents a traceback from
-    asyncore."""
-
-    # Message for compatibility with clients who expect the output from
-    # waitress.serve here
-    if sys.stdin is not None:
-      print( f'serving on http://{ self.effective_host }:'
-             f'{ self.effective_port }' )
-
-    try:
-      self.run()
-    except select.error:
-      if not self.shutdown_requested:
-        raise
-
-
-  def Shutdown( self ):
-    """Properly shutdown the server."""
-    self.shutdown_requested = True
-    # Shutdown waitress threads.
-    self.task_dispatcher.shutdown()
-    # Close asyncore channels.
-    # We use list() here because _map is modified while looping through it.
-    # NOTE: _map is an attribute from the asyncore.dispatcher class, which is a
-    # base class of TcpWSGIServer. This may change in future versions of
-    # waitress so extra care should be taken when updating waitress.
-    for channel in list( self._map.values() ):
-      channel.close()
+  def __init__( self, app, host, port ):
+    super().__init__( ( host, port ), WSGIRequestHandler )
+    self.set_app( app )
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
 [submodule "third_party/bottle"]
 	path = third_party/bottle
 	url = https://github.com/defnull/bottle
-[submodule "third_party/waitress"]
-	path = third_party/waitress
-	url = https://github.com/Pylons/waitress
 [submodule "third_party/jedi"]
 	path = third_party/jedi_deps/jedi
 	url = https://github.com/davidhalter/jedi
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,4 @@
 bottle   >= 0.12.18
 regex    >= 2020.2.20
 jedi     >= 0.16.0
-requests >= 2.22.0
-waitress >= 1.4.3
 watchdog >= 0.10.2
--- a/ycmd/__main__.py
+++ b/ycmd/__main__.py
@@ -200,9 +200,13 @@ def Main():
   CloseStdin()
   handlers.wsgi_server = StoppableWSGIServer( handlers.app,
                                               host = args.host,
-                                              port = args.port,
-                                              threads = 30 )
-  handlers.wsgi_server.Run()
+                                              port = args.port )
+  if sys.stdin is not None:
+    print( f'serving on http://{ handlers.wsgi_server.server_name }:'
+           f'{ handlers.wsgi_server.server_port }' )
+  handlers.wsgi_server.serve_forever()
+  handlers.wsgi_server.server_close()
+  handlers.ServerCleanup()
 
 
 if __name__ == "__main__":
--- a/ycmd/handlers.py
+++ b/ycmd/handlers.py
@@ -306,7 +306,6 @@ def DebugInfo():
 def Shutdown():
   LOGGER.info( 'Received shutdown request' )
   ServerShutdown()
-
   return _JsonResponse( True )
 
 
@@ -369,7 +368,7 @@ def _GetCompleterForRequestData( request_data ):
 def ServerShutdown():
   def Terminator():
     if wsgi_server:
-      wsgi_server.Shutdown()
+      wsgi_server.shutdown()
 
   # Use a separate thread to let the server send the response before shutting
   # down.


bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 0 of 2 LGTMs obtained

___
*[ycmd/wsgi_server.py, line 29 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MNQbHu06diWK-COFRhJ:-MNQbHu15UmYrE77Ay7K:b-vptt9o) ([raw file](https://github.com/ycm-core/ycmd/blob/3448f0d8bcc77479ab117c2ed2ffd324eccdbc69/ycmd/wsgi_server.py#L29)):*
> ```Python
> 
>   shutdown_requested = False
>   daemon_threads = True
> ```

Do we want daemonized threads?
___
*[ycmd/wsgi_server.py, line 54 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MNQb6GG3vBGTUOYNmcD:-MNQb6GG3vBGTUOYNmcE:br0wuf) ([raw file](https://github.com/ycm-core/ycmd/blob/3448f0d8bcc77479ab117c2ed2ffd324eccdbc69/ycmd/wsgi_server.py#L54)):*
> ```Python
>     """Properly shutdown the server."""
>     self.shutdown_requested = True
>     self.server_close()
> ```

This is supposed to clean up before shutdown. As it is written now, with `daemon_threads=True`, this does exactly nothing.

ycmd/wsgi_server.py
```diff
@@ -15,27 +15,34 @@
...
+from wsgiref.simple_server import WSGIServer, WSGIRequestHandler
+from socketserver import ThreadingMixIn
 import select
 
 
-class StoppableWSGIServer( TcpWSGIServer ):
+class StoppableWSGIServer( ThreadingMixIn, WSGIServer ):
   """StoppableWSGIServer is a subclass of the TcpWSGIServer Waitress server
   with a shutdown method. It is based on StopableWSGIServer class from webtest:
   https://github.com/Pylons/webtest/blob/master/webtest/http.py"""
```
inglor, line 26: Update the comment, or better remove it
```suggestion
  """StoppableWSGIServer extends WSGIServer by adding a shutdown method."""
```

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 0 of 2 LGTMs obtained (waiting on @inglor)

___
*[ycmd/wsgi_server.py, line 26 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MNQadCD87bsMWUfBI2J-r2-26:-MN__5-r6vd8A86sWCfJ:bovu45w) ([raw file](https://github.com/ycm-core/ycmd/blob/4e7ce5429a1dcc7702f67daae1f0fe0a64164de0/ycmd/wsgi_server.py#L26)):*
<details><summary><i>Previously, inglor (Leonidas Spyropoulos) wrote…</i></summary><blockquote>

Update the comment, or better remove it
```suggestion
  """StoppableWSGIServer extends WSGIServer by adding a shutdown method."""
```

</blockquote></details>

Comment removed.
___
*[ycmd/wsgi_server.py, line 38 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MN__9Bp466nKsR2lzLU:-MN__9Bp466nKsR2lzLV:b-n80qgl) ([raw file](https://github.com/ycm-core/ycmd/blob/4e7ce5429a1dcc7702f67daae1f0fe0a64164de0/ycmd/wsgi_server.py#L38)):*
> ```Python
>   def Run( self ):
>     """Wrapper of TcpWSGIServer run method. It prevents a traceback from
>     asyncore."""
> ```

Do we need the `Run` wrapper? What even is "asyncore"?

inglor (commented):
> Reviewed 2 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[ycmd/wsgi_server.py, line 29 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MNQbHu06diWK-COFRhJ:-MNbndf35Get1k2Tuo1w:bgiibtx) ([raw file](https://github.com/ycm-core/ycmd/blob/3448f0d8bcc77479ab117c2ed2ffd324eccdbc69/ycmd/wsgi_server.py#L29)):*

Do we want daemonized threads?
</blockquote></details>

If we are shutting down after probably not. You'd only want to daemonize if you care about what happens to them _after_ you shut down.

Might worth ditching completely this class and try by just creating a new `WSGIServer`, what do you think?
```
server = wsgiserver.WSGIServer(my_app)
server.start()
```
___
*[ycmd/wsgi_server.py, line 38 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MN__9Bp466nKsR2lzLU:-MNbmuYO1wkitAlaVC4W:b-nzeeud) ([raw file](https://github.com/ycm-core/ycmd/blob/4e7ce5429a1dcc7702f67daae1f0fe0a64164de0/ycmd/wsgi_server.py#L38)):*
> Do we need the Run wrapper? What even is "asyncore"?

Apart from the `print` statement, which only makes sense if I read the comment above, I don't think this whole class serves a purpose. As for `asyncore` I think it might refer to https://docs.python.org/3/library/asyncore.html (?)

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @inglor)

___
*[ycmd/wsgi_server.py, line 38 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MN__9Bp466nKsR2lzLU:-MOx6sQPAO7JcTIp-pg3:b-kiquu5) ([raw file](https://github.com/ycm-core/ycmd/blob/4e7ce5429a1dcc7702f67daae1f0fe0a64164de0/ycmd/wsgi_server.py#L38)):*
>  I don't think this whole class serves a purpose.

You missed that I'm not only only wrapping `WSGIServer`, but also adding the `ThreadingMixIn`, to have a threaded server.

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @inglor)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MPocHLr3h6a8w6DWsZ6:-MPocHLr3h6a8w6DWsZ7:b36md3m) (no related file):*
I don't really know anything about the background to this stuff, bu tin principle I'm happy to ditch a non-stdlib dependency in favour of a standard library one.

I need to read what the various new things are to understand this a bit better before approval if that' ok

---

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic, @inglor, and @puremourning)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MPocHLr3h6a8w6DWsZ6:-MPohuvoDVrO3X1pCcZd:b-ea5jut) (no related file):*

I don't really know anything about the background to this stuff, bu tin principle I'm happy to ditch a non-stdlib dependency in favour of a standard library one.

I need to read what the various new things are to understand this a bit better before approval if that' ok
</blockquote></details>

Take your time!

---

inglor (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[ycmd/wsgi_server.py, line 54 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MNQb6GG3vBGTUOYNmcD:-MPp-kVuCA0t_0X6tz9-:bb6ywmx) ([raw file](https://github.com/ycm-core/ycmd/blob/3448f0d8bcc77479ab117c2ed2ffd324eccdbc69/ycmd/wsgi_server.py#L54)):*

This is supposed to clean up before shutdown. As it is written now, with `daemon_threads=True`, this does exactly nothing.
</blockquote></details>

LGTM
___
*[ycmd/wsgi_server.py, line 38 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MN__9Bp466nKsR2lzLU:-MPp-tNB48ql8mmTvGMV:b32w0x) ([raw file](https://github.com/ycm-core/ycmd/blob/4e7ce5429a1dcc7702f67daae1f0fe0a64164de0/ycmd/wsgi_server.py#L38)):*

>  I don't think this whole class serves a purpose.

You missed that I'm not only only wrapping `WSGIServer`, but also adding the `ThreadingMixIn`, to have a threaded server.
</blockquote></details>

OK

inglor (commented):
:lgtm:

> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509#-:-MT_WXve3fznd4J9dI4l:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[ycmd/wsgi_server.py, line 29 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MNQbHu06diWK-COFRhJ:-MTa2hoLEoQsxqYxOPC4:b-8ecgfi) ([raw file](https://github.com/ycm-core/ycmd/blob/3448f0d8bcc77479ab117c2ed2ffd324eccdbc69/ycmd/wsgi_server.py#L29)):*
<details><summary><i>Previously, inglor (Leonidas Spyropoulos) wrote…</i></summary><blockquote>

If we are shutting down after probably not. You'd only want to daemonize if you care about what happens to them _after_ you shut down.

Might worth ditching completely this class and try by just creating a new `WSGIServer`, what do you think?
```
server = wsgiserver.WSGIServer(my_app)
server.start()
```

</blockquote></details>

by setting `daemon_threads` we are quitting ycmd without waiting for the actual request threads to exit. won't this lead to flakiness ? I feel like we should have `daemon_threads = False` and that we should be able to shut down cleanly.

Looking down the code we are calling `server_close()` _then_ calling `shutdown()`. `server_close()` is supposed to wait for the threads to complete, but it's `shutdown()` that's supposed to make `serve_forever()` return.

I'm not 100% sure which sequence they need to be called in but it stands to reason that the sequence we have is correct.
___
*[ycmd/wsgi_server.py, line 28 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MTa1gr06hgoTNIRhv4-:-MTa1gr06hgoTNIRhv40:b1on4fp) ([raw file](https://github.com/ycm-core/ycmd/blob/0754a1fc4ce0106c9b7218423d3eee9fefd75c88/ycmd/wsgi_server.py#L28)):*
> ```Python
>   daemon_threads = True
> 
>   def __init__( self, app, host, port, threads ):
> ```

we seem to be ignoring the `threads` argument
___
*[ycmd/wsgi_server.py, line 44 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MTa1qAI2ZdTXKj1NhZX:-MTa1qAI2ZdTXKj1NhZY:b80nbwo) ([raw file](https://github.com/ycm-core/ycmd/blob/0754a1fc4ce0106c9b7218423d3eee9fefd75c88/ycmd/wsgi_server.py#L44)):*
> ```Python
> 
>     try:
>       self.serve_forever()
> ```

maybe we could replace this with

```python
while not self.shutdown_requested:
  self.handle_request()

self.server_close()
```

I think that gets rid of a lot of finicky stuff here like handling a `select` error which doesn't seem to be exactly part of the api ? According to the docs `self.server_close()` is supposed to make `serve_forever()` return, not throw an exception, so when do we hit the `except` here?

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[ycmd/wsgi_server.py, line 29 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MNQbHu06diWK-COFRhJ:-MTa5jBFDH9lLqrmoj2b:bwndzqd) ([raw file](https://github.com/ycm-core/ycmd/blob/3448f0d8bcc77479ab117c2ed2ffd324eccdbc69/ycmd/wsgi_server.py#L29)):*

by setting `daemon_threads` we are quitting ycmd without waiting for the actual request threads to exit. won't this lead to flakiness ? I feel like we should have `daemon_threads = False` and that we should be able to shut down cleanly.

Looking down the code we are calling `server_close()` _then_ calling `shutdown()`. `server_close()` is supposed to wait for the threads to complete, but it's `shutdown()` that's supposed to make `serve_forever()` return.

I'm not 100% sure which sequence they need to be called in but it stands to reason that the sequence we have is correct.



</blockquote></details>

The sequence is correct, as far as I remember. The `server_close()` thing only does something when threads are not daemonized. I'll set `daemon_threads` to `False`.
___
*[ycmd/wsgi_server.py, line 28 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MTa1gr06hgoTNIRhv4-:-MTa5v2c7tTN776iMQAK:bie1prc) ([raw file](https://github.com/ycm-core/ycmd/blob/0754a1fc4ce0106c9b7218423d3eee9fefd75c88/ycmd/wsgi_server.py#L28)):*

we seem to be ignoring the `threads` argument
</blockquote></details>

We are, because it's not settable. Should I drop it?
___
*[ycmd/wsgi_server.py, line 44 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MTa1qAI2ZdTXKj1NhZX:-MTa67yF4WYj7jbCXEB5:b1t3z3t) ([raw file](https://github.com/ycm-core/ycmd/blob/0754a1fc4ce0106c9b7218423d3eee9fefd75c88/ycmd/wsgi_server.py#L44)):*

maybe we could replace this with

```python
while not self.shutdown_requested:
  self.handle_request()

self.server_close()
```

I think that gets rid of a lot of finicky stuff here like handling a `select` error which doesn't seem to be exactly part of the api ? According to the docs `self.server_close()` is supposed to make `serve_forever()` return, not throw an exception, so when do we hit the `except` here?


</blockquote></details>

I have never really understood the `select` error that we have already been handling. I've left it in because "it seems to work". I'll test your suggestion.

However, I'm also not sure what exactly are you suggesting. If I replace the whole `try:`/`except` block with `while not self.shutdown_requested`, where am I supposed to call `self.serve_forever()`?

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @inglor and @puremourning)

___
*[ycmd/wsgi_server.py, line 29 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MNQbHu06diWK-COFRhJ:-MTwJOW72djUjsBCVTQC:b-tfbw2o) ([raw file](https://github.com/ycm-core/ycmd/blob/3448f0d8bcc77479ab117c2ed2ffd324eccdbc69/ycmd/wsgi_server.py#L29)):*

The sequence is correct, as far as I remember. The `server_close()` thing only does something when threads are not daemonized. I'll set `daemon_threads` to `False`.
</blockquote></details>

I can't make ycmd showdown the subservers at all if I set `daemon_threads` to `False`...

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic, @inglor, and @puremourning)

___
*[ycmd/wsgi_server.py, line 29 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MNQbHu06diWK-COFRhJ:-MU067ffC5WKV_LTMMnO:b397wj6) ([raw file](https://github.com/ycm-core/ycmd/blob/3448f0d8bcc77479ab117c2ed2ffd324eccdbc69/ycmd/wsgi_server.py#L29)):*

I can't make ycmd showdown the subservers at all if I set `daemon_threads` to `False`...
</blockquote></details>

hmmm. I wonder if using our own handling loop would make this easier too. Currently we're using some `atexit` hooks to actually trigger the `ServerCleanup` (which calls the shutdown on each subsever).

This has always been kind of brittle. If we were in control of the "main" loop, we can control this our own way, as in, doing something like this in __main__ rather than `server.Run()`

```python
while not shutdown_requested:
  server.handle_request() # handle one request and return (presumably, this despatches to a thread

# Loop finished - we should shutdown
# shutdown request threads
self.server_close()
# shutdown wigs server
self.shutdown()
# shutdown subserves
handlers.CleanupServer()
```

would this work... and simplify things ?
___
*[ycmd/wsgi_server.py, line 28 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MTa1gr06hgoTNIRhv4-:-MU06On46HReBj94lVi_:bbw6wys) ([raw file](https://github.com/ycm-core/ycmd/blob/0754a1fc4ce0106c9b7218423d3eee9fefd75c88/ycmd/wsgi_server.py#L28)):*

We are, because it's not settable. Should I drop it?
</blockquote></details>

so we no longer can control the thread pool size for the server ? That seems... strange.. how many threads does it use?
___
*[ycmd/wsgi_server.py, line 44 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MTa1qAI2ZdTXKj1NhZX:-MU06EtyE9hebqyyIoM3:b-ml6bph) ([raw file](https://github.com/ycm-core/ycmd/blob/0754a1fc4ce0106c9b7218423d3eee9fefd75c88/ycmd/wsgi_server.py#L44)):*

I have never really understood the `select` error that we have already been handling. I've left it in because "it seems to work". I'll test your suggestion.

However, I'm also not sure what exactly are you suggesting. If I replace the whole `try:`/`except` block with `while not self.shutdown_requested`, where am I supposed to call `self.serve_forever()`?
</blockquote></details>

if using a while 'handle request' loop, you don't call server_forever(), as that's what the loop does.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic, @inglor, and @puremourning)

___
*[ycmd/wsgi_server.py, line 29 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MNQbHu06diWK-COFRhJ:-MU0Cymu34Bzzkhm85qh:b-h6elth) ([raw file](https://github.com/ycm-core/ycmd/blob/3448f0d8bcc77479ab117c2ed2ffd324eccdbc69/ycmd/wsgi_server.py#L29)):*

hmmm. I wonder if using our own handling loop would make this easier too. Currently we're using some `atexit` hooks to actually trigger the `ServerCleanup` (which calls the shutdown on each subsever).

This has always been kind of brittle. If we were in control of the "main" loop, we can control this our own way, as in, doing something like this in __main__ rather than `server.Run()`

```python
while not shutdown_requested:
  server.handle_request() # handle one request and return (presumably, this despatches to a thread

# Loop finished - we should shutdown
# shutdown request threads
self.server_close()
# shutdown wigs server
self.shutdown()
# shutdown subserves
handlers.CleanupServer()
```

would this work... and simplify things ?
</blockquote></details>

Doing exactly that does start, but doesn't shut down. I'll push the changes if you want to take a look. In the mean time, I'll be having dinner.
___
*[ycmd/wsgi_server.py, line 28 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MTa1gr06hgoTNIRhv4-:-MU0DYHqEDXJ3SXp8M27:bh5cpt2) ([raw file](https://github.com/ycm-core/ycmd/blob/0754a1fc4ce0106c9b7218423d3eee9fefd75c88/ycmd/wsgi_server.py#L28)):*

so we no longer can control the thread pool size for the server ? That seems... strange.. how many threads does it use?
</blockquote></details>

It doesn't have a limited size. It spawns a new thread and, when calling `_threads.append(new_thread)`, `append()` calls `purge()`, which removes the threads that have finished already.

https://github.com/python/cpython/blob/master/Lib/socketserver.py#L631-L650

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic, @inglor, and @puremourning)

___
*[ycmd/wsgi_server.py, line 29 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MNQbHu06diWK-COFRhJ:-MU0EuxmDXyGx_1I30gz:b7glkg9) ([raw file](https://github.com/ycm-core/ycmd/blob/3448f0d8bcc77479ab117c2ed2ffd324eccdbc69/ycmd/wsgi_server.py#L29)):*

Doing exactly that does start, but doesn't shut down. I'll push the changes if you want to take a look. In the mean time, I'll be having dinner.
</blockquote></details>

Looks like `StoppableWSGIServer.Shutdown()` isn't getting called for some reason.

puremourning (commented):
> Reviewed 1 of 3 files at r1, 1 of 2 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic, @inglor, and @puremourning)

___
*[ycmd/wsgi_server.py, line 29 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MNQbHu06diWK-COFRhJ:-MU0Q5lND8hasROOWutP:b7i9d4w) ([raw file](https://github.com/ycm-core/ycmd/blob/3448f0d8bcc77479ab117c2ed2ffd324eccdbc69/ycmd/wsgi_server.py#L29)):*

Looks like `StoppableWSGIServer.Shutdown()` isn't getting called for some reason.
</blockquote></details>

Shutdown() is supposed to be called by `handlers.ServerShutdown()`. This is called by:

* The /shutdown handler (though in another thread)
* The watchdog (also in another thread)

I'm not seeing an obvious reason why that would stop happening! 
___
*[ycmd/wsgi_server.py, line 28 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MTa1gr06hgoTNIRhv4-:-MU0Qw3b3zGtDYUQAGyk:b4f4dyy) ([raw file](https://github.com/ycm-core/ycmd/blob/0754a1fc4ce0106c9b7218423d3eee9fefd75c88/ycmd/wsgi_server.py#L28)):*

It doesn't have a limited size. It spawns a new thread and, when calling `_threads.append(new_thread)`, `append()` calls `purge()`, which removes the threads that have finished already.

https://github.com/python/cpython/blob/master/Lib/socketserver.py#L631-L650
</blockquote></details>

ok so it's unbounded, and not pooled?

inglor (commented):
@bstaletic Should you also update the `requirements.txt` file to remove `waitress`?

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509#-:-MU353hJAMRrUPenw7A5:b-fkx8u1)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic, @inglor, and @puremourning)

bstaletic (commented):
Done. Thanks for the reminder.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509#-:-MU3HJvr-gMO7I5C3qJB:bwppva0)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @inglor and @puremourning)

___
*[ycmd/handlers.py, line 308 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MU3IzlM-uRanoNS9UfT:-MU3IzlM-uRanoNS9UfU:b-knsj4q) ([raw file](https://github.com/ycm-core/ycmd/blob/b668b6615204bee8f3d02b2656ce45049015e8f7/ycmd/handlers.py#L308)):*
> ```Python
> def Shutdown():
>   LOGGER.info( 'Received shutdown request' )
>   ServerShutdown()
> ```

Still calls `ServerShutown()`.
___
*[ycmd/handlers.py, line 375 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MU3J8Vn21XeX9-nn9Bz:-MU3J8Vn21XeX9-nn9C-:b-az0lur) ([raw file](https://github.com/ycm-core/ycmd/blob/b668b6615204bee8f3d02b2656ce45049015e8f7/ycmd/handlers.py#L375)):*
> ```Python
>   # Use a separate thread to let the server send the response before shutting
>   # down.
>   StartThread( Terminator )
> ```

This actually calls `wsgi_server.shutdown()`. If we don't do it in a separate thread, we don't have the time to respond to `/shutdown` request. Just like the comment says.

However, this still feels racy and it always was. There's no guarantee that the thread in which the handler was called is going to return before the `Terminator` thread requests shutdown and the main server thread shuts down the server. To back up these claims, I remember us changing either YCM or the shutdown tests to handle the case where the response to `/shutdown` doesn't happen.

This feels like a job for `threading.Event`, but there's just no place to put it... unless `bottle` provides some callback invocation thing after it finishes its job?
___
*[ycmd/wsgi_server.py, line 29 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MNQbHu06diWK-COFRhJ:-MU3HOaI-NExA4JIQ4to:b34irdf) ([raw file](https://github.com/ycm-core/ycmd/blob/3448f0d8bcc77479ab117c2ed2ffd324eccdbc69/ycmd/wsgi_server.py#L29)):*

Shutdown() is supposed to be called by `handlers.ServerShutdown()`. This is called by:

* The /shutdown handler (though in another thread)
* The watchdog (also in another thread)

I'm not seeing an obvious reason why that would stop happening! 
</blockquote></details>

After chasing deadlocks and race conditions for a while, I realized a few things. If we want to stick to the `while not shutdown_requested`, we'd have to reinvent the whole `serve_forever()` mechanism.

For that reason, I made the following changes:

- `__main__` is calling `wsgi_server.serve_forever()`.
- `/shutdown` handler calls `wsgi_server.shutdown()`, requesting `serve_forever()` to finish.
- `__main__` is then calling `wsgi_server.close_server()` to clean up and then `handlers.CleanupServer()` to shut down the subservers.

This locally seems to work without problems. CI pending.
___
*[ycmd/wsgi_server.py, line 28 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MTa1gr06hgoTNIRhv4-:-MU0jgcT89a5XXNE50ym:bub4gea) ([raw file](https://github.com/ycm-core/ycmd/blob/0754a1fc4ce0106c9b7218423d3eee9fefd75c88/ycmd/wsgi_server.py#L28)):*

ok so it's unbounded, and not pooled?
</blockquote></details>

Right. Should I replace `ThreadingMixIn` with something based on `ThreadPoolExecutor`?
___
*[ycmd/wsgi_server.py, line 44 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MTa1qAI2ZdTXKj1NhZX:-MU3IYoIEmnyxMkhiCvb:b-rbrkas) ([raw file](https://github.com/ycm-core/ycmd/blob/0754a1fc4ce0106c9b7218423d3eee9fefd75c88/ycmd/wsgi_server.py#L44)):*

if using a while 'handle request' loop, you don't call server_forever(), as that's what the loop does.
</blockquote></details>

Like explained in the other thread, I couldn't figure out hhow t do that without replicating the entire `serve_forever()` mechanism.

puremourning (changes requested):
I like this simplification, thanks!

:lgtm: with one minor comment.

> Reviewed 1 of 3 files at r1, 4 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509#-:-MXI0XdA2GGos_H3rXLH:brovntq)* status: 1 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic)

___
*[ycmd/handlers.py, line 375 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MU3J8Vn21XeX9-nn9Bz:-MXI07On1dT2vh7UczkO:b-2vtz8i) ([raw file](https://github.com/ycm-core/ycmd/blob/b668b6615204bee8f3d02b2656ce45049015e8f7/ycmd/handlers.py#L375)):*

This actually calls `wsgi_server.shutdown()`. If we don't do it in a separate thread, we don't have the time to respond to `/shutdown` request. Just like the comment says.

However, this still feels racy and it always was. There's no guarantee that the thread in which the handler was called is going to return before the `Terminator` thread requests shutdown and the main server thread shuts down the server. To back up these claims, I remember us changing either YCM or the shutdown tests to handle the case where the response to `/shutdown` doesn't happen.

This feels like a job for `threading.Event`, but there's just no place to put it... unless `bottle` provides some callback invocation thing after it finishes its job?
</blockquote></details>

This is fine - no less broken than before. The only option would be to sleep in the Terminator() thread, but meh, why bother.
___
*[ycmd/wsgi_server.py, line 25 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MXI0PagAO7cCEcob1sf:-MXI0PagAO7cCEcob1sg:bwct9pc) ([raw file](https://github.com/ycm-core/ycmd/blob/b668b6615204bee8f3d02b2656ce45049015e8f7/ycmd/wsgi_server.py#L25)):*
> ```Python
>   daemon_threads = False
> 
>   def __init__( self, app, host, port, threads ):
> ```

I think we should remove the `threads` argument. It's not used, so we don't need it.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509)* status: 1 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic and @puremourning)

___
*[ycmd/handlers.py, line 375 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MU3J8Vn21XeX9-nn9Bz:-MXI4yfq3CUQ58N-fcBI:b1f1jku) ([raw file](https://github.com/ycm-core/ycmd/blob/b668b6615204bee8f3d02b2656ce45049015e8f7/ycmd/handlers.py#L375)):*

This is fine - no less broken than before. The only option would be to sleep in the Terminator() thread, but meh, why bother.
</blockquote></details>

It also isn't *more* broken than before.
___
*[ycmd/wsgi_server.py, line 25 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1509#-MXI0PagAO7cCEcob1sf:-MXI5dVTEIGHE8-HvlWf:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/b668b6615204bee8f3d02b2656ce45049015e8f7/ycmd/wsgi_server.py#L25)):*

I think we should remove the `threads` argument. It's not used, so we don't need it.
</blockquote></details>

Done.

bstaletic (commented):
Note to self: Finish a rebase before pushing.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509#-:-MXI7599DLVuUQt_-iNX:b-dly33r)* status: 0 of 2 LGTMs obtained (and 2 stale) (waiting on @bstaletic and @puremourning)

puremourning (approved):
:lgtm:

> Reviewed 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1509#-:-MXI7rPr69VzW1H9nAr8:bnfp4nl)* status: 1 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic and @puremourning)

PR #1515, opened 2020-12-17
[READY] Use raw CPython API in PythonSupport.cpp

bstaletic:
Using the raw C API allows access to objects internals without checking
for errors. This is fine, because ycmd already checks everything
necessary. Plus, pybind11 can easily fall back to generic
implementations - for example, `len()` is `PyObject_Len()`, even if we
know that the object is `PyListObject*`. The C API, thus, allows us to
exploit the knowledge of types that pybind11 doesn't have.

And yes, I understand that this might be a controversial pull request. The performance improvements are from 3% to 10%, depending on the specific benchmark.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1515)

### DIFF
--- a/cpp/ycm/PythonSupport.cpp
+++ b/cpp/ycm/PythonSupport.cpp
@@ -24,33 +24,27 @@
 #include <utility>
 #include <vector>
 
-using pybind11::len;
-using pybind11::str;
-using pybind11::bytes;
-using pybind11::object;
-using pybind11::isinstance;
-using pylist = pybind11::list;
-
 namespace YouCompleteMe {
 
 namespace {
 
 std::vector< const Candidate * > CandidatesFromObjectList(
-  pylist candidates,
-  const std::string &candidate_property ) {
-  size_t num_candidates = len( candidates );
+  const pybind11::list& candidates,
+  pybind11::str candidate_property,
+  size_t num_candidates ) {
   std::vector< std::string > candidate_strings;
   candidate_strings.reserve( num_candidates );
-  // Store the property in a native Python string so that the below doesn't need
-  // to reconvert over and over:
-  str py_prop( candidate_property );
 
-  for ( size_t i = 0; i < num_candidates; ++i ) {
-    if ( candidate_property.empty() ) {
-      candidate_strings.emplace_back( GetUtf8String( candidates[ i ] ) );
-    } else {
-      candidate_strings.emplace_back( GetUtf8String(
-                                        candidates[ i ][ py_prop ] ) );
+  if ( !PyUnicode_GET_LENGTH( candidate_property.ptr() ) ) {
+    for ( size_t i = 0; i < num_candidates; ++i ) {
+        candidate_strings.emplace_back(
+            GetUtf8String( PyList_GET_ITEM( candidates.ptr(), i ) ) );
+    }
+  } else {
+    for ( size_t i = 0; i < num_candidates; ++i ) {
+        auto element = PyDict_GetItem( PyList_GET_ITEM( candidates.ptr(), i ),
+                                       candidate_property.ptr() );
+        candidate_strings.emplace_back( GetUtf8String( element ) );
     }
   }
 
@@ -61,16 +55,15 @@ std::vector< const Candidate * > CandidatesFromObjectList(
 } // unnamed namespace
 
 
-pylist FilterAndSortCandidates(
-  pylist candidates,
-  const std::string &candidate_property,
-  std::string query,
+pybind11::list FilterAndSortCandidates(
+  const pybind11::list& candidates,
+  pybind11::str candidate_property,
+  std::string& query,
   const size_t max_candidates ) {
-  pylist filtered_candidates;
 
-  size_t num_candidates = len( candidates );
+  size_t num_candidates = PyList_GET_SIZE( candidates.ptr() );
   std::vector< const Candidate * > repository_candidates =
-    CandidatesFromObjectList( candidates, candidate_property );
+    CandidatesFromObjectList( candidates, std::move( candidate_property ), num_candidates );
 
   std::vector< ResultAnd< size_t > > result_and_objects;
   {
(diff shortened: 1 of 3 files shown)


puremourning (changes requested):
I don't have a problem with using the c api directly

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1515#-:-MOwWGX41TJWMrSXVRJY:b1p89e)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[cpp/ycm/PythonSupport.h, line 37 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1515#-MOwVOkCCbBIoWbR4U92:-MOwVOkCCbBIoWbR4U93:b-v8d0ko) ([raw file](https://github.com/ycm-core/ycmd/blob/ab9024322e71db3211ca263b47fc94c65ed3a156/cpp/ycm/PythonSupport.h#L37)):*
> ```objc
>   const size_t max_candidates = 0 );
> 
> /// Given a Python object that's supposed to be "string-like", returns a UTF-8
> ```

is this still true ? "python object that's supposed to be 'string-like'" ?

at least should we say something about why it's a handle not an object?
___
*[cpp/ycm/PythonSupport.cpp, line 114 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1515#-MOwVtpR3wrd7gP5hy6q:-MOwVtpR3wrd7gP5hy6r:b-2qfxjz) ([raw file](https://github.com/ycm-core/ycmd/blob/ab9024322e71db3211ca263b47fc94c65ed3a156/cpp/ycm/PythonSupport.cpp#L114)):*
> ```cpp
>     ssize_t size = 0;
>     const char* buffer = nullptr;
>     buffer = PyUnicode_AsUTF8AndSize( value.ptr(), &size );
> ```

how does ownership of the returned pointer get dealt with here? (i.e. the data pointed to by `buffer`)
___
*[cpp/ycm/PythonSupport.cpp, line 127 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1515#-MOwW6oN0yfAg23eoxEu:-MOwW6oN0yfAg23eoxEv:b-a7sjot) ([raw file](https://github.com/ycm-core/ycmd/blob/ab9024322e71db3211ca263b47fc94c65ed3a156/cpp/ycm/PythonSupport.cpp#L127)):*
> ```cpp
>   ssize_t size = 0;
>   const char* buffer =
>       PyUnicode_AsUTF8AndSize( PyObject_Str( value.ptr() ), &size );
> ```

do we have to do any reference counting on the `PyObject_Str()` that's created here?

puremourning (commented):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1515)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1515)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[cpp/ycm/PythonSupport.h, line 37 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1515#-MOwVOkCCbBIoWbR4U92:-MOwqCOfE0_t7b8fvHKn:b88u2zf) ([raw file](https://github.com/ycm-core/ycmd/blob/ab9024322e71db3211ca263b47fc94c65ed3a156/cpp/ycm/PythonSupport.h#L37)):*
> is this still true ?

I *did not* change the behaviour in any way.

> "python object that's supposed to be 'string-like'" ?

Right, that's an odd phrasing, since this function happily deals with lists or basically anything.

I was wondering if we needed that "default" case...

> at least should we say something about why it's a handle not an object?

Because `py::handle` doesn't do reference counting. We're expecting the callers to keep their references alive for the duration of this function's call.
___
*[cpp/ycm/PythonSupport.cpp, line 114 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1515#-MOwVtpR3wrd7gP5hy6q:-MOwrAVl34WATeWOVHsE:b-1eucp7) ([raw file](https://github.com/ycm-core/ycmd/blob/ab9024322e71db3211ca263b47fc94c65ed3a156/cpp/ycm/PythonSupport.cpp#L114)):*

how does ownership of the returned pointer get dealt with here? (i.e. the data pointed to by `buffer`)
</blockquote></details>


- [The docs](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUTF8AndSize)
- [The source](https://github.com/python/cpython/blob/v3.9.0/Objects/unicodeobject.c#L4040)

If I understand correctly, `buffer` points to the `value`'s internal buffer. Then `std::string`'s constructor copies that buffer.

If the internal buffer isn't UTF8, the `value` would be mutated by re-encoding the data. However, unless we have a bug, this branch of cpython won't ever be hit. `GetUtf8String()` is only used in the depths of `ycm_core` where we expect all strings to be UTF8 already.
___
*[cpp/ycm/PythonSupport.cpp, line 127 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1515#-MOwW6oN0yfAg23eoxEu:-MOwuHAfE2ghbaa9CN9n:b6ehg7y) ([raw file](https://github.com/ycm-core/ycmd/blob/ab9024322e71db3211ca263b47fc94c65ed3a156/cpp/ycm/PythonSupport.cpp#L127)):*

do we have to do any reference counting on the `PyObject_Str()` that's created here?
</blockquote></details>

Oh, wow... That object needs to have a `Py_DECREF()` called on it *after* constructing the `std::string`. That's going to be awkward, so I'll do it with pybind11 and see if I can measure the difference.

Back from measuring. Once fixed, pybind11 overhead is ~0.7% on this function only. My current thoughts:

- This branch should be really cold in actual use.
- 0.7% overhead in `GetUtf8String()` should be drowned out by all the other things `FilterAndSortCandidates()` is doing.
- Code with `pybind::str` managing the lifetime of the object is much nicer.
- Our benchmarks don't ever hit this path. They only care about the `PyUnicode_AsUTF8AndSize()` branch.

Therefore, I allowed `pybind11::str` to manage the lifetime of that string object.

bstaletic:
Okay, finally fixed and ready for review.

puremourning (approved):
:lgtm:

> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1515#-:-MPu9Uzi5-8baY2QeEDJ:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[cpp/ycm/PythonSupport.cpp, line 114 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1515#-MOwVtpR3wrd7gP5hy6q:-MPu8aJ61zWFIYK4PCT1:b-y8qm4m) ([raw file](https://github.com/ycm-core/ycmd/blob/ab9024322e71db3211ca263b47fc94c65ed3a156/cpp/ycm/PythonSupport.cpp#L114)):*
You had me at:

> The caller is not responsible for deallocating the buffer
___
*[cpp/ycm/PythonSupport.cpp, line 127 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1515#-MOwW6oN0yfAg23eoxEu:-MPu8EpC7ovtWqNoEAV4:bp7vbga) ([raw file](https://github.com/ycm-core/ycmd/blob/ab9024322e71db3211ca263b47fc94c65ed3a156/cpp/ycm/PythonSupport.cpp#L127)):*

Oh, wow... That object needs to have a `Py_DECREF()` called on it *after* constructing the `std::string`. That's going to be awkward, so I'll do it with pybind11 and see if I can measure the difference.

Back from measuring. Once fixed, pybind11 overhead is ~0.7% on this function only. My current thoughts:

- This branch should be really cold in actual use.
- 0.7% overhead in `GetUtf8String()` should be drowned out by all the other things `FilterAndSortCandidates()` is doing.
- Code with `pybind::str` managing the lifetime of the object is much nicer.
- Our benchmarks don't ever hit this path. They only care about the `PyUnicode_AsUTF8AndSize()` branch.

Therefore, I allowed `pybind11::str` to manage the lifetime of that string object.
</blockquote></details>

sounds good.

bstaletic (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1515#-:-MPuA_ZgEAxoxbqpfCwF:bnfp4nl)* status: :shipit: complete! 2 of 2 LGTMs obtained (waiting on @bstaletic)

puremourning (commented):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1515#-:-MPuHHio7eBUE3NeOYck:b-uwkowd)* status: 1 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic)

PR #1518, opened 2020-12-25
[READY] Add diagnostic support to python completer

bstaletic:
Thanks to @ysobolev for pointing out that `jedi.Script.get_syntax_errors()` exists.

To be honest, jedi's diagnostics are pretty dumb. It reports parsing errors, with the description being the last line of the exception. The start location is meaningful, but the end location is somewhat iffy.

~~No tests yet, mostly because I'm being lazy.~~ Tests added.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1518)

### DIFF
--- a/ycmd/completers/python/python_completer.py
+++ b/ycmd/completers/python/python_completer.py
@@ -17,7 +17,9 @@
 
 from ycmd import extra_conf_store, responses
 from ycmd.completers.completer import Completer, SignatureHelpAvailalability
-from ycmd.utils import ( CodepointOffsetToByteOffset,
+from ycmd.completers.completer_utils import GetFileLines
+from ycmd.utils import ( ByteOffsetToCodepointOffset,
+                         CodepointOffsetToByteOffset,
                          ExpandVariablesInPath,
                          FindExecutable,
                          LOGGER )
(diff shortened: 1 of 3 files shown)


puremourning (commented):
My slight concern with this is that the diagnostics are... kinda weak, and will now start fighting with people using something like neomake/etc. to use flake8 for diagnostics.

Personally, I would not find these Jedi diags useful and would have to go and suppress them somehow to ensure the ti still get my flake8 stuff.

WDYT?

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1518#-:-MPoaVs-4gDQ8lhRVifb:bsw5fcp)* status: 0 of 2 LGTMs obtained

bstaletic (commented):
I am worried about that as well. We *could* provide diagnostics through flake8 as well: https://flake8.pycqa.org/en/latest/user/python-api.html

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1518#-:-MPoi9zwEs_CUYbLC-RB:b-db2d1l)* status: 0 of 2 LGTMs obtained

ysobolev:
Flake8 diagnostics would be preferable. This is not exactly in line with the compiler errors produced for other languages, but I think it far more useful that what jedi outputs. You will likely start getting requests for other linters (such as pylint) afterward, though.

bstaletic:
The thing that worries me with using flake8 API is performance. Without incremental updates that would allow us to run flake8 on a limited set of files, flake8 would be a massive hog.

puremourning:
Indeed and the whole problem that there are just tons of python linters and everyone has their own preferences. Flake8 works for us because that’s what we use but there are many others. We’d end up having to do a lot of work to keep up and be duplicating functionality of other plugins which i don’t really like.

ysobolev:
Well, flake8 would certainly be helpful. However, if that's not going in, I'm not completely sure how important it is to add the jedi stuff. I think it is more nuisance than useful. If there is no flake8 support, the first thing I'm doing is disabling this.

bstaletic:
I think we've decided not to do this.

hassec:
@bstaletic, sorry for being late to the party, but I just found this wondering if YCM could give me diagnostics in python...  

> My slight concern with this is that the diagnostics are... kinda weak, and will now start fighting with people using something like neomake/etc. to use flake8 for diagnostics.  

Couldn't this be solved by just disabling the diagnostics for python in your config? Or  having it disabled by default, but allowing people that don't want to run flake8 or other linters incrementally, to enable the ycmd diagnostics to catch a few errors earlier?

Additionally, from what I gathered from davidhalter/jedi#1488 the `get_syntax_errors()` call is relatively new, but maybe with some feedback from this use case here, these diagnostics could become better?

PR #1519, opened 2020-12-25
[READY] Optimize IdentifierCompleter

bstaletic:
With this PR we can avoid some additional copies. The changes in the
python files allow pybind to preallocate `StringVector` objects.

NOTE: We are now moving the contents of `StringVector` objects once we
call into `ycm_core.so`. This doesn't make stuff blow up, but python
code should be aware that `StringVector` objects are mutable, just like
python lists.

A really ugly part of this PR is the use of mutable l-value references
to move data. This makes the code harder to reason about, but the reason
for it is that pybind11 doesn't support r-value reference parameters.
A possible solution would be:

```c++
template <typename T>
using force_rvalue_to_lvalue_conversion = std::conditional_t<std::is_rvalue_reference<T>, T&, T>;

template <typename Return, typename... Args>
constexpr auto steal_rvalues(Return (*f)(Args...)) {
        return [f](force_lvalue_to_rvalue_conversion<Args> &&... args) { return f(std::move(args)...); };
}
```

And then changing the bindings to:

    m.def("FilterAndSortCandidates", steal_rvalues(FilterAndSortCandidates));

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1519)

### DIFF
--- a/cpp/ycm/IdentifierCompleter.cpp
+++ b/cpp/ycm/IdentifierCompleter.cpp
@@ -42,7 +42,7 @@ IdentifierCompleter::IdentifierCompleter(
 
 
 void IdentifierCompleter::AddIdentifiersToDatabase(
-  std::vector< std::string > new_candidates,
+  std::vector< std::string >& new_candidates,
   std::string& filetype,
   std::string& filepath ) {
   identifier_database_.AddIdentifiers( std::move( new_candidates ),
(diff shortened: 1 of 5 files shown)


puremourning (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1519#-:-MPobRH12sjwVHQbCx03:bg7igkv)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1519#-:-MPog4lNCu0f2xjGt3XY:bnfp4nl)* status: :shipit: complete! 2 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic:
:lgtm_strong:

PR #1522, opened 2020-12-31
[READY] Add very simple documentSymbol support

puremourning:
We re-use the GoToSymbol handling for the "legacy" list-of-symbols type
response from documentSymbol and just populate the location list in the
same way that GoToSymbol does.

This is not exactly a drawer/document outline from a classic IDE but it
servers a purpose of getting a list of all the thigns int the file if
you're looking for something specific or just browsing.

Demo: https://asciinema.org/a/AtogDrr5ZXzfd3EgwKOU6tNZd

WIP cos I should write some tests

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1522)


bstaletic (approved):
:lgtm: 

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1522#-:-MPu5P60CpK9kaF3455P:bg7igkv)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/completers/language_server/language_server_completer.py, line 2420 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1522#-MPu5IvhEgean_znAiwn:-MPu5IvhEgean_znAiwo:bqj5qto) ([raw file](https://github.com/ycm-core/ycmd/blob/527fdb6f6372c9d385fc7b1ab9795d9f2a26ea2c/ycmd/completers/language_server/language_server_completer.py#L2420)):*
> ```Python
>     result = response.get( 'result' ) or []
> 
>     # We should only receive SymbolInformation (not DocumentSymbol)
> ```

Any reason we can't handle `DocumentSymbol`?

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1522)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1522#-MPtd0RF8bAmKqjOthe2:-MPtd0RF8bAmKqjOthe3:bb9110d) (no related file):*
Discussion point:

I named this `GoToDocumentOutline` but I'm not totally sure that's the best name. It's only a GoTo because that's an easy way to implement it.

I thought to call it 'DocumentOutline' but actually that sort of prevents us doing something 'better' in the future and calling it that' Using `GoTo` makes it clear that this is a list of jump locations, not a full document outline, so that's why I stuck with it. 

WDYT?

---
___
*[ycmd/completers/language_server/language_server_completer.py, line 2420 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1522#-MPu5IvhEgean_znAiwn:-MPu67jjAM1O6jucAXaZ:b2xnofd) ([raw file](https://github.com/ycm-core/ycmd/blob/527fdb6f6372c9d385fc7b1ab9795d9f2a26ea2c/ycmd/completers/language_server/language_server_completer.py#L2420)):*

Any reason we can't handle `DocumentSymbol`?
</blockquote></details>

only that the hierarchical version (DocumentSymbol) is just harder to convert to a GoTo list. The flat version is simpler to work with and older (so may be supported by more servers).

I think if we do this "properly" and make some sort of document outline in the UI we'd use the hierarchical version (DocumentSymbol), but for simple list of go-to targets, a flat list is preferred (SymbolInformation)

bstaletic (approved):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1522)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1522#-MPtd0RF8bAmKqjOthe2:-MPuAq-FEDTlNp4jNWap:bhiimog) (no related file):*

Discussion point:

I named this `GoToDocumentOutline` but I'm not totally sure that's the best name. It's only a GoTo because that's an easy way to implement it.

I thought to call it 'DocumentOutline' but actually that sort of prevents us doing something 'better' in the future and calling it that' Using `GoTo` makes it clear that this is a list of jump locations, not a full document outline, so that's why I stuck with it. 

WDYT?
</blockquote></details>

I personally like the response being flat, because we don't need to invent new UI features in vim. I guess something including `GoTo` in the name is fitting, though I'm not sure if calling it `(GoTo)DocumentOutline` is descriptive enough. I also don't think I can figure out a better name for it, right off the bat.

---

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1522)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1522#-MPtd0RF8bAmKqjOthe2:-MPuHMRgBPZCBX4aPDNo:bf8j4c0) (no related file):*

I personally like the response being flat, because we don't need to invent new UI features in vim. I guess something including `GoTo` in the name is fitting, though I'm not sure if calling it `(GoTo)DocumentOutline` is descriptive enough. I also don't think I can figure out a better name for it, right off the bat.
</blockquote></details>

Let's go with GoToDocumentOutline for now. We can probably change it up to the point we update the ycm submodule and update the docs :)

---

bstaletic (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1522)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

___
*[ycmd/completers/language_server/language_server_completer.py, line 2408 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1522#-MPuJdc7AAElniknCeB1:-MPuJdc7AAElniknCeB2:bcgehnl) ([raw file](https://github.com/ycm-core/ycmd/blob/99b2c2ecf374e41c6bf3d2991c130433a8c3f249/ycmd/completers/language_server/language_server_completer.py#L2408)):*
> ```Python
>   def GoToDocumentOutline( self, request_data ):
>     if not self.ServerIsReady():
>       raise RuntimeError( 'Server is initializing. Please wait.' )
> ```

This line should be easy to cover with tests. We already have tests that try to run subcommands with `ServerIsReady()` always returning `False`. This should satisfy codecov.

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1522)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic and @puremourning)

___
*[ycmd/completers/language_server/language_server_completer.py, line 2408 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1522#-MPuJdc7AAElniknCeB1:-MPuK2rB0pvHHhUAmI_z:bda1i54) ([raw file](https://github.com/ycm-core/ycmd/blob/99b2c2ecf374e41c6bf3d2991c130433a8c3f249/ycmd/completers/language_server/language_server_completer.py#L2408)):*

This line should be easy to cover with tests. We already have tests that try to run subcommands with `ServerIsReady()` always returning `False`. This should satisfy codecov.
</blockquote></details>

it should be covered now. I pushed an update to cover it (at least I thought I did)

bstaletic (approved):
Okay, my :lgtm: still stands.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1522#-:-MPuKjjc6fOXZvAf5Swn:b-yruhis)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

puremourning (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1522#-:-MPuLHrnANGNOmnbfZsk:bnfp4nl)* status: :shipit: complete! 2 of 2 LGTMs obtained (waiting on @puremourning)

PR #1523, opened 2021-01-07
[READY] More C++17 niceties

bstaletic:
Three small changes:

1. `GetValueElseInsert` now uses `try_emplace` instead of `insert`. This has two benefits:
1.1. We're not constructing a `pair<K,V>` before moving it into the map. The arguments are passed directly to placement `new`, so there are fewer copies.
1.2. We can `std::move()` the key as well, because `try_emplace` is nice enough not to move before checking that the key is actually not in the map already.
2. `Result` now has padding bytes at the end. That means we can potentially reclaim the padding bytes for `other_object` in `ResultAnd<T>` with `[[no_unique_address]]`.
2.1. To actually see the benefits of this, `vector<ResultAnd<size_t>>` in `PythonSupport.cpp` was changed to `vector<ResultAnd<uint32_t>>`. The size of `ResultAnd<uint32_t>` drops from 48 bytes to 40 bytes.
3. The only use of `ContainsKey` is dropped, because we were looking up a value in an `unordered_set` *twice*.
4. Finally, `IdentifierDatabase` got a few structured bindings.
4.1. Take a look at these more closely. I'm not sure all of them improve readability. Specifically, the one that says:

      auto&& [ _, paths_to_candidates ] = *it;

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1523)

### DIFF
--- a/cpp/ycm/IdentifierDatabase.cpp
+++ b/cpp/ycm/IdentifierDatabase.cpp
@@ -37,13 +37,11 @@ void IdentifierDatabase::AddIdentifiers(
   FiletypeIdentifierMap&& filetype_identifier_map ) {
   std::lock_guard locker( filetype_candidate_map_mutex_ );
 
-  for ( auto&& filetype_and_map : filetype_identifier_map ) {
-    for ( auto&& filepath_and_identifiers : filetype_and_map.second ) {
-      auto filetype = filetype_and_map.first;
-      auto filepath = filepath_and_identifiers.first;
-      AddIdentifiersNoLock( std::move( filepath_and_identifiers.second ),
-                            std::move( filetype ),
-                            std::move( filepath ) );
+  for ( auto&& [ filetype, paths_to_candidates ] : filetype_identifier_map ) {
+    for ( auto&& [ filepath, identifiers ] : paths_to_candidates ) {
+      AddIdentifiersNoLock( std::move( identifiers ),
+                            std::string( filetype ),
+                            std::string( filepath ) );
     }
   }
 }
@@ -88,8 +86,8 @@ std::vector< Result > IdentifierDatabase::ResultsForQueryAndType(
   {
     std::lock_guard locker( filetype_candidate_map_mutex_ );
     auto& paths_to_candidates = *it->second;
-    for ( const auto& path_and_candidates : paths_to_candidates ) {
-      for ( const Candidate * candidate : *path_and_candidates.second ) {
+    for ( const auto& [ _, candidates ] : paths_to_candidates ) {
+      for ( const Candidate * candidate : *candidates ) {
         if ( !seen_candidates.insert( candidate ).second ) {
           continue;
         }
--- a/cpp/ycm/IdentifierDatabase.cpp
+++ b/cpp/ycm/IdentifierDatabase.cpp
@@ -87,12 +87,12 @@ std::vector< Result > IdentifierDatabase::ResultsForQueryAndType(
 
   {
     std::lock_guard locker( filetype_candidate_map_mutex_ );
-    for ( const auto& path_and_candidates : *it->second ) {
+    auto& paths_to_candidates = *it->second;
+    for ( const auto& path_and_candidates : paths_to_candidates ) {
       for ( const Candidate * candidate : *path_and_candidates.second ) {
-        if ( ContainsKey( seen_candidates, candidate ) ) {
+        if ( !seen_candidates.insert( candidate ).second ) {
           continue;
         }
-        seen_candidates.insert( candidate );
 
         if ( candidate->IsEmpty() ||
              !candidate->ContainsBytes( query_object ) ) {
--- a/cpp/ycm/Utils.h
+++ b/cpp/ycm/Utils.h
@@ -61,19 +61,13 @@ YCM_EXPORT inline std::string Lowercase( std::string_view text ) {
 std::vector< std::string > ReadUtf8File( const fs::path &filepath );
 
 
-template <class Container, class Key>
+template <class Container, class Key, typename Value>
 typename Container::mapped_type &
 GetValueElseInsert( Container &container,
-                    const Key &key,
-                    typename Container::mapped_type &&value ) {
-  return container.insert(
-    typename Container::value_type( key, std::move( value ) ) ).first->second;
-}
-
-
-template <class Container, class Key>
-bool ContainsKey( Container &container, const Key &key ) {
-  return container.find( key ) != container.end();
+                    Key&& key,
+                    Value&& value ) {
+  return container.try_emplace(
+    std::forward< Key >( key ), std::forward< Value >( value ) ).first->second;
 }
 
 


bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1523)* status: 0 of 2 LGTMs obtained

___
*[cpp/ycm/IdentifierDatabase.cpp, line 44 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSaHLZEzHvw1J8BOFR:-MQSaHLZEzHvw1J8BOFS:b-rrdpk) ([raw file](https://github.com/ycm-core/ycmd/blob/2392bf90ecc7fd06c7ba8957ee09c2c2f3142c75/cpp/ycm/IdentifierDatabase.cpp#L44)):*
> ```cpp
>       AddIdentifiersNoLock( std::move( identifiers ),
>                             std::string( filetype ),
>                             std::string( filepath ) );
> ```

These structured bindings look good to me.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 80 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSaZBk37u93yZ4-MYM:-MQSaZBk37u93yZ4-MYN:b-gd3zh9) ([raw file](https://github.com/ycm-core/ycmd/blob/2392bf90ecc7fd06c7ba8957ee09c2c2f3142c75/cpp/ycm/IdentifierDatabase.cpp#L80)):*
> ```cpp
>     }
>   }
>   auto&& [ _, paths_to_candidates ] = *it;
> ```

I'm really not sure about this structured binding here... Maybe

    auto paths_to_candidates = it->second;

or just nothing at all and revert to `*it->second` in the outer loop below?
___
*[cpp/ycm/IdentifierDatabase.cpp, line 91 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSantR6ybK6B4LN2F0:-MQSantR6ybK6B4LN2F1:b2j2av8) ([raw file](https://github.com/ycm-core/ycmd/blob/2392bf90ecc7fd06c7ba8957ee09c2c2f3142c75/cpp/ycm/IdentifierDatabase.cpp#L91)):*
> ```cpp
>     for ( const auto& [ path, candidates ] : *paths_to_candidates ) {
>       for ( const Candidate * candidate : *candidates ) {
>         if ( auto&& [ _, new_candidate ] = seen_candidates.insert( candidate ); !new_candidate ) {
> ```

Questionable readability again. Which one is best?

```c++
        if ( auto&& [ _, new_candidate ] = seen_candidates.insert( candidate ); !new_candidate ) {
          continue;
        }
```
vs
```c++
        if ( bool new_candidate = seen_candidates.insert( candidate ).second; !new_candidate ) {
          continue;
        }
```
vs
```c++
        if ( !seen_candidates.insert( candidate ).second ) {
          continue;
        }
```
___
*[cpp/ycm/IdentifierDatabase.cpp, line 95 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSb4uZ1W9eIuPtnIoW:-MQSb4uZ1W9eIuPtnIoX:byg1wgd) ([raw file](https://github.com/ycm-core/ycmd/blob/7aa52117023a4ff5f76d919d6bc2b5f6feba62d8/cpp/ycm/IdentifierDatabase.cpp#L95)):*
> ```cpp
>           continue;
>         }
>         seen_candidates.insert( candidate );
> ```

In combination with `ContainsKey()`, this line was the second identical lookup.
___
*[cpp/ycm/PythonSupport.cpp, line 68 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSbGfgBxDHeQvyrNIn:-MQSbGfgBxDHeQvyrNIo:b-6pcxrq) ([raw file](https://github.com/ycm-core/ycmd/blob/2392bf90ecc7fd06c7ba8957ee09c2c2f3142c75/cpp/ycm/PythonSupport.cpp#L68)):*
> ```cpp
>     CandidatesFromObjectList( candidates, std::move( candidate_property ), num_candidates );
> 
>   std::vector< ResultAnd< uint32_t > > result_and_objects;
> ```

`Result` has 5 bytes of padding, allowing us to fit 4 bytes (plus alignment).
___
*[cpp/ycm/Result.h, line 117 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSbe3c2UlXyVRZKCOn:-MQSbe3c2UlXyVRZKCOo:b-hbv0p) ([raw file](https://github.com/ycm-core/ycmd/blob/2392bf90ecc7fd06c7ba8957ee09c2c2f3142c75/cpp/ycm/Result.h#L117)):*
> ```objc
>   }
> 
>   NO_UNIQUE_ADDRESS Result result_;
> ```

MSVC doesn't know how to reclaim these trailing padding bytes. That's why it unconditionally defines the macro to an empty string.
___
*[cpp/ycm/Utils.h, line 69 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSbuNN5Y3y_nuwhlHH:-MQSbuNN5Y3y_nuwhlHI:b-43k4us) ([raw file](https://github.com/ycm-core/ycmd/blob/2392bf90ecc7fd06c7ba8957ee09c2c2f3142c75/cpp/ycm/Utils.h#L69)):*
> ```objc
>                     Key&& key,
>                     Value&& value ) {
>   return container.try_emplace(
> ```

For the overview of C++17 STL map insertion API, check out
https://www.fluentcpp.com/2018/12/11/overview-of-std-map-insertion-emplacement-methods-in-cpp17/

bstaletic (commented):
Okay, 2nd commit breaks tests...

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1523#-:-MQSdRRw45RYipLCi9fC:btgu8ol)* status: 0 of 2 LGTMs obtained

bstaletic (commented):
Ready! I completely overlooked that we need the candidate name in both the key and the value of our repository's map. Copying into the key and moving into the value is a better idea than moving into the key, copying the key and then moving into the value!

Still, a few copies are saved.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1523#-:-MQSqlM7CHik23YRy1fJ:b9z96i0)* status: 0 of 2 LGTMs obtained

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1523)* status: 0 of 2 LGTMs obtained

___
*[cpp/ycm/Utils.h, line 70 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSrXKVCfRSNCoS6fxr:-MQSrXKVCfRSNCoS6fxs:bkt25f) ([raw file](https://github.com/ycm-core/ycmd/blob/be55966ea56c981862a39ae09ac3694116bcf99d/cpp/ycm/Utils.h#L70)):*
> ```objc
>                     Value&& value ) {
>   return container.try_emplace(
>     std::forward< Key >( key ), std::move( value ) ).first->second;
> ```

In practice, we never move anything on this line. `key` isn't moved because we always pass an l-value reference and `value` isn't moved because `nullptr` and `size_t` can not be moved.

Should we change anything here?

bstaletic:
Green!

puremourning (changes requested):
> Reviewed 3 of 11 files at r1, 2 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1523)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[cpp/ycm/IdentifierDatabase.cpp, line 80 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSaZBk37u93yZ4-MYM:-MQSu3fZ25foK7UVaDze:b-ccj93x) ([raw file](https://github.com/ycm-core/ycmd/blob/2392bf90ecc7fd06c7ba8957ee09c2c2f3142c75/cpp/ycm/IdentifierDatabase.cpp#L80)):*
yeah 

> auto paths_to_candidates = it->second;

this one for me
___
*[cpp/ycm/IdentifierDatabase.cpp, line 91 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSantR6ybK6B4LN2F0:-MQSu0MA10X4ldx5qodQ:b-736ydr) ([raw file](https://github.com/ycm-core/ycmd/blob/2392bf90ecc7fd06c7ba8957ee09c2c2f3142c75/cpp/ycm/IdentifierDatabase.cpp#L91)):*
>         if ( !seen_candidates.insert( candidate ).second ) {
>           continue;

this one for me
___
*[cpp/ycm/Result.h, line 117 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSbe3c2UlXyVRZKCOn:-MQSvANQ48_LA49n12S-:bxrobtq) ([raw file](https://github.com/ycm-core/ycmd/blob/2392bf90ecc7fd06c7ba8957ee09c2c2f3142c75/cpp/ycm/Result.h#L117)):*

MSVC doesn't know how to reclaim these trailing padding bytes. That's why it unconditionally defines the macro to an empty string.
</blockquote></details>

I'm not sure about this padding jiggery pokery. It seems fiddly and easily broken (i.e. we could add a member to result and blow it up again)

how much is it worth in practice ?
___
*[cpp/ycm/Result.h, line 26 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSuRr79cqaU6EokyRJ:-MQSuRr79cqaU6EokyRK:bxbdrg5) ([raw file](https://github.com/ycm-core/ycmd/blob/be55966ea56c981862a39ae09ac3694116bcf99d/cpp/ycm/Result.h#L26)):*
> ```objc
> 
> #if !defined( _MSC ) && __has_cpp_attribute(no_unique_address)
> #    define NO_UNIQUE_ADDRESS [[no_unique_address]]
> ```

aiui no_unique_address is C++ 20 ?
___
*[cpp/ycm/Utils.h, line 70 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSrXKVCfRSNCoS6fxr:-MQSyp5o7wXLd8LeInwC:bvs41ej) ([raw file](https://github.com/ycm-core/ycmd/blob/be55966ea56c981862a39ae09ac3694116bcf99d/cpp/ycm/Utils.h#L70)):*

In practice, we never move anything on this line. `key` isn't moved because we always pass an l-value reference and `value` isn't moved because `nullptr` and `size_t` can not be moved.

Should we change anything here?
</blockquote></details>

isn't it a case of std::forward( value ) rather than std::move( value) ?

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1523)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[cpp/ycm/IdentifierDatabase.cpp, line 80 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSaZBk37u93yZ4-MYM:-MQT-1634EMHd8JqzO8-:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/2392bf90ecc7fd06c7ba8957ee09c2c2f3142c75/cpp/ycm/IdentifierDatabase.cpp#L80)):*

yeah 

> auto paths_to_candidates = it->second;

this one for me

</blockquote></details>

Done.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 91 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSantR6ybK6B4LN2F0:-MQT-9yFEUogxe4wS5ot:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/2392bf90ecc7fd06c7ba8957ee09c2c2f3142c75/cpp/ycm/IdentifierDatabase.cpp#L91)):*

>         if ( !seen_candidates.insert( candidate ).second ) {
>           continue;

this one for me
</blockquote></details>

Done.
___
*[cpp/ycm/Result.h, line 117 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSbe3c2UlXyVRZKCOn:-MQT-pOs5ZQ40GFxfIZh:b-5c8a0i) ([raw file](https://github.com/ycm-core/ycmd/blob/2392bf90ecc7fd06c7ba8957ee09c2c2f3142c75/cpp/ycm/Result.h#L117)):*

I'm not sure about this padding jiggery pokery. It seems fiddly and easily broken (i.e. we could add a member to result and blow it up again)

how much is it worth in practice ?
</blockquote></details>

I wasn't able to measure any performance difference, but this is used in `FilterAndSortCandidates` - one per filtered candidate. 16.67% memory use decrease (`ResultAnd<uint32_t>` only) doesn't seem negligible.

We could add a `static_assert(sizeof(Result) == 40)`?
___
*[cpp/ycm/Result.h, line 26 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSuRr79cqaU6EokyRJ:-MQT-ChF2DE5gPAt_HGr:b-vy693g) ([raw file](https://github.com/ycm-core/ycmd/blob/be55966ea56c981862a39ae09ac3694116bcf99d/cpp/ycm/Result.h#L26)):*

aiui no_unique_address is C++ 20 ?
</blockquote></details>

It is, but all compilers we care about support it in C++17 too.

https://godbolt.org/z/qTffaE

This is protected by `__has_cpp_attribute(no_unique_address)` and works fine as of gcc/clang 9.
___
*[cpp/ycm/Utils.h, line 70 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSrXKVCfRSNCoS6fxr:-MQT0TjBAvlBG_wL8O_b:bjimwps) ([raw file](https://github.com/ycm-core/ycmd/blob/be55966ea56c981862a39ae09ac3694116bcf99d/cpp/ycm/Utils.h#L70)):*

isn't it a case of std::forward( value ) rather than std::move( value) ?
</blockquote></details>

`std::move( value )` was already there for years. The `std::move` would steal from an l-value too as long as it's not `const`. Then again, no moves are actually happening.

I'm not against using `std::forward< Value >( value )` here at all, if you think that's a better idea.

bstaletic:
Rebased, because why not.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1523)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[cpp/ycm/Result.h, line 117 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSbe3c2UlXyVRZKCOn:-MU41c3Q3KoT2ngDa9Sw:b-8cs1av) ([raw file](https://github.com/ycm-core/ycmd/blob/2392bf90ecc7fd06c7ba8957ee09c2c2f3142c75/cpp/ycm/Result.h#L117)):*

I wasn't able to measure any performance difference, but this is used in `FilterAndSortCandidates` - one per filtered candidate. 16.67% memory use decrease (`ResultAnd<uint32_t>` only) doesn't seem negligible.

We could add a `static_assert(sizeof(Result) == 40)`?
</blockquote></details>

I have just done some measurements and calculations. This is saving 8 bytes per candidates. At 65k candidates (max in our benchmarks) that's 512kB. At that same instant, our tests take up ~60MB. Relative memory usage savings: 0.8333%

I am a little surprised it didn't have a more noticeable effect on cache, but this is going through python types, which might explain thing.

puremourning (approved):
> Reviewed 2 of 11 files at r1, 6 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1523)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[cpp/ycm/Utils.h, line 70 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1523#-MQSrXKVCfRSNCoS6fxr:-MXXShmoCtXmKnjjUgj-:b-adcvpn) ([raw file](https://github.com/ycm-core/ycmd/blob/be55966ea56c981862a39ae09ac3694116bcf99d/cpp/ycm/Utils.h#L70)):*

`std::move( value )` was already there for years. The `std::move` would steal from an l-value too as long as it's not `const`. Then again, no moves are actually happening.

I'm not against using `std::forward< Value >( value )` here at all, if you think that's a better idea.
</blockquote></details>

I feel like std::forward better describes what we're doing (which is forwarding the argument to the emplace). This function is basically now just a wrapper for try_emplace, and probably could be eliminated. But 'm fine either way.

bstaletic:
:lgtm:

PR #1534, opened 2021-02-13
[RFC] Do not touch stdin/stdout when they are redirected by the shell  

### LINKED ISSUE #1525: ycmd server crashes when piping into vim or using process substituion
repro:
```
$ ls | vim -
# or
$ vim -d <(echo foo) <(echo bar)
```

vim `:messages` output:
```
The ycmd server SHUT DOWN (restart with ':YcmRestartServer'). Unexpected exit code 1. Type ':YcmToggleLogs ycmd_52525_stderr_cfxdny1s.log' to check the logs.
```

`ycmd.*log` contents:
```
2021-01-13 17:56:17,393 - DEBUG - Global extra conf not loaded or no function YcmCorePreload
2021-01-13 17:56:17,415 - INFO - Completion config: 50, detailing 0 candiates
2021-01-13 17:56:17,416 - INFO - Completion config: 50, detailing 0 candiates
2021-01-13 17:56:17,416 - INFO - Completion config: 50, detailing 0 candiates
2021-01-13 17:56:17,416 - INFO - Completion config: 50, detailing 0 candiates
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/sam/etc/vim/plugged/YouCompleteMe/python/ycm/../../third_party/ycmd/ycmd/__main__.py", line 193, in <module>
    Main()
  File "/home/sam/etc/vim/plugged/YouCompleteMe/python/ycm/../../third_party/ycmd/ycmd/__main__.py", line 184, in Main
    CloseStdin()
  File "/home/sam/etc/vim/plugged/YouCompleteMe/python/ycm/../../third_party/ycmd/ycmd/__main__.py", line 140, in CloseStdin
    sys.stdin.close()
AttributeError: 'NoneType' object has no attribute 'close'
2021-01-13 17:56:17,417 - DEBUG - Global extra conf not loaded or no function VimClose
2021-01-13 17:56:17,417 - DEBUG - Global extra conf not loaded or no function Shutdown
```

bstaletic:
If stdin and/or stdout is redirected by the shell, `sys.stdin` and `sys.stdout` will be `None` and `print()` won't work. The `sys.stdin` part is easy, because we never redirect it. However, we attempt to redirect `sys.stdout` to our log file. Python thinks that succeeds and then we're left with no way to check if we're allowed to call `print()` in `wsgi_server.py`. A workaround is to check if `sys.stdout` is `None` and skip the pointless redirection - `print()` will fail anyway.

Fixes #1525

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1534)

### DIFF
--- a/ycmd/__main__.py
+++ b/ycmd/__main__.py
@@ -152,7 +152,8 @@ def SetupOptions( options_file ):
 
 
 def CloseStdin():
-  sys.stdin.close()
+  if sys.stdin is not None:
+    sys.stdin.close()
   os.close( 0 )
 
 
--- a/ycmd/wsgi_server.py
+++ b/ycmd/wsgi_server.py
@@ -17,6 +17,7 @@
 
 from waitress.server import TcpWSGIServer
 import select
+import sys
 
 
 class StoppableWSGIServer( TcpWSGIServer ):
@@ -32,7 +33,9 @@ class StoppableWSGIServer( TcpWSGIServer ):
 
     # Message for compatibility with clients who expect the output from
     # waitress.serve here
-    print( f'serving on http://{ self.effective_host }:{self.effective_port}' )
+    if sys.stdin is not None:
+      print( f'serving on http://{ self.effective_host }:'
+             f'{ self.effective_port }' )
 
     try:
       self.run()


puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1534#-:-MTSAcYQ1ivo9MsIVROm:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic:
The cleaner alternative, with not redirecting stdout didn't work. This patch is the only thing that seems to work, so :lgtm:

PR #1535, opened 2021-02-13
[READY] Clean up C++ warnings

bstaletic:
Nothing too special. Most noteworthy changes are in `IdentifierUtils.cpp`:

1. Remove unused `StringCompare` class.
2. Remove pairs from `LANG_TO_FILETYPE` where the value is just a lowercase version of the key - that's covered by the fallback branch.
3. To avoid the constructor at global scope, `LANG_TO_FILETYPE`'s type is now `array<pair<string_view, string_view>, N>` and is `constexpr`.
3.1. As a result, `FindWithDefault` had to be rewritten to work with `unordered_map::find` and `std::find`. Damn you libclang completer.
4. The main loop in `IdentifierUtils.cpp` is changed to use iterators instead of indices, which helps with warnings. Also, it allows us to use `std::default_searcher()` which means we don't have to use magic `index + 9` to skip over `"language:"`.
5. The rest are just minor `-Wextra-semi` and `-Wsign-conversion` things. Some of which could be solved nicer with `std::ssize()`, but that's C++20.

That leaves us with the following warnings:

1. `-Wc++98-compat` - yeah, we're at C++17.
2. `-Wc++98-compat-pedantic` - I thought I said C++17?
3. `-Wexit-time-destructor` - the repository singletons are destroyed at exit time. No way around that except leaking them, Google style.
4. `-Wpadding` - yeeeah...

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1535)

### DIFF
--- a/cpp/ycm/CodePoint.cpp
+++ b/cpp/ycm/CodePoint.cpp
@@ -57,7 +57,7 @@ RawCodePoint FindCodePoint( std::string_view text ) {
 
   auto it = std::lower_bound( original.begin(), original.end(), text );
   if ( it != original.end() && text == *it ) {
-    size_t index = std::distance( original.begin(), it );
+    auto index = static_cast< size_t >( std::distance( original.begin(), it ) );
     return { *it,
              code_points.normal[ index ],
              code_points.folded_case[ index ],
@@ -111,6 +111,6 @@ CodePointSequence BreakIntoCodePoints( std::string_view text ) {
 
 const char* UnicodeDecodeError::what() const noexcept {
   return std::runtime_error::what();
-};
+}
 
 } // namespace YouCompleteMe
--- a/cpp/ycm/IdentifierUtils.cpp
+++ b/cpp/ycm/IdentifierUtils.cpp
@@ -18,9 +18,11 @@
 #include "IdentifierUtils.h"
 #include "Utils.h"
 
+#include <array>
 #include <filesystem>
+#include <functional>
 #include <string_view>
-#include <unordered_map>
+#include <utility>
 
 namespace YouCompleteMe {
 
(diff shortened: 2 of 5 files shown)


puremourning (commented):
:lgtm: 

> Reviewed 1 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1535#-:-MU0ZDgFEsIc3Vhhyifr:bg7igkv)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[cpp/ycm/IdentifierUtils.cpp, line 37 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1535#-MU0T8PlCRVS6nBcrCS6:-MU0T8PlCRVS6nBcrCS7:b-njuddb) ([raw file](https://github.com/ycm-core/ycmd/blob/70ee8e3790b2b07678b96ec0f3492bf9e496c576/cpp/ycm/IdentifierUtils.cpp#L37)):*
> ```cpp
> // To map a language name to a filetype, see this file:
> //   :e $VIMRUNTIME/filetype.vim
> // This is a map of const char* and not std::string to prevent issues with
> ```

this comment is now misleading
___
*[cpp/ycm/IdentifierUtils.cpp, line 40 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1535#-MU0YwG555Nf5u3SInlz:-MU0YwG555Nf5u3SInm-:b-37cdgt) ([raw file](https://github.com/ycm-core/ycmd/blob/70ee8e3790b2b07678b96ec0f3492bf9e496c576/cpp/ycm/IdentifierUtils.cpp#L40)):*
> ```cpp
> // static initialization.
> constexpr std::array< std::pair< std::string_view, std::string_view
>       >, 10 > LANG_TO_FILETYPE = {{
> ```

as discussed, I don't love that we have to hard-code the size here; I'd rather use a vector as I don't _think_ we ever _use_ this in a constexpr context. But I don't really mind. This hardly ever changes.
___
*[cpp/ycm/IdentifierUtils.cpp, line 90 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1535#-MU0WJ0DEeYyxehVYhGU:-MU0WJ0DEeYyxehVYhGV:bdzg999) ([raw file](https://github.com/ycm-core/ycmd/blob/70ee8e3790b2b07678b96ec0f3492bf9e496c576/cpp/ycm/IdentifierUtils.cpp#L90)):*
> ```cpp
>     const std::string_view lang_str = "language:";
>     auto searcher = std::default_searcher( lang_str.cbegin(), lang_str.cend() );
>     const auto lang_begin = searcher( path_end + 1, line.cend() ).second;
> ```

aren't you expected to call `std:search` with a `default_searcher` as argument?
___
*[cpp/ycm/Utils.h, line 91 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1535#-MU0XqGLEVaG8uJyLOoX:-MU0XqGLEVaG8uJyLOoY:btah2e2) ([raw file](https://github.com/ycm-core/ycmd/blob/70ee8e3790b2b07678b96ec0f3492bf9e496c576/cpp/ycm/Utils.h#L91)):*
> ```objc
>                  const Key &key,
>                  Ret&& value ) {
>   typename Container::const_iterator it = [&key](auto&& c) {
> ```

I think we need some spaces inside some parentheses and brackets in this function.

bstaletic (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1535#-:-MU0fU5T1ZBq7giA9Lzf:bnfp4nl)* status: 1 of 2 LGTMs obtained (and 1 stale)

___
*[cpp/ycm/IdentifierUtils.cpp, line 37 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1535#-MU0T8PlCRVS6nBcrCS6:-MU0fWBqCMWqjKEKYde8:b187d8j) ([raw file](https://github.com/ycm-core/ycmd/blob/70ee8e3790b2b07678b96ec0f3492bf9e496c576/cpp/ycm/IdentifierUtils.cpp#L37)):*

this comment is now misleading
</blockquote></details>

It was always misleading, since `std::unordered_map` still had a problem with the initialization order fiasco and all that jazz.

Comment removed.
___
*[cpp/ycm/IdentifierUtils.cpp, line 40 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1535#-MU0YwG555Nf5u3SInlz:-MU0fhYT0I4rel1TjvDg:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/70ee8e3790b2b07678b96ec0f3492bf9e496c576/cpp/ycm/IdentifierUtils.cpp#L40)):*

as discussed, I don't love that we have to hard-code the size here; I'd rather use a vector as I don't _think_ we ever _use_ this in a constexpr context. But I don't really mind. This hardly ever changes.
</blockquote></details>

Done.
___
*[cpp/ycm/IdentifierUtils.cpp, line 90 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1535#-MU0WJ0DEeYyxehVYhGU:-MU0fiqH2ovZxefXVb4L:b-wg5442) ([raw file](https://github.com/ycm-core/ycmd/blob/70ee8e3790b2b07678b96ec0f3492bf9e496c576/cpp/ycm/IdentifierUtils.cpp#L90)):*

aren't you expected to call `std:search` with a `default_searcher` as argument?
</blockquote></details>

That's the usual use case, but we need the `end` iterator of the found `"language:"`, not the `begin`. That makes `std::search` inconvenient - it would make us keep `+ 9` or calculate the length of `"language:"`. Calling the searcher's `operator()` returns a begin-end pair so we instantly get what we're looking for - a pointer to the start of the language identifier.
___
*[cpp/ycm/Utils.h, line 91 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1535#-MU0XqGLEVaG8uJyLOoX:-MU0gYJ58_pF2O9qXKvF:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/70ee8e3790b2b07678b96ec0f3492bf9e496c576/cpp/ycm/Utils.h#L91)):*

I think we need some spaces inside some parentheses and brackets in this function.
</blockquote></details>

Done.

puremourning (approved):
:lgtm:

> Reviewed 2 of 5 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1535#-:-MU0gzeHBTxints865h8:bnfp4nl)* status: :shipit: complete! 2 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1538, opened 2021-02-27
[READY] The easy part of performance improvements

bstaletic:
As discussed on gitter, the simplest changes first.

- Changes `reserve()`/`puch_back()` loops to `resize()` and "raw" writes.
- Changes `FiletypeCandidateMap` to a more... sane type.

Details are in the commit messages.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1538)


bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1538)* status: 0 of 2 LGTMs obtained

___
*[cpp/ycm/CandidateRepository.cpp, line 70 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MUYaa2VBy4xsrZ5SW5t:-MUYaa2VBy4xsrZ5SW5u:b-7phi02) ([raw file](https://github.com/ycm-core/ycmd/blob/37bf6a4998ac3360dd2fcdf7b315b7400861a0b5/cpp/ycm/CandidateRepository.cpp#L70)):*
> ```cpp
>       }
> 
>       *it++ = candidate.get();
> ```

I can rewrite this as a `std::transform` call, with a big lambda. Not sure if a big lambda would be more readable than `*it++`.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 102 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MUYau-35lvhT44G-MHp:-MUYau-35lvhT44G-MHq:b-lty2mn) ([raw file](https://github.com/ycm-core/ycmd/blob/37bf6a4998ac3360dd2fcdf7b315b7400861a0b5/cpp/ycm/IdentifierDatabase.cpp#L102)):*
> ```cpp
>   std::unordered_set< const Candidate *,
>                       CandidateHasher,
>                       CandidateCompareEq > seen_candidates;
> ```

This is still storing `const Candidate*`, but we can't use the default hasher and comparison, because we might find the same candidate for filetype X in both filepath A and filepath B. If you have ideas for simplifying this, I'm all ears. Note that `Candidate` and `Word` (the base type) don't have `operator==`.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 163 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MUYbULg6PKquO2wdTXJ:-MUYbULg6PKquO2wdTXK:b69l96x) ([raw file](https://github.com/ycm-core/ycmd/blob/37bf6a4998ac3360dd2fcdf7b315b7400861a0b5/cpp/ycm/IdentifierDatabase.cpp#L163)):*
> ```cpp
>     }
>   }
>   candidate_repository_.GetCandidatesForStrings( std::move( new_candidates ) );
> ```

This part is awful. Since now `IdentifierDatabase` owns Candidates the `IdentifierCompleter` has discovered, without this last line, it would "rob" `CandidateRepository` of the candidates. That may not be that bad. Right now, I'm preserving old behaviour at the expense of insert time. Here are our options:

1. Skip "forwarding" new candidates from `IdentifierCompleter` to the `CandidateRepository`. The `IdentifierCompleter` doesn't care about the repo any more, but an empty repo can negatively impact the performance of the semantic completers.
2. Do what this PR is doing.  Performance problems come from a) copying the entire `new_candidates` vector b) iterating twice and worst of all... c) the `for-find_if-emplace_back` loop is `O(n^2)`.
2.1. We can keep the stored `Candidate`s sorted, so that `lower_bound` can be used instead of `find_if`.
3. Since we don't care about the order of `new_candidates` or the order of `Candidates` in the `IdentifierDatabase`, we could sort both vectors, call `set_difference` and then insert the result. That would be asymptotically better, but I haven't measured.

puremourning (changes requested):
> Reviewed 1 of 9 files at r1.
> Dismissed @bstaletic from a discussion.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1538)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[cpp/ycm/CandidateRepository.cpp, line 70 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MUYaa2VBy4xsrZ5SW5t:-MVY8KrN3eJILXou18wL:b-kj3d7e) ([raw file](https://github.com/ycm-core/ycmd/blob/37bf6a4998ac3360dd2fcdf7b315b7400861a0b5/cpp/ycm/CandidateRepository.cpp#L70)):*

I can rewrite this as a `std::transform` call, with a big lambda. Not sure if a big lambda would be more readable than `*it++`.
</blockquote></details>

*it++ reads perfectly well for me.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 102 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MUYau-35lvhT44G-MHp:-MVYAnCA1SwBk5X-CfyE:b9abie5) ([raw file](https://github.com/ycm-core/ycmd/blob/37bf6a4998ac3360dd2fcdf7b315b7400861a0b5/cpp/ycm/IdentifierDatabase.cpp#L102)):*

This is still storing `const Candidate*`, but we can't use the default hasher and comparison, because we might find the same candidate for filetype X in both filepath A and filepath B. If you have ideas for simplifying this, I'm all ears. Note that `Candidate` and `Word` (the base type) don't have `operator==`.
</blockquote></details>

so previously we were hashing based on the pointer address ? And using a ContainsKey to check if the exact same pointer was in the set ?

Now we're comparing the text using a string hash?

Are they really equivalent ? 
___
*[cpp/ycm/IdentifierDatabase.cpp, line 33 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MVY9ZPF0PIrtC7gIxr6:-MVY9ZPF0PIrtC7gIxr7:b-ac1xzv) ([raw file](https://github.com/ycm-core/ycmd/blob/082d1dfe80ea0ad749c1a11ce314e82c45c893bf/cpp/ycm/IdentifierDatabase.cpp#L33)):*
<details><summary><i alt="summary">Quoted 11 lines of code…</i></summary>

> ```
> struct CandidateHasher {
>        size_t operator() (const Candidate* c) const noexcept{
>                static std::hash<std::string> h;
>                return h(c->Text());
>        }
> };
> struct CandidateCompareEq {
>        bool operator() (const Candidate* a, const Candidate* b) const noexcept{
>                return a->Text() == b->Text();
>        }
> };
> ```
</details>


Indentation looks dodgy in this section. should be 2 spaces and should have whitespace inside parentheses.

there's a script somewhere that applies the style (though I never run it). we should probably set up clang format I guess.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 74 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MWF_Bd4E4qTYh_Jj-rm:-MWF_Bd4E4qTYh_Jj-rn:b-lptqpp) ([raw file](https://github.com/ycm-core/ycmd/blob/ed6c814eb1118bc92a4abfbb7b279dfb13cc192c/cpp/ycm/IdentifierDatabase.cpp#L74)):*
> ```cpp
>   std::lock_guard locker( filetype_candidate_map_mutex_ );
>   auto candidate_pointer = candidate_repository_.GetCandidatesForStrings(
>                   std::move( new_candidates ) )[ 0 ];
> ```

should we at least assert that `new_candidates.size() == 1` ? Still, this interface feels wonky now with "add single identifier" taking a vector of length 1 ?
___
*[ycmd/completers/all/identifier_completer.py, line 70 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MWFYxdn4e1r84NN1OiO:-MWFYxdn4e1r84NN1OiP:bp3h7kq) ([raw file](https://github.com/ycm-core/ycmd/blob/ed6c814eb1118bc92a4abfbb7b279dfb13cc192c/ycmd/completers/all/identifier_completer.py#L70)):*
> ```Python
>     LOGGER.info( 'Adding ONE buffer identifier for file: %s', filepath )
>     self._completer.AddSingleIdentifer(
>       ycm_core.StringVector( [ identifier ] ),
> ```

Seems strange to pass a list of strings to a method that says "AddSingleIdentifier"

bstaletic (commented):
I've fixed the segfault! Dumb mistake.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1538#-:-MWFnRVn3gzIz6r_d3dI:b-ab0a3v)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[cpp/ycm/CandidateRepository.cpp, line 70 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MUYaa2VBy4xsrZ5SW5t:-MWFjDJb6cx01qu-VfVD:b5iwnn1) ([raw file](https://github.com/ycm-core/ycmd/blob/37bf6a4998ac3360dd2fcdf7b315b7400861a0b5/cpp/ycm/CandidateRepository.cpp#L70)):*

*it++ reads perfectly well for me.
</blockquote></details>

I know libc++ doesn't like `*it++`, but [duff's device](https://en.wikipedia.org/wiki/Duff%27s_device) is beautiful.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 102 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MUYau-35lvhT44G-MHp:-MWFmCQr0Ri5bkq7g56h:bu2phlc) ([raw file](https://github.com/ycm-core/ycmd/blob/37bf6a4998ac3360dd2fcdf7b315b7400861a0b5/cpp/ycm/IdentifierDatabase.cpp#L102)):*

so previously we were hashing based on the pointer address ? And using a ContainsKey to check if the exact same pointer was in the set ?

Now we're comparing the text using a string hash?

Are they really equivalent ? 
</blockquote></details>

Good question. On master, all `Candidate` objects are in the `CandidateRepository`. Thus you really can just compare their addresses.

In this PR, `IdentifierDatabase` has its own storage for `Candidate` objects. What's more, the same candidate may be found in `foo.cpp` and `bar.cpp`. Those two `Candidate`s would have different addresses in memory, but would contain the same data. Hashing/comparing addresses is out of the question. Since `Text()` is normalized, hashing/comparing `Text()` allows deduplication of `Candidate` objects, and later `Result` objects, in this function.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 33 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MVY9ZPF0PIrtC7gIxr6:-MWFk1Uf0MAlvfLSW1QO:b-6vr9zy) ([raw file](https://github.com/ycm-core/ycmd/blob/082d1dfe80ea0ad749c1a11ce314e82c45c893bf/cpp/ycm/IdentifierDatabase.cpp#L33)):*

> ```
> struct CandidateHasher {
>        size_t operator() (const Candidate* c) const noexcept{
>                static std::hash<std::string> h;
>                return h(c->Text());
>        }
> };
> struct CandidateCompareEq {
>        bool operator() (const Candidate* a, const Candidate* b) const noexcept{
>                return a->Text() == b->Text();
>        }
> };
> ```

Indentation looks dodgy in this section. should be 2 spaces and should have whitespace inside parentheses.

there's a script somewhere that applies the style (though I never run it). we should probably set up clang format I guess.
</blockquote></details>

The script never worked for me. As for clang-format, I could never get it to work *just right*.

I'd say "done", but I haven't pushed the changes.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 74 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MWF_Bd4E4qTYh_Jj-rm:-MWFkgp28JUgE0_YzxNM:b-w2yui9) ([raw file](https://github.com/ycm-core/ycmd/blob/ed6c814eb1118bc92a4abfbb7b279dfb13cc192c/cpp/ycm/IdentifierDatabase.cpp#L74)):*

should we at least assert that `new_candidates.size() == 1` ? Still, this interface feels wonky now with "add single identifier" taking a vector of length 1 ?
</blockquote></details>

Agreed that it looks odd, but I do have an explanation. A few lines lower, we're comparing `c.Text() == candidate_pointer->Text()`. For that comparison to actually be reasonable, we need text normalized and canonically decomposed. That's what`GetCandidatesForStrings()` is buying us.

`GetCandidatesForStrings()` works on a vector of strings and I didn't feel like reinventing the normalization algorithm for a case of one candidate. This part isn't hard to do:

1. Break `new_candidates[ 0 ]` into code points.
2. Accumulate `CodePoint::Normal()` from those code points.
3. Call `Candidate::Candidate(string_view)` directly.

The other reason to use `GetCandidatesForStrings()` is that it's going through the `CandidateRepository`, so it avoids recreating a `Candidate` that we have already seen.

Ideas for improvement very welcome.
___
*[ycmd/completers/all/identifier_completer.py, line 70 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MWFYxdn4e1r84NN1OiO:-MWFmIGICukpkZYJtg5T:b-lxm9xv) ([raw file](https://github.com/ycm-core/ycmd/blob/ed6c814eb1118bc92a4abfbb7b279dfb13cc192c/ycmd/completers/all/identifier_completer.py#L70)):*

Seems strange to pass a list of strings to a method that says "AddSingleIdentifier"
</blockquote></details>

Agreed. The other comment has a detailed explanation for that.

puremourning (changes requested):
Now that IdentifierDatabase contains actual Candidates, I'm sort of struggling to see the point of CandidateRepository.

Previously we were using a sort of "flyweight" pattern where the assumption was that there were very many duplicated Candidates that are relatively expensive to create, so the CandidateRepository was a holder for the Candidates and other things just held pointers to them.

I think this PR is changing that picture and declaring that previous optimisation to be false? Or are we saying "yes, they are relatively expensive to create, but they are cheap to _copy_ and they are essentially _immutable_". If so, then I totally get the need for the repository and probably rubber-ducked it by writing this comment !

> Reviewed 5 of 9 files at r1, 1 of 1 files at r2, 6 of 9 files at r3, 4 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1538#-:-MXXOdJkCkYrLTHaeuuO:b-mlcuqv)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[cpp/ycm/Candidate.h, line 39 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MXXQC3aEPTMr6t1CbIy:-MXXQC3aEPTMr6t1CbIz:brl5dm8) ([raw file](https://github.com/ycm-core/ycmd/blob/0ae62f4329bd720ed9b5776bf03018313fd413f7/cpp/ycm/Candidate.h#L39)):*
> ```objc
>   Candidate& operator=( const Candidate& ) = default;
> public:
>   Candidate clone() const { return *this; }
> ```

reming me why we need this, rather than just making the copy constructor explicit and public?

then 

```
                    return candidate_ptr->clone();
```

becomes

```
                    return Candidate( *candidate_ptr );
```

and

```
    current_identifier_set.push_back( candidate_pointer->clone() );
```

becomes

```
    current_identifier_set.emplace_back( *candidate_pointer );
```

or am I missing something?
___
*[cpp/ycm/IdentifierCompleter.cpp, line 49 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MXTFqr96zexn4Cz_yCK:-MXTFqr96zexn4Cz_yCL:bttm1mv) ([raw file](https://github.com/ycm-core/ycmd/blob/0ae62f4329bd720ed9b5776bf03018313fd413f7/cpp/ycm/IdentifierCompleter.cpp#L49)):*
> ```cpp
>   std::string& filepath ) {
>   identifier_database_.AddSingleIdentifier( std::move( new_candidates ),
>                                        std::move( filetype ),
> ```

update intentation
___
*[cpp/ycm/IdentifierDatabase.cpp, line 74 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MWF_Bd4E4qTYh_Jj-rm:-MXTJYqC1Lxl68q0aOW_:b-k5gmsj) ([raw file](https://github.com/ycm-core/ycmd/blob/ed6c814eb1118bc92a4abfbb7b279dfb13cc192c/cpp/ycm/IdentifierDatabase.cpp#L74)):*

Agreed that it looks odd, but I do have an explanation. A few lines lower, we're comparing `c.Text() == candidate_pointer->Text()`. For that comparison to actually be reasonable, we need text normalized and canonically decomposed. That's what`GetCandidatesForStrings()` is buying us.

`GetCandidatesForStrings()` works on a vector of strings and I didn't feel like reinventing the normalization algorithm for a case of one candidate. This part isn't hard to do:

1. Break `new_candidates[ 0 ]` into code points.
2. Accumulate `CodePoint::Normal()` from those code points.
3. Call `Candidate::Candidate(string_view)` directly.

The other reason to use `GetCandidatesForStrings()` is that it's going through the `CandidateRepository`, so it avoids recreating a `Candidate` that we have already seen.

Ideas for improvement very welcome.
</blockquote></details>

idea: change this method to take a single string&& "new_candidate"
use `GetCandidatesForStrings( vector<string>{ new_candidate } )[ 0 ]`

At least then the vector-ness is confined to that one call, not exposed to the caller of this method.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 33 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MXTIZJDCLj-FI0PMvMf:-MXTIZJDCLj-FI0PMvMg:buv7g6c) ([raw file](https://github.com/ycm-core/ycmd/blob/0ae62f4329bd720ed9b5776bf03018313fd413f7/cpp/ycm/IdentifierDatabase.cpp#L33)):*
> ```cpp
> namespace {
> struct CandidateHasher {
> <<<<<<< HEAD
> ```

change marker ?
___
*[cpp/ycm/IdentifierDatabase.cpp, line 46 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MXTIjCQ9yt2gY4Txhy9:-MXTIjCQ9yt2gY4TxhyA:beer3m1) ([raw file](https://github.com/ycm-core/ycmd/blob/0ae62f4329bd720ed9b5776bf03018313fd413f7/cpp/ycm/IdentifierDatabase.cpp#L46)):*
> ```cpp
> };
> struct CandidateCompareEq {
>   bool operator() (const Candidate* a, const Candidate* b) const noexcept{
> ```

space after `noexcept`

bstaletic (commented):
> Or are we saying "yes, they are relatively expensive to create, but they are cheap to copy and they are essentially immutable". If so, then I totally get the need for the repository and probably rubber-ducked it by writing this comment !

This!

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1538#-:-MXlAx4KELW7DilcRkAI:b-uyybzw)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[cpp/ycm/Candidate.h, line 39 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MXXQC3aEPTMr6t1CbIy:-MXlBLqzENeiiMgT79xY:b-eqcar) ([raw file](https://github.com/ycm-core/ycmd/blob/0ae62f4329bd720ed9b5776bf03018313fd413f7/cpp/ycm/Candidate.h#L39)):*

reming me why we need this, rather than just making the copy constructor explicit and public?

then 

```
                    return candidate_ptr->clone();
```

becomes

```
                    return Candidate( *candidate_ptr );
```

and

```
    current_identifier_set.push_back( candidate_pointer->clone() );
```

becomes

```
    current_identifier_set.emplace_back( *candidate_pointer );
```

or am I missing something?
</blockquote></details>

Even though `Candidate` is cheaper to copy than to create from scratch, it is still a pretty large data type. 152 bytes, to be exact, which is about 2.5 cache lines.

For that reason, I wanted to make a copy very explicit, so that using the standard library (algorithms, for the most part) doesn't lead to unexpected copies.
___
*[cpp/ycm/IdentifierCompleter.cpp, line 49 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MXTFqr96zexn4Cz_yCK:-MXlF4vo9eKfoqzXx39h:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/0ae62f4329bd720ed9b5776bf03018313fd413f7/cpp/ycm/IdentifierCompleter.cpp#L49)):*

update intentation
</blockquote></details>

Done.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 33 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MVY9ZPF0PIrtC7gIxr6:-MXlFMKu0vlfFWoLNHjU:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/082d1dfe80ea0ad749c1a11ce314e82c45c893bf/cpp/ycm/IdentifierDatabase.cpp#L33)):*

The script never worked for me. As for clang-format, I could never get it to work *just right*.

I'd say "done", but I haven't pushed the changes.
</blockquote></details>

Done.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 74 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MWF_Bd4E4qTYh_Jj-rm:-MXlUEMS1icB9TkjEJg1:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/ed6c814eb1118bc92a4abfbb7b279dfb13cc192c/cpp/ycm/IdentifierDatabase.cpp#L74)):*

idea: change this method to take a single string&& "new_candidate"
use `GetCandidatesForStrings( vector<string>{ new_candidate } )[ 0 ]`

At least then the vector-ness is confined to that one call, not exposed to the caller of this method.
</blockquote></details>

Done.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 33 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MXTIZJDCLj-FI0PMvMf:-MXlFLOSBSY6aiPJY8mE:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/0ae62f4329bd720ed9b5776bf03018313fd413f7/cpp/ycm/IdentifierDatabase.cpp#L33)):*

change marker ?
</blockquote></details>

Done.
___
*[cpp/ycm/IdentifierDatabase.cpp, line 46 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1538#-MXTIjCQ9yt2gY4Txhy9:-MXlFTK1EpkgXx2FQVDX:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/0ae62f4329bd720ed9b5776bf03018313fd413f7/cpp/ycm/IdentifierDatabase.cpp#L46)):*

space after `noexcept`
</blockquote></details>

Done.

bstaletic (commented):
Fixed. CI should be green now.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1538#-:-MXmUDoY5-V6JcfodlNG:b-yfmhlu)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

puremourning (commented):
not green, but :lgtm:  now

> Reviewed 6 of 9 files at r5.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1538#-:-MXn6r3193XyAVAHZ_S1:bg7ikzq)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

bstaletic:
:lgtm:

puremourning (approved):
:lgtm:

> Reviewed 3 of 3 files at r6.
> Dismissed @bstaletic from a discussion.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1538#-:-MXqgRBW8q3oVXGGt7--:bnfp4nl)* status: 1 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic)

PR #1539, opened 2021-03-02
Fix openmp for bundled clangd

vbarrielle:
I have noticed that when a C++ file contains the directive `#include "omp.h"`, the bundled clangd would fail to understand the file as it would encounter a fatal error, failing to find the `omp.h` header. In contrast, the clangd one can obtain on a typical linux distribution will contain this header.

This pull request fixes this issue by shipping the openmp headers found in [the official clang release corresponding to the current version bundled in ycmd](https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.0/clang+llvm-11.0.0-x86_64-linux-gnu-ubuntu-20.04.tar.xz). It's the release files for ubuntu 20.04, but I don't think it should be a problem for these header files given their contents.

This pull request contains two commits:

- https://github.com/ycm-core/ycmd/commit/c45d2928b91a296a70b8aa1b9a675781a4e39ea9 adds a test demonstrating the problem.
- https://github.com/ycm-core/ycmd/commit/12ee1c6d1d9f70e2a93c244f13b318dca7bcbe52 adds the missing openmp headers, fixing the test failure.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1539)

### DIFF
--- a/third_party/clang/lib/clang/11.0.0/include/omp-tools.h
+++ b/third_party/clang/lib/clang/11.0.0/include/omp-tools.h
@@ -1,0 +1,1082 @@
+/*
+ * include/omp-tools.h.var
+ */
+
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef __OMPT__
+#define __OMPT__
+
+/*****************************************************************************
+ * system include files
+ *****************************************************************************/
+
+#include <stdint.h>
+#include <stddef.h>
+
+/*****************************************************************************
+ * iteration macros
+ *****************************************************************************/
+
+#define FOREACH_OMPT_INQUIRY_FN(macro)      \
+    macro (ompt_enumerate_states)           \
+    macro (ompt_enumerate_mutex_impls)      \
+                                            \
+    macro (ompt_set_callback)               \
+    macro (ompt_get_callback)               \
+                                            \
+    macro (ompt_get_state)                  \
+                                            \
+    macro (ompt_get_parallel_info)          \
+    macro (ompt_get_task_info)              \
+    macro (ompt_get_task_memory)            \
+    macro (ompt_get_thread_data)            \
+    macro (ompt_get_unique_id)              \
+    macro (ompt_finalize_tool)              \
+                                            \
+    macro(ompt_get_num_procs)               \
+    macro(ompt_get_num_places)              \
+    macro(ompt_get_place_proc_ids)          \
+    macro(ompt_get_place_num)               \
+    macro(ompt_get_partition_place_nums)    \
+    macro(ompt_get_proc_id)                 \
+                                            \
+    macro(ompt_get_target_info)             \
+    macro(ompt_get_num_devices)
+
+#define FOREACH_OMPT_STATE(macro)                                                                \
+                                                                                                \
+    /* first available state */                                                                 \
+    macro (ompt_state_undefined, 0x102)      /* undefined thread state */                        \
+                                                                                                \
+    /* work states (0..15) */                                                                   \
+    macro (ompt_state_work_serial, 0x000)    /* working outside parallel */                      \
+    macro (ompt_state_work_parallel, 0x001)  /* working within parallel */                       \
+    macro (ompt_state_work_reduction, 0x002) /* performing a reduction */                        \
+                                                                                                \
+    /* barrier wait states (16..31) */                                                          \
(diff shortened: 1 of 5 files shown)


bstaletic (changes requested):
Thanks for the PR. We were aware of this issue, but didn't know how exactly to fix it.

However, the `third_party/clang/lib/clang/<version>/include` shouldn't be edited like this. It's populated from the `update_clang.py` script, which pulls headers from github releases archive.
https://github.com/ycm-core/ycmd/blob/master/update_clang.py#L488

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1539#-:-MUo6Xph-ozkVDttpsBe:b-8m2c3b)* status: 0 of 2 LGTMs obtained (waiting on @vbarrielle)

___
*[ycmd/tests/clangd/openmp_test.py, line 31 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1539#-MUo7v_O1byFkf9dql4l:-MUo7v_O1byFkf9dql4m:b-cyqwss) ([raw file](https://github.com/ycm-core/ycmd/blob/12ee1c6d1d9f70e2a93c244f13b318dca7bcbe52/ycmd/tests/clangd/openmp_test.py#L31)):*
> ```Python
> 
> @IsolatedYcmd()
> def OpenMp_HeaderFound_test( app ):
> ```

I'd also move this test into `diagnostics_test.py` and call it something like `Diagnostics_OpenMpHeaderFound_test`.

vbarrielle:
I see, thanks for pointing out the proper way, I will try to work with this script. Just to be sure I understand the process: once I've managed to fetch the required headers with `update_clang.py`, should I run the script to create a commit adding the headers?

I will move the test to `diagnostics_test.py` as requested.

puremourning:
Thanks for doing this. Much appreciated!

bstaletic:
The script should fetch the LLVM OpenMP archive, extract the headers and put them in the right folder. Then you manually commit the the changes the script has mad.

bstaletic:
This might be tougher than it looks. My distro's LLVM OpenMP package contains the following headers:

```
/usr/include/omp-tools.h
/usr/include/omp.h
/usr/include/ompt-multiplex.h
/usr/include/ompt.h
```

I downloaded the LLVM OpenMP archive and the first thing I noticed is that there's no `ompt.h` at all. Okay, maybe it will be generated when compiling. Nope.

```
bstaletic@Gallifrey openmp-11.0.0.src % find build -name \*.h
build/runtime/src/omp-tools.h
build/runtime/src/kmp_config.h
build/runtime/src/omp.h
```

vbarrielle:
Indeed, I haven't figured out how these are generated when llvm is built. That's why I took the ones in the clang release built by the LLVM project.

But if the files are different depending on the OS, then maybe it's something that needs to be fetched when ycmd is built.

bstaletic:
Maybe. Let's ping @kadircet. Maybe he can enlighten us.

bstaletic:
The upstream binary archives do contain LLVM OpenMP headers. Maybe we should repackage our own archives to contain those headers and, if the user passes `--clang-completer` or `--clangd-completer`, to copy those headers over as well?

kadircet:
> I downloaded the LLVM OpenMP archive and the first thing I noticed is that there's no ompt.h at all. Okay, maybe it will be generated when compiling. Nope.

AFAIK, openmp archives are just source code so you will need to run the build to get the runtime headers (as you noticed). But the default config might not be what you want (e.g. you need to pass some extra flags to spit out `ompt.h`).

> The upstream binary archives do contain LLVM OpenMP headers. Maybe we should repackage our own archives to contain those headers and, if the user passes --clang-completer or --clangd-completer, to copy those headers over as well?

As you've noticed the prebuilt binaries package in releases page indeed contain all the headers you are looking for but I also don't know where the release script lives (and a quick search in all the repos under llvm didn't really reveal much, it is likely that i've missed it).

I suppose fetching prebuilt binary package for relevant OS and merging (or just using it as-is) should do the trick, but oh boy, that package is huge :/ So repacking definitely makes sense, but I am not sure if you can get away from having a package per distro/os, as these are generated headers they are likely to contain platform specific info, but this is just a wild guess I don't really know much about OMP.

Sorry for all of this text which doesn't help with much :D

bstaletic:
Okay, we already do repacakage the LLVM archives, because we were told that LLVM doesn't like when its servers are being used for CI. Currently those packages reside on  our bintray, but that is shutting down in May. Then we can say "if `--clang{,d}-completer` and no third_party/whatever/omp.h` then copy the extra headers over.

Maybe we should put this aside until LLVM12 arrives and do all the work in one go? @puremourning Thoughts?

@kadircet I had a feeling you wouldn't be too familiar with OpenMP, but either way, thanks for the quick response.

puremourning:
I think if we were to do this it would have to be part of update_llvm.py and/or (as Boris suggested) the repackaging that we do of libclang/clangd.

But this PR as it stands doesn't deliver that, and I think it's stalled so, closing for now.

PR #1542, opened 2021-03-07
Fix compiler warnings from -Wreadability-redundant-string-init

fallingdrift:
dunno if this is desired or not but here it is-- default initializers are redundant for `std::string`.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1542)

### DIFF
--- a/cpp/ycm/ClangCompleter/Location.h
+++ b/cpp/ycm/ClangCompleter/Location.h
@@ -27,9 +27,7 @@ namespace YouCompleteMe {
 
 struct Location {
   // Creates an invalid location
-  Location()
-    : line_number_( 0 ),
-      column_number_( 0 ) {
+  Location() {
   }
 
   Location( const std::string &filename,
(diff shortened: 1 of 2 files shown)


bstaletic (commented):
Thanks for the pull request!

Personally, I don't see how this increases readability, but that might be just me. What I do have a problem with is that, with the proposed patch, half of the object is initialized in the init-list and half in the default member initializers. I'd prefer or the other. As far as I know, the "modern" way is to do it all in default member initializers.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1542#-:-MVE-ftKER3PE2RC_8qo:bh7n7xl)* status: 0 of 2 LGTMs obtained

fallingdrift:
Like this?

puremourning (approved):
:lgtm:

Thanks!

> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1542#-:-MngZ7vg2fkqy4X9G8KD:b9benhj)* status: 1 of 2 LGTMs obtained (waiting on @fallingdrift)

PR #1543, opened 2021-03-10
[READY] Assume input of Character constructor is normalized

bstaletic:
The assumption does hold for regular use, but tests don't always go
through `BreakCodePointsIntoCharacters()` and therefore need
`NormalizeInput()`.

Turns out this wasn't that hard. In normal use, after taking an array of identifiers, the following happens:

1. The identifier is broken into its constituent code points.
2. The array of code points is then transformed into an array of normalized strings, where each string represents a "character" (nitpick: technically a grapheme cluster).
3. Finally, we construct a `Character` from each of those strings.
4. To do this and correctly do the matching we want, we need to decompose the `Character`'s into normalized (according to Normalization Form D) code points. Note: These code points could be different than the ones we have started with, even though the accumulated `CodePoint::Normal()` is the same. Especially in cases like ligatures where a single code point can be normalized into dozens. This is where this PR departs from master.
5. Master doesn't assume the input is a normalized string which makes writing tests easy - non-normalized strings will be normalized. This PR takes advantage of input strings already being normalized, as renormalization is expensive. That meant that a ton of stuff in `cpp/ycm/tests` should be explicitly normalized before being passed to either `CharacterRepository::GetCharacters()` or `Character::Character()`.
5.1. Most of the stuff happens to actually be normalized from the start and doesn't really *require* `NormalizeInput()`, but I opted for consistency over minimalizam to reduce the WTF/line ratio.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1543)

### DIFF
--- a/cpp/ycm/Character.cpp
+++ b/cpp/ycm/Character.cpp
@@ -19,6 +19,7 @@
 #include "CodePoint.h"
 
 #include <algorithm>
+#include <cassert>
 
 namespace YouCompleteMe {
 
@@ -65,14 +66,8 @@ CodePointSequence CanonicalSort( CodePointSequence code_points ) {
 // Canonical Decomposition. See
 // https://www.unicode.org/versions/Unicode13.0.0/ch03.pdf#G733
 CodePointSequence CanonicalDecompose( std::string_view text ) {
-  CodePointSequence code_points = BreakIntoCodePoints( text );
-  std::string normal;
-
-  for ( const auto &code_point : code_points ) {
-    normal.append( code_point->Normal() );
-  }
-
-  return CanonicalSort( BreakIntoCodePoints( normal ) );
+  assert( NormalizeInput( text ) == text );
+  return CanonicalSort( BreakIntoCodePoints( text ) );
 }
 
 } // unnamed namespace
@@ -104,6 +99,17 @@ Character::Character( std::string_view character )
         base_.append( code_point->FoldedCase() );
     }
   }
+}
+
+
+std::string NormalizeInput( std::string_view text ) {
+    CodePointSequence code_points = BreakIntoCodePoints( text );
+    std::string normal;
+
+    for ( const auto &code_point : code_points ) {
+      normal.append( code_point->Normal() );
+    }
+    return normal;
 }
 
 } // namespace YouCompleteMe
--- a/cpp/ycm/Character.h
+++ b/cpp/ycm/Character.h
@@ -108,6 +108,8 @@ private:
 };
 
 
+YCM_EXPORT std::string NormalizeInput( std::string_view text );
+
 using CharacterSequence = std::vector< const Character * >;
 
 } // namespace YouCompleteMe
--- a/cpp/ycm/tests/CharacterRepository_test.cpp
+++ b/cpp/ycm/tests/CharacterRepository_test.cpp
@@ -42,7 +42,10 @@ protected:
 
 
 TEST_F( CharacterRepositoryTest, GetCharacters ) {
-  CharacterSequence character_objects = repo_.GetCharacters( { "α", "ω" } );
+  CharacterSequence character_objects = repo_.GetCharacters( {
+    NormalizeInput( "α" ),
+    NormalizeInput( "ω" )
+  } );
 
   EXPECT_THAT( repo_.NumStoredCharacters(), 2 );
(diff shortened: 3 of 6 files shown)


bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1543)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[cpp/ycm/Character.h, line 111 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1543#-MVRc9TyBwx0oRJ3DH09:-MVRc9TyBwx0oRJ3DH0A:blqy6mp) ([raw file](https://github.com/ycm-core/ycmd/blob/96e1287d27c73fc1829c8ed828229e7de0feeb9d/cpp/ycm/Character.h#L111)):*
> ```objc
> 
> 
> YCM_EXPORT std::string NormalizeInput( std::string_view text );
> ```

Should we wrap this up in `#ifdef DEBUG`? It's only used for testing and the tests will already scream at you if you violate this assumption.
___
*[cpp/ycm/Character.cpp, line 70 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1543#-MVRccqi5L2epLq-6L3V:-MVRccqi5L2epLq-6L3W:b-ig4u1p) ([raw file](https://github.com/ycm-core/ycmd/blob/96e1287d27c73fc1829c8ed828229e7de0feeb9d/cpp/ycm/Character.cpp#L70)):*
> ```cpp
> CodePointSequence CanonicalDecompose( std::string_view text ) {
>   assert( NormalizeInput( text ) == text );
>   return CanonicalSort( BreakIntoCodePoints( text ) );
> ```

Considering this assumption, I tried to directly move `text` into `Character::normal_`, but that doesn't work as it skips `CanonicalSort`. An idea would be to use a `span<CodePoints>` and sort them early on, but the problem with that is that the "old" code points need not be constructed from normalized input.

Considering those two problems, I don't think we can move into `normal_` without sacrificing some correctness.

puremourning (approved):
:lgtm:

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1543#-:-MWFYJI8CnEq56X55Ejm:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[cpp/ycm/Character.h, line 111 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1543#-MVRc9TyBwx0oRJ3DH09:-MWFY-m59i36GU-s78bR:bz4yhgu) ([raw file](https://github.com/ycm-core/ycmd/blob/96e1287d27c73fc1829c8ed828229e7de0feeb9d/cpp/ycm/Character.h#L111)):*

Should we wrap this up in `#ifdef DEBUG`? It's only used for testing and the tests will already scream at you if you violate this assumption.
</blockquote></details>

do we build the tests with -DDEBUG ?

I thought we didn't ?

Anyway I'm not worried about this being exposed. It's ycm_core.so is not _really_ a public API anyway.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1543)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[cpp/ycm/Character.h, line 111 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1543#-MVRc9TyBwx0oRJ3DH09:-MWFiharC5XTnQ5pLDcR:bgyjqdb) ([raw file](https://github.com/ycm-core/ycmd/blob/96e1287d27c73fc1829c8ed828229e7de0feeb9d/cpp/ycm/Character.h#L111)):*
> do we build the tests with -DDEBUG ?

> I thought we didn't

You're right. We could still wrap this in `#ifndef NDEBUG`. Would binary size improve measurable? If you don't think it's worth the effort, merge away.

puremourning (approved):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1543)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

puremourning (approved):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1543)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1545, opened 2021-04-01
[READY] Use C++17 parallel algorithms for filtering and sorting of candidates

bstaletic:
Like I said in gitter, GCC 9+ needs Intel TBB for the parallel algorithms.
On the other hand, clang 9+ supports these C++17 additions, but only runs sequentially.
MSVC, for once, is the easier to deal with. It actually works out of the box.

That should explain the decisions in the PR:

- The usage of `std::execution::par_unseq` is guarded by `__has_include(<execution>)`, which makes clang 7 and clang 8 work.
- For gcc 9+, there's a check in the CMakeLists.txt to see if `-ltbb` is valid. If it isn't, we fail the build.
  - That last sentence is seriously questionable. Should we make parallel algorithms opt-in? Opt-out? Should we fetch TBB with `FetchContent` and provide `USE_SYSTEM_TBB`?

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1545)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -280,20 +280,20 @@ endif()
 file( WRITE ${CMAKE_CURRENT_BINARY_DIR}/main.cpp "#include <filesystem>\nint main( int argc, char ** argv ) {\n  std::filesystem::path p( argv[ 0 ] );\n  return p.string().length();\n}" )
 try_compile( STD_FS_NO_LIB_NEEDED ${CMAKE_CURRENT_BINARY_DIR}
              SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
-	     CXX_STANDARD 17
-	     CXX_STANDARD_REQUIRED TRUE
-	     CXX_EXTENSIONS FALSE )
+             CXX_STANDARD 17
+             CXX_STANDARD_REQUIRED TRUE
+             CXX_EXTENSIONS FALSE )
 try_compile( STD_FS_NEEDS_STDCXXFS ${CMAKE_CURRENT_BINARY_DIR}
              SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
-	     CXX_STANDARD 17
-	     CXX_STANDARD_REQUIRED TRUE
-	     CXX_EXTENSIONS FALSE
+             CXX_STANDARD 17
+             CXX_STANDARD_REQUIRED TRUE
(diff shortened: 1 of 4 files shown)


bstaletic:
Regarding performance, at least on the two machines that I have tried, we break even at 256 candidates, after which I'm seeing almost linear scaling in the number of threads. I can provide the numbers.

bstaletic:
Now with 100% fewer race cnoidtions!

The only thing remaining is "what the hell is macOS doing?"
So @puremourning, please tell me if you, on macOS 10.15 have the `<execution>` header. If you do, what the hell is in it? Because the only scenario I have not accounted for is "`<execution>` exists, but is empty".

puremourning (commented):
`<execution>`"


```
// -*- C++ -*-
//===------------------------- execution ---------------------------------===//
//
// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//

#ifndef _LIBCPP_EXECUTION
#define _LIBCPP_EXECUTION

#include <__config>

#if defined(_LIBCPP_HAS_PARALLEL_ALGORITHMS) && _LIBCPP_STD_VER >= 17
#   include <__pstl_execution>
#endif

#endif // _LIBCPP_EXECUTION
```

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1545#-:-MXI1nsoBcCTkPwKCZ-s:bsar5vo)* status: 0 of 2 LGTMs obtained

puremourning (commented):
But it looks like apple's standard library has not been built with LIBCPP_HAS_PARALLEL_ALGORITHMS:



> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1545#-:-MXI2IbK2BI1MAwtwiUP:b-tu348l)* status: 0 of 2 LGTMs obtained

bstaletic (commented):
It also says `#error` in the other branch! I'll guard against it.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1545#-:-MXI5yoXAcX9EtLFciRJ:b-uxjxo4)* status: 0 of 2 LGTMs obtained

bstaletic:
Scrapping this. One would think one could use a C++17 standard library when one's using a C++17 compiler, but noooo...

PR #1554, opened 2021-04-14
[READY] Fix wrong offset when calculating parameter offsets in /signature_help

bstaletic:
Before this commit, we were passing 0-based columns into
`CodepointOffsetToByteOffset` in LSP completer, when constructing the
`/signature_help` response.

That's clearly a violation of `CodepointOffsetToByteOffset`'s API, which
expects 1-based input. The solution is to adjust the offsets on input,
by adding 1, then subtract one from the return value so that we still
return 0-based offsets, as promised in ycmd's API.

This should be our side of https://github.com/ycm-core/YouCompleteMe/issues/3870.
The other part is in https://github.com/phpactor/phpactor/issues/1235

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1554)

### DIFF
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -1529,8 +1529,8 @@ class LanguageServerCompleter( Completer ):
         else:
           begin, end = arg_label
         arg[ 'label' ] = [
-          utils.CodepointOffsetToByteOffset( sig_label, begin ),
-          utils.CodepointOffsetToByteOffset( sig_label, end ) ]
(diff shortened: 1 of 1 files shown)


puremourning (approved):
:lgtm:

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1554#-:-MYFbrub9s6ZX0BhfGxX:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic:
:lgtm:

PR #1562, opened 2021-04-21
Fixes submodule cwd in build.py

AP2008:
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1562)

### DIFF
--- a/build.py
+++ b/build.py
@@ -72,15 +72,6 @@ def _CheckCall( args, **kwargs ):
     sys.exit( error.returncode )
 
 
-for folder in os.listdir( DIR_OF_THIRD_PARTY ):
-  abs_folder_path = p.join( DIR_OF_THIRD_PARTY, folder )
-  if p.isdir( abs_folder_path ) and not os.listdir( abs_folder_path ):
-    print( "Updating submodules:" )
-    CheckCall(
-      [ "git", "submodule", "update", "--init", "--recursive" ],
-      exit_message = "Please run: git submodule update --init --recursive"
-    )
-
 NO_DYNAMIC_PYTHON_ERROR = (
   'ERROR: found static Python library ({library}) but a dynamic one is '
   'required. You must use a Python compiled with the {flag} flag. '
@@ -1162,8 +1153,29 @@ def DoCmakeBuilds( args ):
   BuildWatchdogModule( args )
 
 
+def CheckSubmodules( quiet ):
+  for folder in os.listdir( DIR_OF_THIRD_PARTY ):
+    abs_folder_path = p.join( DIR_OF_THIRD_PARTY, folder )
+    if p.isdir( abs_folder_path ) and not os.listdir( abs_folder_path ):
+      if not quiet:
+        print( "Updating submodules:" )
+      wd = os.getcwd()
+      try:
+        os.chdir( DIR_OF_THIS_SCRIPT )
+        CheckCall(
+          [ "git", "submodule", "update", "--init", "--recursive" ],
+          exit_message = "Please run: git submodule update --init --recursive",
+          quiet = quiet,
+          status_message = "Updating submodules"
+        )
+      finally:
+        os.chdir( wd )
+
+
 def Main():
   args = ParseArguments()
+
+  CheckSubmodules( args.quiet )
 
   if not args.skip_build:
     DoCmakeBuilds( args )


puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1562)* status: 0 of 2 LGTMs obtained (waiting on @AP2008)

___
*[build.py, line 35 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1562#-MYnYN4X-iWhlko5XzgF:-MYnYN4X-iWhlko5XzgG:brcq0td) ([raw file](https://github.com/ycm-core/ycmd/blob/239378234e4b3ea605308d40d2e88eba397fc2db/build.py#L35)):*
> ```Python
> DIR_OF_THIRD_PARTY = p.join( DIR_OF_THIS_SCRIPT, 'third_party' )
> LIBCLANG_DIR = p.join( DIR_OF_THIRD_PARTY, 'clang', 'lib' )
> CUR_DIR = p.abspath( os.curdir )
> ```

rather than this, what we normally do is something inline like:

```python
wd = os.getcwd()
try:
   os.chdir( path_i_need_to_be_in )
   # do stuff
finally:
  os.chdir( wd )
```

AP2008:
@puremourning Updated

AP2008:
@puremourning Updated

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1562)* status: 0 of 2 LGTMs obtained (waiting on @AP2008)

___
*[build.py, line 35 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1562#-MYnYN4X-iWhlko5XzgF:-MYnbwsf94fXkgfilxeC:bw6rpdq) ([raw file](https://github.com/ycm-core/ycmd/blob/239378234e4b3ea605308d40d2e88eba397fc2db/build.py#L35)):*

rather than this, what we normally do is something inline like:

```python
wd = os.getcwd()
try:
   os.chdir( path_i_need_to_be_in )
   # do stuff
finally:
  os.chdir( wd )
```
</blockquote></details>

I think you missed my point. don't save the global curdir in a global variable, use the one that happens to be current when the function is called. That's what the code I posted does.

don't use `os.curdir`, use `os.getcwd()`

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1562)* status: 0 of 2 LGTMs obtained (waiting on @AP2008)

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1562)* status: 0 of 2 LGTMs obtained (waiting on @AP2008)

___
*[build.py, line 78 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1562#-MYngvvxCvffXJd-BXIH:-MYngvvxCvffXJd-BXII:bnuls0m) ([raw file](https://github.com/ycm-core/ycmd/blob/8d1e53d299910d28fc8a61726040256e9703b3be/build.py#L78)):*
> ```Python
>   abs_folder_path = p.join( DIR_OF_THIRD_PARTY, folder )
>   if p.isdir( abs_folder_path ) and not os.listdir( abs_folder_path ):
>     print( "Updating submodules:" )
> ```

don't use `print` here. Instead, add the `status_message` argument to `CheckCall` so that `--quiet` mode looks sane.

Did you test --quiet mode ?

AP2008:
@puremourning There's a slight problem here. The args are being parsed later in the `Main` function. How do I know if --quiet was passed ?

AP2008:
@puremourning I fixed it and now the --print option works.

AP2008:
@puremourning Why does the CodeCov fail ?

puremourning (approved):
no idea. that's standard practice for Codecov.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1562#-:-MYolpC23ZCxt2l6t-bi:b-l2gdd8)* status: 1 of 2 LGTMs obtained (waiting on @AP2008)

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1562#-:-MYsYNUJC2rD-a-mUYFa:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @AP2008)

bstaletic (approved):
Generally looks good. Here's one concern with doing `git submodule update` in general:

1. If we contain it to `ycmd` directory, it won't take care of YCM's submodules. So are we actually helping users in this case?
2. It might be interfering with testing submodule updates. Currently that can only cause annoyances with regex and watchdog submodules. Right now I can update a submodule, compile, run the tests and then commit. With automatic submodule update, that's not the case.

I'd like @puremourning's thoughts on this.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1562#-:-MZ7QIvw7185PD4pqO2u:b-td21j4)* status: 1 of 2 LGTMs obtained (waiting on @AP2008)

puremourning:
1 is a v good point. Though YCM has few submodules now, it still makes this only partially affective. 

2 - I thought of this too then thought ‘it’s ok because it only does this when dir is empty’. But it occurs to me (as it did then) that this change should _omly_ plapply tonempty subdirs ie do it per directory not ‘everywhere’

bstaletic (approved):
1 - If we adopt ycm-core/YouCompleteMe#3809, YCM's only submodule is ycmd. Perhaps this check should be moved to YCM's `install.py`?
2 - good point. We can do it for empty dirs only. 

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1562#-:-MZ7fHyhD1LOnq-su6QR:bjvztoo)* status: 1 of 2 LGTMs obtained (waiting on @AP2008)

puremourning:
what's the status of this PR ?

AP2008:
@puremourning Do I have to make any changes ?

puremourning:
I think just fix the conflicts and we can merge it?

puremourning:
Seems abandoned, and I'm concerned about installations that aren't from git.

PR #1566, opened 2021-04-28
[READY] Silence MSVC warnings

bstaletic:
Abseil produces ~40'000 lines of warnings when ycmd is compiled on Windows. This should silence them.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1566)

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -182,6 +182,11 @@ if ( MSVC )
   set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP /bigobj /utf-8" )
 endif()
 
+# Makes MSVC conform to the standard
+if ( MSVC )
+  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /permissive- /Zc:__cplusplus /Zc:externConstexpr /Zc:inline /Zc:throwingNew" )
+endif()
+
 # Enables Python compilation for 64-bit Windows. Already defined by Python
 # headers when compiling with MSVC.
 if( WIN32 OR CYGWIN OR MSYS AND NOT MSVC AND 64_BIT_PLATFORM )
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -332,11 +332,6 @@ if ( ${ABSL_SUPPORTED} )
   target_compile_definitions( ${PROJECT_NAME} PUBLIC YCM_ABSEIL_SUPPORTED )
 endif()
 
-# Makes MSVC conform to the standard
-if ( MSVC )
-  target_compile_options( ${PROJECT_NAME} PUBLIC "/permissive-" )
-endif()
-
 target_include_directories(
   ${PROJECT_NAME}
   SYSTEM


bstaletic:
Aaaaand it didn't work at all.

puremourning (commented):
should we be passing `/std:c++17`  to msvc too?

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1566#-:-MZU2RQz-SOlN0rnlPXT:b-92lzm6)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1566)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[cpp/CMakeLists.txt, line 187 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1566#-MZU3Huk6agnhGscqHLb:-MZU3Huk6agnhGscqHLc:b-cf4q7r) ([raw file](https://github.com/ycm-core/ycmd/blob/114641ebd95c38431c92215a9422965e131110ae/cpp/CMakeLists.txt#L187)):*
> ```CMake
> # Makes MSVC conform to the standard
> if ( MSVC )
>   set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /permissive- /Zc:__cplusplus /Zc:externConstexpr /Zc:inline /Zc:preprocessor /Zc:throwingNew" )
> ```

like wow. reading the msvc docs for this is like pain after pain. They are like "enable standard conformant behaviour, because our old default behaviour was just bad, or pre standardisation and we don't want to break everyone by "fixing" it "

bstaletic (commented):
> should we be passing /std:c++17  to msvc too?

No, that's taken care of a bit further below where we set `CMAKE_CXX_STANDARD.

&nbsp;

There are some more warnings, but I have no idea how do fix them while staying portable (other than ugly preprocessor magic).

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1566#-:-MZaVOSk7kly7yXjyDju:b-8r79fq)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[cpp/CMakeLists.txt, line 187 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1566#-MZU3Huk6agnhGscqHLb:-MZaVbE-0S46NO7n54e6:b-ict0fl) ([raw file](https://github.com/ycm-core/ycmd/blob/114641ebd95c38431c92215a9422965e131110ae/cpp/CMakeLists.txt#L187)):*

like wow. reading the msvc docs for this is like pain after pain. They are like "enable standard conformant behaviour, because our old default behaviour was just bad, or pre standardisation and we don't want to break everyone by "fixing" it "
</blockquote></details>

So... here's a quick rehash of what each flag does:

1. `/permissive-` enforces two-phase template instantiation (among other things) and VS 2017 can't compile pybind11 in C++17 mode without it. We absolutely need it.
2. Without `/Zc:__cplusplus` MSVC goes "mhmm... `#define __cplusplus 199711L` looks about right." Then it defines `_MSVC_LANG` to the correct value of `__cplusplus`. This was causing 40k lines of warnings from abseil.
3. `/Zc:externConstexpr` fixes linkage of constexpr variables.
4. `/Zc:inline` is supposed to remove unreferenced `inline` functions.
5. `/Zc:throwingNew` says "operator new doesn't return NULLs".

In other words, MSVC isn't a C++ compiler. It is a compiler that compiler something that only quite resembles C++.

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1566#-:-MZaydSPBts8byVYFAAi:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1569, opened 2021-05-12
[READY] Fix Standard library includes with Xcode 12.5

puremourning:
Previously the "platform" didn't include the standard library, so we
always ended up using the standard library from the "toolchain". As of
12.5 the platform includes a copy of the standard library which leads to
2 incompatible ones being included. This leads to lots of errors.

Solution is to only use one or t'other, preferring the platform one.

Tested this with/without clangd on Xcode 12.0.5 and Xcode 12.5 - passes.
Tested iOS project with Xcode 12.0.5 and 12.5 - works with libclang, fails with clangd (but confirmed this is the case irrespective of these changes, so much be a clangd problem)

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1569)

### DIFF
--- a/third_party/generic_server/client/package-lock.json
+++ b/third_party/generic_server/client/package-lock.json
@@ -1,8 +1,264 @@
 {
 	"name": "lsp-sample-client",
 	"version": "0.0.1",
-	"lockfileVersion": 1,
+	"lockfileVersion": 2,
 	"requires": true,
+	"packages": {
+		"": {
+			"name": "lsp-sample-client",
+			"version": "0.0.1",
+			"license": "MIT",
+			"dependencies": {
+				"vscode-languageclient": "^6.1.3"
+			},
+			"devDependencies": {
+				"@types/vscode": "1.43.0",
+				"vscode-test": "^1.3.0"
+			},
+			"engines": {
+				"vscode": "^1.43.0"
+			}
+		},
+		"node_modules/@types/vscode": {
+			"version": "1.43.0",
+			"resolved": "https://registry.npmjs.org/@types/vscode/-/vscode-1.43.0.tgz",
+			"integrity": "sha512-kIaR9qzd80rJOxePKpCB/mdy00mz8Apt2QA5Y6rdrKFn13QNFNeP3Hzmsf37Bwh/3cS7QjtAeGSK7wSqAU0sYQ==",
+			"dev": true
+		},
+		"node_modules/agent-base": {
+			"version": "4.3.0",
+			"resolved": "https://registry.npmjs.org/agent-base/-/agent-base-4.3.0.tgz",
+			"integrity": "sha512-salcGninV0nPrwpGNn4VTXBb1SOuXQBiqbrNXoeizJsHrsL6ERFM2Ne3JUSBWRE6aeNJI2ROP/WEEIDUiDe3cg==",
+			"dev": true,
+			"dependencies": {
+				"es6-promisify": "^5.0.0"
+			},
+			"engines": {
+				"node": ">= 4.0.0"
+			}
+		},
+		"node_modules/balanced-match": {
+			"version": "1.0.0",
+			"resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
+			"integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
+			"dev": true
+		},
+		"node_modules/brace-expansion": {
+			"version": "1.1.11",
+			"resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
+			"integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
+			"dev": true,
+			"dependencies": {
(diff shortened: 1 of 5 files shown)


bstaletic:
Libclang `flags_test.py` is failing. Looks like we end up missing `-isystem /usr/include/c++/v1` from the flags. Was that intentional?

jchrys:
this change fixes issue on my device. thanks for pointing this out.

puremourning (commented):
Fixed the tests.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1569#-:-M_vddyrAm2MW1nQ_URD:b-379xbn)* status: 0 of 2 LGTMs obtained

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1569)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

bstaletic (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1569)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[third_party/generic_server/package.json, line 60 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1569#-M_voG7F5Q1CVKBZZqAD:-M_voG7F5Q1CVKBZZqAE:b-k74s9u) ([raw file](https://github.com/ycm-core/ycmd/blob/df346430d75cd52d67db7c28d524947a62f2a65d/third_party/generic_server/package.json#L60)):*
> ```JSON
> 		"@typescript-eslint/parser": "^2.3.0",
> 		"eslint": "^6.4.0",
> 		"mocha": "^8.1.1",
> ```

Did you change this? Was it intentional?
___
*[third_party/generic_server/server/package-lock.json, line 1 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1569#-M_vpoXkDulef4gWbR_L:-M_vpoXkDulef4gWbR_M:b-1nidbb) ([raw file](https://github.com/ycm-core/ycmd/blob/df346430d75cd52d67db7c28d524947a62f2a65d/third_party/generic_server/server/package-lock.json#L1)):*
> ```JSON
> {
> ```

Are these `package-lock.json` things even supposed to be committed?
___
*[ycmd/completers/cpp/flags.py, line 558 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1569#-M_vpNeN9AbMuht2L-yT:-M_vpNeN9AbMuht2L-yU:b3pei9p) ([raw file](https://github.com/ycm-core/ycmd/blob/df346430d75cd52d67db7c28d524947a62f2a65d/ycmd/completers/cpp/flags.py#L558)):*
> ```Python
>     if os.path.exists( platform_stdlib ):
>       flags.extend( [ '-isystem', platform_stdlib ] )
>     elif toolchain:
> ```

Do we want the else here? Previously, if `toolchain` was true, we'd add both `toolchain/usr/include/c++/v1` and `sysroot/usr/include/c++/v1`. I get that it can be seen as weird to add both, but just wanted to double check.

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1569)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[third_party/generic_server/package.json, line 60 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1569#-M_voG7F5Q1CVKBZZqAD:-M_w4NK9Det8T236sFST:b5cruwf) ([raw file](https://github.com/ycm-core/ycmd/blob/df346430d75cd52d67db7c28d524947a62f2a65d/third_party/generic_server/package.json#L60)):*

Did you change this? Was it intentional?
</blockquote></details>

I just updated it because that's what run_tests.py does and was left spamming in my client.
___
*[third_party/generic_server/server/package-lock.json, line 1 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1569#-M_vpoXkDulef4gWbR_L:-M_w4W4L9Q5VLLbtVslj:b-wg6qac) ([raw file](https://github.com/ycm-core/ycmd/blob/df346430d75cd52d67db7c28d524947a62f2a65d/third_party/generic_server/server/package-lock.json#L1)):*

Are these `package-lock.json` things even supposed to be committed?
</blockquote></details>

yes, they are, apparently
___
*[ycmd/completers/cpp/flags.py, line 558 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1569#-M_vpNeN9AbMuht2L-yT:-M_w4YV-C7QRsZGsbHEi:b-v1a185) ([raw file](https://github.com/ycm-core/ycmd/blob/df346430d75cd52d67db7c28d524947a62f2a65d/ycmd/completers/cpp/flags.py#L558)):*

Do we want the else here? Previously, if `toolchain` was true, we'd add both `toolchain/usr/include/c++/v1` and `sysroot/usr/include/c++/v1`. I get that it can be seen as weird to add both, but just wanted to double check.
</blockquote></details>

the whole purpose of this change is _not_ to add both - that's what was causing the problem. Previously the first one (the sys root one) didn't exist, now it does so you end up with 2 incompatible stdlibs if you add both.

bstaletic (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1569#-:-M_wCFUgDXErlSdR69Pd:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

PR #1572, opened 2021-06-11
[RFC] Manually parallelize sorting and filtering

bstaletic:
This may be just <strike>~100</strike> ~200 lines, but that's deceptive.

Two things are being parallelized independently - the `PartialSort` and the `FilterAndSortCandidates` functions' loops.

### The filtering algorithm

We only do parallelization if we have >=256 candidates to filter. If we do, we create nproc \* `std::vector partial_results`. Each thread gets its own `partial_results` vector and processes `candidates.size() / nproc` candidates.

Then we ~~`join()` all the threads~~ `get()` all the futures in the main thread and accumulate all the `partial_results` vectors into a "big" `results_and_objects` vector that then needs to be sorted.

Note: ~~TSAN complains about the way I'm using `partial_results` vectors. They are defined in the main thread, then populated in different threads (one thread populates one and only one `partial_results` object). Then the threads are `join()`-ed. Finally, once only the main thread is left, it accumulates all the `partial_results` vectors into the final `result_and_objects`. This all sounds fine at first - we populate the vectors independenty and join the threads, then accumulate once there's only one thread left. Naively, this should work without any synchronization. And... it really seems to work, apart from TSAN getting angry. Language lawyer: This is UB, because, without synchronization, there's no guarantee the main thread will ever see writes from the other threads, `thread::join()` be damned!~~

~~The correct way would be:~~

1. ~~Either move the definition of `partial_results` into the spawned threads, do the work, lock a mutex and append to the "big vector" under mutex.~~
2. ~~Or use `std::packaged_task`/`std::future`, so that we can "return" the partial results from each thread.~~

EDIT: TSAN did complain, but I initially read the report completely wrong. `thread::join()` does synchronize things nicely. It's also irrelevant now that we use a thread pool. The actual race was about wrong capture of a loop counter variable.

### The sorting algorithm

Now that we have a `results_and_objects`, we spawn new nproc threads to sort it.\* Initially, each thread does `std::sort` on its portion of the vector. Once the initial sorting is done, we're left with nproc sorted partitions of the vector. Naively, a single thread sort over the whole vector would work, but we can do better.

    {[(0 1) (2 3)] [(4 5) (6 7)]}

The numbers represent a partition index. The different kinds of brackets represent how one would want to call `std::inplace_merge` to sort the whole vector. "Outer" calls depend on "inner" calls and are represented with different kinds of brackets. Independent calls aren't nested and thus:

- From thread that sorted partition 0:
  - First, merge 0 and 1. Make sure 1 has finished before trying to merge 0 and 1.
  - Second, merge 0/1 with 2/3 once 2 has finished its job.
  - Finally, merge 0/1/2/3 with 4/5/6/7 once 4 notifies that it has done its job.
- From the thread that sorted partition 1 (or 3, or 5, or 7):
  - Nothing. At no point in time are these threads/partitions at the left side of a pair of partitions. Simply notify everyone that the job is done.
- From the thread that sorted partition 2 (or 6):
  - First merge 2 with 3, once 3 is done and notify everyone.
- From the thread that sorted partition 4:
  - First merge 4 and 5.
  - Then 4/5 and 6/7. Then notify.

An obvious question - how do we figure out which thread should wait for which other thread(s)? ***EASY!***

First, notice that you have to wait in order. That is T1 needs to wait for T2, then T3, then T5. Similarly for other threads. That means we can't just have a single atomic flag, but need a bool vector.

In other words:

1. `0` waits for `{1, 2, 4}`
2. `1` waits for `{}`
3. `2` waits for `{3}`
4. `3` waits for `{}`
5. `4` waits for `{5, 6}`
6. `5` waits for `{}`
7. `6` waits for `{7}`

In binary:

1. `0b0000` waits `{0b0001, 0b0010, 0b0100}` and would also wait for `8` (`0b1000`) had there been 9 threads.
2. `0b0001` waits `{}`
3. `0b0010` waits `{0b0011}`
4. `0b0011` waits for `{}`
5. `0b0100` waits for `{0b0101, 0b0110}`
6. `0b0101` waits for `{}`
7. `0b0110` waits for `{0b0111}`

So there should be two conditions  for "there's nothing to wait any more:

1. `current_thread_index + next_chunk_to_wait_for > n_threads - 1` - likely to happen with non-power-of-two threads.
2. Courtesy of @YannickJadoul `~current_index & (next_chunk_to_wait_for << 1)`. The idea is that we stop as soon as a set bit of current index matches any of the set bits of the `next_wait` index. I won't pretend that it is clear, but no matter what numbers I pick, it gets the right result. **Auditing very welcome.**
2.1. Tests are passing and TSAN\*\* is says this is all fine.

We have a `condition_variable` for notifying when some thread has finished its job. There's a bit vector. When a thread finishes its job, it sets the bit at the same index as the partition index above. We have a `cv_mutex` that is used for all of:

1. Synchronizing access to the condition_variable.
2. Synchronizing access to the bit vector.

\* Once again, we're only doing things in parallel if `max_candidates >= 256`.

\*\* The initial sorts that create partitions all operate on a single, big vector, there's no mutex, yet TSAN says it's fine. Because it is. Because I'm being careful to not overlap writes from different threads without synchronization. So, as long as each thread stays within its `begin`/`end` there's no problem. Language lawyer helmet: Technically, there's no parallel access to the array of `T`, because the language says that only scalar types can be accessed. I.e. `arr[0]` doesn't access an array, it only accesses the first element. Thus locking/synchronization is only needed once we start merging partitions.

### FQA

Q: Why no thread pool?
A: ~~Because that would mean dragging in a dependency or hand rolling one and I didn't feel like doing either, since we still gain a measurable performance improvement. `std::thread_pool_executor` should be in C++23, hopefully.~~
A: Says who?

Q: Why not parallelize `IdentifierDatabase::ResultsForQueryAndType()`?
A: Because that one is actually much harder to do. It uses `seen_candidates` set to avoid duplicate work which, as far as parallelizing `ResultsFor...`, is shared state. Skipping deduplication is a performance hit and synchronizing access to `seen_candidates` is an even bigger performance hit. Instead, we just leave `ResultsFor...` single threaded.

Q: WTF is that formatting?!
A: You didn't see it before "cleaning up". Hey! The PR is marked RFC, not READY!

Q: What happened to `<execution>`?
A: First, this is more efficient for `FilterAndSortCandidates()` and on par for the identifier completer. Second, LLVM doesn't yet enable their parallel STL. Third, libstdc++'s parallel algorithms depend on Intel's TBB.

Q: Should TSAN be a part of our CI?
A: I don't know. Should it? We can ditch Valgrind and use ASAN + UBSAN, MSAN and TSAN.

Q: Where benchmarks?
A: Right here!

### Benchmark summary

At a request, I'll upload the raw numbers.

1. Identifier completer benchmarks have shown a margin of error difference of ~0.5ms. Inconsistent, so no percentages.
2. Filter and sort with all candidates cached in the repository up front has seen ~10% improvement when sorting all and ~18% when sorting just top 50 candidates.
3. Filter and sort with no up-front caching in the repository has seen between 1% and 2% improvement.

With this, once candidates are cached in the repo, the C++ side of semantic completer is as fast as the C++ side of identifier completer.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1572)

### DIFF
--- a/cpp/ycm/PythonSupport.cpp
+++ b/cpp/ycm/PythonSupport.cpp
@@ -43,9 +43,9 @@ std::vector< const Candidate * > CandidatesFromObjectList(
     }
   } else {
     for ( size_t i = 0; i < num_candidates; ++i ) {
-        auto element = PyDict_GetItem( PyList_GET_ITEM( candidates.ptr(), i ),
-                                       candidate_property.ptr() );
-        *it++ = GetUtf8String( element );
+      auto element = PyDict_GetItem( PyList_GET_ITEM( candidates.ptr(), i ),
+                                     candidate_property.ptr() );
+      *it++ = GetUtf8String( element );
     }
   }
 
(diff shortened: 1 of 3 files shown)


bstaletic:
Updated the PR. Race condition is no more and I added a thread pool. Benchmark summary is in the original post's edit.

To whomever dares review this: read carefully. I've left a bunch of comments in the code that should be discussed.

puremourning:
is this PR int he endless bucket of "Waiting for Ben to actually look at it?"

bstaletic:
Nope, this one is in the "`std::bad_alloc` happens in CI on macOS... over ~22MB" bucket. I have no idea how to fight that.

PR #1574, opened 2021-06-13
Cmake refactor

Borwe:
Just some small refactoring:

1. Use add_compile_definitions instead of add_definitions
2. Shorten the code for checking c++17
3. Use FetchContent_MakeAvailable to make code simpler/shorter.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1574)


bstaletic (changes requested):
Nice! The fewer cmake lines the better.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1574#-:-Mc85tXJ5NOPZFpVIjzG:b-qfo7m4)* status: 0 of 2 LGTMs obtained (waiting on @Borwe)

___
*[cpp/CMakeLists.txt, line 153 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1574#-Mc850k-6bBUXGgOd12t:-Mc850k-6bBUXGgOd12u:bibxjgz) ([raw file](https://github.com/ycm-core/ycmd/blob/7d313d5fac367515841c39047d2d06795b7db4f1/cpp/CMakeLists.txt#L153)):*
> ```CMake
> # Determining the presence of C++17 support in the compiler
> list( FIND CMAKE_CXX_COMPILE_FEATURES cxx_std_17 HAS_CXX17)
> if( NOT HAS_CXX17 EQUAL -1 )
> ```

This loses the compiler version checks. I'd love to be able to do it like this, but... `std::filesystem` arrived late. GCC 7 does have support for C++17, but is lacking `std::filesystem`. Same goes for Clang 6 and MSVC 1910.
___
*[cpp/CMakeLists.txt, line 236 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1574#-Mc85f9-BtytMyixnXuW:-Mc85f9-BtytMyixnXuX:bhib040) ([raw file](https://github.com/ycm-core/ycmd/blob/7d313d5fac367515841c39047d2d06795b7db4f1/cpp/CMakeLists.txt#L236)):*
> ```CMake
>     SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/absl
>   )
>   FetchContent_MakeAvailable( absl )
> ```

A similar change can be made in `cpp/ycm/{tests,benchmarks}/CMakeLists.txt`.

Borwe:
___
*[cpp/CMakeLists.txt, line 153 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1574#-Mc850k-6bBUXGgOd12t:-Mc95DHg7SS-x5yKiGc9:boammlv) ([raw file](https://github.com/ycm-core/ycmd/blob/7d313d5fac367515841c39047d2d06795b7db4f1/cpp/CMakeLists.txt#L153)):*

This loses the compiler version checks. I'd love to be able to do it like this, but... `std::filesystem` arrived late. GCC 7 does have support for C++17, but is lacking `std::filesystem`. Same goes for Clang 6 and MSVC 1910.
</blockquote></details>

Wait, I thaught that was the whole point of the dance try_compile @ `cpp/ycm/CMakeLists.txt` with :

```
# Check if we need to add -lstdc++fs or -lc++fs or nothing
file( WRITE ${CMAKE_CURRENT_BINARY_DIR}/main.cpp "#include <filesystem>\nint main( int argc, char ** argv ) {\n  std::filesystem::path p( argv[ 0 ] );\n  return p.string().length();\n}" )
try_compile( STD_FS_NO_LIB_NEEDED ${CMAKE_CURRENT_BINARY_DIR}
             SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
             CXX_STANDARD 17
             CXX_STANDARD_REQUIRED TRUE
             CXX_EXTENSIONS ${NEEDS_EXTENSIONS} )
try_compile( STD_FS_NEEDS_STDCXXFS ${CMAKE_CURRENT_BINARY_DIR}
             SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
             CXX_STANDARD 17
             CXX_STANDARD_REQUIRED TRUE
             CXX_EXTENSIONS ${NEEDS_EXTENSIONS}
             LINK_LIBRARIES stdc++fs )
try_compile( STD_FS_NEEDS_CXXFS ${CMAKE_CURRENT_BINARY_DIR}
             SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
             CXX_STANDARD 17
             CXX_STANDARD_REQUIRED TRUE
             CXX_EXTENSIONS ${NEEDS_EXTENSIONS}
             LINK_LIBRARIES c++fs )
file( REMOVE ${CMAKE_CURRENT_BINARY_DIR}/main.cpp )

if( ${STD_FS_NEEDS_STDCXXFS} )
  set( STD_FS_LIB stdc++fs )
elseif( ${STD_FS_NEEDS_CXXFS} )
  set( STD_FS_LIB c++fs )
elseif( ${STD_FS_NO_LIB_NEEDED} )
  set( STD_FS_LIB "" )
else()
  message( FATAL_ERROR "Unknown compiler - C++17 filesystem library missing" )
endif()

```

bstaletic (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1574#-:-Mc99STkDyBowID9ubS1:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @Borwe and @puremourning)

___
*[cpp/CMakeLists.txt, line 153 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1574#-Mc850k-6bBUXGgOd12t:-Mc975Fv43Dpi9c134mZ:bsy17ul) ([raw file](https://github.com/ycm-core/ycmd/blob/7d313d5fac367515841c39047d2d06795b7db4f1/cpp/CMakeLists.txt#L153)):*
<details><summary><i>Previously, Borwe (Brian Orwe) wrote…</i></summary><blockquote>

Wait, I thaught that was the whole point of the dance try_compile @ `cpp/ycm/CMakeLists.txt` with :

```
# Check if we need to add -lstdc++fs or -lc++fs or nothing
file( WRITE ${CMAKE_CURRENT_BINARY_DIR}/main.cpp "#include <filesystem>\nint main( int argc, char ** argv ) {\n  std::filesystem::path p( argv[ 0 ] );\n  return p.string().length();\n}" )
try_compile( STD_FS_NO_LIB_NEEDED ${CMAKE_CURRENT_BINARY_DIR}
             SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
             CXX_STANDARD 17
             CXX_STANDARD_REQUIRED TRUE
             CXX_EXTENSIONS ${NEEDS_EXTENSIONS} )
try_compile( STD_FS_NEEDS_STDCXXFS ${CMAKE_CURRENT_BINARY_DIR}
             SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
             CXX_STANDARD 17
             CXX_STANDARD_REQUIRED TRUE
             CXX_EXTENSIONS ${NEEDS_EXTENSIONS}
             LINK_LIBRARIES stdc++fs )
try_compile( STD_FS_NEEDS_CXXFS ${CMAKE_CURRENT_BINARY_DIR}
             SOURCES ${CMAKE_CURRENT_BINARY_DIR}/main.cpp
             CXX_STANDARD 17
             CXX_STANDARD_REQUIRED TRUE
             CXX_EXTENSIONS ${NEEDS_EXTENSIONS}
             LINK_LIBRARIES c++fs )
file( REMOVE ${CMAKE_CURRENT_BINARY_DIR}/main.cpp )

if( ${STD_FS_NEEDS_STDCXXFS} )
  set( STD_FS_LIB stdc++fs )
elseif( ${STD_FS_NEEDS_CXXFS} )
  set( STD_FS_LIB c++fs )
elseif( ${STD_FS_NO_LIB_NEEDED} )
  set( STD_FS_LIB "" )
else()
  message( FATAL_ERROR "Unknown compiler - C++17 filesystem library missing" )
endif()

```
</blockquote></details>

Good point! Here's the history of both pieces:

Back in C++11 days, what came late was `std::regex` in gcc. We had to require gcc 4.9 instead of gcc 4.8.
Then it turned out that even gcc 4.9 had troubles with `std::regex`, so we decided to go back to Boost.Regex and gcc 4.8.
The compiler version checks remained.
Recently we moved to C++17 and my PR initially required Clang 8 and gcc 9, because I didn't want to bother with `-lc++fs`/`-lstdc++fs`/none thing.
Then @puremourning insisted that we should get Clang 7 and gcc 8 to work, which resulted in the code block you've just quoted.
That's how we arrived at "check versions, then check if `-lstdc++fs` is needed".

So the version checks are useful in more cases than just `std::filesystem` check, but you're right that *today* it's pointless.

Considering that the tests are passing, I'm not against the change that you proposed.

I'll leave the discussion open in Reviewable, so that @puremourning can see it.

puremourning (changes requested):
> Reviewed 2 of 2 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1574)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @Borwe and @puremourning)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1574#-MngaJMf3aKG4VT9JZ7O:-MngaJMf3aKG4VT9JZ7P:bn6pck9) (no related file):*
Thanks for the patch. My main concern with this is testing - it's not easy to test with the range of OS/compiler versions that we do support. In particular, the change to the c++17 logic is a little suspect to me - I'm not sure how much I trust cmake to get this right :)

Anyway, this patch changes 3 things. Ideally each change would be a separate PR, which could be independently reviewed and merged. In this case, I think I'm happy with the add_compile_definitions and FetchContent changes, but less confident in the c++17 logic.

---
___
*[cpp/CMakeLists.txt, line 152 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1574#-Mng_99Y33woJj6D3j1U:-Mng_99Y33woJj6D3j1V:b-iit2iz) ([raw file](https://github.com/ycm-core/ycmd/blob/4e345389c624fac3e0530b830884de8e68cccb8b/cpp/CMakeLists.txt#L152)):*
> ```CMake
> 
> # Determining the presence of C++17 support in the compiler
> list( FIND CMAKE_CXX_COMPILE_FEATURES cxx_std_17 HAS_CXX17)
> ```

space before `)`
___
*[cpp/CMakeLists.txt, line 157 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1574#-Mnga7Xi32eFPSE1etMM:-Mnga7Xi32eFPSE1etMN:bw1ez17) ([raw file](https://github.com/ycm-core/ycmd/blob/4e345389c624fac3e0530b830884de8e68cccb8b/cpp/CMakeLists.txt#L157)):*
> ```
>     set( CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LANGUAGE_STANDARD "c++17" )
>     set( CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LIBRARY "libc++" )
> ```

The previous logic set these for non-Apple clang too. we seem to not be doing that. Have we re-tested this build with non-apple clang on non-apple and on apple OS ?

bstaletic (approved):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1574)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @Borwe and @puremourning)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1574#-MngaJMf3aKG4VT9JZ7O:-Mo0dRmQA0pUDBx76Ob7:b-xwjp4n) (no related file):*

Thanks for the patch. My main concern with this is testing - it's not easy to test with the range of OS/compiler versions that we do support. In particular, the change to the c++17 logic is a little suspect to me - I'm not sure how much I trust cmake to get this right :)

Anyway, this patch changes 3 things. Ideally each change would be a separate PR, which could be independently reviewed and merged. In this case, I think I'm happy with the add_compile_definitions and FetchContent changes, but less confident in the c++17 logic.
</blockquote></details>

I was suspicious about the C++17 detection logic too, but then @borwe pointed out that we explicitly test the ability to correctly compile with `std::filesystem`. We should be safe.

---

bstaletic (approved):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1574)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @Borwe and @puremourning)

___
*[cpp/CMakeLists.txt, line 157 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1574#-Mnga7Xi32eFPSE1etMM:-Mo0fEnM7a4r9RsGdwzy:bvckdp) ([raw file](https://github.com/ycm-core/ycmd/blob/4e345389c624fac3e0530b830884de8e68cccb8b/cpp/CMakeLists.txt#L157)):*

> ```
>     set( CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LANGUAGE_STANDARD "c++17" )
>     set( CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LIBRARY "libc++" )
> ```

The previous logic set these for non-Apple clang too. we seem to not be doing that. Have we re-tested this build with non-apple clang on non-apple and on apple OS ?
</blockquote></details>

We should replace these with `set_target_property(TARGET ${TARGET} XCODE_ATTRIBUTE_${PROPERTY_NAME} ${PROPERTY_VALUE})`.

Borwe (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1574)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @Borwe, @bstaletic, and @puremourning)

___
*[cpp/CMakeLists.txt, line 157 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1574#-Mnga7Xi32eFPSE1etMM:-Ms-LXKA4nZ9zVDFqP4N:b-km9x3w) ([raw file](https://github.com/ycm-core/ycmd/blob/4e345389c624fac3e0530b830884de8e68cccb8b/cpp/CMakeLists.txt#L157)):*

We should replace these with `set_target_property(TARGET ${TARGET} XCODE_ATTRIBUTE_${PROPERTY_NAME} ${PROPERTY_VALUE})`.

 
</blockquote></details>

I think it's good for MacOS now, added back the Clang props of xcode. FYI, I don't have a Mac :stuck_out_tongue_closed_eyes:

bstaletic (approved):
What would boost our confidence in this PR is testing it on plain old Ubuntu 18.04, with whatever gcc 8. That setup has given us headaches previously.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1574#-:-Ms1wNN6CuZOF8pjzN6k:b-6se0yh)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @Borwe and @puremourning)

puremourning:
@MergifyIO rebase

puremourning:
@mergifyio rebase

puremourning:
Seems abandoned, and I'm not sure that the risk of this warrants the benefit we get from it.

PR #1575, opened 2021-06-27
[READY] Two minor fixes

bstaletic:
I'm lazy, so a single PR.

First commit discourages `sudo ./build.py`.
Second commit fixes a typo in `def Langauge()`.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1575)

### DIFF
--- a/build.py
+++ b/build.py
@@ -428,6 +428,10 @@ def ParseArguments():
                        help = 'For developers: specify the cmake executable. '
                               'Useful for testing with specific versions, or '
                               'if the system is unable to find cmake.' )
+  parser.add_argument( '--force-sudo',
+                       action = 'store_true',
+                       help = 'Compiling with sudo causes problems. If you'
+                              ' know what you are doing, proceed.' )
 
   # These options are deprecated.
   parser.add_argument( '--omnisharp-completer', action = 'store_true',
@@ -1150,8 +1154,14 @@ def DoCmakeBuilds( args ):
   BuildWatchdogModule( args )
 
 
-def Main():
+def Main(): # noqa: C901
   args = ParseArguments()
+
+  if 'SUDO_COMMAND' in os.environ:
+    if args.force_sudo:
+      print( 'Forcing build with sudo. If it breaks, keep the pieces.' )
+    else:
+      sys.exit( 'This script should not be run with sudo.' )
 
   if not args.skip_build:
     DoCmakeBuilds( args )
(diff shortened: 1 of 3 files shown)


puremourning (changes requested):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1575)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[build.py, line 1161 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1575#-MdCHT_p8zEu0MsK8Ofs:-MdCHT_p8zEu0MsK8Oft:bgy22dr) ([raw file](https://github.com/ycm-core/ycmd/blob/a3f0b40e9d18f0f29aaf901547d371a65adc074d/build.py#L1161)):*
> ```Python
> 
>   if 'SUDO_COMMAND' not in os.environ or args.force_sudo:
>     print( 'Forcing build with sudo. If it breaks, keep the pieces.' )
> ```

I think it will always say this? Even without SUDO_COMMAND in env.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1575)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[build.py, line 1161 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1575#-MdCHT_p8zEu0MsK8Ofs:-MdCKXdlD5z_W1vryl1u:br8obob) ([raw file](https://github.com/ycm-core/ycmd/blob/a3f0b40e9d18f0f29aaf901547d371a65adc074d/build.py#L1161)):*

I think it will always say this? Even without SUDO_COMMAND in env. 
</blockquote></details>

Obviously. What I initially wrote was correct:

```
  if 'SUDO_COMMAND' in os.environ:
    if args.force_sudo:
      print( 'Forcing build with sudo. If it breaks, keep the pieces.' )
    else:
      sys.exit( 'This script should not be run with sudo.' )
```

But flake8 didn't like it because cyclomatic complexity.

Fixed.

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1575#-:-MdCsheo7gl7uC2BbuId:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic:
:lgtm:

PR #1576, opened 2021-07-01
[READY] Document /ready request

bstaletic:
A few questions:

1. Do we want to document the subserver query, since we don't document it for `/healthy`? The comment says `/healthy` doesn't document `subserver` query because it's only intended for testing.
2. Can `/ready` ever return 500? I don't think so.

Thanks to @theli-ua for pointing out the bug in the docs.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1576)

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -208,6 +208,18 @@
                     </tr>
                     <tr>
                             <td class="swagger--summary-path" rowspan="1">
+                                <a href="#path--ready">/ready</a>
+                            </td>
+                        <td>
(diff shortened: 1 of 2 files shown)


puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1576#-:-MdcU8sG2S62ZMbjBxWx:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1579, opened 2021-07-16
[READY] Upgrade jdt.ls

bstaletic:
WIP because of the problem I mentioned in gitter:

```
That branch has an updated tsserver, gopls and jdt. Here's the problem. Retrying isolated tests brings down and starts up ycmd every time, so we're not *really* retrying anything.

This smells like that "pytest retry failures" bug, but we supposedly worked around it with those Dummy_tests. Yet it doesn't work. I can see jdt logging multiple teardowns/setups logged.

So I don't know if the remaining test failure in jdt diagnostics is a change in behaviour or jdt being slow + ycmd not retrying tests.
I'll send a PR, for easier testing.
```

More specifically, the test failures that I see on my laptop are:

1. `FileReadyToParse_Diagnostics_FileNotOnDisk_test` never sees any diagnostics.
2. `Poll_Diagnostics_ProjectWide_Eclipse_test` sees [this diagnostic](https://github.com/ycm-core/ycmd/blob/master/ycmd/tests/java/diagnostics_test.py#L83-L91) at column 1.
3. `FileReadyToParse_Diagnostics_Simple_test` sees the same diagnostic (link above) at the correct column.

Again, both could be symptoms of jdt being slow.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1579)


bstaletic:
Rebased and left only the jdt commit here. There's a problem to be solved regarding isolated ycmd and we already have the discussion here.

I'll pull gopls and tsserver out into a separate pr, as they don't need to be blocked.

puremourning (commented):
I wonder if this will work if rebated on the unittest PR ? If the issues are with retries being flaky, maybe unittest is the solution.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1579#-:-MjL2fPcExyaHaTWHxaq:bdq2v4e)* status: 0 of 2 LGTMs obtained

bstaletic (commented):
Unfortunately, that didn't help. I have a different branch, from today, where I attempted to do this with unittest. The diagnostics tests are still causing the same issues, except:

> FileReadyToParse_Diagnostics_FileNotOnDisk_test never sees any diagnostics.

That was a lie. It does see diagnostic. It's just that... we talk about `some/long/path/not_on_disk.java` and eclipse talks about `some/long/random_path/not_on_disk.java`, where both `path` and `random_path` are existing directories.

I have no idea what it is doing.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1579#-:-MjL4Gmw0vyBlUhmGldS:bvxoox8)* status: 0 of 2 LGTMs obtained

puremourning:
For the record, Boris raised the following 2 issues on jdt.ls:

* https://github.com/eclipse/eclipse.jdt.ls/issues/1940
* https://github.com/eclipse/eclipse.jdt.ls/issues/1815

bstaletic:
We could add `@ExpectedFailure` to these tests, but I'm not a big fan of that idea.

puremourning:
> We could add `@ExpectedFailure` to these tests, but I'm not a big fan of that idea.

I think we should add it. It's not like the regressions are anything we can fix. The "no file on disk" one is mostly just code coverage for us, and the incorrect diagnostics is not really fixable on our side. I know that we try to guarantee not to regress (that's why we have regression tests for the subserves anyway!) but we are so far behind on this one, I think we should just bite the bullet and do the update.

puremourning (approved):
Greeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeen

:lgtm_strong: 

> Reviewed 2 of 4 files at r2, 4 of 5 files at r5, 2 of 3 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1579#-:-MorXI-V8swDWBAj-hkN:b81z8jw)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1581, opened 2021-07-29
Add rust-analyzer test testing if proc macro works

theli-ua:
Failure can be triggered by using RUST_TOOLCHAIN = 'nightly-2021-06-22'

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1581)


theli-ua:
Broken tests should now be fixed

theli-ua:
ugh, I guess they are not https://github.com/theli-ua/ycmd/runs/3198679854?check_suite_focus=true  . Can't reproduce locally though :(

puremourning (changes requested):
> Reviewed 3 of 6 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1581)* status: 0 of 2 LGTMs obtained (waiting on @theli-ua)

___
*[ycmd/tests/rust/conftest.py, line 57 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1581#-MfwRLCO5-YEttclCN7j:-MfwRLCO5-YEttclCN7k:b-6t9fov) ([raw file](https://github.com/ycm-core/ycmd/blob/37edfc31d02066bb4dc915f3a67f90507f0af75c/ycmd/tests/rust/conftest.py#L57)):*
> ```Python
>     for message in PollForMessages( app, request_data ):
>       print( f'Message { pformat( message ) }' )
>       if message.get( 'message', '' ) == \
> ```

we prefer to avoid backslash line continuation. prefer to just wrap the string in `( ... )`

puremourning (commented):
Thanks for this. I re-approved the workflow. Seems I need to do that every time you push, which is super annoying. Feel free to ping me/boris if/when that happens again

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1581#-:-MfwSPOh-Cg7KWxRYUw_:b-jj5ynj)* status: 0 of 2 LGTMs obtained (waiting on @theli-ua)

theli-ua:
>I re-approved the workflow. Seems I need to do that every time you push, which is super annoying. Feel free to ping me/boris if/when that happens again

I realized github runs CI just fine in my fork and since it was still failing there I figured I'd ping you when I can actually make this work

theli-ua:
Ok, this now passes tests in CI https://github.com/theli-ua/ycmd/actions/runs/1085825346

theli-ua:
To clarify - the whole point of adding this test is making sure we're not going to add version which will have proc macro support broken since this is not covered by any existing tests atm.

theli-ua:
>urllib.error.HTTPError: HTTP Error 502: Bad Gateway

Looks like this needs a rerun

puremourning:
Something is flaking with the deps. Maybe too many requests? Not sure.  Will try again tomorrow

puremourning (approved):
:lgtm:

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1581#-:-Mg4tPiSABS0srrA8jw8:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @theli-ua)

bstaletic:
@theli-ua Just wanted to say thanks for working on this!

> I re-approved the workflow. Seems I need to do that every time you push

Only for the first pull request. For @theli-ua's next contribution, Github will trigger automatically.

puremourning:
yeah I thought it was only the first run on the first PR, but it seems to be every run on the first PR which is kinda annoying. but it is what it is. (I also have a sneaky plan to make this less annoying if we get a huge influx of new contributors)

PR #1585, opened 2021-08-18
Fallback to boost_filesystem if no filesystem lib is present

xloem:
On Termux, the standard clang auxiliary c++fs lib is not yet available, but boost is.

This patch provides for building on such systems.  Even if this change is not desired, it can still be useful for others to patch their builds going forward.

- fs:: namespace consolidated into Utils.h
- cmake filesystem test uses a non-inlined function (parent_path)
- detects and uses boost_filesystem if other options fail
- removed PrintTo test filesystem function which appears replaced

edit: force-pushed to resolve two test failures on other platforms caused by typos

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1585)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -277,12 +277,19 @@ if ( UNIX AND NOT ( APPLE OR SYSTEM_IS_OPENBSD OR HAIKU ) )
 endif()
 
 # Check if we need to add -lstdc++fs or -lc++fs or nothing
-file( WRITE ${CMAKE_CURRENT_BINARY_DIR}/main.cpp "#include <filesystem>\nint main( int argc, char ** argv ) {\n  std::filesystem::path p( argv[ 0 ] );\n  return p.string().length();\n}" )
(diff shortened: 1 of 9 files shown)


bstaletic:
First, thanks for the pull request.

Personally, I'm really not a fan of Boost.Filesystem, because of obscure bugs on Windows. Namely https://github.com/ycm-core/ycmd/pull/1080

An attempt at making `std::experimental::filesystem` work has been made here: https://github.com/ycm-core/ycmd/pull/1528

That means that bringing back Boost.Filesystem would require bringing back `YouCompleteMe::NormalizePath`, which is pretty annoying.

As for `ghc::filesystem`, do we know for certain it *doesn't* have obscure bugs that we would have to work around?

xloem:
I'm not sure if it's ever possible to know something for certain!   But one nice thing with individual projects like ghc is bug tracking.  If there's an issue and ghc is a submodule, a patch branch could be linked directly in until merged.

What do you think of simply disabling Boost on Windows systems, as a clear and concise approach?  Maybe with a comment linking to a discussion like this one?

puremourning (commented):
Thanks for the PR and MANY apologies for the delay in responding. I'm afraid that this is too much complexity to support a niche environment that doesn't have a functioning standard library. I feel like the real fix is to keep using the legacy branch until such a time as the target runtime environment supports ycmd's runtime requirements, rather than having a bunch of complex code in ycmd to support different implementations for different environments.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1585#-:-MngPIsd3-TT0nykZ_KA:b-xm3gz1)* status: 0 of 2 LGTMs obtained

PR #1588, opened 2021-08-30
Option to add general completers via ycm_general_completers global var

bilucodota:
[Tabnine](https://www.tabnine.com/) is a GPT2 based AI completer tool.
Currently Tabnine provides vim support through a [fork](https://github.com/codota/tabnine-vim) for YouCompleteMe which is not optimal to say the least.

As currently `ycmd` doesn't provide an API to register custom completer providers - 
I'm propsing to introduce a new global variable `ycm_general_completers` which users can use to add their own custom completers.

e.g: 
```
let g:ycm_general_completers = 
  \ [
  \  {
  \    'name': 'tabnine',
  \    'module': 'ycmd_tabnine.hook' 
  \  }
  \ ]
```

in this example, `ycmd_tabnine.hook` exposes a GetCompleter factory method same as other completers. 
e.g https://github.com/tabnine/tabnine-ycmd 

https://user-images.githubusercontent.com/70566578/131334533-e0daf04a-3b16-46c7-9c19-1ecbc1f74c4b.mp4

Appreciate your feedback

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1588)

### DIFF
--- a/ycmd/completers/general/general_completer_store.py
+++ b/ycmd/completers/general/general_completer_store.py
@@ -15,6 +15,8 @@
 # You should have received a copy of the GNU General Public License
 # along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
 
+from importlib import import_module
+from ycmd import responses
 from ycmd.completers.completer import Completer
 from ycmd.completers.all.identifier_completer import IdentifierCompleter
 from ycmd.completers.general.filename_completer import FilenameCompleter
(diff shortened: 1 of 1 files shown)


puremourning (commented):
Thanks for the PR, and MANY apologies for the delay in responding. Unfortunately, it's not practical for us to document and maintain a public python API at this level. We can't guarantee we won't break it and we have not hardened or documented this internal API to a level that we would be comfortable supporting out-of-tree completers.

if the server API to tabnine was public, it might be possible to integrate support for it using the exsiting completer model, but this idea of making the completer API public has been mooted before and we've concluded that it's not appropriate as an extension point external to the ycmd codebase.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1588#-:-MngO8vS9nivEWhcJyDn:b-g8qas1)* status: 0 of 2 LGTMs obtained

bilucodota:
@puremourning appreciate your feedback. 
I completely understand your concern. 
While Tabnine has a server API, it's somewhat not conventional I would say (due to optimization considerations).
This is why Tabnine provides a prebuilt client (which communicates with Tabnine servers). I've seen similar architectures with other completion providers.

I've implemented the communication layer with Tabnine binary here https://github.com/tabnine/ycmd/blob/master/ycmd/completers/general/tabnine_completer.py

Let me know of your thoughts.

PR #1590, opened 2021-09-06
Go get is deprecated in favor of install

Felixoid:
The current version returns a warning:

```
go get: installing executables with 'go get' in module mode is deprecated.
	Use 'go install pkg@version' instead.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
```

To install the binaries one should use `go install` nowadays.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1590)


bstaletic:
Thanks. We're aware of this warning, but this change will fail with go 1.15.

Felixoid:
Hmm, yeah. It looks like you are right. Then this PR should wait for some better time.

Felixoid:
BTW, just a point, 1.15 is not officially supported since 1.17 is released.

I understand for sure that it's still in use though

PR #1591, opened 2021-09-07
[READY] Migrate to unittest

bstaletic:
Finally ready.

Port to unittest

Things wrong with pytest:

1. It's trying to do too much and ends up getting in the way.
1.1. Most notably, rerunning tests is pretty broken - see
     pytest-dev/pytest-rerunfailures#51
2. Everything needs to be a fixture, instead of being compatible with
   argument-altering decorators.
3. Yet, parametrizing fixtures is an annoying chore.
4. Weird interaction with `@patch` and its arguments - they are in reversed order, which messes up quite a few things.

As a result, we're moving to the standard unittest, which looks similar
to nose. Here are the relevant differences:

- Writing tests:
  - Each test needs to be a member function, named `test_whatever()`.
  - Each test class needs to
    - Inherit from `unittest.TestCase`
    - Be called `ThingTest`
  - Instead of `yield`ing generated tests a `self.subTest()` is called
    in a loop: https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests
- Running tests:
  - There are far fewer command line options and the biggest problem was `--ignore`.
    - We have used `--ignore` to omit libclang tests from that one CI
      run. Now we add a `load_tests()` to `ycmd/tests/clang/__init__.py`. Reference
      https://docs.python.org/3/library/unittest.html#load-tests-protocol

### What to review

1. Non-python files. All of them.
2. All the `__init__.py`, as they are similar enough that every time I revisit them all, there are silly copy-paste bugs.
3. `run_tests.py`
4. `shutdown_test.py` - it's the one that replaced `@valgrind_skip` with `load_tests()`.
5. `test_utils.py`
5. Optionally: `utils_test.py` - it's the one that just dropped `@valgrind_skip`.

### In case you are a particularly brave reviewer

That is, if you want to review everything. First of all, make sure to ignore whitespace changes. Most of this is just a few mechanical changes:

1. Import `TestCase`
2. Put a `class FooTest( TestCase ):` before the first test.
3. Indent the rest of the file.
4. Rename tests from `Module_TestDescription_test` to `test_Module_TestDescription`
5. Add the `self` parameter.
6. Replace the mess of pytest's `@patch` arguments with what's sensible, if there are any of those.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1591)


bstaletic:
Finally. It only took me slightly less than a week. Ready for review, with the update description. Hopefully, the description will help with the review.

puremourning (approved):
:lgtm: with a few minor nits and a couple of questions. in particular the one about the `pattern` with the load tests protocol is worth checking.

> Reviewed 123 of 123 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1591#-:-MjKT3hd9MjYgTM59ru2:byhi9o4)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG1Pjo2VIC1akMoPAJ:-MjG1Pjo2VIC1akMoPAK:brsbbur) (no related file):*
do we need to update .vimspector.json? I can do this later if you want.

---
*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKR7vC8mx33CZvWl35:-MjKR7vC8mx33CZvWl36:b1gh0bx) (no related file):*
(optional) we've added `#noqa` to a number of places where we're importing `setUpModule`. The purpose of the import is to only have to write it once, I get that however, in many cases:

- the imports are added to existing import lists
- the new import lists are > 80 chars

so the `noqa` is suppressing some other errors. I'd prefer to put these imports individually:

```
from blah import setUpModule #noqa
from blah import tearDownModule #noqa
```

this narrows the exclusion and makes it a bit clearer what it's for. 

But this is a bit of a pain and only do it if you have the energy. I really don't mind that much. 

Same for many of the other comments, they are just bikeshedding nits and not that important.

---
___
*[run_tests.py, line 164 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG0mj29aR1StRdCeKg:-MjG0mj29aR1StRdCeKh:b-u85yqv) ([raw file](https://github.com/ycm-core/ycmd/blob/5fdb530008e52be3b057748a4b87529507b7df9a/run_tests.py#L164)):*
> ```Python
>                        action = 'store_true',
>                        help = 'Run tests inside valgrind.' )
>   parser.add_argument( '--no-parallel', action='store_true',
> ```

it would be a shame to lose parallel running. I take it that's not implemented/supported in unittest?
___
*[ycmd/tests/__init__.py, line 33 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG1pRC8N-Exf_O5Ple:-MjG1pRC8N-Exf_O5Plf:b-b927vh) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/__init__.py#L33)):*
> ```Python
>   """Defines a decorator to be attached to tests of this package. This decorator
>   passes the shared ycmd application as a parameter.
>   Do NOT attach it to test generators but directly to the yielded tests."""
> ```

is this terminology right now? I think we don't yield tests anymore but use subTest or something?
___
*[ycmd/tests/__init__.py, line 41 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG21nt07-4XCJ-Xw9b:-MjG21nt07-4XCJ-Xw9c:b-z8r9zy) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/__init__.py#L41)):*
> ```Python
>   def Wrapper( *args, **kwargs ):
>     ClearCompletionsCache()
>     return test( args[ 0 ], shared_app, *args[ 1: ], **kwargs )
> ```

using `args[ 0 ]` here seems odd. better to name the arg in the wrapper spec `def Wrapper( test_case_instance, *args, **kwargs )` ? 
___
*[ycmd/tests/__init__.py, line 50 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG2M7L4e-qKQDBHYeS:-MjG2M7L4e-qKQDBHYeT:buhrqzg) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/__init__.py#L50)):*
> ```Python
>     def Wrapper( *args, **kwargs ):
>       with IsolatedApp( custom_options ) as app:
>         test( args[ 0 ], app, *args[ 1: ], **kwargs )
> ```

here too ? `def Wrapper( test_case, *args, **kwargs )`
___
*[ycmd/tests/identifier_utils_test.py, line 410 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG70qr7cKero7_Jm2B:-MjG70qr7cKero7_Jm2C:b-kh2dzp) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/identifier_utils_test.py#L410)):*
> ```Python
>         [ 'Simple', '1stNonNumber', 'Subidentifier', 'Unicode' ] )
>     for i, args in enumerate( zipped ):
>       with self.subTest( args[ 1 ], ident = args[ 0 ], i = i ):
> ```

it seems a bit pointless to include `i` here as we're including all of (both) of the arguments anyway. the index is largely irrelevant if we know the ident and the ids. no problem with it, though I would have written this like:

```python
for ident, id in zipped:
  with self.subTest( ident=ident, id=id ):
    ...
```

or am I missing something?
___
*[ycmd/tests/shutdown_test.py, line 121 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjK8Jt3Azx8gDR9dr6G:-MjK8Jt3Azx8gDR9dr6H:b-tbopax) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/shutdown_test.py#L121)):*
> ```Python
> 
> 
> def load_tests( loader: unittest.TestLoader, tests, pattern ):
> ```

do we need to apply the `pattern` filter ourselves here or does that happen before or after this is called? Have you tried to `-k test_FromWatchdog*` or something ?
___
*[ycmd/tests/test_utils.py, line 412 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjK99g19tezmKgEtpEG:-MjK99g19tezmKgEtpEH:baixilz) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/test_utils.py#L412)):*
> ```
>   if os.environ.get( 'YCM_TEST_NO_RETRY' ) == 'XFAIL':
>     return expectedFailure
> ```

I think this means that the test will FAIL if the test passes From docs for expectedFailure

> If the test passes, it will be considered a failure.

I think that's wrong as it will make the test flaky if it happens to succeed first time. 

Better to raise `SkipTest` after the first failure
___
*[ycmd/tests/test_utils.py, line 430 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjK9cmj8Iy64ysZuC8P:-MjK9cmj8Iy64ysZuC8Q:babmp5v) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/test_utils.py#L430)):*
> ```Python
>         except Exception as test_exception:
>           run += 1
>           if run == opts[ 'reruns' ]:
> ```

maybe add:

```
 elif os.environ.get( 'YCM_TEST_NO_RETRY', None ) == 'XFAIL':
    raise SkipTest()
```
___
*[ycmd/tests/clangd/signature_help_test.py, line 25 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKFCxa8CJxEf9ugQCp:-MjKFCxa8CJxEf9ugQCq:b-848as3) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/clangd/signature_help_test.py#L25)):*
> ```Python
> 
> from ycmd import handlers
> from ycmd.tests.clangd import PathToTestFile, SharedYcmd, setUpModule, tearDownModule, IsolatedYcmd # noqa
> ```

I think we can make this 80 charcters even if we're STFUing falke8
___
*[ycmd/tests/cs/__init__.py, line 108 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKGMaB0DIgXkrrRtqZ:-MjKGMaB0DIgXkrrRtq_:b-x148si) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/cs/__init__.py#L108)):*
> ```Python
> 
>   if filepath not in shared_filepaths.setdefault( app, [] ):
>     # StartCompleterServer( app, 'cs', filepath )
> ```

why commented out ? if it's not needed, prefer to remove it (perhaps with a comment).
___
*[ycmd/tests/cs/signature_help_test.py, line 30 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKPK_v6a47UGPAWdUe:-MjKPK_v6a47UGPAWdUf:b-m3oa4r) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/cs/signature_help_test.py#L30)):*
> ```Python
>                             IsolatedYcmd,
>                             SharedYcmd,
>                             WrapOmniSharpServer, setUpModule, tearDownModule )
> ```

would prefer stylistically to put these on their own line. I really abhor bin-packing, but I particularly don't like mixing styles in one command :)
___
*[ycmd/tests/go/diagnostics_test.py, line 28 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKPpT3APZuWkfNuNCb:-MjKPpT3APZuWkfNuNCc:b-5026qs) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/go/diagnostics_test.py#L28)):*
> ```Python
> import json
> 
> from ycmd.tests.go import PathToTestFile, SharedYcmd, setUpModule, tearDownModule # noqa
> ```

check 80 chars.
___
*[ycmd/tests/rust/testdata/common/Cargo.lock, line 3 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKScL9ELVg6Fvq8X_Z:-MjKScL9ELVg6Fvq8X__:b9q515o) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/rust/testdata/common/Cargo.lock#L3)):*
> ```TOML
> # This file is automatically @generated by Cargo.
> # It is not intended for manual editing.
> version = 3
> ```

why was this file added? I have no problem, just curious what is creating it.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1591)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic and @puremourning)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKR7vC8mx33CZvWl35:-MjKYs2b10wHNrtv2y1f:btx2u8f) (no related file):*

(optional) we've added `#noqa` to a number of places where we're importing `setUpModule`. The purpose of the import is to only have to write it once, I get that however, in many cases:

- the imports are added to existing import lists
- the new import lists are > 80 chars

so the `noqa` is suppressing some other errors. I'd prefer to put these imports individually:

```
from blah import setUpModule #noqa
from blah import tearDownModule #noqa
```

this narrows the exclusion and makes it a bit clearer what it's for. 

But this is a bit of a pain and only do it if you have the energy. I really don't mind that much. 

Same for many of the other comments, they are just bikeshedding nits and not that important.
</blockquote></details>

Right, my laziness is showing. Done.

---
*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG1Pjo2VIC1akMoPAJ:-MjKf0F_19Nxa1iCFoG_:b-i31co3) (no related file):*

do we need to update .vimspector.json? I can do this later if you want.
</blockquote></details>

Please check the change to `.vimspector.json`, but I think I got it right.

---
___
*[run_tests.py, line 164 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG0mj29aR1StRdCeKg:-MjKZAbA4LqpmXE00Wos:b-1lclrh) ([raw file](https://github.com/ycm-core/ycmd/blob/5fdb530008e52be3b057748a4b87529507b7df9a/run_tests.py#L164)):*

it would be a shame to lose parallel running. I take it that's not implemented/supported in unittest?
</blockquote></details>

Not supported out of the box. We can opt for a different test runner. I tried using pytest as just a runner, but that fails with `self.subTest` context managers.

Maybe nose2? Quick [StackOverflow](https://stackoverflow.com/questions/4710142/can-pythons-unittest-test-in-parallel-like-nose-can) lead me to [testtools](https://github.com/testing-cabal/testtools).

Haven't really looked into alternative test runners beyond that.
___
*[ycmd/tests/__init__.py, line 33 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG1pRC8N-Exf_O5Ple:-MjKZw5I9Wu-iBMX_YKE:b-7k8hfk) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/__init__.py#L33)):*

is this terminology right now? I think we don't yield tests anymore but use subTest or something?
</blockquote></details>

That's what happens when you (read: I) dig through ancient commits to get the old code back. That comment is definitely wrong!

Comments deleted.
___
*[ycmd/tests/__init__.py, line 41 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG21nt07-4XCJ-Xw9b:-MjKby9b7GrSTdeYLG_E:b-1gto2f) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/__init__.py#L41)):*

using `args[ 0 ]` here seems odd. better to name the arg in the wrapper spec `def Wrapper( test_case_instance, *args, **kwargs )` ? 
</blockquote></details>

Definitely reads better. Done.
___
*[ycmd/tests/__init__.py, line 50 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG2M7L4e-qKQDBHYeS:-MjKc7LCB4USUDR1rTdg:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/__init__.py#L50)):*

here too ? `def Wrapper( test_case, *args, **kwargs )`
</blockquote></details>

Done.
___
*[ycmd/tests/identifier_utils_test.py, line 410 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG70qr7cKero7_Jm2B:-MjKcTGK50log9J5HE_8:b-fj4354) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/identifier_utils_test.py#L410)):*

it seems a bit pointless to include `i` here as we're including all of (both) of the arguments anyway. the index is largely irrelevant if we know the ident and the ids. no problem with it, though I would have written this like:

```python
for ident, id in zipped:
  with self.subTest( ident=ident, id=id ):
    ...
```

or am I missing something?
</blockquote></details>

You're overlooking that the index is passed down to `LoopExpectedLongestIdentifier()`, which then passes it over to C++ as `end_index`.

That was the reason these two tests were written in such an ugly way.

In python 2 the following was possible:

```
for i, (descr, identifier) in enumerate(zipped):
```
___
*[ycmd/tests/shutdown_test.py, line 121 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjK8Jt3Azx8gDR9dr6G:-MjKdCHh4fGuGBvQW7fG:b-8t2t7g) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/shutdown_test.py#L121)):*

do we need to apply the `pattern` filter ourselves here or does that happen before or after this is called? Have you tried to `-k test_FromWatchdog*` or something ?
</blockquote></details>

Right, we're completely ignoring the pattern here. Pattern matching isn't just a substring match, so I put it aside. Particularly because `-k` is allowed to match fully qualified module names.

https://docs.python.org/3/library/unittest.html#cmdoption-unittest-k
___
*[ycmd/tests/test_utils.py, line 412 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjK99g19tezmKgEtpEG:-MjKfGWCAYWqyvtvTGJr:be2rfp3) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/test_utils.py#L412)):*

> ```
>   if os.environ.get( 'YCM_TEST_NO_RETRY' ) == 'XFAIL':
>     return expectedFailure
> ```

I think this means that the test will FAIL if the test passes From docs for expectedFailure

> If the test passes, it will be considered a failure.

I think that's wrong as it will make the test flaky if it happens to succeed first time. 

Better to raise `SkipTest` after the first failure
</blockquote></details>

Isn't the point of expected failures to fail if assertion passes.

To be honest, I don't remember when this `YCM_TEST_NO_RETRY == XFAIL` thing was added. Do you remember?
___
*[ycmd/tests/test_utils.py, line 430 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjK9cmj8Iy64ysZuC8P:-MjKfeF0CJOmEpaI23O1:b-1eh3wi) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/test_utils.py#L430)):*

maybe add:

```
 elif os.environ.get( 'YCM_TEST_NO_RETRY', None ) == 'XFAIL':
    raise SkipTest()
```
</blockquote></details>

Sure, done.
___
*[ycmd/tests/clangd/signature_help_test.py, line 25 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKFCxa8CJxEf9ugQCp:-MjKdxzt1UAVgFwB3OA3:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/clangd/signature_help_test.py#L25)):*

I think we can make this 80 charcters even if we're STFUing falke8
</blockquote></details>

Done.
___
*[ycmd/tests/cs/__init__.py, line 108 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKGMaB0DIgXkrrRtqZ:-MjKe2fx7dHa0-5LyA1T:b-yh27p9) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/cs/__init__.py#L108)):*

why commented out ? if it's not needed, prefer to remove it (perhaps with a comment).
</blockquote></details>

Good question. Probably a debugging remnant. Will look into it.
___
*[ycmd/tests/cs/signature_help_test.py, line 30 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKPK_v6a47UGPAWdUe:-MjKeDV8-_lgsjY2HldZ:b-10il2a) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/cs/signature_help_test.py#L30)):*

would prefer stylistically to put these on their own line. I really abhor bin-packing, but I particularly don't like mixing styles in one command :)
</blockquote></details>

As said before, laziness. Done.
___
*[ycmd/tests/go/diagnostics_test.py, line 28 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKPpT3APZuWkfNuNCb:-MjKeMJC0hAfOhYRsTjp:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/go/diagnostics_test.py#L28)):*

check 80 chars.
</blockquote></details>

Done.
___
*[ycmd/tests/rust/testdata/common/Cargo.lock, line 3 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKScL9ELVg6Fvq8X_Z:-MjKeOn_5aPwTX-SeuIO:b-spty8v) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/rust/testdata/common/Cargo.lock#L3)):*

why was this file added? I have no problem, just curious what is creating it.
</blockquote></details>

I think this is an artifact of rust-analyzer running `cargo check` on our test data. We could put it in `.gitignore` if you'd prefer.

Rust docs say that libraries should not commit `Cargo.lock`, but that executables should. We're neither, so we don't need to listen to others.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1591)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

___
*[ycmd/tests/cs/__init__.py, line 108 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKGMaB0DIgXkrrRtqZ:-MjKiT4O8Q1zHDuPqY5p:bb20u88) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/cs/__init__.py#L108)):*

Good question. Probably a debugging remnant. Will look into it.
</blockquote></details>

Turns out it's not my debugging remnant. It was like that before, in `conftest.py`. Probably even before pytest, as I know I didn't write `cs/__init__.py` or `cs/conftest.py`. Well, my name is on them, but I am not the one who originally came up with `WrapOmniSharpServer`.

Considering that things have been working with that thing commented out, I decided to remove the line. Not pushed, because I'm waiting to see if there will be other review comments.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1591)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

___
*[.vimspector.json, line 4 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKkx1p-12_DAVtgMpj:-MjKkx1p-12_DAVtgMpk:bolvtbj) ([raw file](https://github.com/ycm-core/ycmd/blob/5c35f3fd6e30bad45b5903d1a94080bce83be1bf/.vimspector.json#L4)):*
> ```JSON
>   "$schema": "https://puremourning.github.io/vimspector/schema/vimspector.schema.json",
>   "configurations": {
>     "python - launch pytest": {
> ```

This should say `unittest` instead of `pytest`.

puremourning (changes requested):
> Reviewed 65 of 65 files at r2.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1591)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic and @puremourning)

___
*[.vimspector.json, line 37 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKwwzj0iJwLIcBfKQ9:-MjKwwzj0iJwLIcBfKQA:b-qc4bsf) ([raw file](https://github.com/ycm-core/ycmd/blob/5c35f3fd6e30bad45b5903d1a94080bce83be1bf/.vimspector.json#L37)):*
> ```JSON
>         "args": [
>           "-v",
>           "${Test}"
> ```

I'm thinking that we will need to change this to be more useful, as my current compiler plugin will set `${Test}` to `<filename>:<test_function>`. in order to do that we'd need the args to be `-k <test_function> <filename>` I think.

I'll sort that out next time I happen to use it. I suspect the answer is to switch to using a splat `*${TestArgs}` then I'll change my custom stuff to pass the arguments as a string like `-k <test_function> <filename>`. Vimspector will then splice the args array together.
___
*[run_tests.py, line 164 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG0mj29aR1StRdCeKg:-MjKxgZb1hajUvY8Lcym:b-jsqb6w) ([raw file](https://github.com/ycm-core/ycmd/blob/5fdb530008e52be3b057748a4b87529507b7df9a/run_tests.py#L164)):*

Not supported out of the box. We can opt for a different test runner. I tried using pytest as just a runner, but that fails with `self.subTest` context managers.

Maybe nose2? Quick [StackOverflow](https://stackoverflow.com/questions/4710142/can-pythons-unittest-test-in-parallel-like-nose-can) lead me to [testtools](https://github.com/testing-cabal/testtools).

Haven't really looked into alternative test runners beyond that.
</blockquote></details>

Not urgent right now. We can get this back if we find a way in future. We didn't have it for years. It's a shame to lose it, but it's not the end of the world. Stable tests is always the priority.
___
*[ycmd/tests/shutdown_test.py, line 121 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjK8Jt3Azx8gDR9dr6G:-MjKy42z8oVJ1TjsUAU1:bxqtphg) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/shutdown_test.py#L121)):*

Right, we're completely ignoring the pattern here. Pattern matching isn't just a substring match, so I put it aside. Particularly because `-k` is allowed to match fully qualified module names.

https://docs.python.org/3/library/unittest.html#cmdoption-unittest-k
</blockquote></details>

Does this mean it's no longer possible to run a specific test from within this module? That would be kinda sad.
___
*[ycmd/tests/test_utils.py, line 412 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjK99g19tezmKgEtpEG:-MjKyCbDC5yc8BygO2Tz:bqrjta4) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/test_utils.py#L412)):*

Isn't the point of expected failures to fail if assertion passes.

To be honest, I don't remember when this `YCM_TEST_NO_RETRY == XFAIL` thing was added. Do you remember?
</blockquote></details>

I think the idea was, rather than retrying, if the test fails, mark it as expected failure. This just speeds up test runs and allows you to focus on things which are actually failing. However, the current implementation would _require_ the test to fail first time.

If I added the 'XFAIL' value for YCM_TEST_NO_RETRY, I don't remember doing it and I had totally forgotten about it, though I guess I added it for the reason above, speed up full test runs (particularly java) where there are flaky tests that just require re-running a number of times.

Anyway 
___
*[ycmd/tests/test_utils.py, line 430 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjK9cmj8Iy64ysZuC8P:-MjKyl5b0vDTsZNB_aAg:b-rqdrzt) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/test_utils.py#L430)):*

Sure, done.
</blockquote></details>

This is good, but we need to remove the block above, which makes this unreachable.
___
*[ycmd/tests/test_utils.py, line 411 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKzNPZ2V6wZJ1ytvlX:-MjKzNP_-AnmG1C62iq_:b-fze5c9) ([raw file](https://github.com/ycm-core/ycmd/blob/5c35f3fd6e30bad45b5903d1a94080bce83be1bf/ycmd/tests/test_utils.py#L411)):*
> ```Python
>   opts = { 'reruns': 20, 'reruns_delay': 0.5 }
>   opts.update( kwargs )
>   if os.environ.get( 'YCM_TEST_NO_RETRY' ) == 'XFAIL':
> ```

remove these two lines
___
*[ycmd/tests/cs/__init__.py, line 108 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKGMaB0DIgXkrrRtqZ:-MjKywYB48YME-xHY56i:bb73djm) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/cs/__init__.py#L108)):*

Turns out it's not my debugging remnant. It was like that before, in `conftest.py`. Probably even before pytest, as I know I didn't write `cs/__init__.py` or `cs/conftest.py`. Well, my name is on them, but I am not the one who originally came up with `WrapOmniSharpServer`.

Considering that things have been working with that thing commented out, I decided to remove the line. Not pushed, because I'm waiting to see if there will be other review comments.
</blockquote></details>

SGTM
___
*[ycmd/tests/language_server/__init__.py, line 56 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKzxgvDe2ogIlffzYk:-MjKzxgvDe2ogIlffzYl:bmq17do) ([raw file](https://github.com/ycm-core/ycmd/blob/5c35f3fd6e30bad45b5903d1a94080bce83be1bf/ycmd/tests/language_server/__init__.py#L56)):*
> ```Python
>       with IsolatedApp( custom_options ) as app:
>         try:
>           test( args[ 0 ], app, *args[ 1: ], **kwargs )
> ```

test_case_instance
___
*[ycmd/tests/rust/testdata/common/Cargo.lock, line 3 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKScL9ELVg6Fvq8X_Z:-MjKz1yU8v6nypDzDB4F:bf1vu23) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/rust/testdata/common/Cargo.lock#L3)):*

I think this is an artifact of rust-analyzer running `cargo check` on our test data. We could put it in `.gitignore` if you'd prefer.

Rust docs say that libraries should not commit `Cargo.lock`, but that executables should. We're neither, so we don't need to listen to others.
</blockquote></details>

thanks for the explanation, this is fine.

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1591)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic)

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1591)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic and @puremourning)

___
*[.vimspector.json, line 37 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKwwzj0iJwLIcBfKQ9:-MjL0matAexoySalg74u:b-nbpbo0) ([raw file](https://github.com/ycm-core/ycmd/blob/5c35f3fd6e30bad45b5903d1a94080bce83be1bf/.vimspector.json#L37)):*

I'm thinking that we will need to change this to be more useful, as my current compiler plugin will set `${Test}` to `<filename>:<test_function>`. in order to do that we'd need the args to be `-k <test_function> <filename>` I think.

I'll sort that out next time I happen to use it. I suspect the answer is to switch to using a splat `*${TestArgs}` then I'll change my custom stuff to pass the arguments as a string like `-k <test_function> <filename>`. Vimspector will then splice the args array together.
</blockquote></details>

You could try matching `module.name.ClassTest.test_function`. Do you want me to revert these changes? Or...
___
*[run_tests.py, line 164 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjG0mj29aR1StRdCeKg:-MjL-_QpDKLdfnz-4qrQ:bm4htkt) ([raw file](https://github.com/ycm-core/ycmd/blob/5fdb530008e52be3b057748a4b87529507b7df9a/run_tests.py#L164)):*

Not urgent right now. We can get this back if we find a way in future. We didn't have it for years. It's a shame to lose it, but it's not the end of the world. Stable tests is always the priority.
</blockquote></details>

I also found `unittest-parallel`, but it wasn't packaged for my distro, so didn't bother with setting up a virtual env.
___
*[ycmd/tests/shutdown_test.py, line 121 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjK8Jt3Azx8gDR9dr6G:-MjL12lP5ll_n5a6VoRQ:bbzoa7p) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/shutdown_test.py#L121)):*
Might be. Let me check. If the answer is yes, I'll simply match member function names. `-k` matching other parts of `module.Class.test` seems to be inconsistent.

> Changed in version 3.5: Discovery no longer checks package names for matching pattern due to the impossibility of package names matching the default pattern.
___
*[ycmd/tests/test_utils.py, line 412 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjK99g19tezmKgEtpEG:-MjL1dHyA6Bo4UQotfSR:bbo620u) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/test_utils.py#L412)):*

I think the idea was, rather than retrying, if the test fails, mark it as expected failure. This just speeds up test runs and allows you to focus on things which are actually failing. However, the current implementation would _require_ the test to fail first time.

If I added the 'XFAIL' value for YCM_TEST_NO_RETRY, I don't remember doing it and I had totally forgotten about it, though I guess I added it for the reason above, speed up full test runs (particularly java) where there are flaky tests that just require re-running a number of times.

Anyway 
</blockquote></details>

Okay, I see what is the problem with `expectedFailure` here. Also, irrelevant with these two lines removed, in favour of below.
___
*[ycmd/tests/test_utils.py, line 430 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjK9cmj8Iy64ysZuC8P:-MjL0FpxEkDDgH3f6mdr:b-2qsmfa) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/test_utils.py#L430)):*

This is good, but we need to remove the block above, which makes this unreachable.
</blockquote></details>

Done that too.
___
*[ycmd/tests/test_utils.py, line 411 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKzNPZ2V6wZJ1ytvlX:-MjL00yO4idhSevTLRg5:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/5c35f3fd6e30bad45b5903d1a94080bce83be1bf/ycmd/tests/test_utils.py#L411)):*

remove these two lines
</blockquote></details>

Done.
___
*[ycmd/tests/language_server/__init__.py, line 56 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKzxgvDe2ogIlffzYk:-MjL0U5S2VyH4eXWW10F:b-o7vuhc) ([raw file](https://github.com/ycm-core/ycmd/blob/5c35f3fd6e30bad45b5903d1a94080bce83be1bf/ycmd/tests/language_server/__init__.py#L56)):*

test_case_instance
</blockquote></details>

Good catch! Done.
___
*[ycmd/tests/rust/testdata/common/Cargo.lock, line 3 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKScL9ELVg6Fvq8X_Z:-MjL2ETi21lRccZK9L_F:b8fkdvh) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/rust/testdata/common/Cargo.lock#L3)):*

thanks for the explanation, this is fine.
</blockquote></details>

Decided to remove it, because some users complain about drive space.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1591)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

___
*[ycmd/tests/shutdown_test.py, line 121 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjK8Jt3Azx8gDR9dr6G:-MjL3T4y-YO7PfkSTaaG:b-jmp1tn) ([raw file](https://github.com/ycm-core/ycmd/blob/7602601e61d3e9dcb23fe405475653b878cc9318/ycmd/tests/shutdown_test.py#L121)):*

Might be. Let me check. If the answer is yes, I'll simply match member function names. `-k` matching other parts of `module.Class.test` seems to be inconsistent.

> Changed in version 3.5: Discovery no longer checks package names for matching pattern due to the impossibility of package names matching the default pattern.
</blockquote></details>

Okay, somehow the pattern is still being respected:

```
bstaletic@Gallifrey ycmd  (git)-[unittest]-% ./run_tests.py --skip-b --no-f ycmd/tests/shutdown_test.py -k Without -v
Running tests on Python 3.9.7
test_FromHandlerWithoutSubserver (ycmd.tests.shutdown_test.ShutdownTest) ... ok
test_FromWatchdogWithoutSubserver (ycmd.tests.shutdown_test.ShutdownTest) ... ok
/usr/lib/python3.9/subprocess.py:1052: ResourceWarning: subprocess 3250 is still running
  _warn("subprocess %s is still running" % self.pid,
ResourceWarning: Enable tracemalloc to get the object allocation traceback

Ran 2 tests in 4.322s

OK
bstaletic@Gallifrey ycmd  (git)-[unittest]-% ./run_tests.py --skip-b --no-f ycmd/tests/shutdown_test.py -k Watch -v
Running tests on Python 3.9.7
test_FromWatchdogWithSubservers (ycmd.tests.shutdown_test.ShutdownTest) ... ok
test_FromWatchdogWithoutSubserver (ycmd.tests.shutdown_test.ShutdownTest) ... ok

Ran 2 tests in 17.807s

OK
```

I guess nothing to do here.

puremourning (changes requested):
> Reviewed 4 of 5 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1591)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic)

___
*[.vimspector.json, line 37 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjKwwzj0iJwLIcBfKQ9:-MjhhZzoDAol-lslvl19:b-wis04t) ([raw file](https://github.com/ycm-core/ycmd/blob/5c35f3fd6e30bad45b5903d1a94080bce83be1bf/.vimspector.json#L37)):*

You could try matching `module.name.ClassTest.test_function`. Do you want me to revert these changes? Or...
</blockquote></details>

nah, this is fine; I'll patch it up when I need to.
___
*[ycmd/tests/cs/debug_info_test.py, line 131 at r5](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjhhxFqAsjvSHpr11lo:-MjhhxFqAsjvSHpr11lp:b-k9jez1) ([raw file](https://github.com/ycm-core/ycmd/blob/02a10a8b9ce9540ffc247c17ebff689e793680fa/ycmd/tests/cs/debug_info_test.py#L131)):*
> ```Python
>   @SharedYcmd
>   def test_DebugInfo_UsesSuperfolderHint( self, app ):
>     print( app )
> ```

debug code ? do we need to keep it?

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1591)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

___
*[ycmd/tests/cs/debug_info_test.py, line 131 at r5](https://reviewable.io/reviews/ycm-core/ycmd/1591#-MjhhxFqAsjvSHpr11lo:-Mjji0XS2tVPvx-p4eNJ:b31iyus) ([raw file](https://github.com/ycm-core/ycmd/blob/02a10a8b9ce9540ffc247c17ebff689e793680fa/ycmd/tests/cs/debug_info_test.py#L131)):*

debug code ? do we need to keep it?
</blockquote></details>

Probably a leftover of "am I doing what I think I am doing" from when I was rewriting the decorators.

Removed.

puremourning (approved):
:lgtm:

> Reviewed 1 of 1 files at r6, all commit messages.
> Dismissed @bstaletic from a discussion.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1591#-:-Mjx2NojDX6TApkjNEFq:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic:
Manual merge required - I touched `.github/workflows`.

PR #1601, opened 2021-10-30
[READY] Upgrade Omnisharp-Roslyn and add OgranizeImports subcommand

bstaletic:
Let's see how well this goes.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1601)

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -234,6 +234,9 @@ class CsharpCompleter( Completer ):
          self._SolutionSubcommand( request_data,
                                    method = '_GoToSymbol',
                                    args = args ) ),
+      'OrganizeImports'                  : ( lambda self, request_data, args:
+         self._SolutionSubcommand( request_data,
+                                   method = '_OrganizeImports' ) ),
       'RefactorRename'                   : ( lambda self, request_data, args:
          self._SolutionSubcommand( request_data,
                                    method = '_RefactorRename',
@@ -689,6 +692,23 @@ class CsharpSolutionCompleter( object ):
     if not message:
       raise RuntimeError( 'No type info available.' )
     return responses.BuildDisplayMessageResponse( message )
+
+
+  def _OrganizeImports( self, request_data ):
+    request = self._DefaultParameters( request_data )
+    request[ 'WantsTextChanges' ] = True
+    result = self._GetResponse( '/fixusings', request )
+    fixit = responses.FixIt(
+      _BuildLocation(
+        request_data,
+        request_data[ 'filepath' ],
+        request_data[ 'line_num' ],
+        request_data[ 'column_codepoint' ] ),
+      _LinePositionSpanTextChangeToFixItChunks(
+        result[ 'Changes' ],
+        request_data[ 'filepath' ],
+        request_data ) )
+    return responses.BuildFixItResponse( [ fixit ] )
 
 
   def _Format( self, request_data ):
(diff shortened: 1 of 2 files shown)


bstaletic:
@xiguichen Here's a PR, but I don't know how to go about these test failures. Do we need `msbuild` and/or dotnet SDKs?

xiguichen:
@bstaletic  I think we don't need them.From the error, there should be some unit test failed. I made a fork of the project and checked that current master branch is OK, so the problem should not be caused by a dependency error.

bstaletic:
>  I think we don't need them.

I asked because, without `msbuild`, I see this in the omnishar-roslyn log:

      It looks like you have Mono 6.4.0 or greater installed but MSBuild could not be found.
      Try installing MSBuild into Mono (e.g. 'sudo apt-get install msbuild') to enable better MSBuild support, or upgrade to latest Mono: https://www.mono-project.com/download/stable/.

And when I install msbuild, I get this message:

      It looks like you have Visual Studio 2019 installed without .NET Core SDK support which is required by OmniSharp.
       Try updating Visual Studio 2019 installation with .NET Core SDK to enable better MSBuild support.

> From the error, there should be some unit test failed.

Ugh... the C# test failures get really messy... The summary is that Omnisharp-Roslyn, in some cases doesn't respond with diagnostics, so we end up thinking it isn't ready yet. Here's where the exception is coming from: https://github.com/ycm-core/ycmd/blob/master/ycmd/tests/cs/__init__.py#L138

Here's the list of failed tests:

```
ERROR: test_Diagnostics_MaximumDiagnosticsNumberExceeded (diagnostics_test.DiagnosticsTest)
ERROR: test_GetCompletions_Unicode (get_completions_test.GetCompletionsTest)
ERROR: test_Subcommands_GetDoc_Function (subcommands_test.SubcommandsTest)
ERROR: test_Subcommands_GetToImplementation_Unicode (subcommands_test.SubcommandsTest)
 FAIL: test_Diagnostics_WithRange (diagnostics_test.DiagnosticsTest)
```

Once again, thanks for helping us with this.

bstaletic:
Here's a link to a, hopefully, more readable log; https://gist.github.com/bstaletic/874a86f36d601d18b65c151bdb7e79a5

xiguichen:
I tried to reproduce the problem. But seems I get different problem in two of my PC. In one of my PC I setup today, i always have the following problem:

`
    def decode(self, input, final=False):
        # decode input (taking the buffer into account)
        data = self.buffer + input
>       (result, consumed) = self._buffer_decode(data, self.errors, final)
E       UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc9 in position 206: invalid continuation b
`

bstaletic:
If you have a non-utf-8 filepath, ycmd might trip on that. Anyway, now you can see why we haven't upgraded the omnisharp-roslyn dependency in such a long time.

The up side is that users can point YCM to any roslyn server they like with `g:ycm_roslyn_binary_path`.

If you don't know how to get this working, I'll drop the second commit and just leave the `OrganizeImports` part.

xiguichen:
May be you can drop the second commit for now. The failed testcase might be bug of OmniSharp again, I will do more test when I have time.

bstaletic:
Dropped.

puremourning (approved):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1601)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1601#-:-MngN9K_D49SpKevq7Hm:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1609, opened 2021-11-16
ts-completer: Give node more memory to avoid crashes

cmatheson:
I copied this value from vscode, so I think it's reasonable to just set this on behalf of the user, but arguably this could be a new config option?

One interesting thing I discovered while tracking down this  bug is that the node process out of memory error didn't make it into the tsserver logs. I'm assuming that's not something we have a lot of control over though.

### DIFF
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -207,6 +207,9 @@ class TypeScriptCompleter( Completer ):
     environ = os.environ.copy()
     environ[ 'TSS_LOG' ] = tsserver_log
 
+    # TSServer runs out of memory on larger projects. This is the value that VSCode uses.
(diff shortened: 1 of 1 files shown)


puremourning (approved):
Thanks ! :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1609#-:-Moe6lm64sFoTHH9YzNR:b-v9cd3x)* status: 1 of 2 LGTMs obtained (waiting on @cmatheson)

cmatheson:
sorry @puremourning , i had some sloppy formatting in there and failed the tests.  should be fixed now.

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1609#-:-Moeov248ILbOMulqixf:bnfp4nl)* status: :shipit: complete! 2 of 2 LGTMs obtained (waiting on @cmatheson)

PR #1611, opened 2021-11-18
[READY] Update checksums after having fixed ARM archives in ycm-core/llvm

bstaletic:
@puremourning Thanks for noticing my error and helping with figuring this out.

If we set out to do the M1 cross-compiling quickly, we may delay this PR. Also, please test the new archives on your side.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1611)

### DIFF
--- a/build.py
+++ b/build.py
@@ -1052,32 +1052,32 @@ def GetClangdTarget():
   if OnWindows():
     return [
       ( 'clangd-{version}-win64',
-        '5f4f8612af3f10bb52fd97d6196ac1653802f04f179d2069cf2c1fa861e167b5' ),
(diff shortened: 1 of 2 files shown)


puremourning:
successfully installed on CI and on my aarch64 ubuntu 20.04.

let's ship and update the submodule as the change in the hashes breaks installation right now.

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1611#-:-MomQ9WRCYskQsYT_hNF:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1618, opened 2021-12-09
[READY] Don't set a maximum Python version required

### LINKED ISSUE #1617: Python 3.10.* compile support
[cpp/CMakeLists.txt](https://github.com/ycm-core/ycmd/blob/e42cb71/cpp/CMakeLists.txt#L235) says ycmd supports Python 3.10:

```cmake
find_package( Python3 3.6...3.10 REQUIRED COMPONENTS Interpreter Development )
```

While Python 3.10 successfully compiles ycmd, Python 3.10.1 is rejected by this `find_package` check. I wonder if we can **relax this restriction** if later Python 3.10 versions are tested okay. 

### Compile log with [Python 3.10.1](https://www.python.org/ftp/python/3.10.1/python-3.10.1-macos11.pkg) on macOS 12.0.1 (M1)

```sh
$ cd ~/.vim/bundle/YouCompleteMe
$ python3.10 install.py --clang-completer --go-completer --java-completer --ts-completer --system-libclang

Generating ycmd build configuration...-- The C compiler identification is AppleClang 12.0.5.12050022
-- The CXX compiler identification is AppleClang 12.0.5.12050022
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /opt/homebrew/Cellar/cmake/3.22.1/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Python3: Found unsuitable version "3.10.1", required range
  is "3.6...3.10" (found
  /Library/Frameworks/Python.framework/Versions/3.10/bin/python3.10, found
  components: Interpreter Development Development.Module Development.Embed)
Call Stack (most recent call first):
  /opt/homebrew/Cellar/cmake/3.22.1/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:592 (_FPHSA_FAILURE_MESSAGE)
  /opt/homebrew/Cellar/cmake/3.22.1/share/cmake/Modules/FindPython/Support.cmake:3166 (find_package_handle_standard_args)
  /opt/homebrew/Cellar/cmake/3.22.1/share/cmake/Modules/FindPython3.cmake:490 (include)
  CMakeLists.txt:235 (find_package)


-- Configuring incomplete, errors occurred!
See also "/private/var/folders/51/p8vdz3c53gd27dfbgsrswwzh0000gn/T/ycm_build_v2ke2mnf/CMakeFiles/CMakeOutput.log".
See also "/private/var/folders/51/p8vdz3c53gd27dfbgsrswwzh0000gn/T/ycm_build_v2ke2mnf/CMakeFiles/CMakeError.log".

FAILED

ERROR: the build failed.

NOTE: it is *highly* unlikely that this is a bug but rather that this is a problem with the configuration of your system or a missing dependency. Please carefully read CONTRIBUTING.md and if you're sure that it is a bug, please raise an issue on the issue tracker, including the entire output of this script (with --verbose) and the invocation line used to run it.

The installation failed; please see above for the actual error. In order to get more information, please re-run the command, adding the --verbose flag. If you think this is a bug and you raise an issue, you MUST include the *full verbose* output.
```

puremourning:
We require Python 3.6+, but we're checking for 3.6 up to 3.10, which is
apparently interpreted as up to and including 3.10, but not 3.10.1.

Anyway, we have no "maximum" supported Python version, so we should just
check for a minimum of 3.6.

Fixes #1617

Incidentally, the 'version range' check we were doing is not actually
supported in our minimum cmake version (3.14). It was added in 3.19
according to the cmake docs, so this change also fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1618)

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -232,7 +232,7 @@ if ( NOT CPP17_AVAILABLE )
   message( FATAL_ERROR "Your C++ compiler does NOT fully support C++17." )
 endif()
 
-find_package( Python3 3.6...3.10 REQUIRED COMPONENTS Interpreter Development )
+find_package( Python3 3.6 REQUIRED COMPONENTS Interpreter Development )
 
 #############################################################################
 


puremourning:
Docs for reference: https://cmake.org/cmake/help/latest/command/find_package.html#id5

bstaletic:
Initially, I did set the max version, because I though setting just the lower bound would make cmake stick to that specific version.

I guess I thought wrong, since the tests are passing, but it is something to keep in mind, in cases where multiple python versions can be found on the system.

Maybe we should test with two different venv active?

puremourning:
> Initially, I did set the max version, because I though setting just the lower bound would make cmake stick to that specific version.

Not in my testing. I tested with 3.9 and 3.10 versions and they both picked whatever was in my path.

I'm not sure how I can test with multiple available versions.

More people are starting to see this so I'm inclined to merge it and see if anything breaks ?

puremourning:
Tests are failing because omnisharp keeps crashing 

> System.ArgumentException: '<Document65bf3716-8445-4dda-821f-83943de63c3b>' is not part of the workspace.
158

great.

puremourning (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1618#-:-Mq_IIUsAbq7JkiGGbIp:bnfp4nl)* status: 1 of 2 LGTMs obtained

bstaletic:
Great, then. :lgtm:

Sorry for the delay.

PR #1619, opened 2021-12-16
Fix go install for gopls

### LINKED ISSUE #1596: [go-completer] update gopls version and support go install
when build YCM with --go-completer, it's building error with following:

```
go get: installing executables with 'go get' in module mode is deprecated.
	To adjust and download dependencies of the current module, use 'go get -d'.
	To install using requirements of the current module, use 'go install'.
	To install ignoring the current module, use 'go install' with a version,
	like 'go install example.com/cmd@latest'.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
```

My go version
```
go version
go version go1.17.1 linux/amd64
```

the document said:
https://github.com/golang/tools/blob/master/gopls/doc/advanced.md#installing-unreleased-versions

```
# For go1.17 or older, the above `go get` command will build and
# install `gopls`. For go1.18+ or tip, run the following command to install
# using selected versions in go.mod.
go install golang.org/x/tools/gopls
```

However, it would be better to support new command as `go install`

To solve this problem for me change `get` -> `install` but it would break backward-compatible:
https://github.com/ycm-core/ycmd/blob/a7b972b770f04f2070dddf445e660f6fcfb172ee/build.py#L830

andrewstuart:
This should fix #1596, and works very well with go generics in my testing, out
of the box.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1619)

### DIFF
--- a/build.py
+++ b/build.py
@@ -863,7 +863,7 @@ def EnableGoCompleter( args ):
   new_env[ 'GOPATH' ] = p.join( DIR_OF_THIS_SCRIPT, 'third_party', 'go' )
   new_env.pop( 'GOROOT', None )
   new_env[ 'GOBIN' ] = p.join( new_env[ 'GOPATH' ], 'bin' )
-  CheckCall( [ go, 'get', 'golang.org/x/tools/gopls@v0.7.1' ],
+  CheckCall( [ go, 'install', 'golang.org/x/tools/gopls@latest' ],
              env = new_env,
              quiet = args.quiet,
              status_message = 'Building gopls for go completion' )


puremourning (commented):
How does this handle the problem that Boris says here https://github.com/ycm-core/ycmd/issues/1596#issuecomment-944437599

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1619#-:-Mr1FuIl8LX_4q7uE3Hj:b-ma7sl8)* status: 0 of 2 LGTMs obtained

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1619)* status: 0 of 2 LGTMs obtained (waiting on @andrewstuart and @puremourning)

___
*[build.py, line 866 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1619#-Mr1G1St1gb2zXjJ-BYQ:-Mr1G1St1gb2zXjJ-BYR:b-kys4s6) ([raw file](https://github.com/ycm-core/ycmd/blob/0ddc92743d0b7b7c3636d66c39959cd9971e1b30/build.py#L866)):*
> ```Python
>   new_env.pop( 'GOROOT', None )
>   new_env[ 'GOBIN' ] = p.join( new_env[ 'GOPATH' ], 'bin' )
>   CheckCall( [ go, 'install', 'golang.org/x/tools/gopls@latest' ],
> ```

We do not use unpinned server versions. Server vendors have a habit of breaking changes, so we must pin a tested and supported version.

PR #1621, opened 2021-12-31
Use git submodule for absl instead of FetchContent

Chilledheart:
For every run of cmake, FetchContent will remove the existing absl
directory and re-download. It is inconvenient and time-consuming. git
submodule doesn't have such issue.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1621)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,4 +109,3 @@ third_party/regex-build
 # CMake FetchContent output directories
 cpp/ycm/benchmarks/benchmark/
 cpp/ycm/tests/gmock/
-cpp/absl
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "third_party/mrab-regex"]
 	path = third_party/mrab-regex
 	url = https://bitbucket.org/mrabarnett/mrab-regex.git
+[submodule "cpp/absl"]
+	path = cpp/absl
+	url = https://github.com/abseil/abseil-cpp
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -240,17 +240,20 @@ set( CMAKE_POSITION_INDEPENDENT_CODE ON )
 if ( USE_SYSTEM_ABSEIL )
   find_package( absl REQUIRED )
 else()
-  include( FetchContent )
-  FetchContent_Declare(
-    absl
-    GIT_REPOSITORY https://github.com/abseil/abseil-cpp
-    GIT_TAG 3b4a16abad2c2ddc494371cc39a2946e36d35d11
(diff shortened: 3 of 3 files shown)


puremourning (commented):
Got submodule requires downloading the entire history. We have intentionally moved away from them. 

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1621#-:-MsEiWrCCjjpJOeDfP1t:b-4qt9im)* status: 0 of 2 LGTMs obtained

Chilledheart:
I think you need shallow support for git submodule which supported in git 2.10
```
(And git 2.10 Q3 2016 allows to record that with git config -f .gitmodules submodule.<name>.shallow true.
```

And in my case, FetchContent also downloads the full history of the git repository.

puremourning:
Thanks.

The reason we have stepped away from submodules is that users have in the past had (perhaps legitimate) complaints that ycm+ycmd takes up a lot of disk space. And that it is therefore bloaty" or whatever, and that git submodules are largely a pain and often users don't really understand them.

With the switch to FetchContent, we are able to temporarily download the dependencies and then get rid of them. This solves both issues. 

If we used a submodule, we would hold everything the entire time, whereas with the fetch content, we only need to do the download in when a rebuild is required, which is _very rare_ these days. You can check the CORE_VERSION file history - it doesn't update that often, so I worry that this change unfortunately negatively impacts users overall. :/

I wonder if there's a way instead for fetch content to download a tarball and thus not git clone

bstaletic:
> I wonder if there's a way instead for fetch content to download a tarball and thus not git clone

There is. I was thinking of trying that at one point (for the reason @Chilledheart explained - it's annoying on a metered connection). Then I forgot that idea.

We should still check if FetchContent redownloads tarballs.

bstaletic:
```
FetchContent_Declare(
  googletest
  GIT_REPOSITORY https://github.com/google/googletest.git
  GIT_TAG        703bd9caab50b139428cea1aaff9974ebee5742e # release-1.10.0
)

FetchContent_Declare(
  myCompanyIcons
  URL      https://intranet.mycompany.com/assets/iconset_1.12.tar.gz
  URL_HASH MD5=5588a7b18261c20068beabfb4f530b87
)
```

We could grab Abseil LTS: https://github.com/abseil/abseil-cpp/releases/tag/20211102.0

puremourning:
> We could grab Abseil LTS: https://github.com/abseil/abseil-cpp/releases/tag/20211102.0

SGTM.

I know abseil is all "live at head" and stuff, but using a tag seems reasonable to me.

puremourning:
we're not going to go back to submodules for this, and it looks like nobody is interested in implementing the change Boris suggested, so closing.

PR #1622, opened 2021-12-31
Fix rpath to libclang archive on mac

Chilledheart:
tested with `otool -l ycm_core.cpython-39-darwin.so|grep -A3 LC_RPATH`

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1622)

### DIFF
--- a/cpp/ycm/CMakeLists.txt
+++ b/cpp/ycm/CMakeLists.txt
@@ -361,6 +361,13 @@ target_link_libraries( ${PROJECT_NAME}
 
 if( LIBCLANG_TARGET )
   set( LIBCLANG_DIR "${CMAKE_SOURCE_DIR}/../third_party/clang/lib" )
+  # add rpath to libclang and remove rpath to temporary directory
+  if ( NOT MSVC )
+    set_target_properties( ${PROJECT_NAME} PROPERTIES
+      BUILD_WITH_INSTALL_RPATH ON
+      SKIP_BUILD_RPATH FALSE
+      INSTALL_RPATH ${LIBCLANG_DIR} )
+  endif ()
   # Remove all previous libclang libraries.
   file( GLOB LIBCLANG_FILEPATHS "${LIBCLANG_DIR}/libclang*" )
   foreach( FILEPATH ${LIBCLANG_FILEPATHS} )
--- a/ycmd/tests/__init__.py
+++ b/ycmd/tests/__init__.py
@@ -15,10 +15,6 @@
 # You should have received a copy of the GNU General Public License
 # along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
 
-from ycmd.utils import ImportAndCheckCore, CORE_COMPATIBLE_STATUS
-if ImportAndCheckCore() != CORE_COMPATIBLE_STATUS:
-  raise RuntimeError( "Could NOT import core!" )
-
 import functools
 import os
 from ycmd.tests.test_utils import ClearCompletionsCache, IsolatedApp, SetUpApp
--- a/ycmd/tests/clang/comment_strip_test.py
+++ b/ycmd/tests/clang/comment_strip_test.py
@@ -20,10 +20,6 @@ method/variable,etc. headers in order to remove non-data-ink from the raw
 comment"""
 
 
-from ycmd.utils import ImportAndCheckCore, CORE_COMPATIBLE_STATUS
-if ImportAndCheckCore() != CORE_COMPATIBLE_STATUS:
-  raise RuntimeError( "Could NOT import core!" )
-
 # flake8: noqa
 from hamcrest import assert_that, equal_to
 from unittest import TestCase


puremourning (commented):
Thanks for this. 

Could you please summarise what problem is being solved and how this solves it?

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1622#-:-MsEiDCZ3n_EZOUMcUzt:bmyqkrp)* status: 0 of 2 LGTMs obtained

Chilledheart:
Sure. Let me re-state the issue. It happens with `./build.py --clang-completer.py` which download libclang archive and dynamically link with the extracted libclang.dylib at third_party directory named `third_party/clang/lib`. There are two issues in previous code:
1. Firstly, no libclang.dylib is found due to ycm_core is dynamically linked with `@rpath\libclang.dylib` while rpath is a temporary path used in extracting libclang archive. Rpath means the dynamic loader will search the path first.

The grep of rpath for ycm_core is like this:
```
ycmd git:(dev)  otool -l ycm_core.cpython-39-darwin.so|grep -A3 LC_RPATH                 
          cmd LC_RPATH
      cmdsize 96
         path /private/var/folders/g6/rsdttj293_b4y9ftggw7s6dc0000gn/T/ycm_build_02s69rnz/lib (offset 12)
Load command 18
```
Thereforce the import of ycm_core module will get ImportError like
```
Python 3.9.9 (v3.9.9:ccb0e6a345, Nov 15 2021, 13:06:05) 
[Clang 13.0.0 (clang-1300.0.29.3)] on darwin
>>> import ycm_core
  File "<stdin>", line 1, in <module>
ImportError: dlopen(/Users/ko/.vim/plugged/YouCompleteMe/third_party/ycmd/ycm_core.cpython-39-darwin.so, 0x0002): Library not loaded: @rpath/libclang.dylib
  Referenced from: /Users/ko/.vim/plugged/YouCompleteMe/third_party/ycmd/ycm_core.cpython-39-darwin.so
  Reason: tried: '/private/var/folders/g6/rsdttj293_b4y9ftggw7s6dc0000gn/T/ycm_build_02s69rnz/lib/libclang.dylib' (no such file), '/private/var/folders/g6/rsdttj293_b4y9ftggw7s6dc0000gn/T/ycm_build_02s69rnz/lib/libclang.dylib' (no such file), '/usr/lib/libclang.dylib' (no such file)
``` 
2. Secondly, it is a potential security issue that dynamical loader will look at the temporary path first and load malicious code if a forged shared library exists there.

The fix is quite simple:
- remove the rpath of temporary path which exists only at build time
- add the rpath to `third_party/clang/lib`

puremourning:
Thanks for the explanation. We certainly used to have stuff in the CMake scripts to fix the rpath, I'm not sure what happened to it.

> Firstly, no libclang.dylib is found

I guess my confusion is - what exact problem does this cause, and when? Like is there a set of steps I can take on my Mac to repro a problem? We currently load libclang.dylib _manually_ from the correct path: 

https://github.com/ycm-core/ycmd/blob/master/ycmd/utils.py#L32-L34

https://github.com/ycm-core/ycmd/blob/master/ycmd/utils.py#L480-L487

Chilledheart:
Thanks for explanation. It should only appear in some situation.

The first fix doesn't work for me. In my situation (macOS 12.0.1, Xcode 13.2.1 and official universal2 Python 3.9.9), first fix fails because module os doesn't have the attribute `add_dll_directory`:
OS's python3:
```
➜  ycmd git:(dev) ✗ /usr/bin/python3
import os
Python 3.8.9 (default, Oct 26 2021, 07:25:53) 
[Clang 13.0.0 (clang-1300.0.29.30)] on darwin
>>> import os
>>> os.add_dll_directory
  File "<stdin>", line 1, in <module>
AttributeError: module 'os' has no attribute 'add_dll_directory'
```
python.org's python3:
```
➜  ycmd git:(dev) ✗ python3         
Python 3.9.9 (v3.9.9:ccb0e6a345, Nov 15 2021, 13:06:05) 
[Clang 13.0.0 (clang-1300.0.29.3)] on darwin
>>> import os
>>> os.add_dll_directory
  File "<stdin>", line 1, in <module>
AttributeError: module 'os' has no attribute 'add_dll_directory'
```

The second fix works for me (still not sure why the non-found error appeared in my environment, but I cannot reproduce it).
```
Python 3.9.9 (v3.9.9:ccb0e6a345, Nov 15 2021, 13:06:05) 
[Clang 13.0.0 (clang-1300.0.29.3)] on darwin
>>> libclang_path = 'third_party/clang/lib/libclang.dylib'
>>> import ctypes
>>> ctypes.cdll.LoadLibrary( libclang_path )
<CDLL 'third_party/clang/lib/libclang.dylib', handle 20124e060 at 0x100a78670>
>>> import ycm_core
>>> 
(successed)
```
However, it is still a security hole because it will search a rpath to temporary directory.

bstaletic:
> first fix fails because module os doesn't have the attribute add_dll_directory

`add_dll_directory` is a Windows specific hack. It's not supposed to exist anywhere else, but python says that the linked snippet is correct usage. Thanks, CPython...

> We certainly used to have stuff in the CMake scripts to fix the rpath, I'm not sure what happened to it.

I think that was back when LLVM github completely skipped a macOS archive, so we couldn't repackage that way. The workaround was to repackage the one from homebrew, which required fiddling with `otool` and RPATH.

That is, unless you're recalling some other RPATH fiddling?

puremourning:
> That is, unless you're recalling some other RPATH fiddling?

I'm probably thinking that we fix up the rpath of ycm_core.so actually. We probably still do that.

Chilledheart:
I agree with that we should fix RPATH of ycm_core.so.

Just tested with a Linux distribution, the RPATH (or RUNPATH called in Linux) is not targeted to `third_party/clang/lib`
```
ubuntu@10-7-152-224:~/.vim/plugged/YouCompleteMe/third_party/ycmd$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.3 LTS
Release:	20.04
Codename:	focal
ubuntu@10-7-152-224:~/.vim/plugged/YouCompleteMe/third_party/ycmd$ objdump -x ycm_core.cpython-38-x86_64-linux-gnu.so |grep PATH
  RUNPATH              $ORIGIN:/usr/lib/python3.8/config-3.8-x86_64-linux-gnu
```
WIth this patch, the RPATH is correct
```
ubuntu@10-7-152-224:~/.vim/plugged/YouCompleteMe/third_party/ycmd$ objdump -x ycm_core.cpython-38-x86_64-linux-gnu.so |grep PATH
  RUNPATH              /home/ubuntu/.vim/plugged/YouCompleteMe/third_party/ycmd/cpp/../third_party/clang/lib:/usr/lib/python3.8/config-3.8-x86_64-linux-gnu
```

puremourning:
for the record I checked and I'm pretty sure that the 3 vars used here in cmake land are all supported and sensible in our minimum version (3.14)

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1622)* status: 0 of 2 LGTMs obtained (waiting on @Chilledheart)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1622#-MsZMx9yEJAM4u6YVwND:-MsZMx9yEJAM4u6YVwNE:bsnbm9n) (no related file):*
As mentioned, I think we should remove the test lib manually loading this now.

---

puremourning (approved):
:lgtm:

Thanks!

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1622#-:-MsaDmB8Bctuw4RtLgsr:b9benhj)* status: 1 of 2 LGTMs obtained (waiting on @Chilledheart)

bstaletic (approved):
Thanks for the PR. I won't pretend I know enough about macOS to give an opinion here, but @puremourning says it works, therefore :lgtm:

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1622#-:-Msj_AfT3CsSu_h-FfmO:bfapob2)* status: :shipit: complete! 2 of 2 LGTMs obtained (waiting on @Chilledheart)

puremourning:
Sorry for the delay. I've re-run the CI which probably flaked.

puremourning:
@Mergifyio rebase

puremourning:
@mergifyio rebase

puremourning (approved):
> Reviewed 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1622)* status: 0 of 2 LGTMs obtained (and 2 stale) (waiting on @Chilledheart)

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1622#-:-Mzb1zT32gMpEFsoUbtT:bnfp4nl)* status: 1 of 2 LGTMs obtained (and 1 stale) (waiting on @Chilledheart)

PR #1624, opened 2022-01-05
Fix libclang's objc and objcpp support

Chilledheart:
For c++ source, it is correct to use "-x c++" language options but it
doesn't true for objective-c++ and objective-c source code. This patch
will fix it.

Add three unittests, subcommands unittests will be added next time.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1624)

### DIFF
--- a/ycmd/completers/cpp/flags.py
+++ b/ycmd/completers/cpp/flags.py
@@ -369,6 +369,11 @@ def _AddLanguageFlagWhenAppropriate( flags, enable_windows_style_flags ):
           for fl in reversed( flags ) ):
     return flags
 
+  # Libclang will add "-x objective-c" and "-x objective-c++" automatically
+  if any( fl.endswith( '.m' ) or fl.endswith( '.mm' )
+          for fl in reversed( flags ) ):
+    return flags
+
   # NOTE: This is intentionally NOT checking for enable_windows_style_flags.
   #
   # The first flag is now either an absolute path, a Windows style flag or a
--- a/ycmd/tests/clang/diagnostics_test.py
+++ b/ycmd/tests/clang/diagnostics_test.py
@@ -29,7 +29,8 @@ from pprint import pprint
 from unittest import TestCase
 from ycmd.tests.clang import setUpModule # noqa
 from ycmd.tests.clang import SharedYcmd, IsolatedYcmd, PathToTestFile
-from ycmd.tests.test_utils import BuildRequest, LocationMatcher, RangeMatcher
+from ycmd.tests.test_utils import ( BuildRequest, LocationMatcher,
+                                    RangeMatcher )
 from ycmd.utils import ReadFile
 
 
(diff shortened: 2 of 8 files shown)


puremourning:
Thanks for the PR. 

In general when sending a PR to fix an issue it helps to explain what the problem is that's being fixed, and a set of steps that reproduce the problem (in much the same way as when raising an issue on the tracker). This makes it a lot easier to understand the motivation for the change and why it's necessary.

so in this case am I understanding correctly that the compiler invocation is `clang++ <stuff> some_objective_c.m`, and that ycmd is seeing the `clang++` without a `-x` flag and assuming that it's `c++` ?

puremourning (changes requested):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1624)* status: 0 of 2 LGTMs obtained (waiting on @Chilledheart)

___
*[ycmd/completers/cpp/flags.py, line 373 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MseqJv47Oq2-G01r4DH:-MseqJv47Oq2-G01r4DI:b-de7dga) ([raw file](https://github.com/ycm-core/ycmd/blob/6da5c87618b1e635a2cf441768c7d0be2fca5a0c/ycmd/completers/cpp/flags.py#L373)):*
> ```Python
> 
>   if any( fl.endswith( '.mm' ) for fl in reversed( flags ) ):
>     return [ first_flag, '-x', 'objective-c++' ] + flags[ 1: ]
> ```

perhaps instead we should `return flags` like we do for coda.

the reason is that then we let libclang decide based on the filetype, rather than trying to guess.

We're guessing for `clang++ ...` because that's the common use case.
___
*[ycmd/tests/clang/diagnostics_test.py, line 399 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MseqWpk3YBXYIhU-wGe:-MseqWpk3YBXYIhU-wGf:b4u0boc) ([raw file](https://github.com/ycm-core/ycmd/blob/6da5c87618b1e635a2cf441768c7d0be2fca5a0c/ycmd/tests/clang/diagnostics_test.py#L399)):*
> ```Python
> 
> 
>   @MacOnly
> ```

why Mac only? I mean, there's nothing stopping the obj c tests working on other OS is there? I realise that it is unlikely objc is used on other platforms.

Chilledheart:
> Thanks for the PR.
> 
> In general when sending a PR to fix an issue it helps to explain what the problem is that's being fixed, and a set of steps that reproduce the problem (in much the same way as when raising an issue on the tracker). This makes it a lot easier to understand the motivation for the change and why it's necessary.
> 
> so in this case am I understanding correctly that the compiler invocation is `clang++ <stuff> some_objective_c.m`, and that ycmd is seeing the `clang++` without a `-x` flag and assuming that it's `c++` ?

I haven't checked the source code of clang driver but it should depends on the suffix of source code filename if it isn't passed with "-x" flag.

Chilledheart:
> > Reviewed 8 of 8 files at r1, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1624)_ status: 0 of 2 LGTMs obtained (waiting on @Chilledheart)
> 
> _[ycmd/completers/cpp/flags.py, line 373 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MseqJv47Oq2-G01r4DH:-MseqJv47Oq2-G01r4DI:b-de7dga) ([raw file](https://github.com/ycm-core/ycmd/blob/6da5c87618b1e635a2cf441768c7d0be2fca5a0c/ycmd/completers/cpp/flags.py#L373)):_
> 
> > ```python
> >   if any( fl.endswith( '.mm' ) for fl in reversed( flags ) ):
> >     return [ first_flag, '-x', 'objective-c++' ] + flags[ 1: ]
> > ```
> 
> perhaps instead we should `return flags` like we do for coda.
> 
> the reason is that then we let libclang decide based on the filetype, rather than trying to guess.
> 
> We're guessing for `clang++ ...` because that's the common use case.
It is better. I tested with the new code and it works.

> 
> _[ycmd/tests/clang/diagnostics_test.py, line 399 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MseqWpk3YBXYIhU-wGe:-MseqWpk3YBXYIhU-wGf:b4u0boc) ([raw file](https://github.com/ycm-core/ycmd/blob/6da5c87618b1e635a2cf441768c7d0be2fca5a0c/ycmd/tests/clang/diagnostics_test.py#L399)):_
> 
> > ```python
> >   @MacOnly
> > ```
> 
> why Mac only? I mean, there's nothing stopping the obj c tests working on other OS is there? I realise that it is unlikely objc is used on other platforms.

Not tested on a non-Mac Os for objective-c support and I am not sure if libclang works for objective-c there.

bstaletic (changes requested):
Thanks for the pull request. I did leave two small comments, but the change to `flags.py` looks fine.

> Reviewed 6 of 8 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1624#-:-MsjZCvAAGtRASMgyxGH:bg2xgzz)* status: 0 of 2 LGTMs obtained (waiting on @Chilledheart and @puremourning)

___
*[ycmd/tests/clang/diagnostics_test.py, line 399 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MseqWpk3YBXYIhU-wGe:-MsjYFoE8JVeCOQiC8mr:b-8i6ilu) ([raw file](https://github.com/ycm-core/ycmd/blob/6da5c87618b1e635a2cf441768c7d0be2fca5a0c/ycmd/tests/clang/diagnostics_test.py#L399)):*
<details><summary><i>Previously, Chilledheart wrote…</i></summary><blockquote>

Not tested on a non-Mac Os for objective-c support and I am not sure if libclang works for objective-c there.
</blockquote></details>

I agree with @puremourning. If the tests pass on linux/windows we shouldn't use `@MacOnly`.
___
*[ycmd/tests/clang/flags_test.py, line 66 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MsjYk4nAeS3dToNzSvs:-MsjYk4nAeS3dToNzSvt:bot67kh) ([raw file](https://github.com/ycm-core/ycmd/blob/669fb4ecb85449478d7ff2ebfc42b93a39b5d601/ycmd/tests/clang/flags_test.py#L66)):*
> ```Python
> 
>   for to_remove in to_removes:
>     assert_that( [ compiler ] + language_flag + [ '-c', filename ] + expected,
> ```

This seems odd. `flags.py` intentionally removes `-c filename` part, as leaving it in would cause errors during compiler invocation.

Chilledheart:
> Thanks for the pull request. I did leave two small comments, but the change to `flags.py` looks fine.
> 
> > Reviewed 6 of 8 files at r1, 2 of 2 files at r2, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1624#-:-MsjZCvAAGtRASMgyxGH:bg2xgzz)_ status: 0 of 2 LGTMs obtained (waiting on @Chilledheart and @puremourning)
> 
> _[ycmd/tests/clang/diagnostics_test.py, line 399 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MseqWpk3YBXYIhU-wGe:-MsjYFoE8JVeCOQiC8mr:b-8i6ilu) ([raw file](https://github.com/ycm-core/ycmd/blob/6da5c87618b1e635a2cf441768c7d0be2fca5a0c/ycmd/tests/clang/diagnostics_test.py#L399)):_
> 
> Previously, Chilledheart wrote…
> I agree with @puremourning. If the tests pass on linux/windows we shouldn't use `@MacOnly`.

It seems to work on platforms other than mac. Let me remove it. https://github.com/Chilledheart/ycmd/runs/4727423777?check_suite_focus=true

> 
> _[ycmd/tests/clang/flags_test.py, line 66 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MsjYk4nAeS3dToNzSvs:-MsjYk4nAeS3dToNzSvt:bot67kh) ([raw file](https://github.com/ycm-core/ycmd/blob/669fb4ecb85449478d7ff2ebfc42b93a39b5d601/ycmd/tests/clang/flags_test.py#L66)):_
> 
> > ```python
> >   for to_remove in to_removes:
> >     assert_that( [ compiler ] + language_flag + [ '-c', filename ] + expected,
> > ```
> 
> This seems odd. `flags.py` intentionally removes `-c filename` part, as leaving it in would cause errors during compiler invocation.

No, it is still there in flags otherwise you cannot tell if it is a cuda/objc/objcpp file in `flags.py`

bstaletic (approved):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1624)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[ycmd/tests/clang/diagnostics_test.py, line 399 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MseqWpk3YBXYIhU-wGe:-MsjhNkN9Zps-zOkN8GS:bc2056x) ([raw file](https://github.com/ycm-core/ycmd/blob/6da5c87618b1e635a2cf441768c7d0be2fca5a0c/ycmd/tests/clang/diagnostics_test.py#L399)):*
<details><summary><i>Previously, Chilledheart wrote…</i></summary><blockquote>

It seems to work on platforms other than mac. Let me remove it. https://github.com/Chilledheart/ycmd/runs/4727423777?check_suite_focus=true
</blockquote></details>

I still see `@MacOnly` decorators in this file. I'm guessing that was an oversight.
___
*[ycmd/tests/clang/flags_test.py, line 66 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MsjYk4nAeS3dToNzSvs:-MsjgPuZ-Sn1zrD4l5EO:b-g3ydqn) ([raw file](https://github.com/ycm-core/ycmd/blob/669fb4ecb85449478d7ff2ebfc42b93a39b5d601/ycmd/tests/clang/flags_test.py#L66)):*
<details><summary><i>Previously, Chilledheart wrote…</i></summary><blockquote>

No, it is still there in flags otherwise you cannot tell if it is a cuda/objc/objcpp file in `flags.py`
</blockquote></details>

https://github.com/ycm-core/ycmd/blob/master/ycmd/completers/cpp/flags.py#L43

https://github.com/ycm-core/ycmd/blob/master/ycmd/completers/cpp/flags.py#L420

But that happens well after `_AddLanguageFlagWhenAppropriate()`.

Chilledheart:
> > Reviewed 1 of 1 files at r3, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1624)_ status: 0 of 2 LGTMs obtained (waiting on @puremourning)
> 
> _[ycmd/tests/clang/diagnostics_test.py, line 399 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MseqWpk3YBXYIhU-wGe:-MsjhNkN9Zps-zOkN8GS:bc2056x) ([raw file](https://github.com/ycm-core/ycmd/blob/6da5c87618b1e635a2cf441768c7d0be2fca5a0c/ycmd/tests/clang/diagnostics_test.py#L399)):_
> 
> Previously, Chilledheart wrote…
> I still see `@MacOnly` decorators in this file. I'm guessing that was an oversight.

Done.
> 
> _[ycmd/tests/clang/flags_test.py, line 66 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MsjYk4nAeS3dToNzSvs:-MsjgPuZ-Sn1zrD4l5EO:b-g3ydqn) ([raw file](https://github.com/ycm-core/ycmd/blob/669fb4ecb85449478d7ff2ebfc42b93a39b5d601/ycmd/tests/clang/flags_test.py#L66)):_
> 
> Previously, Chilledheart wrote…
> https://github.com/ycm-core/ycmd/blob/master/ycmd/completers/cpp/flags.py#L43
> 
> https://github.com/ycm-core/ycmd/blob/master/ycmd/completers/cpp/flags.py#L420
> 
> But that happens well after `_AddLanguageFlagWhenAppropriate()`.

I see. I thought "-c" flag was a require_argument option for gcc/clang until I looked at the source code of clang:
https://github.com/llvm-mirror/clang/blob/master/include/clang/Driver/Options.td#L549

Let me remove the skippable "-c" flag.

Chilledheart:
Then I noticed the fact that `RemoveUnusedFlags` whose caller is `PrepareFlagsForClang` which is only called when it comes to a compiler database it is not true for the unittest where only `_RemoveFlagsPrecedingCompiler` is called.

https://github.com/ycm-core/ycmd/blob/master/ycmd/completers/cpp/flags.py#L420
https://github.com/ycm-core/ycmd/blob/master/ycmd/completers/cpp/flags.py#L281
https://github.com/ycm-core/ycmd/blob/master/ycmd/completers/cpp/flags.py#L331


Another issue is that if I remove "-c" in the unittests they fail because the method `_RemoveFlagsPrecedingCompiler` will assume the first flag, i.e. filename should be skipped.

bstaletic (approved):
Yeah, I wasn't quite clear. `-c` has to be followed by a positional argument.

For the purposes of this PR, it's fine to leave `-c filename` there to see ycmd reacting to `-x objective-c`. There should just be a comment, to avoid confusions in the future.

> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1624#-:-Mt9Jiw97X1MQl4X4coY:b-6rvsas)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

puremourning (changes requested):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1624)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @Chilledheart)

___
*[ycmd/completers/cpp/flags.py, line 368 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1624#-Mtitr9_3b1it9EAgF0n:-Mtitr9_3b1it9EAgF0o:b-esu6zc) ([raw file](https://github.com/ycm-core/ycmd/blob/5e9433d5af25d4f14174ada9ee9206eeca2f8974/ycmd/completers/cpp/flags.py#L368)):*
> ```Python
> 
>   # Libclang has already added `-xcuda` for us.
>   if any( fl.lower().endswith( '.cu' ) or fl.lower().endswith( '.cuh' )
> ```

checking the lower-case version I think is good. Could we add a test that covers it?
___
*[ycmd/tests/clang/diagnostics_test.py, line 399 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MseqWpk3YBXYIhU-wGe:-MtithoVBo0mLOX4Ybfr:b-fx15it) ([raw file](https://github.com/ycm-core/ycmd/blob/6da5c87618b1e635a2cf441768c7d0be2fca5a0c/ycmd/tests/clang/diagnostics_test.py#L399)):*
<details><summary><i>Previously, Chilledheart wrote…</i></summary><blockquote>

Done.
</blockquote></details>

I still see @MacOnly here?

Chilledheart:
> > Reviewed 1 of 2 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1624)_ status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @Chilledheart)
> 
> _[ycmd/completers/cpp/flags.py, line 368 at r4](https://reviewable.io/reviews/ycm-core/ycmd/1624#-Mtitr9_3b1it9EAgF0n:-Mtitr9_3b1it9EAgF0o:b-esu6zc) ([raw file](https://github.com/ycm-core/ycmd/blob/5e9433d5af25d4f14174ada9ee9206eeca2f8974/ycmd/completers/cpp/flags.py#L368)):_
> 
> > ```python
> >   # Libclang has already added `-xcuda` for us.
> >   if any( fl.lower().endswith( '.cu' ) or fl.lower().endswith( '.cuh' )
> > ```
> 
> checking the lower-case version I think is good. Could we add a test that covers it?

I tried with the lower-case version but it seems the flags parsing route in compilation database couldn't truely handle with them correctly (tests added but all failed). I think it is better to handle it in another patch.

> 
> _[ycmd/tests/clang/diagnostics_test.py, line 399 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1624#-MseqWpk3YBXYIhU-wGe:-MtithoVBo0mLOX4Ybfr:b-fx15it) ([raw file](https://github.com/ycm-core/ycmd/blob/6da5c87618b1e635a2cf441768c7d0be2fca5a0c/ycmd/tests/clang/diagnostics_test.py#L399)):_
> 
> Previously, Chilledheart wrote…
> I still see @MacOnly here?

Done.

puremourning (approved):
> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1624)* status: 0 of 2 LGTMs obtained (waiting on @Chilledheart)

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1624#-:-N9M_n9eEeuvlxM-EYS3:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @Chilledheart)

puremourning:
@Mergifyio rebase

puremourning:
tests are failing

puremourning:
Seems abandoned.

PR #1640, opened 2022-06-07
Add more Refactor* subcommands for python

mnikulin:
I didn't understand why adding the Refactor Inline would be inconsistent and it seemed to me that using the [extract_variable](https://jedi.readthedocs.io/en/latest/docs/api.html#jedi.Script.extract_variable) and extract_function without additional arguments is easy to implement exactly like Refactor Rename, and the arguments **until_line** and **until_column** are used in visual block mode.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1640)

### DIFF
--- a/ycmd/completers/python/python_completer.py
+++ b/ycmd/completers/python/python_completer.py
@@ -293,6 +293,14 @@ class PythonCompleter( Completer ):
                            self._GetDoc( request_data ) ),
       'RefactorRename' : ( lambda self, request_data, args:
                            self._RefactorRename( request_data, args ) ),
+      'RefactorInline' : ( lambda self, request_data, args:
+                           self._RefactorInline( request_data, args ) ),
+      'RefactorExtractVariable' : ( lambda self, request_data, args:
+                                    self._RefactorExtractVariable( request_data,
+                                                                   args ) ),
+      'RefactorExtractFunction' : ( lambda self, request_data, args:
+                                    self._RefactorExtractFunction( request_data,
+                                                                   args ) ),
     }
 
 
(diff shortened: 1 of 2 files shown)


puremourning (commented):
so sorry I missed this, thanks _so much_ for this PR.

it's kind of weird that CI hasn't run though. I'll see if I can kick it...

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1640#-:-N9MYIlW7TxkH790y2dR:bpwysbc)* status: 0 of 2 LGTMs obtained

puremourning:
@MergifyIO rebase

puremourning (changes requested):
Gave this a go.

* I managed to get extract function to work, ish.
* I didn't manage to get extract variable to work. I always get `RefactoringError: Cannot extract a "simple_stmt"`; I guess I;m just holding it wrong, but still I would need to work it out to update the YCM docs.
* refractor inline worked, ish

so yeah, I'm really happy with it, thanks A TON.

Again sorry I've been lax on reviewing ycmd PRs recently, I completely forgot about them!

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1640#-:-N9N0FO_DbWXeFIII2j-:b-6thlv5)* status: 0 of 2 LGTMs obtained (waiting on @mnikulin)

___
*[`ycmd/completers/python/python_completer.py` line 504 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1640#-N9N03Qc7iJsBB_Z26A2:-N9N03Qd-CkYtEz6AZgX:b-b7ntw0) ([raw file](https://github.com/ycm-core/ycmd/blob/9e6b32ba8a90563a53e998d5258b4214b48aec39/ycmd/completers/python/python_completer.py#L504)):*
> ```Python
>         new_name = new_name,
>         until_line = until_line,
>         until_column = until_column )
> ```

So I think Jedi takes offsets as unicode code points, but these offsets in the ycmd api are byte offsets.

I think we need to do CodepointOffsetToByteOffset on this value (and other similar range values in this PR)

puremourning (approved):

PR #1643, opened 2022-06-10
Support for semantic highlighting (experimental)

puremourning:
We decode the strange LSP protocol in ycmd and send a simple one to the client. This might use more data, but means the client code is simpler too, where it probably matters more.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1643)

### DIFF
--- a/ycmd/completers/completer.py
+++ b/ycmd/completers/completer.py
@@ -369,6 +369,10 @@ class Completer( metaclass = abc.ABCMeta ):
     return {}
 
 
+  def ComputeSemanticTokens( self, request_data ):
+    return {}
+
+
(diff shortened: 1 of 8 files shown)


puremourning (commented):
:lgtm:

> Reviewed 5 of 5 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1643#-:-N4IF9BT93fdmzAvZXwK:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

PR #1649, opened 2022-07-14
reduce the requirements for cmake version

wangling12:
I have tested with cmake3.12, and everything seems to be ok.

---
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1649)

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -15,7 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
 
-cmake_minimum_required( VERSION 3.14 )
+cmake_minimum_required( VERSION 3.12 )
 
 project( YouCompleteMe )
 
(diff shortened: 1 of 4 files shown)


puremourning (commented):
what platform are you seeking to support. The reason we chose 3.14 is that is a version that I frequently actually test with on the basis of it being the version in RHEL7 (see https://github.com/ycm-core/ycmd/pull/1513). I'm not comfortable with supporting a version that isn't regularly actually tested, and am not willing to commit to providing support on "arbitrary" legacy OS versions.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1649#-:-N6wVC6P0GQs7RfIwBd_:b-jlnufz)* status: 0 of 2 LGTMs obtained

wangling12:
>  I'm not comfortable with supporting a version that isn't regularly actually tested

I agree with you in this point.

On the other hand, cmake doesn't seem to use features that are strongly version related. By modifying a small number here, we can "free" support more distributions.

puremourning:
This is why I was asking what it was for specifically. I need to understand the why to decide if it is ok (and go and read all the policy changes between the versions to see if anything important changed, etc etc)

PR #1651, opened 2022-08-13
Inlay hints

puremourning:
This is the bare minimum, just forawrding the LSP response, with a few
    fixups for ycmd/LSP differences.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1651)

* we use a cut-down version of the LSP spec (which is over specified for our uses)
* we update the docs
* we update all the servers to version that support it
* manual implementation for typescript, but it's essentially the LSP spec under the hood.
* add a setting to disable tern, as this is helpful for testing and where you have a mix of projects.
* had to hack go completer due to upstream bug in atlas usage

### DIFF
--- a/build.py
+++ b/build.py
@@ -95,7 +95,7 @@ JDTLS_SHA256 = (
   '09650af5c9dc39f0b40981bcdaa2170cbbc5bb003ac90cdb07fbb57381ac47b2'
 )
 
-RUST_TOOLCHAIN = 'nightly-2021-10-26'
+RUST_TOOLCHAIN = 'nightly-2022-08-17'
 RUST_ANALYZER_DIR = p.join( DIR_OF_THIRD_PARTY, 'rust-analyzer' )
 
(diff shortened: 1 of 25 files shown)


puremourning (commented):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1651)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

puremourning (commented):
:lgtm:

> Reviewed 20 of 20 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1651#-:-N9vJ7Ue2X_9iAEdZQ2u:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

PR #1658, opened 2022-09-07
Update to clang 15

puremourning:
Drop support for rebuilds for freebsd as they aren't available upstream and this OS is niche.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1658)


puremourning:
blocked by https://github.com/clangd/clangd/issues/1306

Issue #1583, opened 2021-08-15
Language-model Based Completer


xloem:
Hey, using new models such as EleutherAI's and NovelAI's [Genji](https://huggingface.co/NovelAI/genji-python-6B) or apis such as AI21 or OpenAI, a completer could actually sometimes complete an entire function body.

I might be able to help implement this a little, although I have a serious cognitive disability so it's a weak gamble.

@bilucodota will tabnine be opening a pull request for the ycm changes?  Is my link in the expandable details below what you mean about registering a general completer?

<details>
  <summary>Notes on implementing a completer</summary>
  
  - ycm will find general completers if they are added to the [general completer store](https://github.com/ycm-core/ycmd/blob/master/ycmd/completers/general/general_completer_store.py#L24)
  - ycm has [official documentation for writing new completers](http://ycm-core.github.io/YouCompleteMe/#writing-new-semantic-completers), haven't read it myself
  - if needed, [ycmd's protocol documentation](https://ycm-core.github.io/ycmd/) can clarify the existing ycmd interface
</details>

<details>
  <summary>Notes on using code prediction models</summary>

  - genji is based on gpt-j which is presently being [merged into the mainstream transformers library](https://github.com/huggingface/transformers/pull/13022).  There may be a new form of genji after the merge.  Additionally there are links in that thread towards approaches for lower end systems.  Genji itself has a split form that can work on lower end systems.
  - I have a [simple hack towards a unified interface](https://github.com/xloem/codesynth) for some of the code prediction models out there.  I'm new to machine learning and don't know where to find existing similar work.
  - the community that made genji recommend https://fast.ai/ for learning machine learning.  have not gone through the course myself.
</details>

bstaletic:
We've discussed this idea and aren't completely against the idea. Though I personally do have some concerns.

1. How is the AI being fed the data to learn from? Just the repo the user is editing? The whole github/gitlab/whatever?
2. If former, then is that enough data to actually be useful? That's what some tools try and I've read that people have conflicting experiences with those.
3. If latter, then we have the Microsoft Copilot problem - scanning GPL3 and occasionally regurgitating it even in proprietary codebases.
3.1. Saying that the user just needs to be vigilant, isn't just infeasible, but also conflicts my own ethics.
4. I also have concerns regarding fitting this kind of thing well within my own mental model of how things should work, but that's getting ahead of ourselves.
[...]

PR #1661, opened 2022-09-24
Language Model Completion Example

xloem:
I'm not clear on why #1583 was closed, but I wanted to add code around showing how easy it can be to shoehorn pretrained AI completion into a community backend like this.

I'm still learning ycmd's internals, and I haven't gotten ycmd testing locally yet, so this code is untested, but I copied some boilerplate into a call ( https://github.com/ycm-core/ycmd/pull/1661/files#diff-5cf178b6abf0d5798084fb4f41b39b7871c14f65a7deb45c299ac4b470b41023R36 ) to an old free language model api server. If I (or somebody) can get the ycm boilerplate right, it should be workable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1661)

### DIFF
--- a/ycmd/completers/all/example_language_model_completer.py
+++ b/ycmd/completers/all/example_language_model_completer.py
@@ -1,0 +1,247 @@
+from ycmd.completers.general_completer import GeneralCompleter
+# from ycmd.utils import ImportCore, LOGGER, SplitLines
+from ycmd import responses
+
+import requests
+import time
+
+
+class ExampleLanguageModelCompleter( GeneralCompleter ):
+  # reminder: offsets are _byte_ offsets; one byte may be a partial character
+  def __init__( self, user_options ):
+    super().__init__( user_options )
+    # self._completer = ycm_core.ExampleLanguageModelCompleter()
+    # self._tags_file_last_mtime = defaultdict( int )
+    # self._max_candidates = user_options[ 'max_num_identifier_candidates' ]
+    self.requests = requests.Session()
+    self._ratelimit_duration = 1 # 12
+    self._ratelimit_mark = time.time() - self._ratelimit_duration
+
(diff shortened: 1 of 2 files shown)


puremourning (commented):
most of this code is commented out. I don't understand what I'm reviewing or testing. The description doesn't say what this does or what it's for, so it's hard to review. A change like this needs detailed design elaboration and absolutely bulletproof test coverage. The use of a third-party URL is not something that I particularly want to include in YCM as this has all sorts of IP and license and privacy issues.

further, completion of prose is not in scope for YCM and never has been.

So, while I appreciate the effort of making a pull request, I don't see how we can pursue this.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1661#-:-NCjpJPP-tXXGnbNx6rT:brav9rh)* status: 0 of 2 LGTMs obtained

xloem:
Hey, thank you so much for looking.

> most of this code is commented out. I don't understand what I'm reviewing or testing. The description doesn't say what this does or what it's for, so it's hard to review.

I'm sorry for the state of the example. The commented code can be ignored; it is left over from copying from another file. I have a lot of trouble coding nowadays.

The code is very simple, and simply passes its argument to the third party URL.

The intent was to show how simple it is to add this functionality.

> A change like this needs detailed design elaboration and absolutely bulletproof test coverage. The use of a third-party URL is not something that I 

Yes, it is just a draft example.

> particularly want to include in YCM as this has all sorts of IP and license and privacy issues.

Thank you for looking at it. There are alternatives to the third-party URL, but it is of course simplest to implement.

> further, completion of prose is not in scope for YCM and never has been.

The models that complete prose also complete source code. I think of the prose as a side effect.

> So, while I appreciate the effort of making a pull request, I don't see how we can pursue this.

Thank you for looking.

PR #1663, opened 2022-10-12
Try starting Omnisharp server without solution file

sankhesh:
With this change, ycmd will attempt to start the Omnisharp server with the current working directory when no solution file is found. The runtime check for solution file presence is not required and exception raising is defered to Omnisharp.


puremourning:
@sankhesh  can you given an example of a specific problem that this solves?

@mispencer  WDYT?

sankhesh:
@puremourning 
Currently, ycmd checks for a solution file in the module configuration and current working directory and starts the Omnisharp server *iff* a solution file exists. 

However, it is common for C# projects (Unity projects, for example) to not have a VS solution file to go with. Secondly, Onmisharp doesn't really require a solution file anymore. It just needs a working directory which is what the change provides.

Hth

puremourning:
Thanks,

Can we please add a test directory and some tests that show it works? Hopefully just a simple project without a solution file and basic tests. We have test primitives for things like StartCompleterServerInDirectory or something.

Otherwise, I'm likely to just break this as I don't c# often

sankhesh:
Yeah - the test failure in CI is related to this feature. I'll fix the tests so that it provides sufficient testing for this feature as well.

Btw, can you label this PR as "hacktoberfest-accepted". Might as well :)

puremourning:
> Btw, can you label this PR as "hacktoberfest-accepted". Might as well :)

I don't know what that means but if you need me to do something for you to earn internet points, then let me know and I guess I can do it if it isn't too much effort.

puremourning:
@Mergifyio rebase

sankhesh:
@Mergifyio rebase

Issue #1673, opened 2023-01-10
Invalid protocol data crashes ycmd


sc68cal:
I have configured [ansible-language-server](https://github.com/ansible/ansible-language-server) as a language server, as follows in my `vimrc`

```vim
let g:ycm_language_server = 
  \ [ 
  \   {
  \     'name': 'ansible',
  \     'cmdline': [ 'ansible-language-server', '--stdio' ],
  \     'filetypes': [ 'yaml' ],
  \     'project_root_files': ['ansible.cfg']
  \ ]
```

The problem is that `YouCompleteMe/ycmd` crashes due the fact that when `ansible-language-server` starts, it has a couple `console.debug` statements around [which linting system it will use](https://github.com/ansible/ansible-language-server/blob/30fd958beb4cd8f625fd5f9302390ede55011155/src/providers/validationProvider.ts#L64). `ansible-language-server` is built using Node.js and unfortunately only honors the `--stdio` flag for starting the language server. In Node.js, `console.debug` is just a redirect to [console.log](https://nodejs.org/api/console.html#consoledebugdata-args).

As a result, `ansible-language-server` upon startup emits a line to `stdio` that is not a valid protocol data, which crashes ycmd.

```
2023-01-10 11:21:06,067 - ERROR - Received invalid protocol data from server: b'\nValidating u
sing ansible-lint'
Traceback (most recent call last):
  File "/Users/scolli010/.vim/bundle/YouCompleteMe/third_party/ycmd/ycmd/completers/language_s
erver/language_server_completer.py", line 564, in _ReadHeaders
    key, value = utils.ToUnicode( line ).split( ':', 1 )
ValueError: not enough values to unpack (expected 2, got 1)
2023-01-10 11:21:06,067 - ERROR - The language server communication channel closed unexpectedly. Issue a RestartServer command to recover.
Traceback (most recent call last):
  File "/Users/scolli010/.vim/bundle/YouCompleteMe/third_party/ycmd/ycmd/completers/language_server/language_server_completer.py", line 367, in run
    self._ReadMessages()
  File "/Users/scolli010/.vim/bundle/YouCompleteMe/third_party/ycmd/ycmd/completers/language_server/language_server_completer.py", line 486, in _ReadMessages
    data, read_bytes, headers = self._ReadHeaders( data )
[...]

PR #1674, opened 2023-01-10
Remove raise from _ReadHeaders

sc68cal:
The only caller to this method does not handle exceptions from this method, so instead of crashing, just log the message and continue processing

This is related to #1673

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1674)

### DIFF
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -566,7 +566,6 @@ class LanguageServerConnection( threading.Thread ):
             except Exception:
               LOGGER.exception( 'Received invalid protocol data from server: '
                                  + str( line ) )
-              raise
 
         read_bytes += 1
 


puremourning:
Thanks for sending a PR, but I'm not convinced this is the right thing to do.

The parser will be in an indeterminate state at this point and "just carry on hoping for the best" doesn't seem like the correct result.

For the record:

> The only caller to this method does not handle exceptions from this method

that's not strictly true. Exceptions bubble up to the top of the thread in `LanguageServerConnection.run` where the connection is cleaned up and terminated when we receive invalid protocol data.

puremourning:
I think fixing ansible-language-server achieves those goals too.

I admit that in the past I have modified this code to print out the invalid protocol data so that it's at least clear what was received, but most language servers don't just output junk to the console and expect clients to handle it.

By removing this exception, we're entering into untestable/unspecifiable behaviour territory, and masking other potential problems. If you feel like improving the diagnostics when this happens, I'm all for that, but I'm not keen on just "hoping for the best" by eating the exception, `on error resume next` style.

PR #1676, opened 2023-01-12
TS: add @param and @returns to docs

notEvil:
see https://github.com/ycm-core/ycmd/issues/1675

This adds a table between the signature and the description, listing all `@param` tags and any `@return`/`@returns` tag.
If a return tag exists, it is the last row and is named `return`.

GetDoc:
Screenshot from 2023-01-12 16-51-09
Autocomplete:
Screenshot from 2023-01-12 16-52-29

Missing:
- tests

I would appreciate if anyone could write tests for this because I have 0 knowledge.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1676)

### DIFF
--- a/ycmd/completers/typescript/typescript_completer.py
+++ b/ycmd/completers/typescript/typescript_completer.py
@@ -913,7 +913,7 @@ class TypeScriptCompleter( Completer ):
 
     interface = _GetInterfaceDocumentation(info.get("tags", []))
     interface = '' if interface is None else (interface + '\n\n')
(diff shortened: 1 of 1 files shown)


puremourning:
either way the existing tests are failing (flake8 errors)

puremourning:
> I don't care enough to invest hours

I guess I don't care enough to review and merge this change then.

Thanks for the contribution though. Perhaps someone who does care about this change will pick it up in the future.

notEvil:
Sry, didn't see flake8-ycm coming... I won't try again

Thanks, I hope so too

PR #1679, opened 2023-01-23
Specify CMAKE_BUILD_TYPE manually for ninja generator

### LINKED ISSUE #1678: build.py builds ycm_core in Debug configuration
Today I found out an issue that ycm_core built by build.py is in Debug configuration.

I copied the ycm_core.pyd from one computer to another found out ycmd failed to start by ImportError that some dlls are not installed. By inspecting the pyd binary with LLVM (aka llvm-readobj --coff-imports <path-to-pyd>), I found out it is linked to debug vc runtime that is vcruntime140_1d.dll and ucrtbased.dll which requires a full visual c++ installation. It must be a mistake because not only Debug configuration requires **non-redistributable** debug runtime but also **performance penalty** introduced by built-in assertions and zero compiler optimization.

I was looking into build.py's options with by-pass something like `-DCMAKE_BUILD_TYPE=Release` to cmake but only `--enable-debug` option. Is it that by-design? In my memory, cmake's BUILD_TYPE is `Debug` if not specified, which means `--enable-debug` is default.

Chilledheart:
CMAKE_BUILD_TYPE should be Release but it seems a problem under windows that with ninja generator, where cmake has pre-defined CMAKE_BUILD_TYPE and it passes by the setting line in cpp\CMakeLists.txt. Specify it manually, so it compiles and links correctly with release runtime.

Verified "/O2" instead of "/Od" in cl.exe commandline. Tested on a Windows 10 22H2 machine and VS2019 installation.

TODO I didn't have time to test it with "MSYS Makefiles" or "MINGW Makefiles" generator on windows. We can add this fix if we need it as well.

Fixes #1678.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1679)

### DIFF
--- a/build.py
+++ b/build.py
@@ -613,6 +613,8 @@ def GetCmakeArgs( parsed_args ):
   if parsed_args.enable_debug:
     cmake_args.append( '-DCMAKE_BUILD_TYPE=Debug' )
     cmake_args.append( '-DUSE_DEV_FLAGS=ON' )
+  else:
+    cmake_args.append( '-DCMAKE_BUILD_TYPE=Release' )
 
   # coverage is not supported for c++ on MSVC
   if not OnWindows() and parsed_args.enable_coverage:


puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1679)* status: 0 of 2 LGTMs obtained (waiting on @Chilledheart)

Chilledheart:
> esn't seem to be a good reason to

Okay. I have updated the commit.

PR #1686, opened 2023-03-27
Fix gopls test with Go1.20

zhsj:
Go1.20 changes diagnostics message.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1686)


puremourning:
Thanks for this! Going to try it now

PR #1704, opened 2023-08-21
Implement simple type hierarchy support

bstaletic:
As agreed, this pull request is here for feedback purposes and field testing. Not necessarily to be merged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1704)

PR #1707, opened 2023-09-11
Handle empty object capabilities as truthy where applicable

bstaletic:
Another one of those "be vscode plox" things.
Servers seem to expect things like
```json
{
	"capabilities": {
		"whateverProvider": {}
	}
}
```
to mean that the server is really a provider of whatever. Earliest offender is the swift server being a code action provider.

Please take extra care wen reviewing this thing.

Unresolved things:

1. Are we correctly handling `textDocumentSync`?
2. Type hierarchy needs to get the same treatment once #1704 is accepted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1707)

### DIFF
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -1496,7 +1496,8 @@ class LanguageServerCompleter( Completer ):
     if not self.ServerIsReady():
       return responses.SignatureHelpAvailalability.PENDING
 
-    if bool( self._server_capabilities.get( 'signatureHelpProvider' ) ):
+    if _IsCapabilityProvided( self._server_capabilities,
+                              'signatureHelpProvider' ):
       return responses.SignatureHelpAvailalability.AVAILABLE
     else:
       return responses.SignatureHelpAvailalability.NOT_AVAILABLE
@@ -1506,7 +1507,8 @@ class LanguageServerCompleter( Completer ):
     if not self.ServerIsReady():
       return {}
 
-    if not self._server_capabilities.get( 'signatureHelpProvider' ):
+    if not _IsCapabilityProvided( self._server_capabilities,
+                                  'signatureHelpProvider' ):
       return {}
 
     self._UpdateServerWithCurrentFileContents( request_data )
@@ -1551,8 +1553,9 @@ class LanguageServerCompleter( Completer ):
     if not self._semantic_token_atlas:
       return {}
 
-    range_supported = self._server_capabilities[ 'semanticTokensProvider' ].get(
-      'range', False )
+    range_supported = _IsCapabilityProvided(
+        self._server_capabilities[ 'semanticTokensProvider' ],
+        'range' )
 
     self._UpdateServerWithCurrentFileContents( request_data )
 
@@ -1588,7 +1591,8 @@ class LanguageServerCompleter( Completer ):
     if not self._ServerIsInitialized():
       return []
 
-    if 'inlayHintProvider' not in self._server_capabilities:
+    if not _IsCapabilityProvided( self._server_capabilities,
+                                  'inlayHintProvider' ):
       return []
 
     self._UpdateServerWithCurrentFileContents( request_data )
@@ -1763,7 +1767,8 @@ class LanguageServerCompleter( Completer ):
       )
 
     if ( self._server_capabilities and
-         'callHierarchyProvider' in self._server_capabilities ):
+         _IsCapabilityProvided( self._server_capabilities,
+                                'callHierarchyProvider' ) ):
       commands[ 'GoToCallees' ] = (
         lambda self, request_data, args:
             self.CallHierarchy( request_data, [ 'outgoing' ] )
@@ -1787,9 +1792,10 @@ class LanguageServerCompleter( Completer ):
 
     for providers in provider_list:
       if isinstance( providers, tuple ):
-        if all( capabilities.get( provider ) for provider in providers ):
+        if all( _IsCapabilityProvided( capabilities, provider )
+                for provider in providers ):
           return providers
-      if capabilities.get( providers ):
+      if _IsCapabilityProvided( capabilities, providers ):
         return providers
     return None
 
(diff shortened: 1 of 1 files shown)


puremourning (approved):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1707)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[`ycmd/completers/language_server/language_server_completer.py` line 3546 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1707#-Ne9I54L2RWc5L7KZpd5:-Ne9I54L2RWc5L7KZpd6:bwngmtp) ([raw file](https://github.com/ycm-core/ycmd/blob/12c6c42da1b76ba446de71b5518d2fe4f809a077/ycmd/completers/language_server/language_server_completer.py#L3546)):*
> ```Python
> def _IsCapabilityProvided( capabilities, query ):
>   capability = capabilities.get( query )
>   return bool( capability or capability == {} )
> ```

this looks a little odd, the RHS of `or` is already a bool, so I think `bool( capability ) or capability == {}` maybe.

But perhaps what we're really saying is that we should return True if:

* the query is in the dict
* the value of query in the dict is not False or None (json null)

Interestingly, not _all_ providers are allowed to be `False`, for example:

```
	 * The server provides code lens.
	 */
	codeLensProvider?: CodeLensOptions;
```


This may be:
- not supplied (return False)
- null/None (return False)
- a CodeLensOptions object (return True)

Wheres:

```
	codeActionProvider?: boolean | CodeActionOptions;
```

This may be:
- not supplied (return False)
- null/None (return False)
- True  (return True)
- False (return False)
- a CodeActionOptions object (return True)

sadly, we _may_ need a third argument to this function `allow_boolean` which says if True/False is valid for the entry in the spec.

something like

```python
def _IsCapabilityProvided( c, q, allow_boolean=True ):
   if not q in c: return False
   v = c[q]
   if type( v ) == bool:
     if not allow_boolean: return False
     return v
   if type( v ) == dict:
     return True
   return False
```

maybe maybe?

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1707)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[`ycmd/completers/language_server/language_server_completer.py` line 3546 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1707#-Ne9I54L2RWc5L7KZpd5:-Ne9M1FlCkTHFYTD0AsM:b-kv0644) ([raw file](https://github.com/ycm-core/ycmd/blob/12c6c42da1b76ba446de71b5518d2fe4f809a077/ycmd/completers/language_server/language_server_completer.py#L3546)):*
> `bool( capability ) or capability == {}`

That *is* what I intended to write!

I definitely was not considering getting a `bool` where it is not expected. Or rather I was and thought it would be fine to accept `True` even against the protocol.
Yes, that's questionable.

That last suggestion of yours seems expensive. For semantic tokens and signature help we query capabilities on every request. Maybe we need to start being careful about those?

The following should cover all cases "by the book" without unnecessary type comparisons and dict lookups:
```python
def _IsCapabilityProvided( c, q, allow_boolean=True ):
    v = c.get( q )
    if v is None: return False
    t = type( v )
    if t == bool:
        return v if allow_boolean else False
    return t == dict
```

puremourning (approved):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1707)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[`ycmd/completers/language_server/language_server_completer.py` line 3546 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1707#-Ne9I54L2RWc5L7KZpd5:-Ne9SjUa2IvSnoPvN2qn:b-568k3v) ([raw file](https://github.com/ycm-core/ycmd/blob/12c6c42da1b76ba446de71b5518d2fe4f809a077/ycmd/completers/language_server/language_server_completer.py#L3546)):*

> `bool( capability ) or capability == {}`

That *is* what I intended to write!

I definitely was not considering getting a `bool` where it is not expected. Or rather I was and thought it would be fine to accept `True` even against the protocol.
Yes, that's questionable.

That last suggestion of yours seems expensive. For semantic tokens and signature help we query capabilities on every request. Maybe we need to start being careful about those?

The following should cover all cases "by the book" without unnecessary type comparisons and dict lookups:
```python
def _IsCapabilityProvided( c, q, allow_boolean=True ):
    v = c.get( q )
    if v is None: return False
    t = type( v )
    if t == bool:
        return v if allow_boolean else False
    return t == dict
```
</blockquote></details>

I think you're right. Just fix the bool() expression in your first version and keep it simple. The pedantic-but-correct version is harder to follow, harder to explain and just a bit ugly :D

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1707)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[`ycmd/completers/language_server/language_server_completer.py` line 3546 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1707#-Ne9I54L2RWc5L7KZpd5:-Ne9Uhwt2tLGM1FhzW9p:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/12c6c42da1b76ba446de71b5518d2fe4f809a077/ycmd/completers/language_server/language_server_completer.py#L3546)):*

I think you're right. Just fix the bool() expression in your first version and keep it simple. The pedantic-but-correct version is harder to follow, harder to explain and just a bit ugly :D
</blockquote></details>

Done.

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1707#-:-NeHMrpc7GbjlYuKptdB:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic:
CI is green. I am happy with this PR. Codecov got rate limited again.
@puremourning I'm only not merging this in case you wish to inspect the latest changes.

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1707#-:-NeJws7rC-xKeprFSrN8:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1711, opened 2023-09-27
msvc: Allow building with Visual Studio Preview versions

sankhesh:
This change allows compiling ycmd against latest Visual Studio Preview versions.

Tested against Visual Studio 2022 Community version 17.8.0 Preview 2.0.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1711)


bstaletic:
Thanks for the pull request.
While I am not at all against this, I would like to know what `-prerelease` does.
Apologies, but I do not have a Windows machine handy.

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1711)* status: 0 of 2 LGTMs obtained (waiting on @sankhesh)

___
*[`build.py` line 129 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1711#-NfNJTH682ptyWAy7Yc4:-NfNJTH77B491e1xEWa-:bigjkvb) ([raw file](https://github.com/ycm-core/ycmd/blob/9c5317394360a6a258e79a29785a26eff588aa08/build.py#L129)):*
> ```Python
>       print( "Calling vswhere -latest -installationVersion" )
>     latest_full_v = subprocess.check_output(
>       [ VSWHERE_EXE, '-latest', '-prerelease', '-property', 'installationVersion' ]
> ```

Please can you explain canonically what this does? We don’t want to always use prerelease versions. In fact I’d rather not support them at all as we will never test with them. I’m happy to make it work by chance or by opt-in, but not by default.

sankhesh:
With `-prerelease`,  `vswhere` also returns any prereleases/preview versions along with the release versions.
See doc: https://github.com/microsoft/vswhere/blob/f791b32483204ad3b0ff3cdce57a0c85ee119a8f/src/vswhere.lib/vswhere.lib.rc#L80

The current code defaults to the latest without user opt-in/choice if there are multiple versions. This change just extends it further to allow latest *Preview* versions as well. Note that without this change, if I just have VS Preview version installed (with valid MSVC compiler), ycmd build fails with the error that MSVC could not be found.

puremourning:
> With `-prerelease`,  `vswhere` also returns any prereleases/preview versions along with the release versions.
> 
> See doc: https://github.com/microsoft/vswhere/blob/f791b32483204ad3b0ff3cdce57a0c85ee119a8f/src/vswhere.lib/vswhere.lib.rc#L80
> 
> 
> 
> The current code defaults to the latest without user opt-in/choice if there are multiple versions. This change just extends it further to allow latest *Preview* versions as well. Note that without this change, if I just have VS Preview version installed (with valid MSVC compiler), ycmd build fails with the error that MSVC could not be found.

If I have both a released and pre-release version, which is picked?

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1711)* status: 0 of 2 LGTMs obtained (waiting on @sankhesh)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1711#-NfNO3w86OwB18rwbt8w:-NfNO3w91bzgqAykUiJk:b-pud9wf) (no related file):*
Flake8 error means ci is broken. Line too long. 

___

sankhesh:
> If I have both a released and pre-release version, which is picked?

I am not an expert but IMHO, because of the `-latest` flag, the later of the two versions will be picked i.e. between VS2019 Preview and VS 2022, the latter will be chosen.

> Flake8 error means ci is broken. Line too long.

Thanks, I'll fix that.

bstaletic (approved):
I am in favour of this. MS takes ages between two non-preview releases and I know that some people do use latest preview only. Plus MS likes to write blog posts about how latest preview versions are the bestest.

:lgtm: 

@sankhesh Would you mind using reviewable for the review. It helps keep discussion threads in order and has some nice features.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1711#-:-Nfpn2S51je21MXyIOOD:bcv5aqw)* status: 1 of 2 LGTMs obtained (waiting on @sankhesh)

sankhesh (commented):
Sure :+1:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1711#-:-Nfr26NC9Ec5oB5-mdH2:b-tt8ls0)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic and @puremourning)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1711#-NfNO3w86OwB18rwbt8w:-Nfr11KF2YIzjXEZ-20X:b-896fix) (no related file):*

Flake8 error means ci is broken. Line too long.
</blockquote></details>

Done.

___
___
*[`build.py` line 129 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1711#-NfNJTH682ptyWAy7Yc4:-Nfr1F4YAYv3rhz7-KfN:bb6un1j) ([raw file](https://github.com/ycm-core/ycmd/blob/9c5317394360a6a258e79a29785a26eff588aa08/build.py#L129)):*

Please can you explain canonically what this does? We don’t want to always use prerelease versions. In fact I’d rather not support them at all as we will never test with them. I’m happy to make it work by chance or by opt-in, but not by default.
</blockquote></details>

Done

sankhesh:
Hmm.. I don't know why that CI is failing.

bstaletic:
It happens. Windows jobs are a little flakey.

sankhesh:
@bstaletic @puremourning Good to merge?

bstaletic:
I am ok with this pull request, but @puremourning said he would prefer this to be opt-in.

> I’m happy to make it work by chance or by opt-in, but not by default.

So could you add an option like `--preview-msvc` (or smothing, naming is hard) to the argument parser in `build.py`?

sankhesh:
@bstaletic @puremourning The latest commit adds the `--preview-msvc` option.

puremourning (approved):
:lgtm:

thanks!

> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1711#-:-NlUjJeP5k9tno_yBYoV:b-i7on7d)* status: 1 of 2 LGTMs obtained (and 1 stale) (waiting on @sankhesh)

sankhesh (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1711#-:-NlUnBfu82ri1nsFEy94:bnfp4nl)* status: <span>:shipit:</span> complete! 2 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

sankhesh (commented):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1711)* status: <span>:shipit:</span> complete! 2 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

sankhesh:
@bstaletic @puremourning Anything else needed here?

puremourning:
> @bstaletic @puremourning Anything else needed here?

Well, CI is failing, so we will need to resolve that before merging.

PR #1713, opened 2023-10-06
Make GoToCallers/GoToCallees respect ycmd protocol

bstaletic:
ycmd protocol states that goto responses that contain only a single location will not be a list of locations, but just that location i.e. a json object, not a list of objects.

YCM as the main client uses this type level distinction to decide if it should open a location list or jump right away.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1713)

### DIFF
--- a/ycmd/completers/cpp/clangd_completer.py
+++ b/ycmd/completers/cpp/clangd_completer.py
@@ -313,6 +313,9 @@ class ClangdCompleter( language_server_completer.LanguageServerCompleter ):
 
 
   def GoToAlternateFile( self, request_data ):
+    if not self.ServerIsReady():
+      raise RuntimeError( 'Server is initializing. Please wait.' )
+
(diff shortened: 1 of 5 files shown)


bstaletic (commented):
The behaviour does make sense, but was not what I expected in case of type hierarchy, where I:

- see the inheritance and want to jump straight to the definition, to check its inheritance.
- but YCM opens a quickfix list.

For call hierarchies, I think the existing behaviour is fine. Plus we have decided not to go for the simple type hierarchy support.

I should probably drop functional changes and reduce the pull request to tests only.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1713#-:-Ng4G7qT58jOW23B69Ez:b8e6z4f)* status: 0 of 2 LGTMs obtained

PR #1722, opened 2023-12-15
Only parse the initial file the first time it's opened

puremourning:
When the language_server_completer is first constructed, it's usually a FileReadyToParse callback. But at that time, we can't actually parse anything because we haven't done the initialization exchange with the server, so we queue this up for later.

The way this was implemenented was to register a "on file ready to parse" callback to update the server with the file contents at the time the completer was constructed. In practice, as the completer is usually constructed via a file ready to pase event, this actually happens immediately and we then shunt the request into "OnInitializeComplete" handlers, which are fired when the initialize exchange happens.

Why don't we just put something directly in OnInitializeComplete handlers? Well because in the constructor, we don't have the request_data - we only get that in the OnFileReadyToParse callback, so we have this jank. WCGW?

As a result of this dance, we actually introduced a subtle error. OnFileReadyToParse handlers are called _every time_ we have a file parse event, and - we never _remove_ this "parse the file contents as they were at the time of initialization" request from the "file ready to parse handlers" list. So every time we has a file parse event, we update the server with _stale_ data, then immediately correct it! In practice, this isn't all that noticable, but it is braindead.

Simple solution is to add a "once" flag to the handler creating oneshot handlers, and clear such handlers after firing them (or converting them to initialize handlers). Jank to fix jank. It might be better to find a way to just remove these handler "lists" altogether, as there is only one other place in the codebase where one is registered, and we could always just pass request_data to constructor if we have it...

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1722)

### DIFF
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -1005,7 +1005,8 @@ class LanguageServerCompleter( Completer ):
     self._on_file_ready_to_parse_handlers = []
     self.RegisterOnFileReadyToParse(
       lambda self, request_data:
-        self._UpdateServerWithFileContents( request_data )
+        self._UpdateServerWithFileContents( request_data ),
+      True # once
     )
 
     self._signature_help_disabled = user_options[ 'disable_signature_help' ]
@@ -1898,6 +1899,12 @@ class LanguageServerCompleter( Completer ):
     if not self.ServerIsHealthy():
       return
 
+    def ClearOneshotHandlers():
+      self._on_file_ready_to_parse_handlers = [
+        ( handler, once ) for handler, once
+        in self._on_file_ready_to_parse_handlers if not once
+      ]
+
     # If we haven't finished initializing yet, we need to queue up all functions
     # registered on the FileReadyToParse event and in particular
     # _UpdateServerWithFileContents in reverse order of registration. This
@@ -1905,13 +1912,15 @@ class LanguageServerCompleter( Completer ):
     # messages. This is important because server start up can be quite slow and
     # we must not block the user, while we must keep the server synchronized.
     if not self._initialize_event.is_set():
-      for handler in reversed( self._on_file_ready_to_parse_handlers ):
+      for handler, _ in reversed( self._on_file_ready_to_parse_handlers ):
         self._OnInitializeComplete( partial( handler,
                                              request_data = request_data ) )
+      ClearOneshotHandlers()
       return
 
-    for handler in reversed( self._on_file_ready_to_parse_handlers ):
+    for handler, _ in reversed( self._on_file_ready_to_parse_handlers ):
       handler( self, request_data )
+    ClearOneshotHandlers()
 
     # Return the latest diagnostics that we have received.
     #
@@ -2480,8 +2489,8 @@ class LanguageServerCompleter( Completer ):
     self._on_initialize_complete_handlers.append( handler )
 
 
-  def RegisterOnFileReadyToParse( self, handler ):
-    self._on_file_ready_to_parse_handlers.append( handler )
+  def RegisterOnFileReadyToParse( self, handler, once=False ):
+    self._on_file_ready_to_parse_handlers.append( ( handler, once ) )
 
 
   def GetHoverResponse( self, request_data ):
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -1922,6 +1922,8 @@ class LanguageServerCompleter( Completer ):
       handler( self, request_data )
     ClearOneshotHandlers()
 
+    self._UpdateServerWithFileContents( request_data )
+
     # Return the latest diagnostics that we have received.
     #
     # NOTE: We also return diagnostics asynchronously via the long-polling
--- a/ycmd/tests/rust/__init__.py
+++ b/ycmd/tests/rust/__init__.py
@@ -44,10 +44,11 @@ def tearDownModule():
 
 def StartRustCompleterServerInDirectory( app, directory ):
   app.post_json( '/event_notification',
-                 BuildRequest(
-                   filepath = os.path.join( directory, 'src', 'main.rs' ),
-                   event_name = 'FileReadyToParse',
-                   filetype = 'rust' ) )
+                 BuildRequest( filepath = os.path.join( directory,
+                                                        'src',
+                                                        'main.rs' ),
+                               event_name = 'FileReadyToParse',
+                               filetype = 'rust' ) )
   WaitUntilCompleterServerReady( app, 'rust' )
 
 
--- a/ycmd/tests/test_utils.py
+++ b/ycmd/tests/test_utils.py
@@ -16,6 +16,7 @@
 # along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
 
 
+from ycmd.utils import ReadFile
 from hamcrest import ( assert_that,
                        contains_exactly,
                        contains_string,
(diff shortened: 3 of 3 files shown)


puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1722)* status: 0 of 2 LGTMs obtained

___
*[`ycmd/completers/language_server/language_server_completer.py` line 1918 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1722#-NljRAAm2PjsfTOvTwqM:-NljRAAm2PjsfTOvTwqN:b-kwl87d) ([raw file](https://github.com/ycm-core/ycmd/blob/70000e3326ac36cbbbe1d9d7b0d779096feb2e3f/ycmd/completers/language_server/language_server_completer.py#L1918)):*
> ```Python
>         self._OnInitializeComplete( partial( handler,
>                                              request_data = request_data ) )
>         ClearOneshotHandlers()
> ```

no, do this outside the for loop, muppet

bstaletic (changes requested):
Sorry for taking so long to get to reviewing this.

> we never _remove_ this "parse the file contents as they were at the time of initialization"

Please tell me if I am getting this whole thing right.

Both `ClangdCompleter` and `LanguageServerCompleter` register a certain callback for after initialization handshake is done.
Those callbacks are accumulated in `LanguageServerCompleter._on_file_ready_to_parse_handlers`.
These are then processed in `LanguageServerCompleter.OnFileReadyToParse`.
But! If the handshake is not done, we put the handlers into `_on_initialize_complete_handlers`.
So, we can accumulate stuff in the `_on_initialize_complete_handlers` multiple times even though we only have a single parse request pending.

Currently, `_on_file_ready_to_parse_handlers` is never emptied, which this pull request is trying to solve.

If all that is correct (and I believe it is), then why are we introducing this `once` flag, instead of just treating both sources of `_on_file_ready_to_parse` callbacks as "one-shot"?

> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1722#-:-NmZe3TT4Qltk6fnbloL:b-lx8gzv)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[`ycmd/tests/go/__init__.py` line 49 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1722#-NmZgqgg2bZcZC6VK30G:-NmZgqgg2bZcZC6VK30H:b-dvnq5i) ([raw file](https://github.com/ycm-core/ycmd/blob/6778d575ab88b52fc461f5b13c0b667fba1639af/ycmd/tests/go/__init__.py#L49)):*
> ```Python
>                  BuildRequest(
>                    filepath = main_file,
>                    contents = GetTestFileContents( main_file ),
> ```

And why were these changes necessary? We should be able to handle this even without explicitly getting contents. Right?

bstaletic (changes requested):
To answer my own question:

Clangd is not a true *one shot* callback, unlike the other one. It is a once-per-file callback and that optimization is handled in the callback itself.

So that part is fine.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1722#-:-Nm_py-d2mUiQdE1frd8:b-yuakx7)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[`ycmd/tests/go/__init__.py` line 49 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1722#-NmZgqgg2bZcZC6VK30G:-Nma3J8P-vAk4z-aQePb:bk3ww5g) ([raw file](https://github.com/ycm-core/ycmd/blob/6778d575ab88b52fc461f5b13c0b667fba1639af/ycmd/tests/go/__init__.py#L49)):*

And why were these changes necessary? We should be able to handle this even without explicitly getting contents. Right?
</blockquote></details>

Okay, we do need to read the contents at some point... obviously!

But can we change `BuildRequest` to actually do something lik

    contents = kwrgs.get( 'contents', GetTestFileContents( filepath ) )

?
That seems cleaner than having to remember to always pass `contents` explicitly.

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1722)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[`ycmd/tests/go/__init__.py` line 49 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1722#-NmZgqgg2bZcZC6VK30G:-NmftIfX9kQsj6VK21Ox:b-epq0ri) ([raw file](https://github.com/ycm-core/ycmd/blob/6778d575ab88b52fc461f5b13c0b667fba1639af/ycmd/tests/go/__init__.py#L49)):*

Okay, we do need to read the contents at some point... obviously!

But can we change `BuildRequest` to actually do something lik

    contents = kwrgs.get( 'contents', GetTestFileContents( filepath ) )

?
That seems cleaner than having to remember to always pass `contents` explicitly.
</blockquote></details>

We... could. But I fear that it will cause a billion other things to break.

I'll try it, and see how much it breaks. If it breaks the world, I don't think it's worth the effort to change tons of tests that "assume" its current behaviour.

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1722)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[`ycmd/tests/go/__init__.py` line 49 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1722#-NmZgqgg2bZcZC6VK30G:-NmfzAOJ9rpQZH9Z09zm:b-m1q393) ([raw file](https://github.com/ycm-core/ycmd/blob/6778d575ab88b52fc461f5b13c0b667fba1639af/ycmd/tests/go/__init__.py#L49)):*

We... could. But I fear that it will cause a billion other things to break.

I'll try it, and see how much it breaks. If it breaks the world, I don't think it's worth the effort to change tons of tests that "assume" its current behaviour.
</blockquote></details>

Akshually it works great. Done.

bstaletic (approved):
Two minor comments, but :lgtm_strong: anyway.

> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1722#-:-NmiwfaG9FGil2c4VTso:bys5ic2)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

___
*[`ycmd/tests/test_utils.py` line 93 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1722#-Nmiw6ZF7PubVqj2QWob:-Nmiw6ZF7PubVqj2QWoc:bz74he7) ([raw file](https://github.com/ycm-core/ycmd/blob/32041794cdbf9588e09b2c729903b93daae036f0/ycmd/tests/test_utils.py#L93)):*
> ```Python
>     if 'contents' in kwargs
>     else GetTestFileContents( filepath )
>       if 'filepath' in kwargs
> ```

Do we need this? There's definitely nothing at `/foo` in CI or on my machine.
___
*[`ycmd/tests/rust/__init__.py` line 49 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1722#-NmivnRCAdmLGn3YcT3Y:-NmivnRCAdmLGn3YcT3Z:b-mkwey5) ([raw file](https://github.com/ycm-core/ycmd/blob/32041794cdbf9588e09b2c729903b93daae036f0/ycmd/tests/rust/__init__.py#L49)):*
> ```Python
>                  BuildRequest( filepath = os.path.join( directory,
>                                                         'src',
>                                                         'main.rs' ),
> ```

I'm guessing this also could have been reverted to fit on one line, but oh well.

puremourning (commented):
:lgtm:

> Reviewed 1 of 5 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1722#-:-NmjsNhA9c0ccVOoBSly:bnfp4nl)* status: <span>:shipit:</span> complete! 2 of 2 LGTMs obtained (waiting on @puremourning)

___
*[`ycmd/tests/test_utils.py` line 93 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1722#-Nmiw6ZF7PubVqj2QWob:-Nmjs4R471cNi1F6-ncl:bc2kre6) ([raw file](https://github.com/ycm-core/ycmd/blob/32041794cdbf9588e09b2c729903b93daae036f0/ycmd/tests/test_utils.py#L93)):*

Do we need this? There's definitely nothing at `/foo` in CI or on my machine.
</blockquote></details>

While /foo probably doesn't exist, trying to open it, throwing an exception, handling that exception and returning '' seems wasteful, as this is a very common case. It's also saying what it means, which I prefer.
___
*[`ycmd/tests/rust/__init__.py` line 49 at r3](https://reviewable.io/reviews/ycm-core/ycmd/1722#-NmivnRCAdmLGn3YcT3Y:-NmjsOecEidqsuQiTiff:b8sfeze) ([raw file](https://github.com/ycm-core/ycmd/blob/32041794cdbf9588e09b2c729903b93daae036f0/ycmd/tests/rust/__init__.py#L49)):*

I'm guessing this also could have been reverted to fit on one line, but oh well.
</blockquote></details>

meh, not gonna change it.

PR #1724, opened 2024-01-04
[READY] Make C# completer consistent with other completers regarding GoToDocumentOutline

bstaletic:
All other semantic completers return a list only when more than one item is in the list. Otherwise return that one item.

WIP because I can't be bothered to set up C# on my machine. I will be using CI for that instead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1724)

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -681,7 +681,9 @@ class CsharpSolutionCompleter( object ):
                             ref_line,
                             ref[ 'Column' ] ),
             line ) )
-      return goto_locations
+      if len( goto_locations ) > 1:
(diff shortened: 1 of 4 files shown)


bstaletic:
*nix seems happy with this pull request, so marking it ready.

puremourning (approved):
:lgtm:

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1724#-:-NnL50dX0ivi0_itFfwj:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1725, opened 2024-01-09
Update abseil to the latest LTS.

bstaletic:
This should be enough to fix ycm-core/YouCompleteMe#4212

@Jaywalker, would you mind double checking my choice of abseil commit?

@puremourning Abseil's cmake is warning about this:

https://github.com/abseil/abseil-cpp/blob/lts_2023_08_02/CMakeLists.txt#L73-L78

We should figure out how to set that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1725)

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -247,11 +247,7 @@ else()
     GIT_TAG 3b4a16abad2c2ddc494371cc39a2946e36d35d11
     SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/absl
   )
-  FetchContent_GetProperties( absl )
-  if ( NOT absl_POPULATED )
-    FetchContent_Populate( absl )
-  endif()
-  add_subdirectory( absl )
+  FetchContent_MakeAvailable( absl )
 endif()
 
 add_subdirectory( ycm )
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -237,6 +237,7 @@ find_package( Python3 3.6 REQUIRED COMPONENTS Interpreter Development )
 #############################################################################
 
 set( CMAKE_POSITION_INDEPENDENT_CODE ON )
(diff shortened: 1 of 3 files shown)


puremourning (approved):
:lgtm:

What happens if we enable that flag? Our build is c++17 so abseil should just use that and STFU

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1725#-:-Nnh0l_CE7mFJVS1cWkI:b8225uf)* status: 1 of 2 LGTMs obtained

puremourning (approved):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1725)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

Jaywalker:
An eyes-only review shows it has the patch necessary for the OpenBSD/FreeBSD fix. I'll do full build of it on my OpenBSD machine in a little bit and confirm it works, but I fully expect it will!

bstaletic (commented):
> What happens if we enable that flag? Our build is c++17 so abseil should just use that and STFU

Tried it, it's fine. In general, a dependency saying "needs C++14" means "needs C++14 or newer". Plus abseil knows about C++17 and switches to it after turning the option on.

```
-- Performing Test ABSL_INTERNAL_AT_LEAST_CXX17
-- Performing Test ABSL_INTERNAL_AT_LEAST_CXX17 - Success
```

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1725#-:-NnkXQeA5qtiKSMi3jJh:b6ddths)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1725#-:-No7blFoCzitb9xx33h6:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1726, opened 2024-01-09
[RFC] Do not use a full line as description for symbol search

bstaletic:
This does lose a bit of context, but not much. Methods still have their signatures visible and variables are... well... variables.

The idea is, if a client wants to later do filtering and sorting of symbols, we should not be doing that on the whole line, which might be left padded with tabs and have other punctuation. We already had a report that this does not work well. It almost works if a client `strip()`s the description before using it at all.

LSP completer instead uses `extra_data` to supply identifier name and kind, but omnisharp-roslyn does not provide us with a symbol kind and thus we can only put `ref[ 'Text' ]` in the description.

@teasp00n was the user who pointed out the problems with using the whole line for filtering.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1726)

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -633,31 +633,23 @@ class CsharpSolutionCompleter( object ):
     if quickfixes:
       if len( quickfixes ) == 1:
         ref = quickfixes[ 0 ]
-        ref_file = ref[ 'FileName' ]
-        ref_line = ref[ 'Line' ]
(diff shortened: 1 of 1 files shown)


puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1726#-:-No7bU1z97LHFbSZ1tAu:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1728, opened 2024-01-14
Sort diagnostics according to severity for detailed diag responses

bstaletic:
Users are more likely to be interested in errors than warnings in /detailed_diagnostic responses. That means we need to sort the diagnostics before finding the one nearest to the cursor.

For LSP completers, that's easy as the severity is numerical. For Omnisharp-roslyn and TSServer, the severity is a word and we can't just say `sort( key = severity )`.

Since LSP severity was convenient for comparison, I have opted for sorting C# and TS diags by the equivalent LSP severity.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1728)

### DIFF
--- a/ycmd/completers/cs/cs_completer.py
+++ b/ycmd/completers/cs/cs_completer.py
@@ -316,6 +316,9 @@ class CsharpCompleter( Completer ):
     if not diagnostics:
       raise ValueError( NO_DIAGNOSTIC_MESSAGE )
 
+    # Prefer errors to warnings and warnings to infos.
+    diagnostics.sort( key = _CsDiagnosticToLspSeverity )
+
     closest_diagnostic = None
     distance_to_closest_diagnostic = 999
 
@@ -968,3 +971,11 @@ def _ModifiedFilesToFixIt( changes, request_data ):
         request_data[ 'line_num' ],
         request_data[ 'column_codepoint' ] ),
       chunks )
+
+
+def _CsDiagnosticToLspSeverity( diagnostic ):
+  if diagnostic.kind_ == 'ERROR':
+    return 1
+  if diagnostic.kind_ == 'WARNING':
+    return 2
+  return 3
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -1663,9 +1663,14 @@ class LanguageServerCompleter( Completer ):
       return responses.BuildDisplayMessageResponse(
           'No diagnostics for current file.' )
 
+    # Prefer errors to warnings and warnings to infos.
+    diagnostics.sort( key = lambda d: d[ 'severity' ] )
+
+    # request_data uses 1-based offsets, but LSP diagnostics use 0-based.
+    # It's easier to shift this one offset by -1 than to shift all diag ranges.
     current_column = lsp.CodepointsToUTF16CodeUnits(
         GetFileLines( request_data, current_file )[ current_line_lsp ],
-        request_data[ 'column_codepoint' ] )
+        request_data[ 'column_codepoint' ] ) - 1
     minimum_distance = None
 
     message = 'No diagnostics for current line.'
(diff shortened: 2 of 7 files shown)


bstaletic:
According to the coverage data, we're only ever dealing with errors in our tests, at least for C# and TS.

puremourning (commented):
Change LGTM, but could we at least add a regression test for LSP ? Seems like something that we could easily forget about.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1728#-:-No7Pi0hEgwbwCJ-1jaA:bge8cyc)* status: 0 of 2 LGTMs obtained

puremourning (approved):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1728)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic (commented):
I wanted to, but I couldn't make a test case without pulling in the entirety of ranges-v3. That just feels like a massive dependency for our need.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1728#-:-No7QrD5DPE4bsW0g-iT:b5t3wfj)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

puremourning:
Don't we just need some code which produces a warning and error on the same line.

I think clangd does this. maybe something like:

```
int x;
if ( x = 2 )
{
}
```

With `-Wno-error -Werror=uninitialized`

bstaletic (commented):
Tests added.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1728#-:-NoBh4Ez0FoZPuftk8c4:b-wj4ea6)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

puremourning (changes requested):
> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1728)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[`ycmd/completers/language_server/language_server_completer.py` line 1666 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1728#-NqcV-t06HAgy07GJewQ:-NqcV-t06HAgy07GJewR:bhlzhr5) ([raw file](https://github.com/ycm-core/ycmd/blob/b5d1e22d777afa60b9fc501c4ca6eb0d3ef90844/ycmd/completers/language_server/language_server_completer.py#L1666)):*
> ```Python
>     current_column = lsp.CodepointsToUTF16CodeUnits(
>         GetFileLines( request_data, current_file )[ current_line_lsp ],
>         request_data[ 'column_codepoint' ] - 1 )
> ```

I'm not sure about this.... 


the help doc for this func says the offset is in 1-based units:

```
def CodepointsToUTF16CodeUnits( line_value, codepoint_offset ):
  """Return the 1-based UTF-16 code unit offset equivalent to the 1-based
  unicode codepoint offset |codepoint_offset| in the Unicode string
  |line_value|"""
```

If `current_column` is `0-based` then we should subtract 1 from the result of the call, not the input.

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1728)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[`ycmd/completers/language_server/language_server_completer.py` line 1666 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1728#-NqcV-t06HAgy07GJewQ:-NqemSvz-PJZRzP9lFre:b8sp355) ([raw file](https://github.com/ycm-core/ycmd/blob/b5d1e22d777afa60b9fc501c4ca6eb0d3ef90844/ycmd/completers/language_server/language_server_completer.py#L1666)):*

I'm not sure about this.... 


the help doc for this func says the offset is in 1-based units:

```
def CodepointsToUTF16CodeUnits( line_value, codepoint_offset ):
  """Return the 1-based UTF-16 code unit offset equivalent to the 1-based
  unicode codepoint offset |codepoint_offset| in the Unicode string
  |line_value|"""
```

If `current_column` is `0-based` then we should subtract 1 from the result of the call, not the input.
</blockquote></details>

Without the `-1`, `current_column` is 1-based.
However, we are coparing that column to `diagnostic[ 'range' ]`, later in the for loop, where diagnostics have 0-based character offsets. The comment I added above the call to `CodepointsToUTF16CodeUnits()` should explain that.

I remember going back and forward and not being sure if that `-1` should go before of after `)` of the function call.

Thanks for catching this! Fixed.

puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1728#-:-Nqr1-yV9zt_Gli8AQ5F:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1731, opened 2024-01-21
[READY] Add support for didChangeWorkspaceFolders

bstaletic:
Perhaps it's finally time to add this capability...

There are a couple missing pieces:

1. So far, this does not take into account the extra confs.
2. Clangd completer does not have GetProjectRootFiles() implemented, but also clangd does not want to listen to this notification. I guess it knows better than us?
3. Java completer also does not have GetProjectRootFiles() implemented, but actually wants this notification.
4. Do we need `_project_directory` *and* `_server_workspace_dirs`?
5. Tests... after we figure out the answers to the above!

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1731)

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,8 @@ wouldn't usually know about. The value is a list of dictionaries containing:
 - `capabilities'`: Overrides the default LSP capabilities of ycmd.
   - If you enable `workspace/configuration` support, check the extra conf
     details, relevant to LSP servers.
+- `additional_workspace_dirs`: Specifies statically known workspaces that should
+  be open on LSP server startup.
 - `triggerCharacters`: Override the LSP server's trigger characters for
   completion. This can be useful when the server obnoxiously requests completion
   on every character or for example on whitespace characters.
--- a/ycmd/completers/java/java_completer.py
+++ b/ycmd/completers/java/java_completer.py
@@ -404,6 +404,10 @@ class JavaCompleter( language_server_completer.LanguageServerCompleter ):
     return self._java_project_dir
 
 
+  def GetWorkspaceForFilepath( self, filepath ):
+    return _FindProjectDir( os.path.dirname( filepath ) )
+
+
   def _WipeWorkspace( self, request_data, args ):
     with_config = False
     if len( args ) > 0 and '--with-config' in args:
--- a/ycmd/completers/java/java_completer.py
+++ b/ycmd/completers/java/java_completer.py
@@ -358,9 +358,6 @@ class JavaCompleter( language_server_completer.LanguageServerCompleter ):
       'OrganizeImports': (
         lambda self, request_data, args: self.OrganizeImports( request_data )
       ),
-      'OpenProject': (
-        lambda self, request_data, args: self._OpenProject( request_data, args )
-      ),
       'WipeWorkspace': (
         lambda self, request_data, args: self._WipeWorkspace( request_data,
                                                               args )
@@ -416,25 +413,6 @@ class JavaCompleter( language_server_completer.LanguageServerCompleter ):
     self._RestartServer( request_data,
                          wipe_workspace = True,
                          wipe_config = with_config )
-
-
-  def _OpenProject( self, request_data, args ):
-    if len( args ) != 1:
-      raise ValueError( "Usage: OpenProject <project directory>" )
-
-    project_directory = args[ 0 ]
-
-    # If the dir is not absolute, calculate it relative to the working dir of
-    # the client (if supplied).
-    if not os.path.isabs( project_directory ):
-      if 'working_dir' not in request_data:
-        raise ValueError( "Project directory must be absolute" )
-
-      project_directory = os.path.normpath( os.path.join(
-        request_data[ 'working_dir' ],
-        project_directory ) )
-
-    self._RestartServer( request_data, project_directory = project_directory )
 
 
   def _Reset( self ):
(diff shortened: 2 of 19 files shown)


bstaletic:
This pull request actually allows rust-analyzer to report diagnostics for two projects opened in the same vim instance. For example:

1. `ycmd/tests/rust/testdata/common`
2. `ycmd/tests/rust/testdata/macro`

bstaletic:
Rebased.

This pull request is behaving correctly for rust, go and java. C++ server does not support this notification and we respect that.

Other than that, I have tried it with haskell and it did what I expected, but haskell server does not really do anything with the notification.

puremourning (changes requested):
Looks good!

Answers

1. extra cones... I think you added this?
2. clangd does its own copile_commands.json scanning and does a good job of associating files with dbs, let's leave them to it!
3. java completer is special, and I can see that you've handled it like the special sausage that it is
4. I think we keep both. The project root is still _special_ in that it's the rootUri in the init and it's the CWD of the server
5. I would like to see some tests for the other engines if we can. The java one was interesting that it shows the server is bonkers and seems to duplicate the messages :(

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1731#-:-NqcT87TCk9HLFJQ1Ope:b-jovavj)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[`ycmd/completers/java/java_completer.py` line 421 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1731#-NqcS0WG9SULnPS0K9Pa:-NqcS0WG9SULnPS0K9Pb:bgpqo2r) ([raw file](https://github.com/ycm-core/ycmd/blob/98a541b7f9c5e1d1fe8dfc4d3f9a2c9dc62d2235/ycmd/completers/java/java_completer.py#L421)):*
> ```Python
> 
> 
>   def _OpenProject( self, request_data, args ):
> ```

I wonder if we can just ditch this now? I doubt anyone uses it. I added it for a specific purpose to work around this limitation, but TBH YcmCompleter RestartServer works just as well, and this is just a bit of a hack on top of that.

I vote remove. WDYT?

bstaletic (commented):
Right, I should have updated the opening comment.

> extra cones... I think you added this?

Extra confs can report statically known workspaces. I am unsure if we will need something like `extra_conf.GetWorkspaceForFile(file)`. I am happy to defer that to a point after we get a lot more field experience and thus know whether we need that or not.

> clangd

Agreed.

> java [snip] special sausage

Loved that!

> I would like to see some tests for the other engines if we can.

Originally, I tried to write a test for rust. But it tends to publish one set of diagnostics, change its mind, then publish another set of diagnostics. But not always. Our tests for the diagnostic poll don't deal well with that, so I gave up on testing with rust.

I can try with go...

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1731#-:-Nqc_DUc3KJZKZOK3Bg1:b7tcl1e)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[`ycmd/completers/java/java_completer.py` line 421 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1731#-NqcS0WG9SULnPS0K9Pa:-Nqcabsq1SXflW1Xku_2:bp67wos) ([raw file](https://github.com/ycm-core/ycmd/blob/98a541b7f9c5e1d1fe8dfc4d3f9a2c9dc62d2235/ycmd/completers/java/java_completer.py#L421)):*

I wonder if we can just ditch this now? I doubt anyone uses it. I added it for a specific purpose to work around this limitation, but TBH YcmCompleter RestartServer works just as well, and this is just a bit of a hack on top of that.

I vote remove. WDYT?
</blockquote></details>

I initially stayed away from that because it is technically a breaking change. I can remove it, sure.

bstaletic (commented):
Ha! Gopls works great!

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1731#-:-Nqf3wJI9mrh7LPI2sT-:b-4g2eg0)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic and @puremourning)

___
*[`ycmd/completers/java/java_completer.py` line 421 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1731#-NqcS0WG9SULnPS0K9Pa:-Nqf3nxuEFFuAkEZizDI:bss4a2b) ([raw file](https://github.com/ycm-core/ycmd/blob/98a541b7f9c5e1d1fe8dfc4d3f9a2c9dc62d2235/ycmd/completers/java/java_completer.py#L421)):*

I initially stayed away from that because it is technically a breaking change. I can remove it, sure.
</blockquote></details>

Removed.

bstaletic (commented):
Wiped off the jank sauce. 100% diff coverage hit.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1731#-:-NqgSWd57q75QMIclvNe:b-78qa8j)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

puremourning (approved):
:lgtm:

> Reviewed 9 of 12 files at r1, 10 of 10 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1731#-:-Nqr1R3q9evvoRy12ZML:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1733, opened 2024-02-11
[READY] Implement LSP type/call hierarchies as actual hierarchies

bstaletic:
This is the implementation of the server side changes needed to have useful support for LSP call/type hierarchies.
While the clien side changes deserve more attention, the server side should be fine to review and merge.

What's definitely missing are tests and API documentation.
On top of that, what do we want to do with the old `GoToCollers`/`GoToCallees`? I see three options:

- Remove them completely, which might break some users, but the new alternative should be better.
- Make them an alias for the new subcommands. This option wouldn't break users, but would break clients who would not adapt.
- Do nothing and maintain the old API forever, like with have done with other things in the past.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1733)


puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1733#-:-NzskZfo0F5St-hX5oas:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1733#-Nzsj4nHEM2_dgwL6v0B:-Nzsj4nI4_h_TB_ltbQo:b-cf81dh) (no related file):*
Ideally, we would also update the openapi.yaml for the docs. No need to duplicate anything that's in the LSP spec, but we maybe should document our modifications/normalisattions on our ycmd API.

___

puremourning (commented):
> Reviewed 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1733)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

puremourning (approved):
:lgtm:

> Reviewed 15 of 15 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1733#-:-NzydIWI72xj9lTK_abt:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic (commented):
This should finally be ready.
If only Microsoft had not borked windows-2022 C runtime...

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1733#-:-NzzGcdmB1l7OzK1pUkL:bp3987a)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

*[a discussion](https://reviewable.io/reviews/ycm-core/ycmd/1733#-Nzsj4nHEM2_dgwL6v0B:-NzzGbZF6Cee6rVtu6V1:b-896fix) (no related file):*

Ideally, we would also update the openapi.yaml for the docs. No need to duplicate anything that's in the LSP spec, but we maybe should document our modifications/normalisattions on our ycmd API.
</blockquote></details>

Done.

___

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1733)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic)

___
*[`ycmd/completers/language_server/language_server_completer.py` line 3413 at r5](https://reviewable.io/reviews/ycm-core/ycmd/1733#-O-7gP5e8elE19TvjPvi:-O-7gP5e8elE19TvjPvj:bpuk5zn) ([raw file](https://github.com/ycm-core/ycmd/blob/07d0424e872827ce4463234167f417ab458300ab/ycmd/completers/language_server/language_server_completer.py#L3413)):*
> ```Python
>     file_contents = []
> 
>   range = location.get( 'selectionRange' ) or location[ 'range' ]
> ```

I think this might be surprising when used in other contexts. There are other request that have 'selectionRange' , such as "DocumentSymbol" In that case I'm not 100% sure that 'selectionRange' is better than 'range' for our purposes:

```
	/**
	 * The range enclosing this symbol not including leading/trailing whitespace
	 * but everything else like comments. This information is typically used to
	 * determine if the clients cursor is inside the symbol to reveal in the
	 * symbol in the UI.
	 */
	range: Range;

	/**
	 * The range that should be selected and revealed when this symbol is being
	 * picked, e.g. the name of a function. Must be contained by the `range`.
	 */
	selectionRange: Range;
```

perhaps we should pass a 'range_property="range"' argument to this function instead?

puremourning (commented):
> Reviewed 7 of 7 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1733)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @bstaletic)

bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1733)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

___
*[`ycmd/completers/language_server/language_server_completer.py` line 3413 at r5](https://reviewable.io/reviews/ycm-core/ycmd/1733#-O-7gP5e8elE19TvjPvi:-O-7mfYe7D0TWexdfjn-:bl5nrcy) ([raw file](https://github.com/ycm-core/ycmd/blob/07d0424e872827ce4463234167f417ab458300ab/ycmd/completers/language_server/language_server_completer.py#L3413)):*

I think this might be surprising when used in other contexts. There are other request that have 'selectionRange' , such as "DocumentSymbol" In that case I'm not 100% sure that 'selectionRange' is better than 'range' for our purposes:

```
	/**
	 * The range enclosing this symbol not including leading/trailing whitespace
	 * but everything else like comments. This information is typically used to
	 * determine if the clients cursor is inside the symbol to reveal in the
	 * symbol in the UI.
	 */
	range: Range;

	/**
	 * The range that should be selected and revealed when this symbol is being
	 * picked, e.g. the name of a function. Must be contained by the `range`.
	 */
	selectionRange: Range;
```

perhaps we should pass a 'range_property="range"' argument to this function instead?
</blockquote></details>

I was considering that too. I was not sure about `DocumentSymbol` either.

Implemented your suggestion. That should also generalize to `originSelectionRange` and `targetSelectionRange`, were we ever to need a `LocationLink` LSP type.

puremourning (approved):
:lgtm:

> Reviewed 15 of 15 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1733#-:-O-bjoY_1Puk7Naj0OeD:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[`ycmd/completers/typescript/typescript_completer.py` line 823 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1733#-O-bifv6-MX00Wrk1o2k:-O-bifv73w0r4OW8RCHY:biipa8n) ([raw file](https://github.com/ycm-core/ycmd/blob/253f35c5ae0d2bb309abdeaa8214bbbf474ee02b/ycmd/completers/typescript/typescript_completer.py#L823)):*
> ```Python
> 
>   def _Hierarchy( self, request_data, args ):
>     self._Reload( request_data )
> ```

is this reload needed? I suspect that we can rely on the reload from the prepare step?
___
*[`ycmd/tests/typescript/subcommands_test.py` line 1203 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1733#-O-bjYHW0HilK9batdux:-O-bjYHW0HilK9batduy:b22dbu7) ([raw file](https://github.com/ycm-core/ycmd/blob/253f35c5ae0d2bb309abdeaa8214bbbf474ee02b/ycmd/tests/typescript/subcommands_test.py#L1203)):*
> ```Python
> 
>   @SharedYcmd
>   def test_Subcommands_OutgoingCallHierarchy( self, app ):
> ```

I take it there is no type hierarchy for typescript?

bstaletic (commented):
> Dismissed @puremourning from a discussion.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1733)* status: 0 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

___
*[`ycmd/completers/typescript/typescript_completer.py` line 823 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1733#-O-bifv6-MX00Wrk1o2k:-O-cCEhu0ZYEd19TDgFM:b-oqh2wo) ([raw file](https://github.com/ycm-core/ycmd/blob/253f35c5ae0d2bb309abdeaa8214bbbf474ee02b/ycmd/completers/typescript/typescript_completer.py#L823)):*

is this reload needed? I suspect that we can rely on the reload from the prepare step?
</blockquote></details>

Good point.
I started thinking about a potential scenario where a client would keep the hierarchy tree around and then after editing stuff go back to it, but

1. Then `_Reload()` would be called for a different reason anyway.
2. If the call site changes the hierarchy will be wrong.

Dropped the reload.
___
*[`ycmd/tests/typescript/subcommands_test.py` line 1203 at r6](https://reviewable.io/reviews/ycm-core/ycmd/1733#-O-bjYHW0HilK9batdux:-O-cBjcD0vni1ldx9Iz3:b-ht0ukm) ([raw file](https://github.com/ycm-core/ycmd/blob/253f35c5ae0d2bb309abdeaa8214bbbf474ee02b/ycmd/tests/typescript/subcommands_test.py#L1203)):*

I take it there is no type hierarchy for typescript?
</blockquote></details>

That's right.

PR #1734, opened 2024-02-13
Uninstall default bottle plugins

bstaletic:
By default, bottle installs two plugins: `JSONPlugin` and `TemplatePlugin`.

`TemplatePlugin` is always a no-op, because none of the routes we use (like `/ready`) have a template parameter.

`JSONPlugin` could have been useful were it actually serializing all json messages. Instead it only deals with dictionaries. Ycmd has always had `ycmd.handlers._JsonResponse` and we will need to keep it that way. `_JsonResponse` makes `JSONPlugin` a no-op too, because, by the time we get to the `JSONPlugin.apply()` the response is always a str already.

`Bottle.uninstall()` can be passed:

1. An instance of an installed plugins, to uninstall that one plugin.
2. A type, to uninstall all plugins of that type.
3. `True`, to uninstall all plugins.

Since we are uninstalling all default plugins, we can just do

`handlers.app.uninstall( True )` and then install the plugins we care about - `ycmd.HmacPlugin` and `ycmd.WatchdogPlugin`.


puremourning (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1734#-:-NqWXkYO29zZxCNzJW0s:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1736, opened 2024-02-20
[READY] Drop bottle in favour of a tiny hand-rolled "framework"

bstaletic:
Turns out ycmd needs a tiny part of the whole framework specification from PEP 3333.
Note that this does not fully conform to the specification, but it mostly does.

Currently we are only handling POST and GET requests.
Support for middleware is limited. Inspired by Bottle plugins API, but only `plugin.__call__` and `app.install` are implemented.
Error handling is also limited. Application needs to replace the default error handler that *will* handle all errors.

However, this just about fits ycmd's needs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1736)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "third_party/bottle"]
-	path = third_party/bottle
-	url = https://github.com/defnull/bottle
 [submodule "third_party/jedi"]
 	path = third_party/jedi_deps/jedi
 	url = https://github.com/davidhalter/jedi
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-bottle   >= 0.12.18
 regex    >= 2020.2.20
 jedi     >= 0.16.0
 watchdog >= 0.10.2
--- a/run_tests.py
+++ b/run_tests.py
@@ -17,14 +17,10 @@ DIR_OF_WATCHDOG_DEPS = p.join( DIR_OF_THIRD_PARTY, 'watchdog_deps' )
 LIBCLANG_DIR = p.join( DIR_OF_THIRD_PARTY, 'clang', 'lib' )
 
 python_path = [
-  p.join( DIR_OF_THIRD_PARTY, 'bottle' ),
   p.join( DIR_OF_THIRD_PARTY, 'regex-build' ),
-  p.join( DIR_OF_THIRD_PARTY, 'frozendict' ),
   p.join( DIR_OF_THIRD_PARTY, 'jedi_deps', 'jedi' ),
   p.join( DIR_OF_THIRD_PARTY, 'jedi_deps', 'parso' ),
   p.join( DIR_OF_WATCHDOG_DEPS, 'watchdog', 'build', 'lib3' ),
-  p.join( DIR_OF_WATCHDOG_DEPS, 'pathtools' ),
-  p.join( DIR_OF_THIRD_PARTY, 'waitress' ),
 ]
 if os.environ.get( 'PYTHONPATH' ) is not None:
   python_path.append( os.environ[ 'PYTHONPATH' ] )
--- a/ycmd/__main__.py
+++ b/ycmd/__main__.py
@@ -36,19 +36,10 @@ DIR_OF_WATCHDOG_DEPS = os.path.join( DIR_OF_THIRD_PARTY, 'watchdog_deps' )
 DIR_OF_REQUESTS_DEPS = os.path.join( DIR_OF_THIRD_PARTY, 'requests_deps' )
 sys.path[ 0:0 ] = [
     os.path.join( ROOT_DIR ),
-    os.path.join( DIR_OF_THIRD_PARTY, 'bottle' ),
     os.path.join( DIR_OF_THIRD_PARTY, 'regex-build' ),
-    os.path.join( DIR_OF_THIRD_PARTY, 'frozendict' ),
     os.path.join( DIR_OF_THIRD_PARTY, 'jedi_deps', 'jedi' ),
     os.path.join( DIR_OF_THIRD_PARTY, 'jedi_deps', 'parso' ),
-    os.path.join( DIR_OF_REQUESTS_DEPS, 'requests' ),
-    os.path.join( DIR_OF_REQUESTS_DEPS, 'chardet' ),
-    os.path.join( DIR_OF_REQUESTS_DEPS, 'certifi' ),
-    os.path.join( DIR_OF_REQUESTS_DEPS, 'idna' ),
-    os.path.join( DIR_OF_REQUESTS_DEPS, 'urllib3', 'src' ),
-    os.path.join( DIR_OF_WATCHDOG_DEPS, 'watchdog', 'build', 'lib3' ),
-    os.path.join( DIR_OF_WATCHDOG_DEPS, 'pathtools' ),
-    os.path.join( DIR_OF_THIRD_PARTY, 'waitress' ) ]
+    os.path.join( DIR_OF_WATCHDOG_DEPS, 'watchdog', 'build', 'lib3' ) ]
 sys.path.append( os.path.join( DIR_OF_THIRD_PARTY, 'jedi_deps', 'numpydoc' ) )
 
 import atexit
@@ -203,7 +194,6 @@ def Main():
                                     args.stderr,
                                     args.keep_logfiles )
   atexit.register( handlers.ServerCleanup )
-  handlers.app.uninstall( True )
   handlers.app.install( WatchdogPlugin( args.idle_suicide_seconds,
                                         args.check_interval_seconds ) )
   handlers.app.install( HmacPlugin( hmac_secret ) )
--- a/ycmd/handlers.py
+++ b/ycmd/handlers.py
@@ -15,15 +15,14 @@
 # You should have received a copy of the GNU General Public License
 # along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
 
-import bottle
 import json
 import platform
 import sys
 import time
 import traceback
-from bottle import request
 
 
+import ycmd.web_plumbing
 from ycmd import extra_conf_store, hmac_plugin, server_state, user_options_store
 from ycmd.responses import ( BuildExceptionResponse,
                              BuildCompletionResponse,
@@ -40,18 +39,14 @@ from ycmd.utils import LOGGER, StartThread, ImportCore
 ycm_core = ImportCore()
 
 
-# num bytes for the request body buffer; request.json only works if the request
-# size is less than this
-bottle.Request.MEMFILE_MAX = 10 * 1024 * 1024
-
 _server_state = None
 _hmac_secret = bytes()
-app = bottle.Bottle()
+app = ycmd.web_plumbing.AppProducer()
 wsgi_server = None
 
 
 @app.post( '/event_notification' )
-def EventNotification():
+def EventNotification( request, response ):
   request_data = RequestWrap( request.json )
   event_name = request_data[ 'event_name' ]
   LOGGER.debug( 'Event name: %s', event_name )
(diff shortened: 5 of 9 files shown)


bstaletic (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1736)* status: 0 of 2 LGTMs obtained

___
*[`ycmd/web_plumbing.py` line 73 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr4UCDb45l9syOdk9R5:-Nr4UCDb45l9syOdk9R6:bpq6bcb) ([raw file](https://github.com/ycm-core/ycmd/blob/b8451e6dc72aa08db62b87bc920adac74d3dd055/ycmd/web_plumbing.py#L73)):*
> ```Python
> 
> 
> class Request:
> ```

The request class eagerly processes most things. I thought about some caching, but `hmac_plugin.py` will definitely need everything anyway.

The only potential improvement is `json()` and `query()`, since those are actual function calls.

`json()` is only called once at the start of every POST callback, so caching won't do anything.

`query()` is also only called once at the start of every GET callback, so, again, caching won't do anything.

Maybe we could try to avoid setting `self.body` at all if we are processing a GET request, but I don't think that's a worthwhile effort.
___
*[`ycmd/web_plumbing.py` line 114 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr4TfhW2ZcdjixXFq9a:-Nr4TfhW2ZcdjixXFq9b:bq1axm3) ([raw file](https://github.com/ycm-core/ycmd/blob/b8451e6dc72aa08db62b87bc920adac74d3dd055/ycmd/web_plumbing.py#L114)):*
> ```Python
>     self._routes : Mapping[ str, Mapping[ str, CallbackType ] ] = {
>       'GET': {},
>       'POST': {} }
> ```

Routes are saved as a nested dictionary, so we can access callables as

```python
self._routes[method][path]
```

For performance, maybe we want to do

```python
self._get_routes[path]
self._post_routes[path]
```

Need to measure that...

puremourning (changes requested):
This looks really good. Few minor comments. Very pleased with this.

Also, did you try running the YCM client tests too ? Wettest may be a bit of a cheat factor, so I'd like to know that real client tests are working too. (there's also client_test.py which is a good litmus, and that clearly passes).

> Reviewed 8 of 9 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1736#-:-Nr6bArJBeMkJAYxu3V1:b-wdf8p2)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[`ycmd/__main__.py` line 50 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr6GrqU2NYBSXH8AH4w:-Nr6GrqU2NYBSXH8AH4x:b-bjen8x) ([raw file](https://github.com/ycm-core/ycmd/blob/b63d2e86c62d39c20284badfd4bfed6e8797d134/ycmd/__main__.py#L50)):*
> ```Python
>     os.path.join( DIR_OF_REQUESTS_DEPS, 'urllib3', 'src' ),
>     os.path.join( DIR_OF_WATCHDOG_DEPS, 'watchdog', 'build', 'lib3' ),
>     os.path.join( DIR_OF_WATCHDOG_DEPS, 'pathtools' ),
> ```

is pathtools not still needed by watchdog? (or is this an existing error) ?
___
*[`ycmd/web_plumbing.py` line 60 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr6YGBp5Wwa4ztwXmTg:-Nr6YGBq8PLEoev8t8cK:b-tx1pep) ([raw file](https://github.com/ycm-core/ycmd/blob/418bdc86c46dba8ad4370d91a5acfa91b77979f3/ycmd/web_plumbing.py#L60)):*
> ```Python
> 
>   def __getattribute__( self, query_name : str ) -> str:
>     return super().get( query_name, [ '' ] )[ 0 ]
> ```

it's not immediately obvious why this is expecting a list response, nor why the 0th element is the right one.

Is it essentially because you can have the same query string parameter multiple times and parse_qa just always returns a list [ value ] in the case of only one?
___
*[`ycmd/web_plumbing.py` line 96 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr6Z16I2ZY14k-Vnj_g:-Nr6Z16I2ZY14k-Vnj_h:b-42scfj) ([raw file](https://github.com/ycm-core/ycmd/blob/418bdc86c46dba8ad4370d91a5acfa91b77979f3/ycmd/web_plumbing.py#L96)):*
> ```Python
> 
> 
> def _ApplyPlugins( plugins, callback ):
> ```

It took me way too long to understand what was going on here. Not being familiar with bottle's plugin system, I was like wtf is `callback = plugin( callback )` doing...?

But it's basically creating a set of decorators with decorate the underlying handler (original `callback` value) with wrappers (return values of `plugin(callback)`).

Maybe stick that in a comment, because I _will_ forget....!
___
*[`ycmd/web_plumbing.py` line 196 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr6aHs7DQvnh3kor57n:-Nr6aHs7DQvnh3kor57o:bf1zp6e) ([raw file](https://github.com/ycm-core/ycmd/blob/418bdc86c46dba8ad4370d91a5acfa91b77979f3/ycmd/web_plumbing.py#L196)):*
> ```Python
>       response = Response()
>       out = self._ErrorHandler( e, response )
>     except Exception:
> ```

I think if someone sends an invalid route the returned error will be 500 KeyError or something.
I think that's probably OK, though I was half-expecting 404.

I checked the API docs and 404 is never mentioned, so I think it's fine, While the error message will be confusing to users, it represents a programming error that only ycmd/client devs will ever see.
___
*[`ycmd/web_plumbing.py` line 205 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr6ay9hCjdm8wo3Uhez:-Nr6ay9hCjdm8wo3Uhf-:b-8kck8c) ([raw file](https://github.com/ycm-core/ycmd/blob/418bdc86c46dba8ad4370d91a5acfa91b77979f3/ycmd/web_plumbing.py#L205)):*
> ```Python
>       out = self._ErrorHandler( e, response )
>     out = out.encode( 'utf-8' )
>     response_headers = response.headers
> ```

this assignment looks fishy. It looks like you're trying to copy `response.headers` but actually you're taking a pointer to it, then mutating it on the next line.

I have no issue with mutating it (it's probably correct), but just want to check that's what you were intending.

bstaletic (commented):
I have no ran YCM tests. I'll set up my YCM fork, dispatch CI and report the results.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1736#-:-Nr6dUhG52chm3xjDdhV:b-8b64yi)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[`ycmd/__main__.py` line 50 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr6GrqU2NYBSXH8AH4w:-Nr6cB6f0bPqJF9cKzqW:bw0x48q) ([raw file](https://github.com/ycm-core/ycmd/blob/b63d2e86c62d39c20284badfd4bfed6e8797d134/ycmd/__main__.py#L50)):*

is pathtools not still needed by watchdog? (or is this an existing error) ?
</blockquote></details>

It's been a long time since my pull request that got watchdog off pathtools and onto pathlib, which is the standard one.
We have just never removed that thing from the path.
___
*[`ycmd/web_plumbing.py` line 60 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr6YGBp5Wwa4ztwXmTg:-Nr6cOf82eqx7K_KFTE8:b-exk1zz) ([raw file](https://github.com/ycm-core/ycmd/blob/418bdc86c46dba8ad4370d91a5acfa91b77979f3/ycmd/web_plumbing.py#L60)):*
> Is it essentially because you can have the same query string parameter multiple times and parse\_qa just always returns a list \[ value \] in the case of only one?

Yes.

```
>>> urllib.parse.parse_qs('a=b&a=3')
{'a': ['b', '3']}
>>> urllib.parse.parse_qs('a=b&c=d')
{'a': ['b'], 'c': ['d']}
```

Maybe we rather want index `-1`?
Should I put that in a comment?
___
*[`ycmd/web_plumbing.py` line 96 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr6Z16I2ZY14k-Vnj_g:-Nr6dicUCZHJ6Db7sO7X:bpylzh7) ([raw file](https://github.com/ycm-core/ycmd/blob/418bdc86c46dba8ad4370d91a5acfa91b77979f3/ycmd/web_plumbing.py#L96)):*

It took me way too long to understand what was going on here. Not being familiar with bottle's plugin system, I was like wtf is `callback = plugin( callback )` doing...?

But it's basically creating a set of decorators with decorate the underlying handler (original `callback` value) with wrappers (return values of `plugin(callback)`).

Maybe stick that in a comment, because I _will_ forget....!
</blockquote></details>

Fair. Decorator stuff easily gets into the head-scratcher territory.
Done.
___
*[`ycmd/web_plumbing.py` line 196 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr6aHs7DQvnh3kor57n:-Nr6e3LM9UIz3fMXvU4H:b9t3euw) ([raw file](https://github.com/ycm-core/ycmd/blob/418bdc86c46dba8ad4370d91a5acfa91b77979f3/ycmd/web_plumbing.py#L196)):*

I think if someone sends an invalid route the returned error will be 500 KeyError or something.
I think that's probably OK, though I was half-expecting 404.

I checked the API docs and 404 is never mentioned, so I think it's fine, While the error message will be confusing to users, it represents a programming error that only ycmd/client devs will ever see.
</blockquote></details>

Yeah, I was lazy there. It definitely will return 500 KeyError.

It's a bit awkward to differentiate `KeyError` because the path does not exist and `KeyError` because the callback has thrown that. Latter of which we do actually want as `500`.

Solving that requires something like

```python
try:
  callback = routes[path]
except KeyError:
  # return 404
else:
  callback( request, response )
```

Which is not bad, but two thoughts crossed my mind:

1. When wasa the last time you have seen a `try`/`else`?
2. We should have a test for this 404 response.
___
*[`ycmd/web_plumbing.py` line 205 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr6ay9hCjdm8wo3Uhez:-Nr6d5G60rvBJZeat2Ow:bep0xr6) ([raw file](https://github.com/ycm-core/ycmd/blob/418bdc86c46dba8ad4370d91a5acfa91b77979f3/ycmd/web_plumbing.py#L205)):*

this assignment looks fishy. It looks like you're trying to copy `response.headers` but actually you're taking a pointer to it, then mutating it on the next line.

I have no issue with mutating it (it's probably correct), but just want to check that's what you were intending.
</blockquote></details>

Yeah, that line is silly. I'd prefer

```python
response.add_header(...)
start_response( status, response.headers )
```

Mutating is fine, but as written it does look fishy.

Fixed.

bstaletic (commented):
Good news, everyone!
https://github.com/bstaletic/YouCompleteMe/actions/runs/7978280435

Vim survived this branch.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1736#-:-Nr6pmqJ0rVR_EEBuVMF:b-o0sefe)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

bstaletic (commented):
Let's see if tests will pass. If they do, and the vim tests pass as well (again), I might merge this.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1736#-:-NuKEOTF8Az8L6k4cgvY:b-tk3pg5)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[`ycmd/web_plumbing.py` line 60 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr6YGBp5Wwa4ztwXmTg:-NuJv7O7AlWfsTv1TL8z:bnzyfkr) ([raw file](https://github.com/ycm-core/ycmd/blob/418bdc86c46dba8ad4370d91a5acfa91b77979f3/ycmd/web_plumbing.py#L60)):*

> Is it essentially because you can have the same query string parameter multiple times and parse\_qa just always returns a list \[ value \] in the case of only one?

Yes.

```
>>> urllib.parse.parse_qs('a=b&a=3')
{'a': ['b', '3']}
>>> urllib.parse.parse_qs('a=b&c=d')
{'a': ['b'], 'c': ['d']}
```

Maybe we rather want index `-1`?
Should I put that in a comment?
</blockquote></details>

I have tried a few websites to see what they do regarding repeated query parameters. One took the first value, one took the last value. Leaving as is.
___
*[`ycmd/web_plumbing.py` line 196 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1736#-Nr6aHs7DQvnh3kor57n:-NuJwwMw7JKlzGz4T1OO:bjn2f2d) ([raw file](https://github.com/ycm-core/ycmd/blob/418bdc86c46dba8ad4370d91a5acfa91b77979f3/ycmd/web_plumbing.py#L196)):*

Yeah, I was lazy there. It definitely will return 500 KeyError.

It's a bit awkward to differentiate `KeyError` because the path does not exist and `KeyError` because the callback has thrown that. Latter of which we do actually want as `500`.

Solving that requires something like

```python
try:
  callback = routes[path]
except KeyError:
  # return 404
else:
  callback( request, response )
```

Which is not bad, but two thoughts crossed my mind:

1. When wasa the last time you have seen a `try`/`else`?
2. We should have a test for this 404 response.
</blockquote></details>

Implemented, with a test. Now the message in vim is

    RouteNotFound: '/the_route'

bstaletic (commented):
:lgtm:

> Reviewed 8 of 9 files at r1, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1736#-:-NuKf4fw3enYoyrp8HCH:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

bstaletic:
All review comments have been addressed. Ycmd and YCM tests are passing and a manual test with unicode identifiers passed as well. Merging.

PR #1749, opened 2024-07-31
Java: Fix gradle project root detection

puremourning:
Typical "modern" layout from `gradle init`:

- settings.gradle
- app/build.gradle
- ...

If you open app/src/java/.../Foo.java, we incorrectly pick app/ as the project root because we find first `build.gradle` then keep looking only for that file. Instead we should keep looking for _any_ gradle file.

Now, when searching for the project root, we continue to search the varioius file names, but we tag each file name with the "kind" and keep searching for other files of the same *kind* rather than just the same name.

Meanwhile, we prefer finding the gradle/maven project. The reason for this is that jdt.ls spams .project files all over the codebase which makes any project you've previously opened look like an eclipse project. This is fine if the project root was previously correctly determined and/or the projects properly imported, but means that we never self-correct. So prefer finding the non-native project types so that we pick the correct root in _most_ cases.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1749)


puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1749)* status: 0 of 2 LGTMs obtained

___
*[`ycmd/completers/java/java_completer.py` line 45 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1749#-O37YZ9g-l-MoWp0G30a:-O37YZ9g-l-MoWp0G30b:b-ol68ub) ([raw file](https://github.com/ycm-core/ycmd/blob/71e4d78cb0e359e615f5b47e64489dd5b0970090/ycmd/completers/java/java_completer.py#L45)):*
> ```Python
> PATH_TO_JAVA = None
> 
> PROJECT_FILE_TAILS = {
> ```

making this a dictionary makes the sequence undefined. Either revert to list or use OrderedDict.

puremourning (commented):
Added a test using bare grade init project

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1749#-:-O37mCeZ1LMFJTUxpAzJ:b-z8gyhy)* status: 0 of 2 LGTMs obtained

puremourning (commented):
> Reviewed 26 of 26 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1749)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

bstaletic (approved):
> Reviewed 26 of 26 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1749)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[`ycmd/completers/java/java_completer.py` line 45 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1749#-O37YZ9g-l-MoWp0G30a:-O3SUOhjAzFdXRXubNVc:b3l8ukf) ([raw file](https://github.com/ycm-core/ycmd/blob/71e4d78cb0e359e615f5b47e64489dd5b0970090/ycmd/completers/java/java_completer.py#L45)):*

done
</blockquote></details>

Iteration order of regular dictionaries is defined to be the same as insertion order, since... python 3.7. We only support python 3.8 and newer and even python 3.8 is nearing EOL.

https://mail.python.org/pipermail/python-dev/2017-December/151283.html

A list of tuples is an odd data type, when dict works. If you think we should be more explicit about ordering, then there's still `OrderedDict`.

bstaletic (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1749#-:-O5-sED49bkZn2t0JY1b:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

PR #1754, opened 2024-08-15
[READY] Add support codeAction/resolve requests

bstaletic:
So...
We advertise that we can handle CodeActions that are either missing an edit or a command (or both).
The server chooses whether it supports either of the two. If it does, it advertises itself as a "code action resolve provider".

For servers that are code action resolve providers:
- If either edit or command is missing, we need to try resolving the code action via codeAction/resolve.
  - Unless we actually got a LSP Command, in which case codeAction/resove is skipped.
- After resolving it that way we have a CodeAction in one of these forms:
  - A LSP Command
  - A LSP CodeAction with only an edit.
  - A LSP CodeAction with only a command.
  - A LSP CodeAction with both an edit and a command.
- Edits are WorkspaceEdits and can easily be converted into ycmd FixIts.
- Commands are to be executed, yielding ApplyEdits. A single ApplyEdit can be converted into a ycmd FixIt.

For servers that are not code action resolve providers, the steps are the same, but we skip the codeAction/resolve route.

One thing missing is properly defined handling of fixit resolutions that yield multiple fixits. That can happen in a few ways:

- On /resolve_fixit, by a LSP command yielding multiple ApplyEdits.
- When eagerly resolving a fixit, again by a LSP command yielding multiple ApplyEdits.
- Even if all commands always yield a single ApplyEdit, if a CodeAction has both an edit and a command, that's again two fixits.

The first two above don't seem to be used by any server ever. The LSP specs nudges servers away from doing that, but no promises. We are still not supporting any scenario where resolving a single fixit results in more than one fixit.

Another scenario that does not seem to happen:
- The server is a code action resove provider.
- The received CodeAction has neither an edit nor a command.
- After resolving, we get only a command.
- We then need to execute the command and collect ApplyEdits.

In practice, such code actions resolve to a CodeAction containing an edit.

As for the ycmd<->client side of things... it's a bit ugly on the ycmd side, but we're completely preserving the API, so clients do not need to change a thing.

Previously, clients got `{ 'fixits': [ { 'command': LSPCommand ... } ] }` for unresolved fixits.  However, we have not given any guarantees about the value of `'command'`.  We have only said that it should be returned to ycmd for the purposes of `/resolve_fixit`.  With this pull request, we need to pass the entire CodeAction, but we're still putting it in the `command` key.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1754)

### DIFF
--- a/ycmd/completers/java/java_completer.py
+++ b/ycmd/completers/java/java_completer.py
@@ -24,7 +24,6 @@ import tempfile
 import threading
 
 from ycmd import responses, utils
-from ycmd.completers.language_server import language_server_protocol as lsp
 from ycmd.completers.language_server import language_server_completer
 from ycmd.utils import LOGGER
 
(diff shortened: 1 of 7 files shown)


puremourning (approved):
:lgtm_strong: 

Thanks!

> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1754#-:-O4MtzwNA67hDhZLP-f2:bmrdz0l)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

puremourning (approved):

bstaletic (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1754#-:-O4V10Cy1m1mMRTUTS3f:bnfp4nl)* status: 1 of 2 LGTMs obtained (and 1 stale) (waiting on @puremourning)

bstaletic:
I'm finally happy with this pull request. Codecov says that I should have covered the changes better, but the overall coverage is up.

PR #1756, opened 2024-08-15
Pin go version to avoid gopls issues

puremourning:
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1756)

* latest stable go 1.23 is incompatible with our gopls version
* we can't just upgrade gopls because the new gopls includes code action command that use showDocument requests which are not supported by YCM, and gopls ignores the client capability

So to make CI green we pin the go version. Unfortunately this is temporary as users will eventually start using new go and having gopls problems.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,8 @@ jobs:
       if: matrix.benchmark == false
       uses: actions/setup-go@v5
       with:
-        go-version: stable
+        # HACK : Our gopls is not compatible with Go 1.23+
+        go-version: '<1.23'
         cache: false
     - name: Install mono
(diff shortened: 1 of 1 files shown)


puremourning:
Related gopls bug: https://github.com/golang/go/issues/68904

puremourning (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1756#-:-O4MxQMNAgw9WR0PxWh-:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @puremourning)

bstaletic:
We can close this one.
Still, thanks for trying to fix it.

PR #1760, opened 2024-09-12
[WIP/RFC] Allow code actions containing an edit and a command

### LINKED ISSUE #1692: Support a mix of edit and command in the code action response
The current ycmd implementation expects the code action response to have either edit or command, but not both. This is inconsistent with the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction) that allows both edit and command in the response. If both are supplied, the `edit` is applied first, then the `command` is executed.
https://github.com/ycm-core/ycmd/blob/d3ae5557b65a56d7e8fa2cfcf8736d161af00db3/ycmd/completers/language_server/language_server_completer.py#L2687-L2700

bstaletic:
Fixes #1692 

WIP, because still no response to https://github.com/microsoft/language-server-protocol/issues/2015
But also tests.
Previously, we could not handle two cases:

- code action containing an edit and a command
- commands producing multiple edits.

This pull request currently aims at the former. Manually tested with the Dart server and everything worked fine.

#### LSP

Assuming we get a code action with an edit and a command, the client should do the following:

1. Apply the edit.
2. Send `didChange` notification.
3. Execute the command.

Step 3 is still pretty much dark magic.

#### ycmd

Current implementation, when faced with "mixed" code actions does:

1. Convert the edit to a ycmd FixIt, but setting a new property, `keep_going` to `True`.
2. Stores the command for later.
3. Responds to the client with that one FixIt.
4. Once a client decides to apply a FixIt with `keep_going` set, after applying, it should ask ycmd for the next FixIt in the chain.
4.1. I couldn't figure out a better way than to introduce a new route - `/next_fixit`
5. Upon receiving a `/next_fixit` request, ycmd will execute the previously stored command (step 2).

Obviously, `/next_fixit` and `keep_going` are prone to bikeshedding.
I'm more concerned whether the whole idea about storing  the command for later is a good approach.
It works... but only because `:YcmCompleter FixIt` is synchronous. If it weren't, there'd be room for error.

### DIFF
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -2997,13 +2997,22 @@ class LanguageServerCompleter( Completer ):
             REQUEST_TIMEOUT_COMMAND )[ 'result' ]
 
     result = []
-    if 'edit' in code_action:
+    edit = code_action.get( 'edit' )
+    command = code_action.get( 'command' )
+    assert edit or command
+    if edit and command:
       result.append( self.CodeActionLiteralToFixIt( request_data,
                                                     code_action ) )
+      self._code_action_continuation = command
+      result[ 0 ].keep_going = True
+      return responses.BuildFixItResponse( result )
+    if edit:
+      result.append( self.CodeActionLiteralToFixIt( request_data,
+                                                    code_action ) )
+      return responses.BuildFixItResponse( result )
 
-    if 'command' in code_action:
-      assert not result, 'Code actions with edit and command is not supported.'
-      if isinstance( code_action[ 'command' ], str ):
+    if command:
+      if isinstance( command, str ):
         unresolved_command_fixit = self.CommandToFixIt( request_data,
                                                         code_action )
       else:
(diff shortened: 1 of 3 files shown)


puremourning:
> I'm more concerned whether the whole idea about storing the command for later is a good approach.

honestly, I don't love it. I would prefer to send the "next list" on the response as opaque data and have YCM send it back to us if necessary.



All In all, I'm not convinced that we need to have this complexity, given that there are still only "suggestions" that anyone will ever actually do this in a server. I would be tempted to wait until something actually uses it and see how it breaks in practice..

WDYT?

bstaletic (commented):
> I would prefer to send the "next list" on the response as opaque data and have YCM send it back to us if necessary.

Let me see if I got your idea right.
Since this pull request only implements the case where we get an edit and a command (i.e. still no support for multiple `applyEdit`s from a single command), we actually know the whole chain of fixits that we want to apply.
Therefore we can do away with the `keep_going` thing.
On one hand, that does seem more robust.
On the other, isn't that the opposite of what we are doing with `completionItem/resolve`, where YCM only gets "some number" and not the whole LSP completion item?

> I would be tempted to wait until something actually uses it and see how it breaks in practice..

Dart language server already does this.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1760#-:-OEF_ova8bS89qeowDq8:bp93yhm)* status: 0 of 2 LGTMs obtained

PR #1762, opened 2024-09-19
fix DocumentURI error for gopls@v0.16.1

WinChua:
When use YCM for golang with gopls@v0.16.1, I meet such response
```json
{"jsonrpc":"2.0","error":{"code":-32700,"message":"JSON RPC parse error: DocumentURI scheme is not \'file\': file:/data/home/winchua/test/g"},"id":1}
```
which disable YCM.

After some debugging, it's cause by [the code from gopls ](https://github.com/golang/tools/blob/master/gopls/internal/protocol/uri.go#L149)  which requires uri of local files should starts with 'file://'

So, I make a pr here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1762)

### DIFF
--- a/ycmd/completers/language_server/language_server_protocol.py
+++ b/ycmd/completers/language_server/language_server_protocol.py
@@ -753,7 +753,7 @@ def InlayHints( request_id, request_data ):
 
 
 def FilePathToUri( file_name ):
-  return urljoin( 'file:', pathname2url( file_name ) )
+  return urljoin( 'file://', pathname2url( file_name ) )
 
 
 def UriToFilePath( uri ):


bstaletic (approved):
Thanks for the pull request.
I do not understand when/how did you get this error.
Especially considering that `urljoin('file:', path)` does add the two slashes:

```
>>> urljoin('file:', '/home') == 'file:///home'
True
>>> urljoin('file://', '/home') == urljoin('file:', '/home')
True
```

Can you share the complete stderr log, so we can see what is going on?

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1762#-:-O79zj442hPBI2QbE8pj:ba3mwgd)* status: 0 of 2 LGTMs obtained (waiting on @WinChua)

WinChua:
o, I see. I compiled python from source code yestoday. [This commit ](https://github.com/python/cpython/pull/123273)
change the behavior of urljoin . 
```python
>>> urljoin("file:", "/home")
'file:/home'
```
Not so sure such behavior is expected.

bstaletic:
Since python3.13 has been released and still `urljoin('file:', '/home') == 'file:///home'`, I will close this.
@WinChua if you run into a problem again, feel free to yell at me.

PR #1764, opened 2024-10-07
Make building regex module optional

bstaletic:
Two things:

1. Make regex optional with `--no-regex`/`--regex`
2. Move `BuildRegexModule()` to the end.

Both done as a way to allow ycmd to work even if regex does not compile on some system.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1764)

### DIFF
--- a/build.py
+++ b/build.py
@@ -568,6 +568,10 @@ def ParseArguments():
   parser.add_argument( '--js-completer', action = 'store_true',
                        help = argparse.SUPPRESS )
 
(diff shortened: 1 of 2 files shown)


puremourning (approved):
:lgtm: nice, thanks!

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1764#-:-O8rpxOu80IzYDehSr2m:b4mua48)* status: 1 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1765, opened 2024-10-10
Upgrade to unicode 16

bstaletic:
No changes to the algorithm, but some combining mark codepoints got a non-zero "indic conjunct break property".
Specification is no longer a series of pdfs, so I updated those links too.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1765)


bstaletic:
@puremourning approved this in gitter, so I'm merging.

PR #1766, opened 2024-10-13
[RFC] Add support for file operations in CodeActions

bstaletic:
Tests will certainly fail, but I'd like to gather feedback before proceeding with this work.

Eventually fixes https://github.com/ycm-core/YouCompleteMe/issues/4267

And of course, JDT was a special snowflake and had to break the protocol again.

### DIFF
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -3590,8 +3590,8 @@ def WorkspaceEditToFixIt( request_data,
   if not workspace_edit:
     return None
 
+  chunks = []
   if 'changes' in workspace_edit:
-    chunks = []
     # We sort the filenames to make the response stable. Edits are applied in
     # strict sequence within a file, but apply to files in arbitrary order.
     # However, it's important for the response to be stable for the tests.
@@ -3599,12 +3599,26 @@ def WorkspaceEditToFixIt( request_data,
       chunks.extend( TextEditToChunks( request_data,
                                        uri,
                                        workspace_edit[ 'changes' ][ uri ] ) )
-  else:
-    chunks = []
+  if 'documentChanges' in workspace_edit:
     for text_document_edit in workspace_edit[ 'documentChanges' ]:
-      uri = text_document_edit[ 'textDocument' ][ 'uri' ]
-      edits = text_document_edit[ 'edits' ]
-      chunks.extend( TextEditToChunks( request_data, uri, edits ) )
+      kind = text_document_edit.get( 'kind', '' )
+      if 'edits' in text_document_edit:
+        uri = text_document_edit[ 'textDocument' ][ 'uri' ]
+        edits = text_document_edit[ 'edits' ]
+        chunks.extend( TextEditToChunks( request_data, uri, edits ) )
+      elif kind == 'rename':
+        chunks.append(
+            responses.RenameChunk(
+              old_filepath = lsp.UriToFilePath( text_document_edit[ 'oldUri' ] ),
+              new_filepath = lsp.UriToFilePath( text_document_edit[ 'newUri' ] ) ) )
+      elif kind == 'delete':
+        chunks.append(
+            responses.DeleteChunk(
+              lsp.UriToFilePath( text_document_edit[ 'uri' ] ) ) )
+      elif kind == 'create':
+        chunks.append(
+            responses.CreateChunk(
+              lsp.UriToFilePath( text_document_edit[ 'uri' ] ) ) )
   return responses.FixIt(
     responses.Location( request_data[ 'line_num' ],
                         request_data[ 'column_num' ],
--- a/ycmd/completers/language_server/language_server_protocol.py
+++ b/ycmd/completers/language_server/language_server_protocol.py
@@ -307,7 +307,10 @@ def Initialize( request_id,
       'didChangeWatchedFiles': {
         'dynamicRegistration': True
       },
-      'workspaceEdit': { 'documentChanges': True, },
+      'workspaceEdit': {
+        'resourceOperations': [ 'create', 'rename', 'delete'],
+        'documentChanges': True,
+      },
       'symbol': {
         'symbolKind': {
           'valueSet': list( range( 1, len( SYMBOL_KIND ) ) ),
--- a/ycmd/responses.py
+++ b/ycmd/responses.py
@@ -268,6 +268,22 @@ class FixIt:
     self.kind = kind
 
 
+class DeleteChunk:
+  def __init__( self, filepath ):
+    self.filepath = filepath
+    self.kind = 'delete'
+
+class CreateChunk:
+  def __init__( self, filepath ):
+    self.filepath = filepath
+    self.kind = 'create'
+
+class RenameChunk:
+  def __init__( self, old_filepath, new_filepath ):
+    self.old_filepath = old_filepath
+    self.new_filepath = new_filepath
+    self.kind = 'rename'
+
 class FixItChunk:
   """An individual replacement within a FixIt (aka Refactor)"""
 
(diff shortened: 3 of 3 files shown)


puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1766)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

___
*[`ycmd/responses.py` line 334 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1766#-O95Ao9TABMYU_v2zC1b:-O95Ao9TABMYU_v2zC1c:b4yo9gv) ([raw file](https://github.com/ycm-core/ycmd/blob/86df022691890b27c93fc04312615cf527f59acd/ycmd/responses.py#L334)):*
> ```Python
> 
>   def BuildFixitChunkData( chunk ):
>     if hasattr( chunk, 'replacement_text' ):
> ```

minor, but maybe we should actually put the "to ycmd protocol" as a instance method on the chunk type.
___
*[`ycmd/completers/language_server/language_server_completer.py` line 3593 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1766#-O95CU0QD2B6ge1egDHA:-O95CU0QD2B6ge1egDHB:bh7evcf) ([raw file](https://github.com/ycm-core/ycmd/blob/86df022691890b27c93fc04312615cf527f59acd/ycmd/completers/language_server/language_server_completer.py#L3593)):*
so previously we only ever processed one of these 'changes' vs 'documentChanges'.

Now we seem to include _both_ if present.

This seems against the protocol which says:

> A workspace edit represents changes to many resources managed in the workspace. The edit should either provide `changes` or `documentChanges`. ***If the client can handle versioned document edits and if `documentChanges` are present, the latter are preferred over `changes`.***

So I think that we should prefer `documentChanges`.
___
*[`ycmd/completers/language_server/language_server_completer.py` line 3612 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1766#-O95CCEi4maPVeLyBiIU:-O95CCEi4maPVeLyBiIV:b-10wdvl) ([raw file](https://github.com/ycm-core/ycmd/blob/86df022691890b27c93fc04312615cf527f59acd/ycmd/completers/language_server/language_server_completer.py#L3612)):*
> ```Python
>         chunks.append(
>             responses.RenameChunk(
>               old_filepath = lsp.UriToFilePath( text_document_edit[ 'oldUri' ] ),
> ```

If it recall correctly, this throws `InvalidUriException` in case that the server sends some bullshit URI like `jdt://` and we have to catch it everywhere.
___
*[`ycmd/completers/language_server/language_server_protocol.py` line 310 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1766#-O95DYp77ZTE6jhfcxfF:-O95DYp77ZTE6jhfcxfG:bpi8kk8) ([raw file](https://github.com/ycm-core/ycmd/blob/86df022691890b27c93fc04312615cf527f59acd/ycmd/completers/language_server/language_server_protocol.py#L310)):*
> ```Python
>         'dynamicRegistration': True
>       },
>       'workspaceEdit': {
> ```

we should consider specifying our `FailureHandlingKind`

 ```
	/**
	 * The failure handling strategy of a client if applying the workspace edit
	 * fails.
	 *
	 * @since 3.13.0
	 */
	failureHandling?: FailureHandlingKind;
```

I guess `	export const Abort: FailureHandlingKind = 'abort';` is what we have.

**Review by bstaletic — COMMENTED** [review]
Regarding "options", like these:

1. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#createFileOptions
2. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#renameFileOptions
3. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#deleteFileOptions

Should we just shove `options` with the chunks and pass them on to YCM?

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1766#-:-O963QHB511DCyw4gKht:b-9l5b0x)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

___
*[`ycmd/responses.py` line 334 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1766#-O95Ao9TABMYU_v2zC1b:-O961fwG3JtvMPJLXANb:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/86df022691890b27c93fc04312615cf527f59acd/ycmd/responses.py#L334)):*

minor, but maybe we should actually put the "to ycmd protocol" as a instance method on the chunk type.
</blockquote></details>

Done.
___
*[`ycmd/completers/language_server/language_server_completer.py` line 3593 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1766#-O95CU0QD2B6ge1egDHA:-O961inK9LG4yhWR_Xjo:b-caf1wg) ([raw file](https://github.com/ycm-core/ycmd/blob/86df022691890b27c93fc04312615cf527f59acd/ycmd/completers/language_server/language_server_completer.py#L3593)):*

so previously we only ever processed one of these 'changes' vs 'documentChanges'.

Now we seem to include _both_ if present.

This seems against the protocol which says:

> A workspace edit represents changes to many resources managed in the workspace. The edit should either provide `changes` or `documentChanges`. ***If the client can handle versioned document edits and if `documentChanges` are present, the latter are preferred over `changes`.***

So I think that we should prefer `documentChanges`.
</blockquote></details>

This is against the protocol, like I mentioned here:
https://github.com/ycm-core/YouCompleteMe/issues/4267#issuecomment-2408831137

If you really think we should prefer `documentChanges`, I will do it.

JDT seems to send an empty `changes` and a populated `documentChanges`.
___
*[`ycmd/completers/language_server/language_server_completer.py` line 3612 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1766#-O95CCEi4maPVeLyBiIU:-O9628sR-gzk850v0bIm:b-ck7vtl) ([raw file](https://github.com/ycm-core/ycmd/blob/86df022691890b27c93fc04312615cf527f59acd/ycmd/completers/language_server/language_server_completer.py#L3612)):*

If it recall correctly, this throws `InvalidUriException` in case that the server sends some bullshit URI like `jdt://` and we have to catch it everywhere.
</blockquote></details>

JDT has made that opt-in long ago.
___
*[`ycmd/completers/language_server/language_server_protocol.py` line 310 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1766#-O95DYp77ZTE6jhfcxfF:-O963-hN3CsGaMisk-UQ:b-896fix) ([raw file](https://github.com/ycm-core/ycmd/blob/86df022691890b27c93fc04312615cf527f59acd/ycmd/completers/language_server/language_server_protocol.py#L310)):*

we should consider specifying our `FailureHandlingKind`

 ```
	/**
	 * The failure handling strategy of a client if applying the workspace edit
	 * fails.
	 *
	 * @since 3.13.0
	 */
	failureHandling?: FailureHandlingKind;
```

I guess `	export const Abort: FailureHandlingKind = 'abort';` is what we have.
</blockquote></details>

Done.

puremourning (commented):
> Should we just shove `options` with the chunks and pass them on to YCM?

yeah why not. we might as well just use the LSP syntax and semantics for these; why swim against the tide.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1766#-:-O96La-a8Hhvh5AfppXN:bjq14jm)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

puremourning (commented):
actually, looking at the LSP spec, it's bonkers:

```
	/**
	 * Overwrite existing file. Overwrite wins over `ignoreIfExists`
	 */
	overwrite?: boolean;

	/**
	 * Ignore if exists.
	 */
	ignoreIfExists?: boolean;
```

erm... don't they mean the same thing!

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1766#-:-O96O12q3muKTIkQW_jp:b-m99x6d)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

bstaletic (commented):
No, they mean opposite things. If `overwrite`, then do not care that `$FILE` is there, just overwrite it. If `ignoreIfExists`, ignore *the operation* and leave that file alone.

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1766#-:-O96TEgNAOPqgiotsgF3:b-fippig)* status: 0 of 2 LGTMs obtained (waiting on @puremourning)

puremourning (commented):
yeah, but I don't see how that's different from having just one of them. What am I missing!?

> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1766#-:-O96eM9U7qw3kao8Ntvu:bdd7f5u)* status: 0 of 2 LGTMs obtained (waiting on @bstaletic)

PR #1772, opened 2024-12-02
Upgrades for jdt.ls, gopls and tsserver

bstaletic:
These three did not need any changes in the tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1772)


puremourning:
LGTM if the tests pass

PR #1773, opened 2024-12-12
[READY] Correct where WithRetry decorator is applied

bstaletic:
A TestCase.subTest() context manager will swallow exceptions, so applying @WithRetry on the entire test does not work. Instead, it needs to be applied on whatever the subTest is doing.

I have not run the tests manually. I'm hoping CI will be merciful.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1773)


puremourning (approved):
LGTM

PR #1777, opened 2024-12-26
Fix GoToDefinition issues in non-compliant repositories lacking go.sum in GOMODCACHE

ZhuzhuNo3:
This commit addresses the problem where the GoToDefinition feature fails to execute in certain repositories located within the GOMODCACHE path. These specific repositories were not compliant due to the absence of a go.sum file, which is essential for ensuring module integrity and correct dependency resolution.

Add `let g:ycm_gopls_excluded_workspace_paths = null` or `let g:ycm_gopls_excluded_workspace_paths = ['PATH_TO_GOMODCACHE', 'NO_GOSUM_REPO_PATH']` to enable this change.

### DIFF
--- a/ycmd/completers/go/go_completer.py
+++ b/ycmd/completers/go/go_completer.py
@@ -15,6 +15,8 @@
 # You should have received a copy of the GNU General Public License
 # along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
 
+from pathlib import Path
+import subprocess
 import json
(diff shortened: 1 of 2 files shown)


ZhuzhuNo3:
Reproduction Steps: In a personal Golang project, use GoToDefinition to navigate to a cached library that does not include go.sum. In the newly opened file, GoToDefinition cannot be used.

PR #1788, opened 2025-07-14
Support jumping to jdt:// URIs with GoTo

christolis:
This pull request addresses a limitation in ycmd's integration with jdt.ls, where jumping to definitions in external JAR files (e.g., project dependencies) resulted in errors like "Cannot jump to location." (look at the before vs after videos below). By enabling the `classFileContentsSupport` extended capability during initialization, ycmd signals to jdt.ls that it can handle jdt:// URIs.

More information on the commit messages.

<b>Note</b>! This PR should be merged alongside [#4302](https://github.com/ycm-core/YouCompleteMe/pull/4302) on the [YouCompleteMe](https://github.com/ycm-core/YouCompleteMe) repository.

# Before vs After

https://github.com/user-attachments/assets/5d6d4e92-d52d-45a7-a964-ca0b1d367c42

https://github.com/user-attachments/assets/960465b6-f447-4882-bafd-9c99d0e65708

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1788)

### DIFF
--- a/ycmd/completers/java/java_completer.py
+++ b/ycmd/completers/java/java_completer.py
@@ -25,7 +25,9 @@ import threading
 from collections import OrderedDict
 
 from ycmd import responses, utils
-from ycmd.completers.language_server import language_server_completer
+from ycmd.completers.java import java_utils
+from ycmd.completers.language_server import ( language_server_protocol as lsp,
+                                             language_server_completer )
 from ycmd.utils import LOGGER
 
 NO_DOCUMENTATION_MESSAGE = 'No documentation available for current context'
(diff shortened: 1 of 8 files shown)


puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1788)* status: 0 of 2 LGTMs obtained

___
*[`ycmd/tests/language_server/language_server_completer_test.py` line 97 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1788#-OVEYLeq6vBHcZ1YVzC4:-OVEYLeq6vBHcZ1YVzC5:bhle7dv) ([raw file](https://github.com/ycm-core/ycmd/blob/40dc16e51910530252a66efafd3c8cca07e9fe26/ycmd/tests/language_server/language_server_completer_test.py#L97)):*
> ```Python
> 
> 
> class MockJavaCompleter( MockCompleter, JavaCompleter ):
> ```

these tests don't belong here. should be testing this feature in tests/java/ by actually triggering the code with a real completion engine rather than mocks.
___
*[`ycmd/completers/java/java_completer.py` line 333 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1788#-OVEUrVi6pvES-g6p-H7:-OVEUrVi6pvES-g6p-H8:busbc66) ([raw file](https://github.com/ycm-core/ycmd/blob/40dc16e51910530252a66efafd3c8cca07e9fe26/ycmd/completers/java/java_completer.py#L333)):*
> ```Python
>     return {
>       'bundles': self._bundles,
>       'capabilities': {
> ```

this seems... strange. why would we send definitionProvider: true to the server? and why in this way? what motivated this part of the change?
___
*[`ycmd/completers/java/java_completer.py` line 673 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1788#-OVEW66GBFMlCAvc_giW:-OVEW66GBFMlCAvc_giX:b-kzzpbb) ([raw file](https://github.com/ycm-core/ycmd/blob/40dc16e51910530252a66efafd3c8cca07e9fe26/ycmd/completers/java/java_completer.py#L673)):*
> ```Python
> 
> 
>   def GoTo( self, request_data, handlers ):
> ```

is this mostly a copy/paste of the base implementation? it doesn't seem completely focussed on supporting different URL schemes.

Better to add a hook in the base class rather than just copy paste the implementation.
___
*[`ycmd/completers/java/java_completer.py` line 685 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1788#-OVEV10B-_U9MUyUUrzn:-OVEV10B-_U9MUyUUrzo:b-vb38vy) ([raw file](https://github.com/ycm-core/ycmd/blob/40dc16e51910530252a66efafd3c8cca07e9fe26/ycmd/completers/java/java_completer.py#L685)):*
> ```Python
>     # flake8 doesn't like long lines so we have to split it up
>     def HandlerCondition( result, request_data ):
>       return result and \
> ```

prefer parentheses or split it into 2 ifs

```
return ( 
    result
    and lip._CursorInsideLocation(....)
)
```
___
*[`ycmd/completers/java/java_completer.py` line 686 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1788#-OVEVId9B95EeKOu2iF4:-OVEVId9B95EeKOu2iF5:bhv9oty) ([raw file](https://github.com/ycm-core/ycmd/blob/40dc16e51910530252a66efafd3c8cca07e9fe26/ycmd/completers/java/java_completer.py#L686)):*
> ```Python
>     def HandlerCondition( result, request_data ):
>       return result and \
>              not language_server_completer._CursorInsideLocation( request_data,
> ```

don't use internal functions. If it needs to be public make it so by renaming to remove the leading `_`
___
*[`ycmd/completers/java/java_completer.py` line 700 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1788#-OVEVsLa7Lta5QNfiSyf:-OVEVsLa7Lta5QNfiSyg:b-3lkyqq) ([raw file](https://github.com/ycm-core/ycmd/blob/40dc16e51910530252a66efafd3c8cca07e9fe26/ycmd/completers/java/java_completer.py#L700)):*
> ```Python
>       raise RuntimeError( 'Cannot jump to location' )
> 
>     first_result = result[ 0 ]
> ```

why is the first result special? why is GoTo special?

we should be handling any URI provided by JDT as jdt://. there are many other requests which return file paths which might not be file://

we need to not special case GoTo and rather implement handling for jdt:// URIs more generally.
___
*[`ycmd/completers/java/java_completer.py` line 727 at r1](https://reviewable.io/reviews/ycm-core/ycmd/1788#-OVEWOoZ75LpjKZqRqrt:-OVEWOoZ75LpjKZqRqru:b8jvwix) ([raw file](https://github.com/ycm-core/ycmd/blob/40dc16e51910530252a66efafd3c8cca07e9fe26/ycmd/completers/java/java_completer.py#L727)):*
> ```Python
> 
> 
>   def GetClassFileContents( self, request_data ):
> ```

I would have expected this to be implemented as a completer subcommand that takes the URI as an argument. We have a whole infrastructure for that, and in the YCM client we have the ability to get "raw" output from the command request for these sorts of things. 


See for example how I hacked some Rust-specific stuff here: https://github.com/puremourning/ycmd-1/commit/961926872a02b4403e683cd97a7cab949d409cb0#diff-b59e92be42fece00d4f5c415d99c3a53d145a7f4a816cb6f7839481830dbf3c8

and in the client, there's the command_request and various ways that it's used.

PR #1791, opened 2025-09-11
Use tsserver to drive vue-typescript-plugin (fix Vue file completion)

Cheaterman:
Everything in the title ; as it currently is, completion in `*.vue` files is not working - not at all out of the box, and very broken if trying to use modern `vue-language-server` the same way one used to use `volar` or `vls` - I'm not 100% sure but despite it being the same project as `volar`/`vls` (or a fork), I suspect it really expects to be driven through `tsserver` nowadays - this seems to work perfectly without the user even needing to change anything in their `.vimrc`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1791)

EDIT: I'm sure in an ideal world there would be a fully separate `build.py` argument for Vue, so that we wouldn't need to "pollute" the `tsserver/package.json` file with Vue-specific stuff, but I figured it might be overkill and doesn't make a whole lot of difference anyway (~4MB in `node_modules`).

### DIFF
--- a/third_party/tsserver/package.json
+++ b/third_party/tsserver/package.json
@@ -1,6 +1,8 @@
 {
   "description": "ycmd tsserver runtime area with required typescript version and plugins",
   "dependencies": {
-    "typescript": "5.7.2"
+    "@vue/language-server": "^3.0.7",
+    "@vue/typescript-plugin": "^3.0.7",
+    "typescript": "^5.9.2"
   }
 }
--- a/ycmd/completers/vue/hook.py
+++ b/ycmd/completers/vue/hook.py
@@ -1,0 +1,26 @@
+# Copyright (C) 2020 ycmd contributors
+#
+# This file is part of ycmd.
+#
+# ycmd is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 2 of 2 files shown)


Cheaterman:
I think the `package.json` in the lsp-examples repo needs the very same changes I put here, lemme take a look.

EDIT: So, the `snippet.vim` seems no longer necessary ; I didn't need to edit anything in the example other than adding the missing `tsconfig.json` - VLS is now dead in favor of `@vue/typescript-plugin`, including for non-typescript-enabled codebases ; see my before and after screenshots attached, and the associated PR to the `lsp-examples` repo [here](https://github.com/ycm-core/lsp-examples/pull/54).

Before :
scrot_006

After :
scrot_009

<details>
  <summary>More screenshots with extra usage examples</summary>
EDIT2 : Just one more screenshot for good luck :
scrot_010

EDIT3: And in fact a couple more just to demonstrate that it works both with TypeScript enabled and disabled :
Auto-completing with TypeScript disabled (same as above screenshot, except no `lang="ts"`) :
scrot_011

Making an invalid TypeScript call to `String.prototype.replace()` doesn't show any issues in JS mode :
scrot_012

But obviously TypeScript isn't happy about it :
scrot_013
</details>

Thanks for your patience :grin:

PR #1794, opened 2025-10-09
Update gopls; 0.16 was incompatible with go 1.25

sjakobs:
update to gopls 0.20.0

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1794)


puremourning:
I fixed the test failure; gopls now supported TypeHierarchy and our tests were there to remind us to test it, but like I CBA so let's just move on.

PR #1809, opened 2026-01-11
Add feature to define LSP settings in g:ycm_language_server

Felixoid:
In https://github.com/ycm-core/YouCompleteMe/issues/4272, I asked if it's possible to define global LSP settings in `.vimrc`

Here's the patch that adds this feature, so the `settings` option will be global with overrides from `.ycm_extra_conf.py`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1809)

### DIFF
--- a/ycmd/completers/language_server/generic_lsp_completer.py
+++ b/ycmd/completers/language_server/generic_lsp_completer.py
@@ -16,6 +16,7 @@
 # along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
 
 import string
+from copy import deepcopy
 from ycmd import responses, utils
 from ycmd.completers.language_server import language_server_completer
 
@@ -131,6 +132,6 @@ class GenericLSPCompleter( language_server_completer.LanguageServerCompleter ):
 
 
   def DefaultSettings( self, request_data ):
-    settings = super().DefaultSettings( request_data ).copy()
-    return utils.UpdateDict( settings,
-                             self._server_settings.get( 'settings', {} ) )
+    """ Override parent method to provide default
+    settings from server config. Return a copy to avoid mutation. """
+    return deepcopy( self._server_settings.get( 'settings', {} ) )
--- a/ycmd/completers/language_server/generic_lsp_completer.py
+++ b/ycmd/completers/language_server/generic_lsp_completer.py
@@ -132,5 +132,5 @@ class GenericLSPCompleter( language_server_completer.LanguageServerCompleter ):
 
   def DefaultSettings( self, request_data ):
     settings = super().DefaultSettings( request_data ).copy()
-    settings.update( self._server_settings.get( 'settings', {} ) )
-    return settings
+    return utils.UpdateDict( settings,
+                             self._server_settings.get( 'settings', {} ) )
--- a/ycmd/completers/language_server/generic_lsp_completer.py
+++ b/ycmd/completers/language_server/generic_lsp_completer.py
@@ -128,3 +128,9 @@ class GenericLSPCompleter( language_server_completer.LanguageServerCompleter ):
   def GetTriggerCharacters( self, server_trigger_characters ):
     return self._server_settings.get( 'triggerCharacters',
                                       server_trigger_characters )
+
+
+  def DefaultSettings( self, request_data ):
+    settings = super().DefaultSettings( request_data ).copy()
+    settings.update( self._server_settings.get( 'settings', {} ) )
+    return settings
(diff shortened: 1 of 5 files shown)


ycmd/completers/language_server/generic_lsp_completer.py
```diff
@@ -128,3 +128,9 @@ def WorkspaceConfigurationResponse( self, request ):
   def GetTriggerCharacters( self, server_trigger_characters ):
     return self._server_settings.get( 'triggerCharacters',
                                       server_trigger_characters )
+
+
+  def DefaultSettings( self, request_data ):
+    settings = super().DefaultSettings( request_data ).copy()
+    settings.update( self._server_settings.get( 'settings', {} ) )
```
puremourning: I think this shallow clone and shallow update is problematic. In particular I think that a lot of language servers have deeply nested configuration, e.g. see https://rust-analyzer.github.io/book/configuration.html and https://github.com/ycm-core/YouCompleteMe/wiki/FAQ#how-do-i-set-java-options

I think that settings from extra cons and ycm_language_server cannot set keys lower down than the top level without overwriting each other.

indeed I thin your example in the readme has this exact problem, and the tests don't disprove it.

I think we  should use `utils.UpdateDict` which is a recursive `dict.update` and add tests to prove it works.

puremourning (changes requested):
Thanks for the PR!

I think this is OK and inline with how our built-in completes can set default config, so I'm happy with the general approach, just there's quite a big bug with the implementation doing a shallow copy and only allowing limited layering.

ycmd/completers/language_server/generic_lsp_completer.py
Felixoid: Yes, indeed. And `dict.update` was also used in `_GetSettingsFromExtraConf`, so the new test caught it too.

ycmd/completers/language_server/language_server_completer.py
```diff
@@ -1901,7 +1901,8 @@ def _GetSettingsFromExtraConf( self, request_data ):
 
       # Merge any user-supplied 'ls' settings with the defaults
       if 'ls' in user_settings:
-        merged_ls_settings.update( user_settings[ 'ls' ] )
+        merged_ls_settings = utils.UpdateDict( merged_ls_settings,
```
puremourning: the assignment is redundant, as updatedict updates merged_ls_settings in place

ycmd/completers/language_server/generic_lsp_completer.py
```diff
@@ -128,3 +128,9 @@ def WorkspaceConfigurationResponse( self, request ):
   def GetTriggerCharacters( self, server_trigger_characters ):
     return self._server_settings.get( 'triggerCharacters',
                                       server_trigger_characters )
+
+
+  def DefaultSettings( self, request_data ):
+    settings = super().DefaultSettings( request_data ).copy()
```
puremourning: I think we can do without the copy() here. the contract of DeafultSettings is that it returns a new dict, which it does in all cases so far. Moreover, the copy is shallow, so it only partially protects us from scribbles (there's actually a possible scribble bug in java_completer which returns `self._bundles` ref in the DefaultSettings dict, but I'm not super worried about that.

I think we can basically do: `return utils.UpdateDict( super().DefaultSettings( request_data ), ... )` here

puremourning (approved):
thanks! just a couple of nits actually.

ycmd/completers/language_server/generic_lsp_completer.py
puremourning: haha! thanks. hoist by my own petard...

ycmd/completers/language_server/generic_lsp_completer.py
Felixoid: Leaving it as `copy` would keep us on the safe side. Just in case `DefaultSettings` would change to something else.

Maybe the right way would be just returning `self._server_settings.get( 'settings', {} ).copy()`?

Felixoid: Well, rather `deepcopy` to preserve the original settings

puremourning (commented):
:lgtm:

> @puremourning reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1809#-:-OitBniBAT6eFgi_NVGR:bnfp4nl)* status: 1 of 2 LGTMs obtained (waiting on @Felixoid).

PR #1813, opened 2026-02-14
Define completer settings in $VIMRC

Felixoid:
This is a continuation of #1809 to define global settings for predefined completers.

`g:ycm_completer_settings` allows us to define the `ls` language server settings globally for any repository without repeating `.ycm_extra_conf.py` in every repository.

Tested it as the following snippet in a `~/.vimrc`, and checked `:YcmDebugInfo`

```vim
let g:ycm_completer_settings = {
\  'rust': {
\    'diagnostics': {
\      'disabled': [
\        'inactive-code',
\      ]
\    },
\  },
\}
```

```
Printing YouCompleteMe debug information...
-- Resolve completions: Up front
-- Client logfile: /tmp/ycm_mgn2hki1.log
-- Server Python interpreter: /usr/bin/python3
-- Server Python version: 3.14.2
-- Server has Clang support compiled in: False
-- Clang version: None
-- No extra configuration file found
-- Rust completer debug information:
--   Rust Language Server running
--   Rust Language Server process ID: 1619087
.....
--   Rust Language Server Settings: {
--   "diagnostics": {
--     "disabled": [
--       "inactive-code"
--     ]
--   }
-- }
--   Rust Language Server Project State: ready
```

If it had a green light, I'd like to prepare documentation for the YouCompleteMe repo as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1813)

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,8 @@ After it starts up, ycmd will _delete_ the settings file you provided after
 it reads it.
 
 The settings file is something your editor should produce based on values your
-user has configured. There's also an extra file (`.ycm_extra_conf.py`) your user
+user has configured. There's also `completer_settings` configuration parameter and
+an extra file (`.ycm_extra_conf.py`) your user
 is supposed to provide to configure certain semantic completers. More
 information on it can also be found in the [corresponding section of YCM's _User
 Guide_][extra-conf-doc].
@@ -269,6 +270,33 @@ def Settings( **kwargs ):
              # `config_sections` is not used for java...
              'config_sections': { 'section0': {} }
 ```
+
+#### Configuring Predefined Completers Globally
+
+You can configure global settings for ycmd's predefined completers (Java, Go,
+Rust, C-family) using the `g:ycm_completer_settings` option (analogue to `ls` key
+in `Settings()` function from `.ycm_extra_conf.py` for a completer):
+
+```vim
+let g:ycm_completer_settings = {
+\  'go': {
+\    'hoverKind': 'SynopsisDocumentation'
+\  },
+\  'rust': {
+\    'diagnostics': {
+\      'disabled': [
+\        'inactive-code',
+\      ]
+\    },
+\  }
+\}
+```
+
+**Supported keys**: `java`/`jdtls`, `go`/`gopls`, `rust`/`rust-analyzer`,
+`cpp`/`c`/`objc`/`objcpp`/`cuda`/`clangd`
+
+**Settings priority**: hardcoded defaults < `g:ycm_completer_settings` <
+`.ycm_extra_conf.py`
 
 ##### `language_server` configuration
 
@@ -385,7 +413,7 @@ def Settings( **kwargs ):
 ```
 
 A number of LSP completers are currently supported without `language_server`,
-usch as:
+such as:
 
 - Java
 - Rust
--- a/README.md
+++ b/README.md
@@ -292,11 +292,11 @@ let g:ycm_completer_settings = {
 \}
 ```
 
-**Supported keys**: `java`/`jdtls`, `go`/`gopls`, `rust`/`rust-analyzer`,
-`cpp`/`c`/`objc`/`objcpp`/`cuda`/`clangd`
+**Supported keys**: `java`, `go`, `rust`, `cpp`/`c`/`objc`/`objcpp`/`cuda`.
 
-**Settings priority**: hardcoded defaults < `g:ycm_completer_settings` <
-`.ycm_extra_conf.py`
+**Settings priority**: hardcoded defaults have the lowest priorities,
+then to set user's defaults `g:ycm_completer_settings`, and finally the
+settings from the `ls` key returned by the `Settings()` function in `.ycm_extra_conf.py`
 
 ##### `language_server` configuration
 
--- a/ycmd/completers/language_server/language_server_completer.py
+++ b/ycmd/completers/language_server/language_server_completer.py
@@ -1880,18 +1880,68 @@ class LanguageServerCompleter( Completer ):
     return {}
 
 
+  def GetBinaryName( self ):
+    """Returns the binary/alias name for this completer to use as a lookup key.
+
+    Subclasses should override this to return the server binary name
+    (e.g., 'gopls', 'rust-analyzer', 'jdtls', 'clangd').
+
+    Returns None by default (no alias support).
+    """
+    return None
+
+
+  def GetCompleterSettingsLookupKeys( self ):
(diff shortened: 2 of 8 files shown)


ycmd/completers/language_server/language_server_completer.py
```diff
@@ -1880,18 +1880,68 @@ def DefaultSettings( self, request_data ):
...
-    # If there is no extra-conf, the total settings are just the defaults:
+    # Layer 2: Merge with global settings from g:ycm_completer_settings
+    lookup_keys = self.GetCompleterSettingsLookupKeys()
+    completer_settings = self.user_options.get( 'completer_settings', {} )
+
+    # Find first matching key and detect duplicates
+    matched_keys = [ key for key in lookup_keys if key in completer_settings ]
+    if matched_keys:
+      if len( matched_keys ) > 1:
+        LOGGER.warning( 'Multiple settings found for %s completer: %s. '
```
Felixoid: This remains unnoticed, because it ends up only in an stderr log. I'll try to find how to improve it.

Felixoid:
Hey-hey, @puremourning, does it fit the code base?

It looks like a missing feature to configure global settings for the completers using `$VIMRC`.

ycmd/completers/language_server/language_server_completer.py
puremourning: I think we can just write it to the log and rely on the filetypes only for matching, as this is what ~100% of Settings() implementations do.

puremourning (changes requested):
Thanks. I think I can accept this feature, but I think that it might be a bit more complicated than necessary. in Settings, we just rely on typically `if kwargs[ 'language' ] == '<filetype>'` so lets use that as the key in the dictionary. we know the filetypes of the source file that triggersd the LS to start and we rely on that elsewhere. Using the filetypes (list) in sequence  is well-defined. In vim, the first filetype in the list is primary, and the others are subordinate  (as in `typescript.react` or something) but generally there's only one filetype for _most_ cases anyway.

> @puremourning made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1813#-:-Om-B-1c3SVJQkhHbh9s:b1snpod)* status: 0 of 2 LGTMs obtained (waiting on Felixoid).

___
*[`README.md` line 276 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1813#-Om-AFOy6seL4remkc3q:-Om-AFOy6seL4remkc3r:b-z70u6k) ([raw file](https://github.com/ycm-core/ycmd/blob/4f677d720090ee39245648b632affa9ae41b7896/README.md#L276)):*
> ```Markdown
> #### Configuring Predefined Completers Globally
> 
> You can configure global settings for ycmd's predefined completers (Java, Go,
> ```

does this not work also for generic LSP completer?
___
*[`README.md` line 295 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1813#-Om-A1Hf8W_NMrjHJlaz:-Om-A1HgAg2u8vOLnqSf:b-j5f11r) ([raw file](https://github.com/ycm-core/ycmd/blob/4f677d720090ee39245648b632affa9ae41b7896/README.md#L295)):*
> ```Markdown
> ```
> 
> **Supported keys**: `java`/`jdtls`, `go`/`gopls`, `rust`/`rust-analyzer`,
> ```

I don't think we should have multiple ways here. I think the key should be the 'filetype' (as in the `filetype` argument passed to `Setttings`)
___
*[`README.md` line 298 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1813#-Om-A9sw30uJ29LF_rh8:-Om-A9sw30uJ29LF_rh9:bkajz1n) ([raw file](https://github.com/ycm-core/ycmd/blob/4f677d720090ee39245648b632affa9ae41b7896/README.md#L298)):*
> ```Markdown
> `cpp`/`c`/`objc`/`objcpp`/`cuda`/`clangd`
> 
> **Settings priority**: hardcoded defaults < `g:ycm_completer_settings` <
> ```

write this in prose, as `<` is ambiguous in my opinion 
___
*[`ycmd/completers/cpp/clangd_completer.py` line 242 at r2](https://reviewable.io/reviews/ycm-core/ycmd/1813#-Om-AQQe0fbfRuswWbxI:-Om-AQQe0fbfRuswWbxJ:b-gz0ttw) ([raw file](https://github.com/ycm-core/ycmd/blob/4f677d720090ee39245648b632affa9ae41b7896/ycmd/completers/cpp/clangd_completer.py#L242)):*
> ```Python
> 
> 
>   def GetBinaryName( self ):
> ```

I don't like this. I prefer to use `filetype`

ycmd/completers/language_server/language_server_completer.py
Felixoid: > I don't like this. I prefer to use `filetype`

When I thought of this, the first approach was the same. But then I found that `clangd` is used in many file types. Sure, it can be simplified to filetypes, but on the other hand it will cause duplicates in the config.

If you say it's a blocker, I'll remove the whole `GetBinaryName` logic. 

Wdyt?

puremourning: in `g:ycm_semantic_triggers` we have a scheme were the keys are actually a list of filetypes. if you really think it's necessary to specify: 

```
"c, cpp": { "lfoo": "bar" }
```

rather than:

```
let c = { "foo": "bar" }
let g:ycm_completer_settings = {
   "c": c,
  "cpp": c
}
```

then maybe do that but I think this is good enough, as I expect this feature to be rarely used.

Felixoid: Yeah, I thought of `let c = { "foo": "bar" }` in sense of DRY.

Looks good enough :+1:

README.md
```diff
@@ -270,6 +271,33 @@ def Settings( **kwargs ):
              'config_sections': { 'section0': {} }
 ```
 
+#### Configuring Predefined Completers Globally
+
+You can configure global settings for ycmd's predefined completers (Java, Go,
```
Felixoid, line 276: > does this not work also for generic LSP completer?

That's a surprise improvement. I think, for this case, I'll keep the config warning. Now we can have `g:ycm_language_server` with few a filetypes and `g:ycm_completer_settings` with duplicates

Felixoid:
The issue in https://github.com/ycm-core/ycmd/actions/runs/22263576043/job/64405540477?pr=1813 looks like an infrastructure issue. Could you rerun the job, please?

Felixoid:
A kind reminder? @puremourning

Felixoid:
@Valloric, could you review this PR, please?

PR #1815, opened 2026-03-01
Fix %-encoding in UriToFilePath

### LINKED ISSUE #1814: Incorrect paths reported with url encoded URIs from LSP
If paths contain %-encoded values, they can be truncated and break operations,

```
b'{"filepath": "/Users/ben/Development/YouCompleteMe/tmp/hash/test#foo/main.c", "line_num": 1, "column_num": 43, "working_dir": "/Users/ben/Development/YouCompleteMe/tmp/hash/test#foo", "options": {"tab_size": 2, "insert_spaces": true}, "command_arguments": ["GoTo"]}'
2026-03-01 15:16:08,730 - DEBUG - RX (run_completer_command): 200
b'{"line_num":1,"column_num":1,"filepath":"/Users/ben/Development/YouCompleteMe/tmp/hash/test","file_only":false}'
```

Repro steps from @aster89 👍 

```
mkdir ~/foo\#bar
cd ~/foo#bar
echo 'int main(int argc, char *argv[]) { return argc; }' > main.cpp
vim -Nu /path/to/YouCompleteMe/vimrc_ycm_minimal main.cpp
```

then do a GoTo on argc. Expect: works, cursor placed in main.cpp; Actual, random buffer created with empty contents.

---

Seems to be caused by UriToFilePath. #1119 seems to have originally put in a workaround for broken python sodlib, but stdlib seems fixed now.

puremourning:
In #1119 a change was made to "unquote" the URI before converting to a file path, ostensibly to support windows. It seems that either that change was broken, or that Python standard library has since been fixed and this has caused legitimate %-encoded paths (like foo#bar/test.c) to now truncate at the #.

The tests added for UriToFilePath with Windows paths still pass unchanged without the hack on Python 3.14 so I'm happy to just remove the hack.

Fixes #1814

- - -
This change is [Reviewable](https://reviewable.io/reviews/ycm-core/ycmd/1815)

### DIFF
--- a/ycmd/completers/language_server/language_server_protocol.py
+++ b/ycmd/completers/language_server/language_server_protocol.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with ycmd.  If not, see <http://www.gnu.org/licenses/>.
 
+import sys
 import collections
 import os
 import json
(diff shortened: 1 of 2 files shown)


puremourning:
hmm it does seem that the change was in Python 3.14: https://docs.python.org/3/library/urllib.request.html#urllib.request.url2pathname

And require_scheme was only added in Python 3.14 so we can't merge this as is.

